PR #71, opened 2017-05-23
Sync status

ajlopez:
Minor improvements:

- Better send status to all
- Only process direct children, not nephews when adding a block

I think the send status to all could be better, but now, baby steps


donequis:
Changes will be added on different PR

PR #72, opened 2017-05-24
use SecureRandom.getInstanceStrong to use /dev/random instead of /dev…

y0ke:
Currently,  class `KeyCrypterScrypt` uses `new SecureRandom()` for random number generation. 
This is used to encrypt data of accounts.
`new SecureRandom()` end up selecting NativePRNG which uses `/dev/urandom` in the case of mac/Linux.
`SecureRandom.getInstanceStrong` fix it though, because of a chance that running out of entropy,Another solution might be needed.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/crypto/KeyCrypterScrypt.java
+++ b/rskj-core/src/main/java/co/rsk/crypto/KeyCrypterScrypt.java
@@ -71,7 +71,7 @@ public class KeyCrypterScrypt implements KeyCrypter {
     public static final int BLOCK_LENGTH = 16;  // = 128 bits.
 
     static {
-        secureRandom = new SecureRandom();
+        secureRandom = SecureRandom.getInstanceStrong();
     }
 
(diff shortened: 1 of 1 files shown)


juli:
The `KeyCrypterScrypt` class, and the rest of the wallet encryption code, is in the process of being rewritten (see #2 ). The risk of weak SecureRandom implementations being provided by the default classpath is known and will be addressed. (see #3 )

The change that you are proposing in this PR does not satisfy the #3 requirement of using a strong `SecureRandom` implementation independently of the default system configuration.

The actual `SecureRandom` implementation returned by `SecureRandom.getInstanceStrong` depends on the value set in the `java.security` file, that is usually `NativePRNGBlocking:SUN` in the platforms where rskj is being used but it could be different in other platforms or could change in future releases of the same JRE.  `NativePRNGBlocking:SUN` reads from `/dev/random` but as you probably know using `/dev/random` (that may block) is considered as safe as using `/dev/urandom` because they use the same source of random bits under the hood. Thanks!

PR #77, opened 2017-05-30
Flush every N blocks

oscarguindzberg:


rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
```diff
@@ -159,6 +159,11 @@ public boolean isFlushEnabled() {
                 config.getBoolean("blockchain.flush") : true;
     }
 
+    public int flushNumberOfBlocks() {
+        return config.hasPath("blockchain.flushNumberOfBlocks") ?
+                config.getInt("blockchain.flushNumberOfBlocks") : 20;
```
donequis: We shuold validate that config.getInt("blockchain.flushNumberOfBlocks") > 0

oscarguindzberg: done

PR #83, opened 2017-05-31
Sending status

ajlopez:
Improving sending status to all when syncing

### DIFF
--- a/rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
+++ b/rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
@@ -142,7 +142,7 @@ public class NodeMessageHandler implements MessageHandler, Runnable {
                 MessageTask task;
 
                 if (this.queue.isEmpty())
-                    task = this.queue.poll(60, TimeUnit.SECONDS);
+                    task = this.queue.poll(10, TimeUnit.SECONDS);
                 else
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -152,7 +152,9 @@ public void run() {
                     logger.trace("Start task");
                     this.processMessage(task.getSender(), task.getMessage());
                     logger.trace("End task");
-                } else
+                } else if (this.blockProcessor != null && this.blockProcessor.hasBetterBlockToSync())
```
donequis: Sending status should not depend on whether a status message was received recently or not.

This can be an alternative:

            try {
                logger.trace("Get task");

                final MessageTask task = this.queue.poll(10, TimeUnit.SECONDS);

                loggerMessageProcess.debug("Queued Messages: {}", this.queue.size());

                if (task != null) {
                    logger.trace("Start task");
                    this.processMessage(task.getSender(), task.getMessage());
                    logger.trace("End task");
                } else {
                    logger.trace("No task");
                }

                //Refresh status to peers every 20 seconds or so
                Long now = System.currentTimeMillis();
                if (now - lastStatusSent > 20000) {
                    this.blockProcessor.sendStatusToAll();
                    lastStatusSent = now;
                }

PR #86, opened 2017-05-31
These files should not be part of the repository

aeidelman:


oscarguindzberg:
Change PR and commit message to "These files..."

PR #97, opened 2017-06-02
Using block store cache

ajlopez:
Adding an in-memory block store cache

(use with blockchain.flush = true; to test in some nodes)

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/config/DefaultConfig.java
+++ b/rskj-core/src/main/java/org/ethereum/config/DefaultConfig.java
@@ -114,11 +114,11 @@ public class DefaultConfig {
 
 
         IndexedBlockStore cache = new IndexedBlockStore();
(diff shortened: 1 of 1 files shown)


donequis:
What do we expect with this change?

rskj-core/src/main/java/org/ethereum/config/DefaultConfig.java
```diff
@@ -114,11 +114,11 @@ public BlockStore blockStore() {
 
 
         IndexedBlockStore cache = new IndexedBlockStore();
-        cache.init(new HashMap<Long, List<IndexedBlockStore.BlockInfo>>(), new HashMapDB(), null, null);
+        cache.init(new HashMap<>(), new HashMapDB(), null, null);
```
m-sossich: Consider the repercussion of not using a "concurrent" map implementation here.

donequis:
Create PR against ginger

aeidelman:
This pull request is replaced by PR #119 and will not be merged.

PR #109, opened 2017-06-09
Load total difficulty from best block hash

ajlopez:
Modified algorithm in case the best number maps to a fork


donequis:
When does blockStore.getTotalDifficulty() return a valid value? We should fix / document / remove the method depending on how it does behave.

ajlopez:
When the getBestBlock method returns a block with number == getMaxNumber

The method getTotalDifficulty is not called from production code, after this modification. It is only called by some tests

PR #112, opened 2017-06-09
extending blocks returned by peers to 100

donequis:
this will in the short term increase bandwidth, but
that issue will be solved later


donequis:
This will reduce the chances of a node getting stuck at a given height

PR #114, opened 2017-06-11
Blocks by number

ajlopez:
New Web3 method

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
@@ -36,6 +36,7 @@ import org.ethereum.config.blockchain.RegTestConfig;
 import org.ethereum.core.*;
 import org.ethereum.crypto.ECKey;
 import org.ethereum.crypto.HashUtil;
+import org.ethereum.db.BlockInformation;
 import org.ethereum.db.TransactionInfo;
 import org.ethereum.facade.Ethereum;
 import org.ethereum.listener.CompositeEthereumListener;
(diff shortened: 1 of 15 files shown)


rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -686,6 +695,28 @@ public BlockResult getBlockResult(Block b, boolean fullTx) {
...
+        try {
+            blockNumber = StringHexToBigInteger(number).longValue();
+        } catch (NumberFormatException | StringIndexOutOfBoundsException e) {
+            throw new JsonRpcInvalidParamException("invalid blocknumber " + number);
+        }
+
+        List<BlockInformationResult> result = new ArrayList<>();
+        Blockchain blockchain = this.worldManager.getBlockchain();
+
+        synchronized (blockchain) {
```
donequis: Do not sync over the blockchain object. Check accessLock object in blockchainImpl

ajlopez:
The original BlockInfo is used with Java serialization by the current index store, so I created the new one to don't alter in any way the current behavior, and have the new method exposing data that is not related only with IndexedBlockStore

rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -686,6 +695,26 @@ public BlockResult getBlockResult(Block b, boolean fullTx) {
         return br;
     }
 
+    public BlockInformationResult[] eth_getBlocksByNumber(String number) {
+        long blockNumber;
+
+        try {
+            blockNumber = StringHexToBigInteger(number).longValue();
```
donequis: Consider the method StringNumberAsBigInt() instead of StringHexToBigInteger()

donequis:
Create PR against ginger

PR #118, opened 2017-06-17
Send no uncle in processGetBlock

ajlopez:
Now, we don't need the uncles to honor get block messages.


bcodesido:
Please, check code coverage

PR #119, opened 2017-06-18
Block cache

ajlopez:
Remove original IndexedBlockStore cache

Add a Block cache by hash, with eviction policy


bcodesido:
co.rsk.core.bc.BlockChainImplTest > addInvalidBlockOneBadStateRoot FAILED

Issue #75, opened 2017-05-30
Inheritance/Function Modifiers seems not working properly


y0ke:
My contract uses Inheritance and Function Modifiers. Which is something like below

```
contract Bar is Foo {
  ....
  function setFooAddress(address _fooAddress) onlyWorking {
    ....
  }
}
```

`onlyWorking` comes from 

```
contract Foo {
    function ownerValidate(address addr) constant returns (bool);
    bool public isWorking;

    function Owned() {
        isWorking = true;
    
    modifier onlyWorking {
        if (!isWorking) throw;
        //_;
        _
    }

    function setWorking(bool _isWorking) onlyOwner {
        isWorking = _isWorking;
    }
}
```

It seems works properly until once `isWorking` is changed, contract Bar's almost all functions returns wrong value after `isWorking` is changed.
the contract works great in Ethereum.

I'm wondering `setWorking` changes wrong position of the storage data though I'm not sure.

ajlopez:
Thanks for your reporting!

We reproduced the issue, with and without using modifier. Can you check the value of isWorking, AFTER CHANGING it?

In our issue reproduction, after the change of isWorking variable,  we get the value and it was not changed. Is it your case?

y0ke:
A state remains the same after calling `setWorking`.
I haven't checked the corresponding data position in its storage.So I'm not sure whether it updates proper position of storage or not. But one thing for sure is that I can't get any other variable's value after changing it. as far as I know, other variables starts returning null after calling `setWorking`. of course they have non-null value originally.

ajlopez:
We are working on the issue. We found one root cause. The solution will be included in an upcoming release (I'm not sure about the release date, yet), But it is related to our storage algorithms,

Your report helped us a lot, thanks!

PR #120, opened 2017-06-18
Fix change variable

ajlopez:
Fix to change variable to 0, related to issue #75

### DIFF
--- a/rskj-core/src/main/java/co/rsk/trie/TrieImpl.java
+++ b/rskj-core/src/main/java/co/rsk/trie/TrieImpl.java
@@ -1,31 +1,17 @@
-/*
- * This file is part of RskJ
- * Copyright (C) 2017 RSK Labs Ltd.
- *
- * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


ajlopez:
I was unable to fix the A1 issue (discussed in email list) using the original code

rskj-core/src/main/java/co/rsk/trie/TrieImpl.java
```diff
@@ -1,31 +1,17 @@
-/*
```
bcodesido: License was removed.

rskj-core/src/main/java/co/rsk/trie/TrieImpl.java
```diff
@@ -40,7 +26,7 @@
 import static org.ethereum.util.ByteUtil.EMPTY_BYTE_ARRAY;
 
 /**
- * Trie is the trie node.
+ * NewTrie is the trie node.
```
bcodesido: Why newtrie?

PR #121, opened 2017-06-20
Set next nonce to null if tx list is empty in TxsPerAccount

ajlopez:
And remove txsPerAccount when remove nonce and the list is empty


rskj-core/src/main/java/co/rsk/net/handler/TxHandlerImpl.java
```diff
@@ -107,7 +107,10 @@ private void cleanTxs() {
                 if (txsPerAccount != null)
                 {
                     txsPerAccount.removeNonce(new BigInteger(1, txt.tx.getNonce()));
+                    if (txsPerAccount.getTransactions().isEmpty())
```
bcodesido: Refactor this code to not nest more than 3 if/for/while/switch/try statements.

PR #126, opened 2017-06-23
Add oscar as author

oscarguindzberg:


rskj-core/src/main/java/co/rsk/util/peg/ReadBridgeState.java
```diff
@@ -0,0 +1,61 @@
+package co.rsk.util.peg;
```
bcodesido: This file is correctly committed?

rskj-core/src/main/resources/rskj.conf
```diff
@@ -0,0 +1,326 @@
+peer.discovery = {
```
bcodesido: This file is correctly committed?

rskj-core/src/main/java/co/rsk/util/peg/ReadBridgeState.java
oscarguindzberg: removed the file from the PR

rskj-core/src/main/resources/rskj.conf
oscarguindzberg: removed the file from the PR, sorry

oscarguindzberg:
@aeidelman 
1)  I just commited a change so @author is not mixed with "created by"
2) It is very common classes have several authors. See this example in the jdk source code
http://hg.openjdk.java.net/jdk7/jdk7/jdk/file/00cd9dc3c2b5/src/share/classes/java/util/ArrayList.java

PR #127, opened 2017-06-23
Accept advanced blocks afterntime

aeidelman:
This will let accept higher blocks (>1000) if block sync does not make progress in X time

### DIFF
--- a/rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
+++ b/rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
@@ -49,6 +49,7 @@ public class NodeMessageHandler implements MessageHandler, Runnable {
     private static final Logger loggerMessageProcess = LoggerFactory.getLogger("messageProcess");
     public static final int MAX_NUMBER_OF_MESSAGES_CACHED = 5000;
     public static final long RECEIVED_MESSAGES_CACHE_DURATION = TimeUnit.MINUTES.toMillis(2);
+    public static final long WAIT_TIME_ACCEPT_ADVANCED_BLOCKS = TimeUnit.MINUTES.toMillis(10);
     private final BlockProcessor blockProcessor;
     private final ChannelManager channelManager;
     private final PendingState pendingState;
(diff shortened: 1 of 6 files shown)


rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -61,6 +62,7 @@
     private LinkedBlockingQueue<MessageTask> queue = new LinkedBlockingQueue<>();
     private Set<ByteArrayWrapper> receivedMessages = Collections.synchronizedSet(new HashSet<ByteArrayWrapper>());
     private long cleanMsgTimestamp = 0;
+    private long lastImportedBestBlock;
```
donequis: Explicitly initialize value

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -246,14 +248,28 @@ private void processBlockMessage(@Nonnull final MessageSender sender, @Nonnull f
...
             result.logResult(block.getShortHash(), time);
 
         Metrics.processBlockMessage("blockProcessed", block, sender.getNodeID());
 
+        long currentTimeMillis = System.currentTimeMillis();
+
+        if (result.anyImportedBestResult())
+            lastImportedBestBlock = currentTimeMillis;
+
+        if (currentTimeMillis - lastImportedBestBlock > WAIT_TIME_ACCEPT_ADVANCED_BLOCKS)
```
donequis: If the first block ever is not imported best, this would be triggered

aeidelman: Fixed

marian151: Nothing works here

donequis: What do youbmean, marian151?

marian151: Faucet System never works

aeidelman: @marian151 we appreciate your report about the faucet, but you're reporting in places that don't have to do anything with this.

PR #138, opened 2017-07-13
Ipv6 bootnodes

m-sossich:


rskj-core/src/main/java/co/rsk/util/IpUtils.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Created by mario on 13/07/17.
+ */
+public class IpUtils {
+    private static final Logger logger = LoggerFactory.getLogger(IpUtils.class);
+
+    private static final String IPV6_INPUT_FORMAT = "^\\[(.*)\\]:([0-9]{1,})";
```
juli: * Square brackets are not mandatory for IPv6 addresses if they are not part of an URL or if a port number is not specified. But if there won't be a default port for nodes then it is correct.
* Second group of regex for port number could limit to 5 digits and allow IPs without port

rskj-core/src/test/java/co/rsk/util/IpUtilsTest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.List;
+
+/**
+ * Created by mario on 13/07/17.
+ */
+public class IpUtilsTest {
+
+    private static final String IPV6_WITH_PORT = "[2001:db8:85a3:8d3:1319:8a2e:370:7348]:443";
+    private static final String IPV6_NO_PORT = "[2001:db8:85a3:8d3:1319:8a2e:370:7348]";
+    private static final String IPV6_INVALID = "2001:db8:85a3:8d3:1319:8a2e:370:7348";
```
juli: This IPv6 address is valid if square brackets are not mandatory.

PR #139, opened 2017-07-17
Db pool

ajlopez:
The opened data sources are closed after a period of time.

The time is specified in seconds:

´´´
# unused datasource is closed after n seconds
# default value: 0 (no close)
datasource.close = 600
´´´


ajlopez:
The pool supports concurrent access. But the datasources are not thread-safe.

The close of unused datasources is only triggered from the DataDetailsStore, from a flush operation that is called synchronized.

donequis:
The problem I see is that someone can access a DataSource instance while the closeUnused is called.

tinchou:
Can we close or move forward with fixing this PR?

PR #145, opened 2017-07-23
Refactor block in blockchain logic, to avoid reading the full blocks

ajlopez:
Refactor to improve block already in blockchain detection. Given the block height, it avoids the read of the full blocks at that height, using directly the known hashes.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockUtils.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockUtils.java
@@ -44,7 +44,7 @@ public class BlockUtils {
         final List<BlockInformation> blocks = blockChain.getBlocksInformationByNumber(blockNumber);
 
         for (final BlockInformation bi : blocks) {
-            if (new ByteArrayWrapper(bi.getHash()).equals(key)) {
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/core/bc/BlockUtils.java
```diff
@@ -35,20 +36,15 @@
...
-        if (block == null)
-            return false;
-
-        final ByteArrayWrapper key = new ByteArrayWrapper(block.getHash());
-        final List<Block> blocks = blockChain.getBlocksByNumber(block.getNumber());
-
-        for (final Block b : blocks) {
-            if (new ByteArrayWrapper(b.getHash()).equals(key)) {
+        for (final BlockInformation bi : blocks) {
+            if (new ByteArrayWrapper(bi.getHash()).equals(key)) {
```
donequis: ByteArrayWrapper unwraps the array when calling equals. It'd be nice to make a new method on the class called something like "equalsToByteArray(byte[] array)" that would do the FastByteComparisons.compareTo() and clal it from equals later.

donequis:
I'll add the suggestions to the PR.

PR #151, opened 2017-07-27
Scoring

ajlopez:
Base functionality for Peer Scoring


rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -134,8 +137,10 @@ public void postMessage(MessageSender sender, Message message) throws Interrupte
             }
             this.queue.offer(new MessageTask(sender, message));
         } else {
+            recordEvent(sender, EventType.REPEATED_MESSAGE);
```
donequis: The same message may come from different peers

ajlopez: It's the intention. The count of repeated message is not related to the SAME node, but to the receipt of messages that was already sent by the node OR BY OTHER NODES. Currently, it is not used in any punishment or scoring calculation. But it is useful to know using RPC what is behavior of message, without using the logs. It could be removed, or improved, but when the specification of scoring is better defined

donequis: Good enough for me.

PR #154, opened 2017-08-03
Immutable txs

ajlopez:
The transactions that are recreated using RLP-encoding are immutable


donequis:
Just asking if it was considered. An inmutable and a mutable transaction may be equal when calling the equals() method.

rskj-core/src/main/java/org/ethereum/core/ImmutableTransaction.java
```diff
@@ -0,0 +1,28 @@
+package org.ethereum.core;
+
+import org.ethereum.crypto.ECKey;
+
+/**
+ * Created by ajlopez on 02/08/2017.
+ */
+public class ImmutableTransaction extends Transaction {
```
tinchou: I think it would be best to have this reversed: `Transaction` inherits from `ImmutableTransaction` and adds the `sign` and `setGasLimit`. This way, instead of having exceptions thrown (i.e. runtime checks), we have the compiler check things for us (compile time checks).

rskj-core/src/test/java/co/rsk/test/builders/TransactionBuilder.java
```diff
@@ -80,6 +87,10 @@ public TransactionBuilder gasLimit(BigInteger gasLimit) {
     public Transaction build() {
         Transaction tx = Transaction.create(receiver != null ? Hex.toHexString(receiver.getAddress()) : (receiverAddress != null ? Hex.toHexString(receiverAddress) : null), value, nonce, gasPrice, gasLimit, data);
         tx.sign(sender.getEcKey().getPrivKeyBytes());
+
+        if (this.immutable)
+            return new ImmutableTransaction(tx.getEncoded());
```
tinchou: .toImmutableTransaction()

rskj-core/src/test/java/org/ethereum/core/TransactionTest.java
```diff
@@ -178,7 +178,7 @@ public void test3() throws Exception {
     @Ignore
     @Test
     public void testTransactionFromSignedRLP() throws Exception {
-        Transaction txSigned = new Transaction(Hex.decode(RLP_ENCODED_SIGNED_TX));
+        Transaction txSigned = new ImmutableTransaction(Hex.decode(RLP_ENCODED_SIGNED_TX));
```
tinchou: change type to ImmutableTransaction

rskj-core/src/main/java/org/ethereum/core/ImmutableTransaction.java
donequis: Sounds like this adds overhead to the PR while not adding that much benefit. Though this change would be welcome.

PR #155, opened 2017-08-03
Sealed block

ajlopez:
When a block is sealed, it cannot be modified. The seal includes the header.

If an unsealed block is sent to the blockchain, a panic is logged.

The blocks that are created using RLP are sealed.

The blocks that are completed by MinerServerImpl are sealed before they are added to the system.}

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
@@ -184,6 +184,11 @@ public class BlockChainImpl implements Blockchain, org.ethereum.facade.Blockchai
      */
     @Override
     public ImportResult tryToConnect(Block block) {
+        if (block == null)
+            return ImportResult.INVALID_BLOCK;
+        
+        block.seal();
+
         if (blockRecorder != null)
             blockRecorder.writeBlock(block);
 
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
@@ -186,8 +186,11 @@ public class BlockChainImpl implements Blockchain, org.ethereum.facade.Blockchai
     public ImportResult tryToConnect(Block block) {
         if (block == null)
             return ImportResult.INVALID_BLOCK;
-        
-        block.seal();
+
+        if (!block.isSealed()) {
+            panicProcessor.panic("unsealedblock", String.format("Unsealed block %s %s", block.getNumber(), Hex.toHexString(block.getHash())));
+            block.seal();
+        }
 
         if (blockRecorder != null)
             blockRecorder.writeBlock(block);
--- a/rskj-core/src/main/java/org/ethereum/core/Block.java
+++ b/rskj-core/src/main/java/org/ethereum/core/Block.java
@@ -70,7 +70,7 @@ public class Block {
 
     private Trie txsState;
 
-    /* Indicates if this block cannot be changed */
+    /* Indicates if this block can or cannot be changed */
     private volatile boolean sealed;
 
     public Block(byte[] rawData) {
--- a/rskj-core/src/main/java/org/ethereum/core/Block.java
+++ b/rskj-core/src/main/java/org/ethereum/core/Block.java
@@ -70,9 +70,12 @@ public class Block {
 
     private Trie txsState;
 
+    protected boolean sealed;
+
     /* Constructors */
     public Block(byte[] rawData) {
         this.rlpEncoded = rawData;
+        this.sealed = true;
     }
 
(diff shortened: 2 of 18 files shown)


rskj-core/src/main/java/org/ethereum/core/Block.java
```diff
@@ -182,6 +185,26 @@ public Block(byte[] parentHash, byte[] unclesHash, byte[] coinbase, byte[] logsB
...
+        return this.sealed;
+    }
+
+    // Clone this block allowing modifications
+    public Block cloneBlock() {
+        Block clone = new Block(this.getEncoded());
+        clone.sealed = false;
+        clone.parseRLP();;
+        // Allow modifications to header
+        clone.header.unseal();
```
donequis: clone header

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -197,6 +201,18 @@ public BlockHeader(byte[] parentHash, byte[] unclesHash, byte[] coinbase,
         this.uncleCount = uncleCount;
     }
 
+    public boolean isSealed() {
+        return this.sealed;
+    }
+
+    public void unseal() {
```
donequis: delete this if possible

donequis:
getEncoded() may modify the BlockHeader

donequis:
A possible approach is to get the encoding when sealing and caching the value on a transient variable

rskj-core/src/main/java/org/ethereum/core/Block.java
ajlopez: The header is a new one, created from RLP. It is only referenced by this clone block

rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
```diff
@@ -184,6 +184,14 @@ public long getSize() {
      */
     @Override
     public ImportResult tryToConnect(Block block) {
+        if (block == null)
+            return ImportResult.INVALID_BLOCK;
+
+        if (!block.isSealed()) {
+            panicProcessor.panic("unsealedblock", String.format("Unsealed block %s %s", block.getNumber(), Hex.toHexString(block.getHash())));
+            block.seal();
```
martinmedina: If an unsealed block arrives here is because something out of the expected flow happened. I understand that it is why the panic is logged. 
What I don't get is why the block is sealed and accepted **instead** of returning INVALID_BLOCK?

rskj-core/src/main/java/org/ethereum/core/Block.java
```diff
@@ -660,4 +711,10 @@ public void flushRLP() {
         this.rlpEncoded = null;
         this.parsed = true;
     }
+
+    public static class SealedBlockException extends RuntimeException {
```
martinmedina: Creating this exception on it's own file would be better. That's a way to avoid unnecessary growth on this already too long class file.

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -506,4 +590,10 @@ public String getShortHashForMergedMining() {
     public byte[] getHashForMergedMining() {
         return HashUtil.sha3(getEncoded(false));
     }
+
+    public static class SealedBlockHeaderException extends RuntimeException {
```
martinmedina: Creating this exception on it's own file would be better. That's a way to avoid unnecessary growth on this already too long class file.

rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
donequis: This is just for checking that everything is OK, but we don't want the behavior to change on testnet.

martinmedina: Ok, let's remember to delete this check once the inmutable block change is applied

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -96,11 +97,14 @@
     private byte[] minimumGasPrice;
     private int uncleCount;
 
-    public BlockHeader(byte[] encoded) {
-        this((RLPList) RLP.decode2(encoded).get(0));
+    /* Indicates if this block header cannot be changed */
+    private volatile boolean sealed;
+
+    public BlockHeader(byte[] encoded, boolean sealed) {
```
donequis: I'd vote for a method that only receives the rlp data and defaults sealed to true.

```
public BlockHeader(byte[] encoded) {
    this(encoded, true);
}
```

tinchou:
I would've tried to implement the sealed block as a separate class, but this seems to be OK.

PR #156, opened 2017-08-04
Replace Jetty with Netty as JSON-RPC Server

diega:


rskj-core/src/main/java/org/ethereum/rpc/JsonRpcWeb3ServerHandler.java
```diff
@@ -0,0 +1,98 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class JsonRpcWeb3ServerHandler extends SimpleChannelInboundHandler<FullHttpRequest> {
+
+    private static Logger LOGGER = LoggerFactory.getLogger("jsonrpc");
```
bauna: missing final?

rskj-core/src/main/java/org/ethereum/rpc/JsonRpcWeb3ServerHandler.java
```diff
@@ -0,0 +1,98 @@
...
+        int readerIndex = body.readerIndex();
+        body.getBytes(readerIndex, rawBody);
+
+        ByteBuf content = Unpooled.buffer();
+        HttpResponseStatus responseStatus = HttpResponseStatus.OK;
+        try (ByteBufOutputStream os = new ByteBufOutputStream(content)){
+            JsonNode jsonRpcRequest = mapper.readTree(rawBody);
+            int result = jsonRpcServer.handleNode(jsonRpcRequest, os);
+            if (result != 0) {
+                if (result == -32700 || result == -32602 || result == -32603 || (result <= -32000 && result >= -32099)) {
```
donequis: Improve error codes names based on doc

http://www.jsonrpc.org/specification#error_object

rskj-core/src/main/java/org/ethereum/rpc/JsonRpcWeb3ServerHandler.java
```diff
@@ -0,0 +1,123 @@
...
+        this.corsConfiguration = corsConfiguration;
+        this.jsonRpcServer = new JsonRpcFilterServer(service, service.getClass(), filteredModules);
+        jsonRpcServer.setErrorResolver(new MultipleErrorResolver(new RskErrorResolver(), AnnotationsErrorResolver.INSTANCE, DefaultErrorResolver.INSTANCE));
+    }
+
+    @Override
+    protected void channelRead0(ChannelHandlerContext ctx, FullHttpRequest request) throws Exception {
+        HttpMethod httpMethod = request.getMethod();
+        HttpResponse response;
+        if (HttpMethod.OPTIONS.equals(httpMethod)) {
```
m-sossich: Maybe you could implement a sort of Chain Of Responsibility to handle the different possible methods of the http Request and get the response like that, instead of having this if/else/if/else structure

rskj-core/src/main/java/org/ethereum/rpc/JsonRpcWeb3ServerHandler.java
```diff
@@ -0,0 +1,123 @@
...
+            int readerIndex = body.readerIndex();
+            body.getBytes(readerIndex, rawBody);
+
+            ByteBuf content = Unpooled.buffer();
+            HttpResponseStatus responseStatus = HttpResponseStatus.OK;
+            try (ByteBufOutputStream os = new ByteBufOutputStream(content)){
+                JsonNode jsonRpcRequest = mapper.readTree(rawBody);
+                int result = jsonRpcServer.handleNode(jsonRpcRequest, os);
+                if (result != 0) {
+                    if (result == JSON_RPC_PARSE_ERROR_CODE ||
```
m-sossich: Maybe, Instead of doing this you could model the response codes with an Enum, and each response knows to which http status gets mapped, so you could get something like 

responseStatus = SarasaEnum.velueOf(result).httpStatus()

Or something like that...

diega: That's what the latest versions of *jsonrpc4j* does. If we bump the library version we can handle it in the same way they do it for servlets (ref: [JsonRpcServer.java](https://github.com/briandilley/jsonrpc4j/blob/master/src/main/java/com/googlecode/jsonrpc4j/JsonRpcServer.java#L150) )

rskj-core/src/main/java/org/ethereum/rpc/JsonRpcWeb3ServerHandler.java
diega: Do you know if is possible to implement it on top of Netty ChannelHandlers?

rskj-core/src/main/java/org/ethereum/rpc/JsonRpcNettyServer.java
```diff
@@ -0,0 +1,72 @@
...
+                    p.addLast(new HttpRequestDecoder());
+                    p.addLast(new HttpObjectAggregator(1024 * 1024 * 5));
+                    p.addLast(new HttpResponseEncoder());
+                    p.addLast(new HttpContentCompressor());
+                    if (corsConfiguration.hasHeader()) {
+                        p.addLast(new CorsHandler(
+                            CorsConfig
+                                .withOrigin(corsConfiguration.getHeader())
+                                .allowedRequestHeaders(HttpHeaders.Names.CONTENT_TYPE)
+                                .allowedRequestMethods(HttpMethod.POST)
```
juli: * Content-type header is allowed but the type is not enforced (application/json).
 * Previous code allowed GET and OPTIONS
 * Previous code added Vary: Origin header to avoid caching problems
 * Previous code set Access-Control-Max-Age to 600

rskj-core/src/main/java/org/ethereum/rpc/JsonRpcNettyServer.java
```diff
@@ -0,0 +1,72 @@
...
+        b.option(ChannelOption.SO_REUSEADDR, reuseAddress);
+        b.group(bossGroup, workerGroup)
+            .channel(NioServerSocketChannel.class)
+            .handler(new LoggingHandler(LogLevel.INFO))
+            .childHandler(new ChannelInitializer<SocketChannel>() {
+                @Override
+                protected void initChannel(SocketChannel ch) throws Exception {
+                    ChannelPipeline p = ch.pipeline();
+                    p.addLast(new HttpRequestDecoder());
+                    p.addLast(new HttpObjectAggregator(1024 * 1024 * 5));
```
juli: The 5MB limit should be documented in the JSON RPC docs

rskj-core/src/main/java/org/ethereum/rpc/JsonRpcNettyServer.java
```diff
@@ -0,0 +1,72 @@
...
+                    p.addLast(new HttpResponseEncoder());
+                    p.addLast(new HttpContentCompressor());
```
juli: Compression should be cofigurable (CPU vs Bandwidth decision)

rskj-core/build.gradle
```diff
@@ -142,10 +142,11 @@ dependencies {
     compile "com.typesafe:config:1.3.1"
     compile "org.mapdb:mapdb:2.0-beta13"
     compile "co.rsk.bitcoinj:bitcoinj-thin:${bitcoinjVersion}"
-    compile "org.eclipse.jetty:jetty-server:9.3.16.v20170120"
-    compile "org.eclipse.jetty:jetty-servlet:9.3.16.v20170120"
+    compile 'javax.servlet:javax.servlet-api:3.1.0'
```
juli: missing checksum?

PR #161, opened 2017-08-15
Validate gaslimit on pendingstate

donequis:


rskj-core/src/main/java/org/ethereum/core/Block.java
```diff
@@ -701,6 +701,10 @@ public BigInteger getMinGasPriceAsInteger() {
         return (this.getMinimumGasPrice() == null) ? null : new BigInteger(1, this.getMinimumGasPrice());
     }
 
+    public BigInteger getGasLimitAsInteger() {
```
tinchou: Please use `BigIntegers.fromUnsignedByteArray`

rskj-core/src/main/java/co/rsk/net/handler/txvalidator/TxValidatorGasLimitValidator.java
```diff
@@ -32,6 +33,6 @@
     @Override
     public boolean validate(Transaction tx, AccountState state, BigInteger gasLimit, BigInteger minimumGasPrice, long bestBlockNumber) {
         BigInteger txGasLimit = tx.getGasLimitAsInteger();
-        return gasLimit.compareTo(txGasLimit) >= 0;
+        return  txGasLimit.compareTo(gasLimit) <= 0;
```
tinchou: Please use `BIUtil.isMoreThan(gasLimit, txGasLimit)` or `!BIUtil.isLessThan(txGasLimit, gasLimit)`

aeidelman: Semantic of those methods is not very clear IMHO

rskj-core/src/main/java/org/ethereum/core/Block.java
aeidelman: Done

rskj-core/src/main/java/co/rsk/net/handler/txvalidator/TxValidatorGasLimitValidator.java
tinchou: Would be great to have extension methods

PR #162, opened 2017-08-17
Fixes RLP border cases and adds more checks

tinchou:
This fixes some border cases in RLP, and includes a refactor with multiple bound checks.

We fixed an issue previously known in EthJ: https://github.com/ethereum/ethereumj/issues/636
We also discovered and fixed a bug in the `decodeBigInteger` implementation: https://github.com/ethereum/ethereumj/issues/917.


tinchou:
Thanks @donequis! I'm a little worried that changing those constants will make it harder for people coming from the EthJ codebase, but it should be OK.

PR #164, opened 2017-08-21
Copyonwrite (not removed)

ajlopez:
Comments added.

Minor improvement: transactions and uncles list in block are exposed as read-only collections


tinchou:
It caught my attention that you didn't change the return type of the methods after switching to `unmodifiableList`. I come from a non-Java background and that seemed wrong from a static typing point of view. I see that some people agree with me: http://wiki.c2.com/?UnmodifiableListIsStupidAndItBreaksLsp

Regardless, if there's no better solution and all our tests are passing (meaning no one was changing the list), I'm OK with using `unmodifiableList`.

PR #174, opened 2017-09-05
limit queue size on outbound messages

donequis:


donequis:
Log level is probably temporary

PR #189, opened 2017-09-11
Uses unmodifiableList for transactions in Block.getTransactionsList

tinchou:
This adds some basic immutability checks. It uncovers some tech debt in the bridge tests.

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/Block.java
+++ b/rskj-core/src/main/java/org/ethereum/core/Block.java
@@ -21,10 +21,9 @@ package org.ethereum.core;
 
 import co.rsk.panic.PanicProcessor;
 import co.rsk.remasc.RemascTransaction;
-import org.apache.commons.collections4.CollectionUtils;
-import org.ethereum.crypto.SHA3Helper;
 import co.rsk.trie.Trie;
 import co.rsk.trie.TrieImpl;
+import org.ethereum.crypto.SHA3Helper;
 import org.ethereum.util.ByteUtil;
 import org.ethereum.util.RLP;
 import org.ethereum.util.RLPElement;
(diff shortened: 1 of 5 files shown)


tinchou:
Also, removes `CopyOnWrite` usage, which would be inefficient if anyone ends up using it.

rskj-core/src/main/java/org/ethereum/core/Block.java
```diff
@@ -416,6 +421,8 @@ public String toFlatString() {
     }
 
     private void parseTxs(RLPList txTransactions) {
+        // TODO(mc) check if we can avoid copying existing transactions
+        ArrayList<Transaction> builder = new ArrayList<>(this.transactionsList);
```
ajlopez: Reconsider use case: the transaction list is expected to be empty

tinchou:
ping @ajlopez

tinchou:
@diega would you like to review?

rskj-core/src/main/java/org/ethereum/core/Block.java
```diff
@@ -455,6 +458,8 @@ public String toFlatString() {
     }
 
     private void parseTxs(RLPList txTransactions) {
+        ArrayList<Transaction> builder = new ArrayList<>();
```
diega: It's recommended to use an interface whenever you can, in this case _builder_ could be  `List<Transaction>`. Also I think the name of the variable could be improved, what about _parsedTransactions_?

rskj-core/networkStateExporterTest.json
```diff
@@ -1,14 +0,0 @@
-{
```
diega: what's this removal?

tinchou: Thanks for spotting!

PR #197, opened 2017-09-27
Replace solidity in ignored tests

ajlopez:
SolidityCompiler reference and use was removed from three ignored tests. Instead of using the node solidity compiler in those tests, the compiled code and ABI is used.

Solidity compiler reference was removed from other tests, where it was not being used.

Out of scope: remove solidity compiler fromt the node, to be discussed.


rskj-core/src/test/java/org/ethereum/core/TransactionTest.java
```diff
@@ -589,86 +578,70 @@ public void multiSuicideTest() throws IOException, InterruptedException {
...
-            Assert.assertNotNull(contract1.parseInvocation(callData).toString());
-
-            try {
-                contract1.parseInvocation(new byte[32]);
-                Assert.fail();
-            }
-            catch (RuntimeException ex) {
-            }
+        CallTransaction.Contract contract1 = new CallTransaction.Contract(abi);
+        byte[] callData = contract1.getByName("multipleHomocide").encode();
```
tinchou: Please change to "homicide", since the "homocide" typo might be offensive.

rskj-core/src/test/java/org/ethereum/core/TransactionTest.java
```diff
@@ -589,86 +578,70 @@ public void multiSuicideTest() throws IOException, InterruptedException {
...
+                throw;
+            }
+        }
+
+        contract TestEventInvoker {
+            event externalEvent();
+
+            function doIt(address invokedAddress) {
+                externalEvent();
+                //TestEventInvoked(invokedAddress).doIt();
```
tinchou: Why does the Solidity program have commented lines?

PR #198, opened 2017-09-27
Configure windows

bcodesido:
Configure.sh with changes for working on windows


configure.sh
```diff
@@ -43,6 +43,8 @@ function platform() {
 		PLATFORM='linux'
 	elif [[ "$unamestr" == 'Darwin' ]]; then
 		PLATFORM='mac'
+	elif [[ "$unamestr" =~ 'MINGW' ]]; then
+		PLATFORM='windows'
 	fi
```
tinchou: How about adding an `else` and exiting with an error if the platform is not recognized?

configure.sh
```diff
@@ -43,6 +43,8 @@ function platform() {
...
 }
```
tinchou: nit: it would be great if instead of using a global variable we returned a value from `platform`. Then passed it to `downloadJar` like: `downloadJar(getPlatform())`

configure.sh
bcodesido: Done

PR #201, opened 2017-10-02
Fix some sonar bugs

ajlopez:
Fix some sonar issues reported in Quality Gate

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
+++ b/rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
@@ -47,6 +47,7 @@ import java.net.Socket;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
+import java.util.concurrent.TimeUnit;
 
 import static org.ethereum.crypto.SHA3Helper.sha3;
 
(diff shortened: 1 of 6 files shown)


rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -690,7 +690,7 @@ public int scoringAddressesPunishmentIncrement() {
 
     public long scoringAddressesPunishmentMaximumDuration() {
         // default value: 1 week
-        return getLong("scoring.addresses.maximum", 60 * 24 * 7)  * 60000L;
+        return getLong("scoring.addresses.maximum", 60L * 24 * 7)  * 60000L;
```
tinchou: Is it possible to use methods that represent these conversions? I think we could use something like:

```java
return TimeUnit.MINUTES.toMillis(getLong("scoring.addresses.maximum", TimeUnit.DAYS.toMinutes(7)));
```

I don't think it's the best, but it does make clear that:

1. `"scoring.addresses.maximum"` is a duration in minutes
1. `scoringAddressesPunishmentMaximumDuration` returns a duration in millis

PR #202, opened 2017-10-02
Fix get logs

ajlopez:
The JSON parser used to parse input JSONs in Web3 JSON RPC methods, parses an array of Strings as List<String>. Previous code expected String[]


rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -1134,9 +1108,9 @@ public String eth_newFilter(FilterRequest fr) throws Exception {
 
             if (fr.address instanceof String) {
                 logFilter.withContractAddress(stringHexToByteArray((String) fr.address));
-            } else if (fr.address instanceof String[]) {
+            } else if (fr.address instanceof List<?> && !((List)fr.address).isEmpty() && ((List)fr.address).get(0) instanceof String) {
```
tinchou: I think it would be better to allow an instance of any `Collection`, and also filter to only allow objects that are an instance of the `String` class. This might be an implementation:

```java
            } else if (fr.address instanceof Collection<?>) {
                Collection<?> iterable = (Collection<?>) fr.address;
                byte[][] addr = iterable.stream()
                        .filter(String.class::isInstance).map(String.class::cast)
                        .map(TypeConverter::stringHexToByteArray)
                        .toArray(byte[][]::new);
                logFilter.withContractAddress(addr);
            }
```

rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -1148,9 +1122,9 @@ public String eth_newFilter(FilterRequest fr) throws Exception {
                         logFilter.withTopic(null);
                     } else if (topic instanceof String) {
                         logFilter.withTopic(new DataWord(stringHexToByteArray((String) topic)).getData());
-                    } else if (topic instanceof String[]) {
+                    } else if (topic instanceof List<?> && !((List)topic).isEmpty() && ((List)topic).get(0) instanceof String) {
```
tinchou: same as above

PR #205, opened 2017-10-03
Improve INetAddress related classes

ajlopez:
Fix code smells reporder by Sonar Quality Gate

### DIFF
--- a/rskj-core/src/main/java/co/rsk/scoring/InetAddressBlock.java
+++ b/rskj-core/src/main/java/co/rsk/scoring/InetAddressBlock.java
@@ -1,7 +1,6 @@
 package co.rsk.scoring;
 
 import com.google.common.annotations.VisibleForTesting;
-import org.spongycastle.pqc.math.linearalgebra.ByteUtils;
 
 import java.net.InetAddress;
 import java.util.Arrays;
--- a/rskj-core/src/main/java/co/rsk/scoring/InetAddressBlock.java
+++ b/rskj-core/src/main/java/co/rsk/scoring/InetAddressBlock.java
@@ -1,6 +1,7 @@
 package co.rsk.scoring;
 
 import com.google.common.annotations.VisibleForTesting;
+import org.spongycastle.pqc.math.linearalgebra.ByteUtils;
 
 import java.net.InetAddress;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


rskj-core/src/main/java/co/rsk/scoring/InetAddressBlock.java
```diff
@@ -94,7 +95,7 @@ public boolean equals(Object obj) {
 
     @VisibleForTesting
     public byte[] getBytes() {
-        return this.bytes;
+        return ByteUtils.clone(this.bytes);
```
tinchou: Please use a native alternative such as `bytes.clone()`: https://stackoverflow.com/a/36513254.

rskj-core/src/main/java/co/rsk/scoring/InetAddressUtils.java
```diff
@@ -26,10 +28,7 @@ public static boolean hasMask(String text) {
 
         String[] parts = text.split("/");
 
-        if (parts.length != 2 || parts[0].length() == 0 || parts[1].length() == 0)
-            return false;
-
-        return true;
+        return !(parts.length != 2 || parts[0].length() == 0 || parts[1].length() == 0);
```
tinchou: How about some De Morgan? `return parts.length == 2 && ...`

rskj-core/src/main/java/co/rsk/scoring/InetAddressUtils.java
```diff
@@ -11,6 +11,8 @@
  * Created by ajlopez on 15/07/2017.
  */
 public class InetAddressUtils {
```
tinchou: Maybe declare the class final? https://stackoverflow.com/questions/7486012/static-classes-in-java

rskj-core/src/main/java/co/rsk/scoring/InetAddressUtils.java
```diff
@@ -40,11 +39,11 @@ public static boolean hasMask(String text) {
      *
      * @return  the text converted to an InetAddress
      */
-    public static InetAddress getAddressForBan(String name) throws InvalidInetAddressException {
-        if (name == null)
+    public static InetAddress getAddressForBan(String hostname) throws InvalidInetAddressException {
```
tinchou: Maybe we can annotate the `hostname` parameter with `@CheckForNull`, just for the sake of being future-proof :)

PR #207, opened 2017-10-05
Constant moved to config, multiplier and divisor

ajlopez:
Instead of using "2 *" in REMASC selection rule, use a multiplier and divisor from configuration

### DIFF
--- a/rskj-core/src/main/java/co/rsk/config/RemascConfig.java
+++ b/rskj-core/src/main/java/co/rsk/config/RemascConfig.java
@@ -45,6 +45,14 @@ public class RemascConfig {
 
     private long lateUncleInclusionPunishmentDivisor = 20;
 
+    // Multiplier and Divisor for paid fees comparison in selection rule
+    private long paidFeesMultiplier = 2;
+    private long paidFeesDivisor = 1;
+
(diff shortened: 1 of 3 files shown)


rskj-core/src/main/java/co/rsk/config/RemascConfig.java
```diff
@@ -45,6 +45,14 @@
 
     private long lateUncleInclusionPunishmentDivisor = 20;
 
+    // Multiplier and Divisor for paid fees comparison in selection rule
+    private long paidFeesMultiplier = 2;
+    private long paidFeesDivisor = 1;
+
+    public long getPaidFeesMultiplier() { return paidFeesMultiplier; }
+
+    public long getPaidFeesDivisor() { return paidFeesDivisor; }
```
donequis: Divisor is always 1 ?

rskops:
Can one of the admins verify this patch?

rskj-core/src/main/java/co/rsk/config/RemascConfig.java
ajlopez: Now, yes. It was discussed with Mr. Lerner

PR #210, opened 2017-10-06
Write log for release BTC transaction

diega:


rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -81,6 +82,7 @@
     public static final CallTransaction.Function GET_FEDERATION_ADDRESS = CallTransaction.Function.fromSignature("getFederationAddress", new String[]{}, new String[]{"string"});
     // Returns the minimum amount of satoshis a user should send to the federation.
     public static final CallTransaction.Function GET_MINIMUM_LOCK_TX_VALUE = CallTransaction.Function.fromSignature("getMinimumLockTxValue", new String[]{}, new String[]{"int"});
+    public static final DataWord RELEASE_BTC_TOPIC = new DataWord("release_btc_topic".getBytes());
```
donequis: getBytes() is platform dependant

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -500,7 +486,13 @@ private void processSigning(long executionBlockNumber, BtcECKey federatorPublicK
             logger.info("Tx fully signed {}. Hex: {}", btcTx, Hex.toHexString(btcTx.bitcoinSerialize()));
             removeUsedUTXOs(btcTx);
             provider.getRskTxsWaitingForSignatures().remove(new Sha3Hash(rskTxHash));
-            provider.getRskTxsWaitingForBroadcasting().put(new Sha3Hash(rskTxHash), new ImmutablePair(btcTx, executionBlockNumber));
+            logs.add(
+                new LogInfo(
+                    getFederationAddress().getHash160(),
```
donequis: Shouldn't this be the current contract address? i.e. bridge

rskj-core/src/main/java/co/rsk/peg/Bridge.java
diega: Good catch, thanks!

rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -81,6 +83,7 @@
     public static final CallTransaction.Function GET_FEDERATION_ADDRESS = CallTransaction.Function.fromSignature("getFederationAddress", new String[]{}, new String[]{"string"});
     // Returns the minimum amount of satoshis a user should send to the federation.
     public static final CallTransaction.Function GET_MINIMUM_LOCK_TX_VALUE = CallTransaction.Function.fromSignature("getMinimumLockTxValue", new String[]{}, new String[]{"int"});
+    public static final DataWord RELEASE_BTC_TOPIC = new DataWord("release_btc_topic".getBytes(StandardCharsets.UTF_8));
```
amendelzon: Do we want to use variable-length character encodings such as `UTF-8` for things we store in the blockchain? Or maybe  fixed-length (but obviously more restrictive) encodings such as `US_ASCII`?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -500,7 +490,13 @@ private void processSigning(long executionBlockNumber, BtcECKey federatorPublicK
             logger.info("Tx fully signed {}. Hex: {}", btcTx, Hex.toHexString(btcTx.bitcoinSerialize()));
             removeUsedUTXOs(btcTx);
             provider.getRskTxsWaitingForSignatures().remove(new Sha3Hash(rskTxHash));
-            provider.getRskTxsWaitingForBroadcasting().put(new Sha3Hash(rskTxHash), new ImmutablePair(btcTx, executionBlockNumber));
+            logs.add(
+                new LogInfo(
+                    contractAddress.getBytes(StandardCharsets.UTF_8),
```
amendelzon: Same question as before.

amendelzon: Also, it would be very nice if we could abstract `bytesFromString` (or whatever you want to call it) somewhere so that we don't end up passing in the specific encoding over and over.

rskj-core/src/main/java/co/rsk/peg/BridgeState.java
```diff
@@ -128,11 +117,10 @@ public String toString() {
         byte[] rlpBtcBlockchainBestChainHeight = RLP.encodeBigInteger(BigInteger.valueOf(this.btcBlockchainBestChainHeight));
         byte[] rlpBtcTxHashesAlreadyProcessed = RLP.encodeElement(BridgeSerializationUtils.serializeSet(btcTxHashesAlreadyProcessed));
         byte[] rlpBtcUTXOs = RLP.encodeElement(BridgeSerializationUtils.serializeList(btcUTXOs));
-        byte[] rlpRskTxsWaitingForBroadcasting = RLP.encodeElement(BridgeSerializationUtils.serializePairMap(rskTxsWaitingForBroadcasting));
         byte[] rlpRskTxsWaitingForConfirmations = RLP.encodeElement(BridgeSerializationUtils.serializeMap(rskTxsWaitingForConfirmations));
         byte[] rlpRskTxsWaitingForSignatures = RLP.encodeElement(BridgeSerializationUtils.serializeMap(rskTxsWaitingForSignatures));
 
-        return RLP.encodeList(rlpBtcBlockchainBestChainHeight, rlpBtcTxHashesAlreadyProcessed, rlpBtcUTXOs, rlpRskTxsWaitingForBroadcasting, rlpRskTxsWaitingForConfirmations, rlpRskTxsWaitingForSignatures);
+        return RLP.encodeList(rlpBtcBlockchainBestChainHeight, rlpBtcTxHashesAlreadyProcessed, rlpBtcUTXOs, rlpRskTxsWaitingForConfirmations, rlpRskTxsWaitingForSignatures);
```
amendelzon: I'm guessing this isn't used in the actual production code, otherwise it'd imply a hard fork. Am I right?

rskj-core/src/main/java/co/rsk/peg/Bridge.java
diega: What would be the issue on storing variable-length characters in the blockchain?

amendelzon: We'd have to be a bit more careful with the 32 byte limit for data words.

amendelzon: That said, UTF-8 is broadly used in the node for encoding of strings, but I thought it might be a point to bring into discussion.

donequis: For ascii characters there's no difference

rskj-core/src/main/java/co/rsk/peg/BridgeState.java
donequis: This PR implies a HF

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
donequis: > bytesFromString (or whatever you want to call it) somewhere so that we don't end up passing in the specific encoding over and over.

I agree. There's a TypeConverter class on rpc that we can use and move to a common package.

rskj-core/src/main/java/co/rsk/peg/Bridge.java
amendelzon: I know, that's why I suggest `US_ASCII` as a safe alternative to `UTF-8` in case of typos.

rskops:
Can one of the admins verify this patch?

PR #213, opened 2017-10-11
Fix read config twice

tinchou:
First: sorry for the number of file changes! I changed every usage of `SystemProperties.CONFIG` and  `RskSystemProperties.RSKCONFIG` to `RskSystemProperties.CONFIG`.

The problem was having those two `CONFIG` and `RSKCONFIG` static fields which were always instantiated. `RskSystemProperties` didn't change any base behavior, so the best option was to just use it everywhere.


rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -164,22 +123,58 @@ public SystemProperties(Config apiConfig) {
...
+        logger.info("Config ( {} ): user properties from -Drsk.conf.file file '{}'", cmdLineConfigFile.entrySet().isEmpty() ? NO : YES, file);
+        return javaSystemProperties
+                .withFallback(cmdLineConfigFile)
+                .withFallback(testUserConfig)
+                .withFallback(testConfig)
+                .withFallback(userConfig)
+                .withFallback(userDirConfig)
+                .withFallback(cmdLineConfigRes)
+                .withFallback(referenceConfig);
+    }
```
diega: Maybe is a good time to tweak this a little so we don't create so many `ConfigFactory.empty()`. Something like
```java
Config javaSystemProperties = ConfigFactory.load("no-such-resource-only-system-props");
// ...
String res = System.getProperty("rskj.conf.res");
if (res != null) {
    javaSystemProperties = javaSystemProperties.withFallback(ConfigFactory.parseResources(res));
    logger.info("Config ( YES ): user properties from -Drskj.conf.res resource '{}'", res);
}
// ...
return javaSystemProperties;
```
but this is just a detail

aeidelman: Removing of those ignored tests is fine.

tinchou:
Ship it! :shipit:

PR #225, opened 2017-10-20
Feature/rfs 36 change federation

amendelzon:
The active federation is now stored in the Bridge contract. A new active federation can potentially be saved for each block. There is no "genesis" federation stored in the genesis block (the store will return `null`), but in that case the Bridge pulls the genesis federation from the BridgeConstants (i.e., the genesis federation is precompiled).

This also adds a few methods to the bridge so that the currently active federation can be queried:
- `getFederationSize()`
- `getFederationThreshold()`
- `getFederationCreationTime()`
- `getFederatorPublicKey(index)`

At the moment a running node with this version runs exactly as the previous version, since there is no way to set a new federation. The idea is to build the add/remove functionality on top of this.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
@@ -62,13 +62,13 @@ public class BridgeUtils {
         }
     }
 
-    public static boolean isLockTx(BtcTransaction tx, TransactionBag wallet, BridgeConstants bridgeConstants) {
+    public static boolean isLockTx(BtcTransaction tx, Federation federation, TransactionBag wallet, BridgeConstants bridgeConstants) {
         // First, check tx is not a typical release tx (tx spending from the federation address and
         // optionally sending some change to the federation address)
         int i = 0;
         for (TransactionInput transactionInput : tx.getInputs()) {
             try {
-                transactionInput.getScriptSig().correctlySpends(tx, i, bridgeConstants.getFederationPubScript(), Script.ALL_VERIFY_FLAGS);
+                transactionInput.getScriptSig().correctlySpends(tx, i, federation.getScript(), Script.ALL_VERIFY_FLAGS);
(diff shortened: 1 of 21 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -114,9 +114,13 @@ public static boolean isFreeBridgeTx(org.ethereum.core.Transaction rskTx, long b
...
+        // must be the genesis federation.
+        // Once the original federation changes, txs are always paid.
         return StringUtils.equals(Hex.toHexString(receiveAddress), PrecompiledContracts.BRIDGE_ADDR) &&
-               blockchainConfig.getConfigForBlock(blockNumber).areBridgeTxsFree() &&
-               rskTx.getSignature() != null &&
-               blockchainConfig.getCommonConstants().getBridgeConstants().getFederatorPublicKeys().contains(co.rsk.bitcoinj.core.BtcECKey.fromPublicOnly(rskTx.getKey().getPubKey()));
+                blockchainConfig.getConfigForBlock(blockNumber).areBridgeTxsFree() &&
+                rskTx.getSignature() != null &&
+                blockchainConfig.getCommonConstants().getBridgeConstants().getGenesisFederation()
+                        .getPublicKeys().contains(co.rsk.bitcoinj.core.BtcECKey.fromPublicOnly(rskTx.getKey().getPubKey()));
```
diega: Maybe we can increase readability by extracting some of this return expression variables

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        if (other instanceof Federation) {
+            return this.equalsFederation((Federation) other);
+        }
+
+        return false;
+    }
+
+    public boolean equalsFederation(Federation other) {
```
diega: Why do we need this two public methods instead of just the _equals_ from Object?

PR #228, opened 2017-10-25
Fix config test

adauto82:


rskj-core/src/test/java/org/ethereum/cli/ClIInterfaceTest.java
```diff
@@ -52,8 +53,10 @@ public void callTestWithCorrectInputParams() {
     public void callTestWithRpcCors() {
         String[] inputParams = { "-rpccors", "*.ethereum.io" };
         Map<String, String> result = CLIInterface.call(inputParams);
-
         Assert.assertEquals(result.get(SystemProperties.PROPERTY_RPC_CORS), "*.ethereum.io");
+        String[] restoreParams = { "-rpccors", CorsConfigurationTest.EXPECTED_CORS_CONFIG };
```
tinchou: I don't think this is testing anything. It's the same call with two different parameters. What are you trying to accomplish?

adauto82: No. It is not testing anything new. It's a workaround in case the test executes before CorsConfigurationTest. Because configuration is global. Thats why is it call "restoreParams"

tinchou: Then remove Ethereum's `inputParams` and leave only RSK's

adauto82: Then we might enter the case that CLIInterface is not working, but we don't detect it because we set the same value that it has in the case that CorsConfigurationTest has runned before CliIntefaceTest

tinchou: Great, I understand now. Maybe change the signature to `public void changeAndRestoreRpcCorsConfig` and let's merge this.

adauto82: done

PR #235, opened 2017-11-04
Add the packaging metadata to build the rskj snap

elopio:
This package will let you publish the latest rskj in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and Linux distributions. You just have to go to https://build.snapcraft.io and enable the automated continuous delivery.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,23 @@
+name: rskj
+version: git
+summary: RSKj is a Java implementation of the RSK protocol
+description: |
+  RskJ is a Java implementation of the RSK node. For more information about
+  RSK, visit www.rsk.co
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: devmode # use 'strict' once you have the right plugs and slots
+
+apps:
+  rskj:
+    command: java -cp $SNAP/jar/rskj-core-0.2.5-GINGER-all.jar "$@" co.rsk.Start
+  gennodekeyid:
+    command: java -cp $SNAP/jar/rskj-core-0.2.5-GINGER-all.jar co.rsk.GenNodeKeyId
+
+parts:
+  rskj:
+    source: .
+    plugin: gradle
+    prepare: ./configure.sh
+    gradle-options: [build, shadow, -x, test]
+    gradle-output-dir: rskj-core/build/libs


diega:
Would be possible to enable it just for release tags? Maybe changing some variable during our release+tag process?

elopio:
@diega well, yes, you can organize your release pipeline in any way you want. But the idea with https://build.snapcraft.io is that every change that lands on master is released to the edge channel in the store. The store has four channels: edge, beta, candidate and stable. Only stable is visible for user searches, which means that you can use the other three channels for a combination of manual, automated and crowd testing to help you find bugs early on your way to stable. Once you and your community of early adopters are confident with a release, you move it to the stable channel and all your users will get an automatic update. I wrote a little about how we are using this to [test and release the IPFS snap](http://elopio.net/blog/ipfs-crowdtesting/), and you can see [parity](https://github.com/paritytech/parity#install-from-the-snap-store), for example, testing in edge atm.

If this proposed pipeline doesn't work, you could use [travis ci](https://tutorials.ubuntu.com/tutorial/continuous-snap-delivery-from-travis-ci), [circle ci](https://circleci.com/blog/build-test-publish-snap-packages/) or other system to release just on certain conditions.

Thanks for reviewing! pura vida.

bcodesido:
Thanks for your collaboration.

We don’t think this will work because of how the jar is built. We have a [reproducible build procedure](https://github.com/rsksmart/rskj/wiki/Reproducible-Build) that is not compatible with your solution.
At the same time, implementing this solution requires several changes from our side to make it work.

If you’re interested, we already have our [Ubuntu PPA repository](https://launchpad.net/~rsksmart/+archive/ubuntu/rskj)

elopio:
@bcodesido you could dump the jar that you get out of the reproducible build into the snap, instead of building it again. Or run snapcraft in the buildmachine:latest docker container making sure that it uses the same gradle invocation to get a reproducible jar.

Here at ubuntu we stopped recommending people to use PPAs, because they are insecure. When users add a PPA to their system, they are effectively giving root access to the developer of that PPA, which is a lot more than what they should ever trust anybody because even honest mistakes happen a lot. 

We are using what we learned after years of working with PPAs to design a better system. snaps are first class citizens, you don't have to install extra repositories to install them. And their first goal is security, they are not installed, they are mounted read-only so the developers can't execute maintainer scripts that affect the machine. They are fully confined to the calls that the user allows them to do and to it's own private storage space.

But, if you are happy with your current solution, maybe it's not the time. If you change your mind in the future and want to make rskj available for 1 click/1 command installs, just give me a ping.

pura vida.

bcodesido:
Probably in a near future we are going to look into this. Thanks

PR #254, opened 2017-11-17
Update circle.yml

ida3m0n:
Stop mongod and mysqld from server in order to save memory


aeidelman:
I close this PR since it's duplicated with #255

PR #260, opened 2017-11-21
Migrating Circle CI configuration to version 2.

tinchou:
We now use the official [openjdk](https://hub.docker.com/r/_/openjdk/) Docker image that doesn't come with the bloat of the previous CI configuration (we had `mysql`, `mongo` and other unused services consuming memory).


.circleci/config.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  build:
+    docker:
+      - image: openjdk:8-jdk
+    environment:
+      _JAVA_OPTIONS: "-Xmx3G -Xms2G"
+      SONAR_VERSION: "3.0.3.778"
+    steps:
+      - checkout
+      - run: apt-get update -y && apt-get install -y apt gnupg2
```
bcodesido: Why `apt-get` installs `apt`?

.circleci/config.yml
```diff
@@ -0,0 +1,30 @@
...
+      - run: curl -sSL https://secchannel.rsk.co/release.asc | gpg2 --import -
+      - run: curl -O -J -L https://sonarsource.bintray.com/Distribution/sonar-scanner-cli/sonar-scanner-cli-$SONAR_VERSION-linux.zip
```
bcodesido: `curl -OJL` is more pretty

.circleci/config.yml
```diff
@@ -0,0 +1,30 @@
...
+      - run: unzip sonar-scanner-cli-$SONAR_VERSION-linux.zip
+      - run: gpg2 --verify SHA256SUMS.asc && sha256sum --check SHA256SUMS.asc
```
bcodesido: This should be run before download the `sonar-scanner-cli`

PR #261, opened 2017-11-22
Dupn swapn txindex

ajlopezrsk:
Updated to Bamboo code base


rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -937,10 +950,40 @@ protected void doDUP() {
         program.step();
     }
 
+    protected void doDUPN() {
+        spendOpCodeGas();
+        // EXECUTION PHASE
+        program.step();
+
+        int n = stack.pop().intValueCheck() + 1;
```
SergioDemianLerner: what if this integer addition overflows?
E.g. stack.pop().intValueCheck() == MaxInteger.
Then n becomes negative. verifyStackSize won't catch it. verifyStackOverflow won't catch it. Not good.

SergioDemianLerner: Suggested:  First do Math.addExact(stack.pop().intValueCheck(), 1); and on exception do  throw Program.Exception.tooSmallStack(stackSize, stack.size());

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -937,10 +950,40 @@ protected void doDUP() {
...
+        stack.swap(stack.size() - 1, stack.size() - n);
+        program.step();
+    }
+
+    protected void doSWAPN(){
+        spendOpCodeGas();
+        // EXECUTION PHASE
+        program.step();
+
+        int n = stack.pop().intValueCheck() + 2;
```
SergioDemianLerner: Same as before: first check for overflow in +2

rskj-core/src/main/java/org/ethereum/vm/VM.java
SergioDemianLerner: Also, what happens if stack.pop().intValueCheck() throws an ArithmeticException exception ? Is this correctly catched and OOG is thrown ?

adauto82: From last to first.

if `intValueCheck()` throws an ArithmeticException then is catched on `executeOpcode()` in with this catch:

```catch (RuntimeException e) {
                logger.warn("VM halted: [{}]", e);
                panicProcessor.panic("vm", String.format("VM halted: [%s]", e.getMessage()));
                program.spendAllGas();
                program.resetFutureRefund();
                program.stop();
                throw e;
        } finally {
            if (isLogEnabled) // this must be prevented because it's slow!
                program.fullTrace();
        }```

PR #262, opened 2017-11-22
Remove unused methods

aeidelman:


aeidelman:
@ajlopezrsk that was weird, looking into it.

tinchou:
We're writing a new config (https://github.com/rsksmart/rskj/pull/260) that doesn't download `solc`. You can wait if you want.

PR #293, opened 2017-12-03
Prevent funds migration when there is no balance left

diega:


donequis:
What are the exceptions we are waiting here?

diega:
@donequis the only expected exception here is `co.rsk.bitcoinj.wallet.Wallet.DustySendRequested`, nevertheless I think is good to catch any exception there, b/c otherwise the _retiring federation_ may not be ever cleared, preventing the creation of upcoming federations

PR #300, opened 2017-12-04
added coinbase.secret property

brakmic:
I was following the [tutorial](https://github.com/rsksmart/rskj/wiki/How-to-initialize-RSK-node-configuration-file-settings) but couldn't find the **<COINBASE_SECRET>** in this sample config. So I added it under **peer**-options and according to my local tests it seems to be working. However, still no sure, if this is the proper place for it.


aeidelman:
@brakmic thanks. coinbase.secret is not mandatory anymore, you don't need it to run an RSK node.

brakmic:
Ok 😄 
Many thanks!

PR #324, opened 2017-12-11
Add messageProcessor to logback file

bcodesido:
Set messageProcessor logger to INFO


tinchou:
I think the name should be `messageProcess`.

PR #379, opened 2018-01-03
Fallback mining

SergioDemianLerner:
NOTE:
The two private keys in rskj\rskj-core\src\main\java\org\ethereum\config\Constants.java
    public byte[] privateMiningPubKey0 = Hex.decode("04a043...3b037368d7");
    public byte[] privateMiningPubKey1 = Hex.decode("04774ae7f..4c9c953c61b");

Must be replaced by keys generated with the utility:
rskj\rskj-core\src\test\java\co\rsk\blockchain\utils\GeneratePrivateKey.java

This class has a main() method and when called creates two files:
pubkey.bin
and
privkey.bin

Both files are BINARY. The utility also prints in the console the pubkey in hexa, where it can be copied and the pasted in the Constants file.

Note that "Pubkey: " string printed is not part of the pubkey.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/mine/MinerServer.java
+++ b/rskj-core/src/main/java/co/rsk/mine/MinerServer.java
@@ -27,13 +27,27 @@ public interface MinerServer {
 
     void start();
 
+    void stop();
+
+    boolean isRunning();
+
     SubmitBlockResult submitBitcoinBlock(String blockHashForMergedMining, co.rsk.bitcoinj.core.BtcBlock bitcoinMergedMiningBlock);
+
+    boolean generateFallbackBlock();
+
+    void setFallbackMining(boolean p);
+
+    void setAutoSwitchBetweenNormalAndFallbackMining(boolean p);
+
+    boolean isFallbackMining();
 
     byte[] getCoinbaseAddress();
 
     MinerWork getWork();
 
     void buildBlockToMine(@Nonnull Block newParent, boolean createCompetitiveBlock);
+
+    void setExtraData(byte[] extraData);
 
     long getCurrentTimeInSeconds();
 
--- a/rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
+++ b/rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
@@ -20,6 +20,7 @@ package co.rsk.mine;
 
 import co.rsk.config.MiningConfig;
 import co.rsk.config.RskMiningConstants;
+import co.rsk.config.RskSystemProperties;
 import co.rsk.core.DifficultyCalculator;
 import co.rsk.core.bc.BlockExecutor;
 import co.rsk.core.bc.FamilyUtils;
@@ -34,12 +35,14 @@ import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.ArrayUtils;
 import org.ethereum.core.*;
+import org.ethereum.crypto.ECKey;
 import org.ethereum.crypto.HashUtil;
 import org.ethereum.db.BlockStore;
 import org.ethereum.db.ByteArrayWrapper;
 import org.ethereum.facade.Ethereum;
 import org.ethereum.listener.EthereumListenerAdapter;
 import org.ethereum.rpc.TypeConverter;
+import org.ethereum.util.RLP;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.spongycastle.crypto.digests.SHA256Digest;
@@ -50,7 +53,10 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Component;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
+import java.io.FileInputStream;
+import java.io.IOException;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.*;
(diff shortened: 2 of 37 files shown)


rskj-core/src/main/java/co/rsk/mine/MinerServer.java
```diff
@@ -26,15 +26,28 @@
...
+
+    boolean isRunning();
 
     SubmitBlockResult submitBitcoinBlock(String blockHashForMergedMining, co.rsk.bitcoinj.core.BtcBlock bitcoinMergedMiningBlock);
 
+    boolean generatePrivateBlock();
+
+    void setPrivateMining(boolean p);
+
+    void setAutoSwitchBetweenNormalAndPrivateMining(boolean p);
```
ajlopez: Why to have change of state?

The private miners generates blocks, at any rate.

All the private mined blocks has a fixed assigned difficulty.

When normal mining pools are active, those blocks are rejected by low difficulty

When normal mining pool is missing, the network switch to low difficulty. Then, these private mined blocks will start to win the block fight.

Considerations:

- The private mined blocks override any delta rule (PoW previous block vs PoW current block)
- The private mined blocks skip Proof of work rule

rskj-core/src/main/java/co/rsk/mine/MinerServer.java
```diff
@@ -26,15 +26,28 @@
...
     void start();
+    void stop();
+
+    boolean isRunning();
 
     SubmitBlockResult submitBitcoinBlock(String blockHashForMergedMining, co.rsk.bitcoinj.core.BtcBlock bitcoinMergedMiningBlock);
 
+    boolean generatePrivateBlock();
+
+    void setPrivateMining(boolean p);
```
ajlopezrsk: Why the miner server is aware of private mining?

Now, it is not aware if it is used from a external mining pool, or from MinerClientImpl. Maybe, we could implement a new PrivateMinerClientImpl, that use the new RSK signature.

A future implementation could  be PrivateHSMMinerClientImpl. The implementation to use depends on config (now, the use of MinerClientImpl depends on config)

rskj-core/src/main/java/co/rsk/mine/MinerServer.java
ajlopezrsk: I subscribe this review :-)

rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
```diff
@@ -137,35 +151,211 @@ public MinerServerImpl(Ethereum ethereum,
...
+
+    public int getPrivateBlocksGenerated() {
+     return privateBlocksGenerated;
+    }
+
+    public boolean isPrivateMining() {
+        return privateMining;
+    }
+
+    public void setPrivateMiningState() {
```
ajlopezrsk: Alternative implemention: move all the implementation to a new PrivateMinerClientImpl, so MinerServerImpl is agnostic of mining method. Maybe it could expose an alternative method to submitWork, to receive specially private mined blocks

rskj-core/src/main/java/co/rsk/mine/MinerServer.java
ajlopezrsk: Also, this approach allows the switch off all private miner nodes, and then, the normal CPU miiners could have a change of produce a block. 

The use of private mining is up to our decision, without further changes to the distributed code.

SergioDemianLerner: > Why to have change of state?
Because not all miners will be running the fallback mining system. Only some trusted miners. Then these miners will have auto-switch enabled: they can activate fallback mining when the blockchain difficulty is low.

> The private miners generates blocks, at any rate.
This is how it was implemented.

> All the private mined blocks has a fixed assigned difficulty.
This is not how it's implemented because fallback mined blocks can be intermingled  with PoW mined blocks, the difficulty must stay as real as possible. Fallback mined blocks can increase/decrease the difficulty depending on the block timestamp. Fallback miners will always mine blocks at one every 15 seconds (+1 seconds of the average interval) to keep the difficulty at the minimum, when possible.

> When normal mining pools are active, those blocks are rejected by low difficulty
Not that easy, because we have the rule that after 10 minutes difficulty drops. This was necessary in order to create the fallback system as fast as possible. The ProofOfWorkRule does not know the parent block timestamp, and therefore cannot act on the parent block difficulty. 

> When normal mining pool is missing, the network switch to low difficulty. Then, these private mined blocks will start to win the block fight.
This is how it was implemented.

rskj-core/src/main/java/co/rsk/mine/MinerServer.java
SergioDemianLerner: Currently, to make things as simple as possible, the MinerServer is aware and in fact it automatically generates fallback mined blocks when mining is active and difficulty is low.
I agree that in the future we can move this to the client side (in fact there was a comment of mine about using an HSM, but now the comment was removed by Martin)

SergioDemianLerner: The first implementation (now replaces) was client-side

rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
SergioDemianLerner: Yes, this is something to do in the future to improve the code.

rskj-core/src/main/resources/logback.xml
```diff
@@ -16,7 +16,7 @@
             <pattern>%date{yyyy-MM-dd-HH:mm:ss.SSSS} %p [%c{1}]  %m%n</pattern>
         </encoder>
         <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
-            <level>OFF</level>
+            <level>DEBUG</level>
```
bcodesido: This must stay off

PR #404, opened 2018-01-08
Make the node to stop gracefully on startup error

diega:
Thread initialization is now outside constructors

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/RskImpl.java
+++ b/rskj-core/src/main/java/co/rsk/core/RskImpl.java
@@ -29,12 +29,16 @@ import org.ethereum.manager.WorldManager;
 import org.ethereum.net.server.ChannelManager;
 import org.ethereum.net.server.PeerServer;
 import org.ethereum.vm.program.invoke.ProgramInvokeFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
 
+@Component
 public class RskImpl extends EthereumImpl implements Rsk {
 
     private boolean isplaying;
     private final NodeBlockProcessor nodeBlockProcessor;
 
+    @Autowired
     public RskImpl(WorldManager worldManager,
                    ChannelManager channelManager,
                    PeerServer peerServer,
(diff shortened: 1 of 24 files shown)


tinchou:
:four::zero::four: PR NOT FOUND

rskj-core/src/main/java/co/rsk/core/RskImpl.java
```diff
@@ -29,12 +29,16 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
 
+@Component
 public class RskImpl extends EthereumImpl implements Rsk {
 
     private boolean isplaying;
     private final NodeBlockProcessor nodeBlockProcessor;
 
+    @Autowired
```
lsebrie: Is this autowired necessary?

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -73,6 +73,7 @@
 
     private TxHandler txHandler;
 
+    @Autowired
```
lsebrie: same goes here, why would use @Autowired with @Component on same class?

rskj-core/src/main/java/co/rsk/core/RskImpl.java
diega: Yes, before this changes we needed to perform some initalization on `co.rsk.core.RskImpl` right after the instance was created and all this logic was part of `co.rsk.core.RskFactory`. Now all that complexity is gone and we just need to instantiate this class and inject the proper dependencies which is what accomplished here through the combination of *@Component* and *@Autowired* (in the constructor)

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
diega: Same situation as in the previous comment, before this changes all was created and initialized in the `co.rsk.core.RskFactory`, now we would just need to create a new instance passing all the dependencies in the constructor and just return it, that's exactly what *@Component* and *@Autowired* do

diega:
This doesn't pass the integration tests

PR #406, opened 2018-01-10
RskAddress class

tinchou:
This introduces an `RskAddress` class. Some of the benefits include:
* stop passing around `String`s and `byte[]`s
* ensure all addresses are valid (20 bytes long)
* type safety (e.g. you can't mix up parameters when passing multiple `byte[]`s)
* `equals` and `hashCode` allows using with standard collections


rskj-core/src/main/java/co/rsk/rpc/modules/personal/PersonalModuleWalletEnabled.java
```diff
@@ -63,7 +64,7 @@ public void init(RskSystemProperties properties) {
         // wallets in production for security reasons.
         Account coinbaseAccount = properties.localCoinbaseAccount();
         if (coinbaseAccount != null) {
-            this.wallet.addAccount(coinbaseAccount);
+            this.wallet.addAccount(coinbaseAccount).getBytes();
```
diega: I think this `.getBytes()` call should be removed

rskj-core/src/main/java/co/rsk/validators/BlockTxsValidationRule.java
```diff
@@ -76,21 +76,20 @@ public boolean isValid(Block block, Block parent) {
...
             }
 
-            byte[] txSender = tx.getSender();
-            ByteArrayWrapper key = new ByteArrayWrapper(txSender);
-            BigInteger expectedNonce = curNonce.get(key);
+            RskAddress sender = tx.getSender();
+            BigInteger expectedNonce = curNonce.get(sender);
             if (expectedNonce == null) {
-                expectedNonce = parentRepo.getNonce(txSender);
+                expectedNonce = parentRepo.getNonce(sender.getBytes());
```
diega: Maybe `org.ethereum.core.Repository` could be refactored as part of this work, it has a lot of `byte[]` referring RskAddresses

rskj-core/src/main/java/org/ethereum/core/PendingTransaction.java
```diff
@@ -64,7 +64,7 @@ public long getBlockNumber() {
 
     public byte[] getSender() {
```
diega: I think this method is not used anywhere, in that case the `byte[] sender` field can be removed

rskj-core/src/main/java/org/ethereum/core/Transaction.java
```diff
@@ -615,4 +615,13 @@ public static Transaction create(byte[] nonce, Web3.CallArguments args){
 
         return new Transaction(nonce, hexArgs.getGasPrice(), hexArgs.getGasLimit(), hexArgs.getToAddress(), hexArgs.getValue(), hexArgs.getData());
     }
+
+    public static RskAddress parseRskAddress(@Nullable byte[] bytes) {
```
diega: doesn't make sense to make it part of the `RskAddress` constructor?

rskj-core/src/main/java/org/ethereum/vm/program/invoke/ProgramInvokeFactoryImpl.java
```diff
@@ -51,15 +51,15 @@ public ProgramInvoke createProgramInvoke(Transaction tx, int txindex, Block bloc
...
 
         /***         ORIGIN op       ***/
         // YP: This is the sender of original transaction; it is never a contract.
-        byte[] origin = tx.getSender();
+        byte[] origin = tx.getSender().getBytes();
 
         /***         CALLER op       ***/
         // YP: This is the address of the account that is directly responsible for this execution.
-        byte[] caller = tx.getSender();
+        byte[] caller = tx.getSender().getBytes();
```
diega: For extra points you could refactor the `org.ethereum.vm.program.invoke.ProgramInvokeImpl` constructor as well :)

rskj-core/src/main/java/co/rsk/core/RskAddress.java
```diff
@@ -0,0 +1,101 @@
...
+     * This is the size of an RSK address in bytes.
+     */
+    private static final int LENGTH_IN_BYTES = 20;
+
+    private static final RskAddress NULL_ADDRESS = new RskAddress(new byte[LENGTH_IN_BYTES]);
+
+    /**
+     * This compares using the lexicographical order of the sender unsigned bytes.
+     */
+    public static final Comparator<RskAddress> COMPARATOR = Comparator.comparing(
```
diega: Maybe this class could just implement `Comparable`

tinchou: We discussed this offline and didn't agree on this. My view is that addresses don't have a natural order (this implies implementing `Comparable`), but just an order for a specific feature (e.g. serializing always in the same order).
I think it's a minor issue, maybe we should leave this out for now.

rskj-core/src/main/java/co/rsk/rpc/modules/personal/PersonalModuleWalletEnabled.java
tinchou: :+1:

rskj-core/src/main/java/co/rsk/validators/BlockTxsValidationRule.java
tinchou: I noticed that too, but the number of changes was getting out of control. Just created #410 to track that.

rskj-core/src/main/java/org/ethereum/core/PendingTransaction.java
tinchou: Just removed the whole class!

rskj-core/src/main/java/org/ethereum/core/Transaction.java
tinchou: I thought about it, but I want to make sure that anyone using this constructor is very conscious about what data they're passing. This is just a wrapper around **a valid address**. The `Transaction` class has the responsibility of sanitizing the input here, with the business rules it has defined.

rskj-core/src/main/java/org/ethereum/vm/program/invoke/ProgramInvokeFactoryImpl.java
tinchou: :sos:

tinchou:
@martinmedina created the constructor as suggested and created https://github.com/rsksmart/rskj/issues/411

PR #434, opened 2018-01-23
Remove BlockGenerator singleton usage

diega:


diega:
@tinchou I already removed most of them, thanks!

PR #460, opened 2018-02-02
Remove JUnit from compile dependencies

tinchou:
It should only be a `testCompile` dependency.

### DIFF
--- a/rskj-core/build.gradle
+++ b/rskj-core/build.gradle
@@ -124,7 +124,6 @@ run {
 }
 
 dependencies {
-    compile "junit:junit:${junitVersion}"
     compile "io.netty:netty-all:4.0.30.Final"
     compile "com.madgag.spongycastle:core:${scastleVersion}" // for SHA3 and SECP256K1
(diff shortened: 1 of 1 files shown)


rskj-core/build.gradle
```diff
@@ -201,7 +199,6 @@ dependencyVerification {
         'org.ethereum:leveldbjni-all:18da00444c77080d4422b16c9d4750c4addabda350b702b4a6d628b86658e585',
         'org.fusesource.hawtjni:hawtjni-runtime:74fe9764e1fb1ef20b159dbca2d29abd6de292082ce3fcf538f81ac912390416',
         'org.fusesource.leveldbjni:leveldbjni:05fe3a006d030aaf8d1e43f6c640a85f9f6b967c4499ce1ad5055ac236c3b944',
-        'org.hamcrest:hamcrest-core:66fdef91e9739348df7a096aa384a5685f4e875584cce89386a7a47251c4d8e9',
         'org.hamcrest:hamcrest-library:711d64522f9ec410983bd310934296da134be4254a125080a0416ec178dfad1c',
```
diega: I would say this should also be removed, hamcrest is a library for being used along with junit

PR #470, opened 2018-02-08
Contributing guidelines

tinchou:
As this very document says, guidelines are open for suggestions.

This might be a good starting point for something we can all agree on.


diega:
What about a rule for only allowing a single constructor setting field values and make all others to call each others through `this()`?

tinchou:
@diega I think it's a great idea! I can't find any Sonar rule to enforce it, but I would definitely add it as a style suggestion.

Let's see what the rest has to say about it.

aeidelman:
This is a good starting point to have rskj Contribution Guidelines, there is a lot to add, but there is no reason not to make this available and work from this version.

PR #487, opened 2018-02-20
Add a smoke test for the basic object graph

tinchou:
This catches some issues that were previously only detected when spinning up a node.


tinchou:
CI can't handle the DB creation operation. Closing for now and adding to the backlog.

PR #512, opened 2018-03-19
Add support for new configuration keys with default value

diega:
A new semantic has been added to the migration file. Now you can put an non existing key in the origin and it will be added to the destination with the value as its default value. For example
```properties
rpc.host = ["localhost"]
```


diega:
This was a flawed implementation

PR #543, opened 2018-04-25
[RFS-156] Improve logging

diega:


rskj-core/src/main/java/co/rsk/net/handler/TxValidator.java
```diff
@@ -90,7 +82,7 @@ public TxValidator(RskSystemProperties config, Repository repository, Blockchain
 
             for (TxValidatorStep step : validatorSteps) {
                 if (!step.validate(tx, state, blockGasLimit, minimumGasPrice, bestBlockNumber, basicTxCost == 0)) {
-                    logger.info("Tx validation failed: validator {} tx={}", step.getClass().getName(), tx.getHash());
+                    logger.debug("[tx={}] {} failed", tx.getHash(), step.getClass());
```
pmprete: i think validation fail is a bit more important than debug, i would go for info or warning

amendelzon: Agreed. I think `info` would be more suitable given that we probably don't want to give a warning on a triggerable condition, although I'm not sure how we're currently treating warnings, so that counts too. Makes sense?

rskj-core/src/main/java/co/rsk/net/handler/TxValidator.java
```diff
@@ -50,7 +47,6 @@
     private final Blockchain blockchain;
 
     private final List<TxValidatorStep> validatorSteps = new LinkedList<>();
-    private final List<TxFilter> txFilters = new LinkedList<>();
```
amendelzon: Was this unused? Do we know why/if it should actually have been used?

diega: This was unused for sure (it was reported by Sonar). About how this should have been used, maybe @donequis could tell us more

tinchou: It's fine. We're tracking this issue and fixing it as part of this sprint.

PR #562, opened 2018-05-18
Jacoco plugin

diega:
This puts in the `rskj-core/build/reports/jacoco/test/html/` directory the HTML output of the coverage report. Let me know if Jenkins needs the output in a different format (available options: xml, csv)


bcodesido:
We need the report in an exec file

bcodesido:
LGTM

PR #569, opened 2018-05-22
Add --devnet parameter

tinchou:
This also makes it work again when you specify a configuration file with `blockchain.config.name = "devnet"`


rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
```diff
@@ -109,8 +110,16 @@ private Config getNetworkDefaultConfig(Config userConfig) {
...
+                return ConfigFactory.load(DEVNET_RESOURCE_PATH);
             } else {
-                logger.warn("Invalid network '{}', using mainnet by default", network);
+                String exceptionMessage = String.format(
+                        "%s is not a valid network name (%s property)",
+                        network,
+                        SystemProperties.PROPERTY_BC_CONFIG_NAME
+                );
+                logger.warn(exceptionMessage);
+                throw new IllegalArgumentException(exceptionMessage);
```
diega: If the `blockchain.config.name` value is **mainnet** it'll fail, please modify this

PR #575, opened 2018-05-28
Add network id to peer discovery

joaquin-fdt:
Add network id to the peer discovery network.


rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
```diff
@@ -109,6 +110,12 @@ public void setUDPChannel(UDPChannel udpChannel) {
 
     public void handleMessage(DiscoveryEvent event) {
         DiscoveryMessageType type = event.getMessage().getMessageType();
+        //If this is not from my network ignore it. But if the messages do not
+        //have a networkId in the message yet, then just let them through, for now.
+        if (event.getMessage().getNetworkId().isPresent() &&
+                event.getMessage().getNetworkId().getAsInt() != this.networkId.getAsInt()) {
+            return;
```
diega: Maybe a `DEBUG` or `WARN` log here?

PR #582, opened 2018-06-01
Change bind_address to 0.0.0.0

tinchou:


rskj-core/src/main/resources/reference.conf
```diff
@@ -136,7 +136,7 @@ rpc {
             cors = "localhost"
             http {
                 enabled = true
-                bind_address = localhost
+                bind_address = 0.0.0.0
```
juli: JSON RPC must not be exposed to the network by default.

rskj-core/src/main/resources/reference.conf
```diff
@@ -146,7 +146,7 @@ rpc {
             }
             ws {
                 enabled = false
-                bind_address = localhost
+                bind_address = 0.0.0.0
```
juli: JSON RPC must not be exposed to the network by default.

PR #592, opened 2018-06-13
Update explorer clean up period

lsebrie:
Update period and clean up period are separated

### DIFF
--- a/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
+++ b/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
@@ -52,7 +52,13 @@ import java.util.List;
 public class RskSystemProperties extends SystemProperties {
     private static final Logger logger = LoggerFactory.getLogger("config");
 
-    public static final int PD_DEFAULT_REFRESH_PERIOD = 60000;
+    /** while timeout period is lower than clean period it doesn't affect much since
+    requests will be checked after a clean period.
+     **/
+    private static final int PD_DEFAULT_CLEAN_PERIOD = 15000; //miliseconds
+    private static final int PD_DEFAULT_TIMEOUT_MESSAGE = PD_DEFAULT_CLEAN_PERIOD - 1; //miliseconds
+    private static final int PD_DEFAULT_REFRESH_PERIOD = 60000; //miliseconds
+
     public static final int BLOCKS_FOR_PEERS_DEFAULT = 100;
     private static final String MINER_REWARD_ADDRESS_CONFIG = "miner.reward.address";
     private static final String MINER_COINBASE_SECRET_CONFIG = "miner.coinbase.secret";
@@ -261,7 +267,7 @@ public class RskSystemProperties extends SystemProperties {
 
     public long peerDiscoveryMessageTimeOut() {
         return configFromFiles.hasPath("peer.discovery.msg.timeout") ?
-                configFromFiles.getLong("peer.discovery.msg.timeout") : 30000;
+                configFromFiles.getLong("peer.discovery.msg.timeout") : PD_DEFAULT_TIMEOUT_MESSAGE;
     }
 
     public long peerDiscoveryRefreshPeriod() {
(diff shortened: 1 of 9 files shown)


rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
```diff
@@ -259,20 +261,29 @@ public NeighborsPeerMessage sendNeighbors(InetSocketAddress nodeAddress, List<No
...
+        removeExpiredChallenges(oldPingRequests);
+        resendExpiredPing(oldPingRequests);
+        removeConnections(oldPingRequests.stream().
+                filter(r -> r.getAttemptNumber() >= RETRIES_COUNT).collect(Collectors.toList()));
 
         removeExpiredRequests(this.pendingFindNodeRequests);
     }
 
+    @VisibleForTesting
     public void cleanAndUpdate() {
```
juli: `public` but only used by tests now.

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorerCleaner.java
```diff
@@ -30,11 +30,13 @@
...
+    private long cleanPeriod;
     private boolean running = false;
 
-    public PeerExplorerCleaner(PeerExplorer peerExplorer, long updatePeriod) {
+    public PeerExplorerCleaner(PeerExplorer peerExplorer, long updatePeriod, long cleanPeriod) {
         this.peerExplorer = peerExplorer;
         this.updatePeriod = updatePeriod;
+        this.cleanPeriod = cleanPeriod;
 
         this.updateTask = Executors.newSingleThreadScheduledExecutor(r -> new Thread(r, "PeerExplorerCleaner"));
```
juli: I suggest adding a comment here that single thread is important because the rest of the code assumes a single thread.

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
```diff
@@ -261,7 +267,7 @@ public void enableRemasc() {
 
     public long peerDiscoveryMessageTimeOut() {
         return configFromFiles.hasPath("peer.discovery.msg.timeout") ?
```
tinchou: Please take this chance to move the default to `reference.conf` and remove this `hasPath` check.

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
```diff
@@ -356,4 +371,9 @@ private void loadInitialBootNodes(List<String> nodes) {
 
         return ret;
     }
+
+    @VisibleForTesting
+    public NodeChallengeManager getChallengeManager() {
```
tinchou: Injecting this would be cleaner. You should tackle it if it doesn't interfere with your other tasks.

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
diega: Now this is only used in the `co.rsk.net.discovery.UDPServerTest#run3NodesFullTest`, I would remove it and replace it with the explicit call for both methods (it's only 6 usages)

lsebrie: Removed :+1:

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorerCleaner.java
lsebrie: :+1:

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
lsebrie: I will address this on another PR

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
lsebrie: I will address this on another PR

PR #621, opened 2018-07-16
patch-2: Add tests for Keccak256 compare

rrcossio:


bcodesido:
Is using env variables for testing that they are exploited by modifications to the circle script.

We are not going to merge this. Thanks for the collaboration.

PR #668, opened 2018-10-17
Added timestamp to evm_mine

ignaciopulicedonatto:
evm_mine now takes one optional parameter, it timestamp will be the timestamp block to be setup as the mining time.


tinchou:
Closing because we figured Open Zeppelin tests don't need it.

Feel free to open a new issue if anyone need this feature.

PR #671, opened 2018-10-19
Restore difficulty drop testnet

martinmedina:
Big sudden changes in hashing power occur and difficulty adjustment algorithm does not react fast enough to avoid the network for stopping mining activity for minutes or even hours (depending on the hashing power drop)
Furthermore, chances that this happens in testnet are way bigger than in mainnet because hashing power is more unstable in the former network.
To prevent the pictured situation from happening, difficulty drop is a feature added in this PR that sets difficulty to its minimum value after 10 minutes without new blocks.


rskj-core/src/main/java/org/ethereum/config/blockchain/testnet/TestNetDifficultyDropEnabledConfig.java
```diff
@@ -0,0 +1,36 @@
...
+
+import co.rsk.core.BlockDifficulty;
+import org.ethereum.core.BlockHeader;
+
+public class TestNetDifficultyDropEnabledConfig extends TestNetAfterBridgeSyncConfig {
+
+    @Override
+    public BlockDifficulty calcDifficulty(BlockHeader currentBlockHeader, BlockHeader parentBlockHeader) {
+        // If more than 10 minutes, reset to minimum difficulty
+        if (currentBlockHeader.getTimestamp() >= parentBlockHeader.getTimestamp() + 600) {
```
ignaciopulicedonatto: Could you move 600 to a variable. Is it in seconds right ?

rskj-core/src/main/java/org/ethereum/config/net/TestNetConfig.java
```diff
@@ -30,6 +31,7 @@ public TestNetConfig() {
         add(0, new TestNetBeforeBridgeSyncConfig());
         // 21 days of 1 block every 14 seconds.
         // On blockchain launch blocks will be faster until difficulty is adjusted to available hashing power.
-        add(129600, new TestNetAfterBridgeSyncConfig());
+        add(129_600, new TestNetAfterBridgeSyncConfig());
+        add(501_726, new TestNetDifficultyDropEnabledConfig());
```
ignaciopulicedonatto: Why you added _ ?

ignaciopulicedonatto:
Could you add some test ?

rskj-core/src/main/java/org/ethereum/config/net/TestNetConfig.java
martinmedina: Because it improves the readability of the code

rskj-core/src/main/java/org/ethereum/config/blockchain/testnet/TestNetDifficultyDropEnabledConfig.java
martinmedina: Done!

PR #675, opened 2018-10-23
Optimized Performance

tehsunnliu:
The get(Sha256Hash hash) functions try to get storedBlock from the cache first. If storedBlock is not found then it computes the storedBlock from byteArrayToStoredblock(ba) which is slower.  The current function processes the BridgeSync Transaction much faster.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/RepositoryBlockStore.java
+++ b/rskj-core/src/main/java/co/rsk/peg/RepositoryBlockStore.java
@@ -77,30 +77,20 @@ public class RepositoryBlockStore implements BtcBlockstoreWithCache {
 
     @Override
     public synchronized StoredBlock get(Sha256Hash hash) throws BlockStoreException {
-        byte[] ba = repository.getStorageBytes(contractAddress, new DataWord(hash.toString()));
+        return getFromCache(hash);
+    }
 
+    public synchronized StoredBlock getFromCache(Sha256Hash hash) throws BlockStoreException {
+        byte[] ba = repository.getStorageBytes(contractAddress, new DataWord(hash.toString()));
         if (ba==null) {
             return null;
         }
         
-        StoredBlock storedBlock = byteArrayToStoredBlock(ba);
-        knownBlocks.put(hash, storedBlock);
-        return storedBlock;
-    }
-
-    public synchronized StoredBlock getFromCache(Sha256Hash hash) throws BlockStoreException {
         StoredBlock storedBlock = knownBlocks.get(hash);
-
         if (storedBlock != null) {
             return storedBlock;
         }
-
-        byte[] ba = repository.getStorageBytes(contractAddress, new DataWord(hash.toString()));
-
-        if (ba==null) {
-            return null;
-        }
-
+        
         storedBlock = byteArrayToStoredBlock(ba);
 
         knownBlocks.put(hash, storedBlock);


tinchou:
Can you support your claim that this is faster? What kind of tests did you do?

Thanks for contributing!

tehsunnliu:
I conducted the following Test 
1st with `hashTwice` algorith
2nd with `CryptoNight` algorith 

The above two algorithms are executed whenever `byteArrayToStoredBlock` function is called. In the first test results we can see little difference in performance. However, to further maximize the difference I changed the hashing algorithm to CryptoNight which is a CPU and Memory intensive algorithm. 
In the following test case table "Before" is the unchanged `get(Sha256Hash hash)` function and "After" is the `get(Sha256Hash hash)` function calling `getFromCache(Sha256Hash hash)` function. For both the test cases I ran the test three times. And the "Height" is the height where the BridgeSync Transaction is mined. 

Also in `hashTwice` algorith there are 500 Block headers and in `CryptoNight` algorith there are 100 Block headers

 ## hashTwice
**Test1**  

| Height | Before (nano) | After (nano) |
| ---: | ---: | ---: |
| 457 | 240483533 | 338324043 |
| 475 | 120512208 | 213888641 |
| 493 | 124291081 | 106676819 |
| 511 | 105201041 | 133850102 |
| 579 | 117867313 | 196447802 |
| 615 | 71908117 | 67147849 |


**Test2**  

| Height | Before (nano) | After (nano) |
| ---: | ---: | ---: |
| 457 | 261373416 | 167337280 |
| 475 | 157188824 | 117226227 |
| 493 | 148522151 | 100251942 |
| 511 | 155171354 | 107776905 |
| 579 | 190953423 | 98200690 |
| 615 | 84434429 | 57745078 |



**Test3** 

| Height | Before (nano) | After (nano) |
| ---: | ---: | ---: |
| 457 | 203510850 | 169159836 |
| 475 | 158782755 | 119540352 |
| 493 | 158578562 | 113244978 |
| 511 | 200858673 | 97642660 |
| 579 | 311053122 | 111437890 |
| 615 | 55776857 | 71358079 |

## CryptoNight

**Test1**

Height | Before (nano) | After (nano)
| ---: | ---: | ---: |
| 196 | 70331433244 | 2161022835 |
| 224 | 83657903720 | 2035568618 |
| 250 | 85484197288 | 2049362893 |
| 279 | 87729059909 | 2205829078 |
| 302 | 86900383926 | 2125755995 |
| 327 | 92108301133 | 2018344526 |

**Test2**

| Height | Before (nano) | After (nano) |
| ---: | ---: | ---: |
| 196 | 77061621957 | 2303699148 |
| 224 | 86647183301 | 2074486931 |
| 250 | 82803222770 | 2077104002 |
| 279 | 86801668969 | 2167518483 |
| 302 | 87730020854 | 2170018876 |
| 327 | 89968488962 | 2062491179 |

**Test3**

| Height | Before (nano) | After (nano) |
| ---: | ---: | ---: |
| 196 | 67254580516 | 2208397927 |
| 224 | 80335990165 | 2059704012 |
| 250 | 82951364260 | 2499840561 |
| 279 | 83758036080 | 2634639973 |
| 302 | 86855085076 | 2306679739 |
| 327 | 89257737330 | 2086025130 |



**NOTE** I've taken `CryptoNight` algorith just for testing, so that we can see the difference. But if bitcoins `hashTwice` is used we cannot see much of a difference. But the difference is still noticable from the above tests.

Hope this helps. 
Thank you.

tehsunnliu:
I didn't remove the `knownBlocks.get(hash)` it is still there. The functions look like this

**New**
``` java
@Override
public synchronized StoredBlock get(Sha256Hash hash) throws BlockStoreException {
	return getFromCache(hash);
}

public synchronized StoredBlock getFromCache(Sha256Hash hash) throws BlockStoreException {
	byte[] ba = repository.getStorageBytes(contractAddress, new DataWord(hash.toString()));
	if (ba==null) {
		return null;
	}
	
	StoredBlock storedBlock = knownBlocks.get(hash);
	if (storedBlock != null) {
		return storedBlock;
	}
	
	storedBlock = byteArrayToStoredBlock(ba);

	knownBlocks.put(hash, storedBlock);

	return storedBlock;
}
```

Unlike before, this function just returns the `storedBlock` if it is available in the cache and does not compute `byteArrayToStoredBlock`.

**Old**
``` java
@Override
    public synchronized StoredBlock get(Sha256Hash hash) throws BlockStoreException {
        byte[] ba = repository.getStorageBytes(contractAddress, new DataWord(hash.toString()));

        if (ba==null) {
            return null;
        }
        
        StoredBlock storedBlock = byteArrayToStoredBlock(ba);
        knownBlocks.put(hash, storedBlock);
        return storedBlock;
    }

    public synchronized StoredBlock getFromCache(Sha256Hash hash) throws BlockStoreException {
        StoredBlock storedBlock = knownBlocks.get(hash);

        if (storedBlock != null) {
            return storedBlock;
        }

        byte[] ba = repository.getStorageBytes(contractAddress, new DataWord(hash.toString()));

        if (ba==null) {
            return null;
        }

        storedBlock = byteArrayToStoredBlock(ba);

        knownBlocks.put(hash, storedBlock);

        return storedBlock;
    }
```

Here, in the old code even if the `storedBlock` is available in the cache and if we just call `get` it does not return the cached `storedBlock` instead it computes the `storedBlock` again and then returns the `storedBlock`. And `knownBlocks.put(hash, storedBlock)`  in `put` function rewrites the storedBlock in the cache even though if it is available in the cache. 
Updated code puts the `storedBlock` in cache only once and reads it from the cache whenever required.

tinchou:
I see now, I was confused by the diff. Sorry about that.

I searched for usages of the `get` method and it is only used for testing in our codebase. We are using `getFromCache` in production, so the cache shouldn't be misused. The code should be refactored to make this clearer, but that's another story.

image

So now I wonder, does changing `getFromCache` to call `repository.getStorageBytes` before `knownBlocks.get` improve the performance in any way?

tehsunnliu:
Actually calling `repository.getStoredBytes` before `knownBlocks.get` doesn't improve any performance. That is called before to `return null` if `ba==null` for the normal working of Bridge Sync. If we place it like the old code the bridge sync doesn't sync the Btc Headers. 

However, the  `get` method is called from external library bitcoinj-thin from `VersionTally.java` `initialize` function whenever bitcoin block headers are sent to Rsk. Furthermore `initialize` function calls `getPrev` function 1000 times for each bitcoin block headers for mainnet and 100 times for testnet. The `getPrev` function again calls get function Overridden at `RepositoryBlockStore`.  

VersionTally.java
image

StoredBlock.java
b8d476021b700c8080c0a3396170035


image

Hope this helps. 
Thank you.

PR #684, opened 2018-11-01
Fix to allow mining reward address setting to be used in regtest

martinmedina:
Config files are additive, that means that if more settings (not present in default config file) are needed to run a certain configuration those settings will override/add to the ones of the base config. For example, regtest.conf is the base config file while my_regtest.conf can be added with some custom settings.
The previous generates a problem because settings in regtest.conf can be overrided/added but not removed. For using `mining.reward.address` the setting `mining.coinbase.secret` should not exist (code validations check that). Since `mining.coinbase.secret` is set on regtest.conf there is no way to use `mining.reward.address` because both settings exist and validations fail.
This PR contemplates that special case in the settings validations so either `mining.coinbase.secret` or `mining.reward.address` can be used in regtest.
Another option that was considered was to remove `mining.coinbase.secret` from regtest.conf but that would have force users to create their own config and set one of the two variables. I believe it goes against usability and against the current use case (user runs the node with `--regtest` and is good to go)

### DIFF
--- a/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
+++ b/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
@@ -60,6 +60,8 @@ public class RskSystemProperties extends SystemProperties {
     private static final int PD_DEFAULT_TIMEOUT_MESSAGE = PD_DEFAULT_CLEAN_PERIOD - 1; //miliseconds
     private static final int PD_DEFAULT_REFRESH_PERIOD = 60000; //miliseconds
 
+    private static final String REGTEST_BLOCKCHAIN_CONFIG = "regtest";
+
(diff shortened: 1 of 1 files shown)


rskops:
SonarQube analysis reported 1 issue

1. ![INFO][INFO] [RskSystemProperties.java#L110](https://github.com/rsksmart/rskj/blob/e278d1781d0dcbdefb09542f32fd2352092b5c41/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java#L110): Method co.rsk.config.RskSystemProperties.localCoinbaseAccount() makes literal string comparisons passing the literal as an argument [rule](https://jenkins.rsklabs.io/sonar/coding_rules#rule_key=fb-contrib%3ALSC_LITERAL_STRING_COMPARISON)

[INFO]: https://sonarsource.github.io/sonar-github/severity-info.png 'Severity: INFO'

PR #690, opened 2018-11-06
Include a info logger with RSK Address.

pamgonzalez:
Add a logger for the RFS233.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -478,6 +478,9 @@ public class BridgeSupport {
                         sender,
                         co.rsk.core.Coin.fromBitcoin(totalAmount)
                 );
+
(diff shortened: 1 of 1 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -478,6 +478,9 @@ public void registerBtcTransaction(Transaction rskTx, byte[] btcTxSerialized, in
                         sender,
                         co.rsk.core.Coin.fromBitcoin(totalAmount)
                 );
+
+                logger.info("Tranfering from BTC Address {}. RSK Address: {}.", senderBtcAddress, sender);
```
josedahlquist: there is a typo here: 
```suggestion
                logger.info("Transfering from BTC Address {}. RSK Address: {}.", senderBtcAddress, sender);
```

amendelzon: I think it's "Transferring"

PR #697, opened 2018-11-14
Wrong condition in setLockWhitelistDisableBlockDelay

yeousunn:
The condition `bestChainHeight < 0` will always give false as `bestChainHeight` is taken from bitcoin blockchain best chain height.
It can never be less than 0 as genesis is always 0. The condition should check if `disableBlockDelay < 0`.


yeousunn:
okey, done rebase. @pmprete

thank you.

pmprete:
This changes where merged in #723 thanks @yeousunn  for your colaboration

PR #702, opened 2018-11-19
StringBuilder to concatenate

yeousunn:
Using StringBuilder to concatenate string instead of + operator.


tinchou:
This class is unused and will be deleted in #709. Thanks for spotting this anyway!

https://github.com/rsksmart/rskj/pull/709/files#diff-66aceffb8220a44879ecd280246bb118

PR #715, opened 2018-12-04
Allow node to start without build-info.properties information

diega:
build -> *fed:fix_build_info*

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/RskFactory.java
+++ b/rskj-core/src/main/java/co/rsk/core/RskFactory.java
@@ -67,6 +67,7 @@ import org.ethereum.net.server.PeerServerImpl;
 import org.ethereum.rpc.Web3;
 import org.ethereum.solidity.compiler.SolidityCompiler;
 import org.ethereum.sync.SyncPool;
+import org.ethereum.util.BuildInfo;
 import org.ethereum.vm.program.invoke.ProgramInvokeFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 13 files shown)


rskj-core/src/main/java/co/rsk/core/RskFactory.java
```diff
@@ -415,7 +415,13 @@ public TransactionGateway getTransactionGateway(
     public BuildInfo getBuildInfo(ResourceLoader resourceLoader) throws IOException {
         Properties props = new Properties();
         Resource buldInfoFile = resourceLoader.getResource("classpath:build-info.properties");
-        props.load(buldInfoFile.getInputStream());
+        try {
```
tinchou: This is the 🥔, great work with the previous refactoring. Please remove the now unnecessary exception!

rskj-core/src/main/java/co/rsk/core/RskFactory.java
```diff
@@ -415,7 +415,13 @@ public TransactionGateway getTransactionGateway(
...
+            props.load(buldInfoFile.getInputStream());
+        } catch (IOException ioe) {
+            logger.warn("build-info.properties file missing from classpath");
```
tinchou: Nice logging pattern.

rskj-core/src/main/java/org/ethereum/util/BuildInfo.java
```diff
@@ -22,36 +22,21 @@
...
-            buildBranch = props.getProperty("build.branch");
-        } catch (IOException e) {
-            logger.error("Error reading /build-info.properties", e);
-        }
+    public String getBuildHash() {
+        return buildHash;
     }
 
-    public static void printInfo(){
+    public void printInfo(Logger logger){
```
tinchou: Could this be a private logger?

diega: I found this method so superfluous that I wanted to give it _some_ kind of meaningfulness. This way you can log in the context of the caller and thus avoid handling an additional Logger in your configuration

rskj-core/src/main/java/co/rsk/core/RskFactory.java
diega: Honestly I'm not quite sure this is the better way. I think you can attach different pattern layouts through Logback configuration according the [top] level enabled but I didn't investigate further. We have a deep debt on how we are handling logging, we can add this as an improvement path when we face it

PR #723, opened 2018-12-10
Remove unnecessary  check for negative bestChainHeight

pmprete:
This was detected by @yeousunn at #697, thanks for your colaboration.
Also add tests for setLockWhitelistDisableBlockDelay and some refactoring.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -1822,14 +1822,13 @@ public class BridgeSupport {
         if (!isLockWhitelistChangeAuthorized(tx)) {
             return LOCK_WHITELIST_GENERIC_ERROR_CODE;
         }
-
         LockWhitelist lockWhitelist = provider.getLockWhitelist();
         if (lockWhitelist.isDisableBlockSet()) {
             return -1;
         }
         int disableBlockDelay = disableBlockDelayBI.intValueExact();
         int bestChainHeight = getBtcBlockchainBestChainHeight();
-        if (bestChainHeight < 0 || Integer.MAX_VALUE - disableBlockDelay < bestChainHeight) {
+        if (disableBlockDelay + bestChainHeight <= bestChainHeight) {
             return -2;
         }
         lockWhitelist.setDisableBlockHeight(bestChainHeight + disableBlockDelay);
--- a/rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
@@ -2926,6 +2926,158 @@ public class BridgeSupportTest {
     }
 
     @Test
+    public void setLockWhitelistDisableBlockDelay_ok() throws IOException, BlockStoreException {
+        Transaction mockedTx = mock(Transaction.class);
+        byte[] senderBytes = ECKey.fromPublicOnly(Hex.decode(
+            // Public key hex of the authorized whitelist admin in regtest, taken from BridgeRegTestConstants
+            "04641fb250d7ca7a1cb4f530588e978013038ec4294d084d248869dd54d98873e45c61d00ceeaeeb9e35eab19fa5fbd8f07cb8a5f0ddba26b4d4b18349c09199ad"
+        )).getAddress();
+        RskAddress sender = new RskAddress(senderBytes);
+        when(mockedTx.getSender()).thenReturn(sender);
+        LockWhitelist mockedWhitelist = mock(LockWhitelist.class);
+        when(mockedWhitelist.isDisableBlockSet()).thenReturn(false);
+        int bestChainHeight = 10;
+        BtcBlockstoreWithCache btcBlockStore = mock(BtcBlockstoreWithCache.class);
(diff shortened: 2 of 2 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1822,17 +1822,21 @@ public Integer setLockWhitelistDisableBlockDelay(Transaction tx, BigInteger disa
...
         if (lockWhitelist.isDisableBlockSet()) {
             return -1;
         }
-        int disableBlockDelay = disableBlockDelayBI.intValueExact();
-        int bestChainHeight = getBtcBlockchainBestChainHeight();
-        if (bestChainHeight < 0 || Integer.MAX_VALUE - disableBlockDelay < bestChainHeight) {
+        try {
+            int disableBlockDelay = disableBlockDelayBI.intValueExact();
+            int bestChainHeight = getBtcBlockchainBestChainHeight();
+            if (bestChainHeight < 0 || disableBlockDelay < 0 ) {
```
amendelzon: Why would `bestChainHeight` be negative?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -2925,6 +2925,182 @@ public void addLockWhitelistAddress_invalidAddress() throws IOException {
...
+        BridgeSupport bridgeSupport = getBridgeSupportWithMocksAndBtcBlockstoreForWhitelistTests(mockedWhitelist,btcBlockStore);
+
+        BigInteger disableBlockDelayBI = BigInteger.valueOf(Integer.MAX_VALUE / 2);
+
+        Assert.assertEquals(-2, bridgeSupport.setLockWhitelistDisableBlockDelay(mockedTx, disableBlockDelayBI).intValue());
+        verify(mockedWhitelist, never()).put(any(), any());
+    }
+
+    @Test
+    public void setLockWhitelistDisableBlockDelay_maxValueDisableBlockBI() throws IOException, BlockStoreException {
```
amendelzon: This is a duplicated test case I think.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -2925,6 +2925,182 @@ public void addLockWhitelistAddress_invalidAddress() throws IOException {
...
+        BridgeSupport bridgeSupport = getBridgeSupportWithMocksAndBtcBlockstoreForWhitelistTests(mockedWhitelist,btcBlockStore);
+
+        BigInteger disableBlockDelayBI = BigInteger.valueOf(Integer.MAX_VALUE);
+
+        Assert.assertEquals(-2, bridgeSupport.setLockWhitelistDisableBlockDelay(mockedTx, disableBlockDelayBI).intValue());
+        verify(mockedWhitelist, never()).put(any(), any());
+    }
+
+    @Test
+    public void setLockWhitelistDisableBlockDelay_negativeBestChain() throws IOException, BlockStoreException {
```
amendelzon: I don't think we need to cover this case.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
pmprete: one is the max int value, the other one is the max long value.
The max long test the exception given when converting from long number, the max int value test the edge case

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
pmprete: my bad, i thougth that #697 was merged already

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
pmprete: good catch

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -2925,6 +2925,158 @@ public void addLockWhitelistAddress_invalidAddress() throws IOException {
...
+        LockWhitelist mockedWhitelist = mock(LockWhitelist.class);
+        when(mockedWhitelist.isDisableBlockSet()).thenReturn(false);
+        int bestChainHeight = (Integer.MAX_VALUE / 2) + 2;
+        BtcBlockstoreWithCache btcBlockStore = mock(BtcBlockstoreWithCache.class);
+        StoredBlock storedBlock = mock(StoredBlock.class);
+        when(storedBlock.getHeight()).thenReturn(bestChainHeight);
+        when(btcBlockStore.getChainHead()).thenReturn(storedBlock);
+        BridgeSupport bridgeSupport = getBridgeSupportWithMocksAndBtcBlockstoreForWhitelistTests(mockedWhitelist,btcBlockStore);
+
+        BigInteger disableBlockDelayBI = BigInteger.valueOf(Integer.MAX_VALUE / 2);
```
juli: If you want to test the case when current block height + delay is negative then use `Integer.MAX_VALUE`  otherwise it won't be negative until the Bitcoin block height is `Integer.MAX_VALUE / 2 + 1` and that will be in 40,000 years.

PR #737, opened 2018-12-20
Supporting Multiple Topics Logs

ajlopezrsk:
Now, the topics in log filter could be a two-level arrays of topics: first level refers to topics by position; second level is an OR filter for that position. Null matches all values


tinchou:
This is outdated. See b73de35d452592d6fa146d9aa198374857ae7b72.

PR #738, opened 2018-12-20
Fix traceListener creation in Program

ajlopezrsk:
The trace listener was created at the wrong time. Now, it is created before the creation of memory, storage and other artifacts inside Program


diega:
Would be possible to build the `programTraceListener` instance outside the `Program` constructor? The less dependencies we add to the _config_ object, the better

ajlopezrsk:
> Would be possible to build the `programTraceListener` instance outside the `Program` constructor? The less dependencies we add to the _config_ object, the better

It would be out of scope in this pull request. 

Anyway, this change will be subsumed by an upcoming pull request with an improved vmtrace output, more aligned with geth debug_traceTransaction JSON output

PR #740, opened 2018-12-20
Proper creation of blockchain builder for Remasc tests

diega:
They had the environment config missing


tinchou:
Could the builder be a field and set up in a `@Before` method?

diega:
@tinchou sure, thanks god it's Thursday!

PR #802, opened 2019-04-04
Most Ethereum inherited state tests re-included

SergioDemianLerner:


SergioDemianLerner:
> Looks good, but it's a pity we still support ignoring the Remasc contract: it's a fundamental piece of RSK and it shouldn't be ignored in my opinion.

This could be a temporal measure. 
However it brings many benefits: we can take Ethereum tests exactly as they are and run them on RSK. If we enable paying fees to REMASC contract then all tests fail because they check which are the resulting contracts, and the REMASC is not part of this set.

PR #805, opened 2019-04-05
Refactor tests that use World

collheesi:
I started reworking some tests that use the old World class.


collheesi:
I'll do this some other time.

PR #828, opened 2019-05-06
Validate Hashes size to be lower than 256 (uint8)

pmprete:
As described in https://github.com/bitcoin/bips/blob/master/bip-0037.mediawiki
If not throws InvalidMerkleBranchException

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/bitcoin/MerkleBranch.java
+++ b/rskj-core/src/main/java/co/rsk/peg/bitcoin/MerkleBranch.java
@@ -30,6 +30,11 @@ public class MerkleBranch {
         this.hashes = Collections.unmodifiableList(hashes);
         this.path = path;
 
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/peg/bitcoin/MerkleBranch.java
```diff
@@ -30,6 +30,11 @@ public MerkleBranch(List<Sha256Hash> hashes, int path) {
         this.hashes = Collections.unmodifiableList(hashes);
         this.path = path;
 
+        //We validate that the number of hashes is uint8 as described in https://github.com/bitcoin/bips/blob/master/bip-0037.mediawiki
+        if (hashes.size() > 256) {
+            throw new InvalidMerkleBranchException("The number of hashes must be lower than 256");
```
josedahlquist:
```suggestion
            throw new InvalidMerkleBranchException("The number of hashes must be lower than or equal to 256");
```

PR #857, opened 2019-05-22
[RSKIP-110] Fork Detection Data

martinmedina:
**Feature** 
https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP110.md

**Implementation changes**
The only implementation changes made to the RSKIP 110 document original version is that:
- Calculation of the CPV  for `ForkDetectionData`: a border case for block height multiple of 64 has been detected and the formula changed to cover that case.
- Calculation of number of uncles for the `ForkDetectionData`: now it is counted starting on the best block and not from the block being built to mined.

**About this version**
It is a first version. There are some performance improvements that can be applied to some data structures. These changes are identified as not necessary for this version and can be included in a forthcoming version.


rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
```diff
@@ -0,0 +1,117 @@
...
+            return new byte[0];
+        }
+
+        byte[] forkDetectionData = new byte[12];
+
+        byte[] commitToParentsVector = buildCommitToParentsVector(mainchainBlockHeaders);
+        System.arraycopy(commitToParentsVector, 0, forkDetectionData, 0, 7);
+
+        short numberOfUncles = getNumberOfUncles(mainchainBlockHeaders);
+        forkDetectionData[7] = ByteBuffer.allocate(2).putShort(numberOfUncles).array()[1];
```
SergioDemianLerner: Why so much pain to get (byte) numberOfUncles ? It doesn't guard against any overflow. Is there anything I'm missing here?

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
```diff
@@ -0,0 +1,117 @@
...
+
+        byte[] blockBeingMinedHeight = getBlockBeingMinedHeight(mainchainBlockHeaders);
+        System.arraycopy(blockBeingMinedHeight, 0, forkDetectionData, 8, 4);
+
+        return forkDetectionData;
+    }
+
+    private byte[] buildCommitToParentsVector(List<BlockHeader> mainchainBlocks) {
+        NetworkParameters params = RegTestParams.get();
```
SergioDemianLerner: How are you sure that RegTestParams work for Bitcoin MainNet here ?

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
```diff
@@ -0,0 +1,117 @@
...
+        new Context(params);
+
+        long bestBlockHeight = mainchainBlocks.get(0).getNumber();
+        long blockBeingMinedHeight = bestBlockHeight + 1;
+        long cpvStartHeight = (isMultipleOf64(blockBeingMinedHeight) ?
```
SergioDemianLerner: This has the same effect as the simpler:

 long cpvStartHeight = ((blockBeingMinedHeight - 1) / CPV_JUMP_FACTOR) * CPV_JUMP_FACTOR;

No need to use Math.floor() or isMultipleOf64()

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
```diff
@@ -0,0 +1,117 @@
...
+                                * CPV_JUMP_FACTOR;
+
+        byte[] commitToParentsVector = new byte[7];
+        for(int i = 0; i < CPV_SIZE; i++){
+            long currentCpvElement = bestBlockHeight - cpvStartHeight + i * CPV_JUMP_FACTOR;
+            BlockHeader blockHeader = mainchainBlocks.get((int)currentCpvElement);
+            byte[] bitcoinBlock = blockHeader.getBitcoinMergedMiningHeader();
+
+            byte[] bitcoinBlockHash = params.getDefaultSerializer().makeBlock(bitcoinBlock).getHash().getBytes();
+            byte leastSignificantByte = bitcoinBlockHash[bitcoinBlockHash.length - 1];
```
SergioDemianLerner: Make sure the LSB of bitcoinBlockHash is not the MSB for PoW target calculations. I remember the transaction hash or the blockhash was reversed, but I don't remember which of the two was.
If it's the MSB, it will be always 0x00 for Bitcoin mainnet.

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
```diff
@@ -0,0 +1,117 @@
...
+        // int to short is a safe cast since number of uncles is max 7 and blocks evaluated are at most 32.
+        // Hence, 7 * 32 = 224 and 224 < 255 (max number that fits on a short type variable)
+        return (short)IntStream
+                .range(0, NUMBER_OF_UNCLES)
+                .map(i -> mainchainBlocks.get(i).getUncleCount()).sum();
+    }
+
+    private byte[] getBlockBeingMinedHeight(List<BlockHeader> mainchainBlocks) {
+        long blockBeingMinedHeight = mainchainBlocks.get(0).getNumber() + 1;
+        return ByteBuffer.allocate(4).putInt((int)blockBeingMinedHeight).array();
```
SergioDemianLerner: The RSKIP doesn't state if this value should be stored in little or big endian. How are RLP values stored ? Are we using the same standard? In any case, if you are going to submit an improvement of the RSKIP110, then specifying the endianess used would be great.

SergioDemianLerner: Also note that Bitcoin header values endianess may differ from RLP's

rskj-core/src/main/java/co/rsk/validators/ForkDetectionDataRule.java
```diff
@@ -0,0 +1,133 @@
...
+            return false;
+        }
+
+        return true;
+    }
+
+    private boolean hasEnoughBlocksToCalculateForkDetectionData(BlockHeader header, List<BlockHeader> headersView) {
+        if (headersView.size() != requiredBlocksForForkDetectionDataCalculation) {
+            logger.error("Missing blocks to calculate fork detection data. Block hash {} ", header.getShortHash());
+            panicProcessor.panic(
```
SergioDemianLerner: This would be a completely abnormal situation. Shouldn't it be better to just shutdown the node?

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -94,6 +95,9 @@
     private byte[] bitcoinMergedMiningMerkleProof;
     /* The bitcoin protobuf serialized coinbase tx for merged mining */
     private byte[] bitcoinMergedMiningCoinbaseTransaction;
+
+    private byte[] miningForkDetectionData;
```
SergioDemianLerner: Why is miningForkDetectionData stored separately if it's already stored inside bitcoinMergedMiningMerkleProof ? 
A lot of code has been added because of this (e.g. new parsing in BlockFactory).
If this is because of a performance issue, then it could be computed a single time and then cached, but not received in the constructor.
Changing BlockFactory has implications in BLOCK_MESSAGE which affects the network protocol. I strongly think this may bring unintended consequences and protocol incompatibilities we are not aware of. Take into account there is no RSKIP that this network packed change.

rskj-core/src/test/java/org/ethereum/rpc/Web3ImplLogsTest.java
```diff
@@ -99,6 +102,7 @@
     public void setUp() {
         RskTestFactory factory = new RskTestFactory();
```
ignaciopulicedonatto: Tincho: RskTestFactory is deprecated use  RskTestContext instead which builds the same graph of dependencies than the productive code

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
tinchou: May I suggest extending `co.rsk.core.types.ints.Uint16`?

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
tinchou: Just so you know:

> The initial order of a byte buffer is always `BIG_ENDIAN`.

https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -233,6 +232,16 @@ public Blockchain getBlockchain() {
         return blockchain;
     }
 
+    public MiningMainchainView getMiningMainchainView() {
+        if (miningMainchainView == null) {
+            miningMainchainView = new MiningMainchainViewImpl(
+                    getBlockchain(),
+                    MiningConfig.REQUIRED_NUMBER_OF_BLOCKS_FOR_FORK_DETECTION_CALCULATION);
```
tinchou: I know this is a silly request before we have a linter in place, but could you stick to the current code style of this file? (the `);` bit goes in a new line).

rskj-core/src/main/java/co/rsk/mine/MinerManager.java
```diff
@@ -18,14 +18,13 @@
...
 
-import org.ethereum.core.Blockchain;
 
 /**
  * Created by ajlopez on 15/04/2017.
  */
 public class MinerManager {
-    public void mineBlock(Blockchain blockchain, MinerClient minerClient, MinerServer minerServer) {
-        minerServer.buildBlockToMine(blockchain.getBestBlock(), false);
+    public void mineBlock(MinerClient minerClient, MinerServer minerServer) {
```
tinchou: Isn't this a good time to get rid of this useless class? Or give it some responsibilities.

rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
```diff
@@ -451,31 +470,23 @@ private boolean getNotify(Block block, Keccak256 parentHash) {
...
         @Override
-        /**
-         * onBlock checks if we have to mine over a new block. (Only if the blockchain's best block changed).
-         * This method will be called on every block added to the blockchain, even if it doesn't go to the best chain.
-         * TODO(???): It would be cleaner to just send this when the blockchain's best block changes.
-         * **/
         // This event executes in the thread context of the caller.
         // In case of private miner, it's the "Private Mining timer" task
-        public void onBlock(Block block, List<TransactionReceipt> receipts) {
+        public void onBestBlock(Block newBestBlock, List<TransactionReceipt> receipts) {
```
tinchou: Could this change be done in a separate commit? Changing the hook is something I've been meaning to do for a while, but was never sure if could cause unintended consequences.
Having it as a separate commit (or even a separate PR) would be very helpful.

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainView.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.List;
+
+public abstract class MiningMainchainView implements MainchainView {
+
+    public abstract void addBestBlock(Block bestBlock);
+
+    public abstract List<Block> get();
+
+    public abstract Block getBestBlock();
```
tinchou: Given this is a brand new interface, I suggest having a `Optional<Block>` return type to signal this might return an empty value. Same below.

tinchou:
I would suggest extracting non-critical changes and refactors into separate PRs so it's easier to review the actual RSKIP logic.

E.g.: the abstraction for a main chain view is a great idea and it must be done, but having it here ultimately undermines this PR. A separate PR:

* benefits the whole team because we can start building on top of your abstractions
* benefits you because you don't have to rebase if other changes collide
* makes reviewing the actual RSKIP much easier
* code improvements reach the `master` branch even if the whole RSKIP doesn't

We've found this is a great strategy while working in `unitrie` for the past semester.

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
martinmedina: Because here we are creating blocks in isolation and not running any validations on them. Furthermore, these blocks do not even have transactions. It's basically a creation of a block based on its hex header value.
I agree it would be a good idea to use the right network config and I'm creating an issue to change that on a next version.

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
martinmedina: Your suggestion is not the exact same formula but it works for what's intended. Change applied!

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
martinmedina: Extended `Uint8` to be used on this case. This new type checks for overflow on its constructor so that is also covered now.

rskj-core/src/test/java/org/ethereum/rpc/Web3ImplLogsTest.java
martinmedina: I did not change that line while implementing this feature.I consider this is out of the scope of this PR.
Still I quick tried your suggestion and it did not work. That means that `RskTestFactory` can not be replaced that easily in this particular case. 
I agree it would be great to have that changed, let's do that someday on a new branch :)

rskj-core/src/main/java/co/rsk/RskContext.java
martinmedina: Sure, np!

rskj-core/src/main/java/co/rsk/mine/MinerManager.java
martinmedina: I did reduce the number of parameters on the signature of `mineBlock` which I consider a step forward in getting rid of this class. I don't think this PR is the place to remove it. 
Since I agree it should be removed, I'll create an issue for doing that after this PR is merged to master.

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
martinmedina: That's a good one, fortunately I did take that into consideration :)
In the [tests](https://github.com/rsksmart/rskj/pull/857/files#diff-e90903819d3632606b72146875bf115dR282) I'm using a valid bitcoin mainnet header and if you debug them you can see that the value is not 0.

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
martinmedina: RLP values are stored in Big Endian.
By default, as @collheesi says, Java byte buffer data structure uses Big Endian.
Hence, there are no code changes to do since it's already big endian and RLP compatible.
I would specify the endianness in RSKIP110 after this PR revision finishes.

rskj-core/src/main/java/co/rsk/validators/ForkDetectionDataRule.java
martinmedina: I agree it's an abnormal situation but I'm not sure if we are handling this kind of cases with a shutdown.
Let me check that and I'll make the change if necessary.

rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
martinmedina: I always meant to do this refactor separately but it ended up being a must for adding `MiningMainchainView` to mining flow. The reason for the former is that there was no easy way to add it using the `onBlock` hook. That explains why the change belongs to that commit.
As far as I'm aware, it's not causing any unintended consequences.

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainView.java
martinmedina: It is not supposed to return an empty value.

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
martinmedina: It can not be removed from constructor because it must be provided on block building flow.
I did extract the value from `bitcoinMergedMiningCoinbaseTransaction` 👍

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
```diff
@@ -63,8 +64,8 @@
         byte[] commitToParentsVector = buildCommitToParentsVector(mainchainBlockHeaders);
         System.arraycopy(commitToParentsVector, 0, forkDetectionData, 0, 7);
 
-        short numberOfUncles = getNumberOfUncles(mainchainBlockHeaders);
-        forkDetectionData[7] = ByteBuffer.allocate(2).putShort(numberOfUncles).array()[1];
+        Uint8 numberOfUncles = getNumberOfUncles(mainchainBlockHeaders);
+        forkDetectionData[7] = numberOfUncles.asByte();
```
collheesi: Aren't you re-implementing encoding here? I would suggest having an `encode(array, offset)` overload for this use case (and many more to come!) and remove `asByte`.

martinmedina: The `asByte` method is what I need for this case and it makes total sense to have it since Uint8 is just a byte wrapper.

rskj-core/src/main/java/co/rsk/mine/MinerManager.java
collheesi: Thanks!

rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
collheesi: Good to know

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainView.java
collheesi: I thought I saw a null check, nevermind!

rskj-core/src/main/java/co/rsk/core/bc/MainchainView.java
```diff
@@ -0,0 +1,23 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package co.rsk.core.bc;
+
+public interface MainchainView {
```
collheesi: Is this interface still needed? It's empty and no one uses it.

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainView.java
```diff
@@ -0,0 +1,34 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package co.rsk.core.bc;
+
+import org.ethereum.core.Block;
+
+import java.util.List;
+
+public abstract class MiningMainchainView implements MainchainView {
```
collheesi: Let me suggest not having this abstract class at all. When you only have one implementor, the team consensus is not to add an abstraction that might not be needed. Information hiding is already achieved through `public/private` methods

rskj-core/src/main/java/co/rsk/core/bc/ConsensusValidationMainchainView.java
```diff
@@ -0,0 +1,32 @@
...
+
+package co.rsk.core.bc;
+
+import co.rsk.crypto.Keccak256;
+import org.ethereum.core.BlockHeader;
+
+import java.util.List;
+import java.util.Map;
+
+public abstract class ConsensusValidationMainchainView implements MainchainView {
```
collheesi: Same as `MiningMainchainView`

rskj-core/src/main/java/co/rsk/core/bc/MainchainView.java
martinmedina: It's a [Marker Interface](https://en.wikipedia.org/wiki/Marker_interface_pattern) intended to group all the views interfaces. All the views use it.

rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
```diff
@@ -189,46 +184,33 @@ public BlockHeader decodeHeader(RLPList rlpHeader, boolean sealed) {
...
-        int headerElementsParsed = 15;
+        int r = 15;
 
         int uncleCount = 0;
-        if (rlpHeader.size() == RLP_HEADER_SIZE_WITH_MERGED_MINING_AND_FORK_DATA ||
-                rlpHeader.size() == RLP_HEADER_SIZE_WITH_MERGED_MINING ||
-                rlpHeader.size() == RLP_HEADER_SIZE_WITH_FORK_DATA ||
-                rlpHeader.size() == RLP_HEADER_SIZE) {
-            byte[] ucBytes = rlpHeader.get(headerElementsParsed++).getRLPData();
+        if ((rlpHeader.size() == 19) || (rlpHeader.size() == 16)) {
```
ignaciopulicedonatto: Can we add some name to the numbers 16 and 19 ?

martinmedina: Sure, I lost that on a recent rollback. Fixed!

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainView.java
martinmedina: As we agreed on our talk on another channel, this class gets to stay as `MainchainView` flew away. Changes applied to comply with that!

rskj-core/src/main/java/co/rsk/core/bc/ConsensusValidationMainchainView.java
martinmedina: Same reply as `MiningMainchainView`

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -471,7 +498,30 @@ public String getParentShortHash() {
...
+                List<Byte> hashForMergedMiningPrefixAsList = Arrays.asList(ArrayUtils.toObject(hashForMergedMiningPrefix));
+                List<Byte> coinbaseAsList = Arrays.asList(ArrayUtils.toObject(coinbaseTransaction));
+
+                int position = Collections.lastIndexOfSubList(coinbaseAsList, hashForMergedMiningPrefixAsList);
+                if (position == -1) {
+                    return new byte[12];
+                }
+
+                int from = position + HASH_FOR_MERGED_MINING_PREFIX_LENGTH;
+                return Arrays.copyOfRange(coinbaseTransaction, from, from + FORK_DETECTION_DATA_LENGTH);
```
SergioDemianLerner: I see many calls to getMiningForkDetectionData() around the code. I would suppose that the first time it is called the miningForkDetectionData is extracted, but the following calls, if the miningForkDetectionData field is non-null, it would be returned (as a cache), instead of being re-computed.
Is there a reason for continuous recomputation?

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -471,7 +498,30 @@ public String getParentShortHash() {
...
+
+                byte[] hashForMergedMiningPrefix = Arrays.copyOfRange(hashForMergedMining, 0, HASH_FOR_MERGED_MINING_PREFIX_LENGTH);
+                byte[] coinbaseTransaction = getBitcoinMergedMiningCoinbaseTransaction();
+
+                List<Byte> hashForMergedMiningPrefixAsList = Arrays.asList(ArrayUtils.toObject(hashForMergedMiningPrefix));
+                List<Byte> coinbaseAsList = Arrays.asList(ArrayUtils.toObject(coinbaseTransaction));
+
+                int position = Collections.lastIndexOfSubList(coinbaseAsList, hashForMergedMiningPrefixAsList);
+                if (position == -1) {
+                    return new byte[12];
```
SergioDemianLerner: Sometimes this method returns new byte[0] and some times new byte[12] (zero-filled). What is the reason for not returning always the same? Does the calling code distinguished between these two values? If these are error conditions, wouldn't be better to rise an exception ?

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -471,7 +498,30 @@ public String getParentShortHash() {
...
+                byte[] hashForMergedMining = HashUtil.keccak256(encodedBlock);
+
+                byte[] hashForMergedMiningPrefix = Arrays.copyOfRange(hashForMergedMining, 0, HASH_FOR_MERGED_MINING_PREFIX_LENGTH);
+                byte[] coinbaseTransaction = getBitcoinMergedMiningCoinbaseTransaction();
+
+                List<Byte> hashForMergedMiningPrefixAsList = Arrays.asList(ArrayUtils.toObject(hashForMergedMiningPrefix));
+                List<Byte> coinbaseAsList = Arrays.asList(ArrayUtils.toObject(coinbaseTransaction));
+
+                int position = Collections.lastIndexOfSubList(coinbaseAsList, hashForMergedMiningPrefixAsList);
+                if (position == -1) {
```
SergioDemianLerner: If we're parsing the mining fields, then maybe it would be wise to cache also the RSKTAG and the ProofOfWork rule wouldn't need to compute this again (it would only call header.getRSKTAG() or something). 
I don't like much that the tag-handling code be duplicated in two classes, being this critical to the consensus and security of the platform.

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
```diff
@@ -0,0 +1,111 @@
...
+        return forkDetectionData;
+    }
+
+    private byte[] buildCommitToParentsVector(List<BlockHeader> mainchainBlocks) {
+        NetworkParameters params = RegTestParams.get();
+        new Context(params);
+
+        long bestBlockHeight = mainchainBlocks.get(0).getNumber();
+        long blockBeingMinedHeight = bestBlockHeight + 1;
+        long cpvStartHeight = ((blockBeingMinedHeight - 1) / CPV_JUMP_FACTOR) * CPV_JUMP_FACTOR;
```
juli: Ok, I prefer `cpvStartHeight = bestBlockHeight - (bestBlockHeight % CPV_JUMP_FACTOR)` but the result is the closest multiple of CPV_JUMP_FACTOR lower or equal current height.

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
```diff
@@ -0,0 +1,111 @@
...
+
+        byte[] commitToParentsVector = new byte[7];
+        for(int i = 0; i < CPV_SIZE; i++){
+            long currentCpvElement = bestBlockHeight - cpvStartHeight + i * CPV_JUMP_FACTOR;
```
juli: Now you subtract two absolute heights what will give you `bestBlockHeight % CPV_JUMP_FACTOR' a value between 0 and 63. And you add a multiple of `CPV_JUMP_FACTOR` to it. The result does not increment with block height.

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private byte[] buildCommitToParentsVector(List<BlockHeader> mainchainBlocks) {
+        NetworkParameters params = RegTestParams.get();
+        new Context(params);
+
+        long bestBlockHeight = mainchainBlocks.get(0).getNumber();
+        long blockBeingMinedHeight = bestBlockHeight + 1;
+        long cpvStartHeight = ((blockBeingMinedHeight - 1) / CPV_JUMP_FACTOR) * CPV_JUMP_FACTOR;
+
+        byte[] commitToParentsVector = new byte[7];
```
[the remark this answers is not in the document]
martinmedina: 👍

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
martinmedina: The case of `new byte[12]` is an error condition, I changed it to throw an exception. 
It is ok to return `new byte[0]` because its consistent with the rest of the code for the case when that value can not be calculated.

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
martinmedina: Worst case scenario, it is called only twice in a row. All the other calls are done to different instances. I think twice re computation does not need a cache, specially in a case where the computational complexity is very low.
Anyway, I added a cache as you asked :)

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
martinmedina: In this case, the value extracted is not the same as the one in PowRule. 
I think that BlockHeader may be responsible for knowing how to extract the values used in PowRule (this a topic I consider open for discussion). 
On this feature implementation, a big one, I'd rather not change PowRule and leave that for a future version. We all know that PowRule needs a huge refactor.

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
```diff
@@ -0,0 +1,111 @@
...
+
+        return forkDetectionData;
+    }
+
+    private byte[] buildCommitToParentsVector(List<BlockHeader> mainchainBlocks) {
+        NetworkParameters params = RegTestParams.get();
+        new Context(params);
+
+        long bestBlockHeight = mainchainBlocks.get(0).getNumber();
+        long blockBeingMinedHeight = bestBlockHeight + 1;
```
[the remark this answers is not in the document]
martinmedina: The algorithm is not wrong, it does what the RSKIP says. 
Refactor applied to add one substract one 👍

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
martinmedina: Replied in first comment of the series.

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
martinmedina: Replied in first comment of the series.

rskj-core/src/test/java/co/rsk/core/BlockHeaderTest.java
```diff
@@ -136,7 +136,7 @@ public void getMiningForkDetectionDataNoDataCanBeFound() {
         header.setBitcoinMergedMiningCoinbaseTransaction(coinbase);
         header.seal();
 
-        assertThat(new byte[12], is(header.getMiningForkDetectionData()));
+        header.getMiningForkDetectionData();
```
collheesi: is this intentional?

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
```diff
@@ -0,0 +1,114 @@
...
+
+        return forkDetectionData;
+    }
+
+    private byte[] buildCommitToParentsVector(List<BlockHeader> mainchainBlocks) {
+        long bestBlockHeight = mainchainBlocks.get(0).getNumber();
+        long cpvStartHeight = (bestBlockHeight / CPV_JUMP_FACTOR) * CPV_JUMP_FACTOR;
+
+        byte[] commitToParentsVector = new byte[CPV_SIZE];
+        for(int i = 0; i < CPV_SIZE; i++){
```
collheesi: For the record: we've seen that this only reads 1 out of 64 blocks. Please consider improving the caching design in a v2 or v3.

rskj-core/src/test/java/co/rsk/core/BlockHeaderTest.java
martinmedina: It is because the handling of that case changed from returning array to throwing an exception

rskj-core/src/main/java/co/rsk/mine/ForkDetectionDataCalculator.java
martinmedina: Sure, we are aware of that and will be improved!

martinmedina:
Please find some improvements on [v2](https://github.com/rsksmart/rskj/pull/873)

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
```diff
@@ -66,24 +83,93 @@ public void addBest(BlockHeader bestHeader) {
...
+     * Given a source list take it as the new mainchain and refill it with as many block headers are needed or trim it
+     * to reach the desired depth/height
+     *
+     * @param sourceList
+     */
+    private void buildMainchainFromList(List<BlockHeader> sourceList) {
+        int sourceSize = sourceList.size();
+
+        if (sourceSize == 0) {
+            return;
```
ignaciopulicedonatto: In this case should we add something like mainchain = new List()

PR #862, opened 2019-05-23
New RepositoryBlockStore that has cache of the latest blocks

pmprete:
Use a cache of the latest blocks inside the Repository BlockStore


rskj-core/src/main/java/co/rsk/peg/BtcBlockStoreWithCache.java
```diff
@@ -1,6 +1,6 @@
 /*
  * This file is part of RskJ
- * Copyright (C) 2018 RSK Labs Ltd.
```
lsebrie: Retrocedió en el tiempo!

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -332,6 +317,23 @@ public void registerBtcTransaction(Transaction rskTx, byte[] btcTxSerialized, in
...
+            throw new BridgeIllegalArgumentException("PartialMerkleTree doesn't have expected size");
+        }
+
+        Sha256Hash merkleRoot;
+        try {
+            PartialMerkleTree pmt = new PartialMerkleTree(bridgeConstants.getBtcParams(), pmtSerialized, 0);
+            List<Sha256Hash> hashesInPmt = new ArrayList<>();
+            merkleRoot = pmt.getTxnHashAndMerkleRoot(hashesInPmt);
+            if (!hashesInPmt.contains(btcTxHash)) {
+                logger.warn("Supplied tx is not in the supplied partial merkle tree");
```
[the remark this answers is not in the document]
pmprete: good catch

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1130,23 +1132,20 @@ public Integer getBtcTransactionConfirmations(Sha256Hash btcTxHash, Sha256Hash b
...
+            // Make sure it belongs to the best chain
+            if (!blockInfo.isMainChain()){
+                return BTC_TRANSACTION_CONFIRMATION_BLOCK_NOT_IN_BEST_CHAIN_ERROR_CODE;
+            }
+        } catch (BlockStoreException e) {
             logger.warn(String.format(
-                    "Illegal state trying to get block at height %d with hash %s",
-                    block.getHeight(),
-                    Hex.toHexString(block.getHeader().getHash().getBytes())
+                    "Illegal state trying to get block with hash %s",
```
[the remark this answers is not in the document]
tinchou: Also, `Hex.toHexString(block.getHeader().getHash().getBytes())` is redundant when you can call `block.getHeader().getHash().toHexString()` (or even better let the logger interpolate `block.getHeader().getHash()`)

tinchou:
The security issue in `MerkleBranch` is a false positive. It might go away if you mark the field `final` (which by the way we should always be doing!).

rskj-core/src/main/java/co/rsk/peg/RepositoryBtcBlockStoreWithCache.java
```diff
@@ -0,0 +1,214 @@
...
+        return ba;
+    }
+
+    private StoredBlock byteArrayToStoredBlock(byte[] ba) {
+        ByteBuffer byteBuffer = ByteBuffer.wrap(ba);
+        return StoredBlock.deserializeCompact(btcNetworkParams, byteBuffer);
+    }
+
+    public static class Factory implements BtcBlockStoreWithCache.Factory {
+        private final Map<Sha256Hash, StoredBlock> cacheBlocks = new MaxSizeHashMap<>(MAX_SIZE_MAP_STORED_BLOCKS, true);
```
collheesi: The idea is that all block stores created with a specific factory share the same cache. I love it.

One thing that isn't a problem now because we don't have parallel execution is concurrent access to this cache, maybe leaving a comment is worth it.

pmprete: good idea

aeidelman:
@pmprete do we still need this PR ?

PR #867, opened 2019-05-28
After validation fail, create again the block header without uncles

ignaciopulicedonatto:
We are not completely sure if the condition (validationRules.isValid()) is ever false. (The 80% of the validations are related to uncle rules and some of those validations doesn't have sense because how we got those uncles, I mean it's like a double check.)
This is a compromise solution to avoid creating an inconsistent block if validation goes wrong. 
So we create again the block with empty uncle list as before but also we create again the header with empty uncle list.


rskj-core/src/test/java/co/rsk/mine/MineBlockWithUncleTest.java
```diff
@@ -43,36 +43,31 @@
...
-import static org.junit.Assert.*;
+import static org.hamcrest.Matchers.hasSize;
+import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.*;
 
 @RunWith(PowerMockRunner.class)
-@PrepareForTest(FamilyUtils.class)
-public class BlockMineTest {
+@PrepareForTest({FamilyUtils.class})
+public class MineBlockWithUncleTest {
```
martinmedina: Change the name of this class to `BlockToMineBuildetTest.java` to comply with tests classes naming of the project.

collheesi: The test sets up a very specific context which won't really extend to other `BlockToMineBuilder` tests, and I suggested the name.
I guess it's a matter of preference but I thought I'd leave this here for reference.

martinmedina: I think it would apply to other `BlockToMineBuilder` tests since with the mocks it covers the whole flow of block creation.

PR #868, opened 2019-05-30
Change Trie::save() method

julianlen:


tinchou:
Consider changing `NodeReference#save` to:

```java
public void save() {
    if (lazyNode != null && !isEmbeddable()) {
        lazyNode.save();
    }
}
```

rskj-core/src/main/java/co/rsk/trie/NodeReference.java
```diff
@@ -55,7 +54,11 @@ public boolean isEmpty() {
      */
     public void save() {
         if (lazyNode != null) {
-            lazyNode.save();
+            if (!isEmbeddable()) {
+                lazyNode.save();
+            } else {
+                store.saveValue(lazyNode);
```
collheesi: This is called even when the node doesn't have a long value. I think this is a bug.

rskj-core/src/main/java/co/rsk/trie/Trie.java
```diff
@@ -503,6 +507,10 @@ public Trie delete(String key) {
      * @return a byte array with the serialized info
      */
     public byte[] toMessage() {
+        if (encoded != null) {
+            return cloneArray(encoded);
```
SergioDemianLerner: Doesn't cloning affect performance ? Trie.gets() also clone. Can't we just tag immutable arrays with large banners in comments and trust humans to follow the specified semantic ?

rskj-core/src/main/java/co/rsk/trie/Trie.java
```diff
@@ -563,7 +571,13 @@ public Trie delete(String key) {
...
         }
 
-        return buffer.array();
+        encoded = buffer.array();
+
+        return cloneArray(encoded);
+    }
+
+    public int getMessageLength() {
+        return toMessage().length;
```
SergioDemianLerner: unnecessary clone here.

rskj-core/src/main/java/co/rsk/trie/Trie.java
collheesi: I don't think we can trust humans. We plan to take the Pantheon approach and have immutable objects that can be safely passed around (`BytesValue`), but we can't do it until now without interfering with other humans' schedules.

See: https://github.com/PegaSysEng/pantheon/blob/master/util/src/main/java/tech/pegasys/pantheon/util/bytes/BytesValue.java

rskj-core/src/main/java/co/rsk/trie/Trie.java
lsebrie: We made this change to address this clone https://github.com/rsksmart/rskj/pull/874

PR #870, opened 2019-06-04
Add fix mechanism for patching broken databases on migration

diega:


collheesi:
Looks good so far. We should discuss where to host this file.

rskj-core/src/main/resources/reference.conf
```diff
@@ -84,6 +84,7 @@ database {
     reset = false
     unitrie-migration {
         minimum-height: 0
+        missing-keys-url: "https://github.com/rsksmart/unitrie-migration/blob/master/migration-extras?raw=true"
```
collheesi: Did you hear from QA about the key override issue?

diega: finally it wasn't an override issue, it was just me retrieving the wrong key

PR #873, opened 2019-06-05
Feature rskip110 improvements - Armadillo v2

ignaciopulicedonatto:


rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
```diff
@@ -34,79 +34,77 @@
...
-            blocksByHash.put(bestBlock.getHash(), bestBlock);
-            addToBlockByNumberMap(bestBlock);
+            addHeaderToMaps(bestHeader);
+
+            BlockHeader previousBest = mainchain.get(0);
+            if (previousBest.isParentOf(bestHeader)) {
+                mainchain.add(0, bestHeader);
+            } else {
+                // new block is not on the current chain, so we need to recalculate it to keep it consistent
+                fillInternalStore(bestHeader);
```
ignaciopulicedonatto: So we basically avoid recalculating the blockchain every time a block arrives right? Beisdes that is it correct recalculate the blockchain if the top is not the parent?

ignaciopulicedonatto: Is this the memory leak that Juliano was worried?

martinmedina: What would be ideal, in terms of BIg O performance, is to search the existent blockchain until a common parent is found. This is an extension of the case that checks if you are adding a block on the tip of the chain. 
Be careful because depending on the kind of reorganization, it may be also needed to add blocks at the end. That allows to complete the number of blocks the view holds as invariant.

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
```diff
@@ -42,13 +41,13 @@
...
     private List<BlockHeader> mainchain;
 
-    public MiningMainchainViewImpl(Blockchain blockchain,
+    public MiningMainchainViewImpl(BlockStore blockStore,
                                    int height) {
         this.height = height;
-        this.blockchain = blockchain;
+        this.blockStore = blockStore;
         this.blocksByHash = new HashMap<>();
         this.blocksByNumber = new HashMap<>();
```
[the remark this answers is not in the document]
M-Picco: Addressed at 53e88ec

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
M-Picco: See 4705e52 for the changes

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
```diff
@@ -19,102 +19,200 @@
...
-        long currentBlockNumber = blockToAdd.getNumber();
-        if (blocksByNumber.containsKey(currentBlockNumber)) {
-            blocksByNumber.get(currentBlockNumber).add(blockToAdd);
+    private void addToBlockHashesByNumberMap(BlockHeader headerToAdd) {
+        long blockNumber = headerToAdd.getNumber();
+        if (blockHashesByNumber.containsKey(blockNumber)) {
+            blockHashesByNumber.get(blockNumber).add(headerToAdd.getHash());
         } else {
-            blocksByNumber.put(blockToAdd.getNumber(), new ArrayList<>(Collections.singletonList(blockToAdd)));
+            blockHashesByNumber.put(headerToAdd.getNumber(), new ArrayList<>(Collections.singletonList(headerToAdd.getHash())));
```
ignaciopulicedonatto: We can use put(currentBlockNumber) here

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
```diff
@@ -19,102 +19,200 @@
...
             }
+        }
+
+        mainchain = Stream.concat(mainchain.stream(), missingHeaders.stream())
+                .collect(Collectors.toList());
+    }
 
-            if(currentBlock.isGenesis()) {
+    /**
+     * Given a start block header and a chain length, retrieve a List of said lenght consisting of
```
ignaciopulicedonatto: typo

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
```diff
@@ -19,102 +19,200 @@
...
+import org.ethereum.core.BlockHeader;
+import org.ethereum.db.BlockStore;
 
 import javax.annotation.concurrent.GuardedBy;
 import java.util.*;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class MiningMainchainViewImpl implements MiningMainchainView {
```
ignaciopulicedonatto: Can we add @ThreadSafe  annotation? , is just to let the world know that we were worried about concurrency

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
```diff
@@ -73,24 +79,84 @@ public void addBest(BlockHeader bestHeader) {
...
+
+        BlockHeader currentHeader = header;
+        for(int i = 0; i < chainLength; i++) {
+
+            // genesis has no parent
             if(currentHeader.isGenesis()) {
                 break;
             }
 
             currentHeader = blockStore.getBlockByHash(currentHeader.getParentHash().getBytes()).getHeader();
```
ignaciopulicedonatto: Remember to check if currentHeader is != null

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
```diff
@@ -73,24 +79,84 @@ public void addBest(BlockHeader bestHeader) {
...
+
+        List<BlockHeader> newMainchain = Stream.concat(
+                Arrays.asList(bestHeader).stream(),
+                commonAncestorChain.stream())
+                .collect(Collectors.toList());
+
+        mainchain = newMainchain;
+
+        fillMissingHeaders();
+        deleteHeaders(staleHeaders);
```
martinmedina: Why do you use `deleteHeaders` here if after `deleteEntriesOutOfBoundaries` is called ? 
I understand these methods do the same.

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
M-Picco: Fixed

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
M-Picco: Should be covered by 50f4ceb

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
```diff
@@ -107,11 +173,46 @@ private void addToBlockHashesByNumberMap(BlockHeader headerToAdd) {
...
+
     private void deleteEntriesOutOfBoundaries(long bestBlockNumber) {
         long blocksHeightToDelete = bestBlockNumber - height;
         if(blocksHeightToDelete >= 0 && blockHashesByNumber.containsKey(blocksHeightToDelete)) {
             blockHashesByNumber.get(blocksHeightToDelete).forEach(blockHashToDelete -> blocksByHash.remove(blockHashToDelete));
             blockHashesByNumber.remove(blocksHeightToDelete);
         }
     }
+
+    private int findParentIndex(BlockHeader header) {
```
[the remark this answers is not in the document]
M-Picco: Cool, didn't know about that type. Addresses in cb95de6

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
M-Picco: Indeed, the methods do more or less the same, but the former is used to delete the stale headers after the chain is rebuilt based on a specific parent from the chain. However, these same stale headers would be eventually removed by `deleteEntriesOutOfBoundaries`.

I'll change it to leave the headers deletion to the later method since I doubt there'd be memory issues.

rskj-core/src/main/java/co/rsk/core/bc/MiningMainchainViewImpl.java
```diff
@@ -20,101 +20,180 @@
...
-        Block currentBlock = bestBlock;
-        for(int i = 0; i < height; i++) {
-            newBlockchain.add(currentBlock);
+        for (BlockHeader header : missingHeaders) {
+            if(!blocksByHash.containsKey(header.getHash())) {
+                addHeaderToMaps(header);
+            }
+        }
+
+        mainchain = Stream.concat(sourceList.stream(), missingHeaders.stream())
```
SergioDemianLerner: expensive copying of items when missingHeaders is just a few items. 
In the future a circular buffer would work reduce most unnecessary element copying. Also you can reduce copying by reversing the list (best header last) and moving the "first" pointer, with periodic removals of initial items (e.g. every 1000 elements added).

SergioDemianLerner: It also can be used to remove old headers only once in a while, with reduced cost (amortized).

rskj-core/src/main/java/co/rsk/rpc/ExecutionBlockRetriever.java
```diff
@@ -82,7 +88,7 @@ public Block getExecutionBlock(String bnOrId) {
         }
 
         if (executionBlockNumber.isPresent()) {
-            Block executionBlock = miningMainchainView.getBlockByNumber(executionBlockNumber.get());
+            Block executionBlock = blockchain.getBlockByNumber(executionBlockNumber.get());
```
SergioDemianLerner: what was the motivation for replacing miningMainchainView with blockchain?

martinmedina: Because `ExecutionBlockRetriever` uses Blocks and now `MiningMainchainView` stores `BlockHeaders`. 
In the future, this class will be refactored to use either the `BlockStore` or a specific view. 
We thought it would be better to not add that refactor to RSKIP 110 feature which is already adding several code changes.

PR #874, opened 2019-06-06
Avoid cloning array for message length use in Trie

lsebrie:


lsebrie:
> I suggest that `internalToMessage` actually returns the message

I considered that option but to avoid doing encoded var assignment on each function that uses it I chose to dismiss this version

PR #880, opened 2019-06-13
Created Solidity test for EXTCODESIZE call after suicide

smcsicardi:
Similar to the `create contract after suicide` test, a new DSL test was implemented to check if the code size of a contract destroyed on another transaction on the same block is zero.

Solidity files are included in folder with the same name of the DSL .txt file


aeidelman:
@smcsicardi please rebase

smcsicardi:
@aeidelman rebased!

PR #882, opened 2019-06-14
Mock the getParentHashJsonString() method as it should be @Notnull

pmprete:
Check if refreshWorkTimer is null at stop() to avoid a NPE

Mock the getParentHashJsonString() method as it should be @Notnull (test failing in some versions of intellij for this)

### DIFF
--- a/rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
+++ b/rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
@@ -156,8 +156,10 @@ public class MinerServerImpl implements MinerServer {
         synchronized (lock) {
             started = false;
             ethereum.removeListener(blockListener);
-            refreshWorkTimer.cancel();
-            refreshWorkTimer = null;
+            if (refreshWorkTimer != null) {
+                refreshWorkTimer.cancel();
+                refreshWorkTimer = null;
+            }
         }
     }
 
--- a/rskj-core/src/test/java/co/rsk/mine/MinerServerTest.java
+++ b/rskj-core/src/test/java/co/rsk/mine/MinerServerTest.java
@@ -673,12 +673,14 @@ public class MinerServerTest extends ParameterizedNetworkUpgradeTest {
         when(block1.getHashForMergedMining()).thenReturn(TestUtils.randomHash().getBytes());
         when(block1.getHash()).thenReturn(TestUtils.randomHash());
         when(block1.getDifficulty()).thenReturn(BlockDifficulty.ZERO);
+        when(block1.getParentHashJsonString()).thenReturn(TestUtils.randomHash().toJsonString());
 
         Block block2 = mock(Block.class);
         when(block2.getFeesPaidToMiner()).thenReturn(new Coin(BigInteger.valueOf(24)));
         when(block2.getHashForMergedMining()).thenReturn(TestUtils.randomHash().getBytes());
         when(block2.getHash()).thenReturn(TestUtils.randomHash());
         when(block2.getDifficulty()).thenReturn(BlockDifficulty.ZERO);
+        when(block2.getParentHashJsonString()).thenReturn(TestUtils.randomHash().toJsonString());
 
         BlockToMineBuilder builder = mock(BlockToMineBuilder.class);
         BlockResult blockResult = mock(BlockResult.class);


rskj-core/src/test/java/co/rsk/mine/MinerServerTest.java
```diff
@@ -664,12 +664,14 @@ public void gasUnitInDollarsIsInitializedOkAtConstructor() {
         when(block1.getHashForMergedMining()).thenReturn(TestUtils.randomHash().getBytes());
         when(block1.getHash()).thenReturn(TestUtils.randomHash());
         when(block1.getDifficulty()).thenReturn(BlockDifficulty.ZERO);
+        when(block1.getParentHashJsonString()).thenReturn(TestUtils.randomHash().toJsonString());
```
martinmedina: It is true that this value should be not null. I don't think this is the cause of the failing test since the value is assigned and retrieved (but not accessed). For that matter it is the same if it's null or any other `String` valid value.
Furthermore, `@NotNull` does not have any other meaning than documentation so that can not be affecting the test or any other execution path in anyway.
That was just to help you track the reason of your not passing test. 
I agree that having this mocked improves the test, thanks!

rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
```diff
@@ -178,6 +177,7 @@ public void start() {
 
             refreshWorkTimer = new Timer("Refresh work for mining");
             refreshWorkTimer.schedule(new RefreshBlock(), DELAY_BETWEEN_BUILD_BLOCKS_MS, DELAY_BETWEEN_BUILD_BLOCKS_MS);
+            started = true;
```
martinmedina: I have to disagree with this change. Suppose something fails in the initialization of the components and you haven't set `started = true` at the begging. When you call `stop()` it won't got further than 
```
if (!started) {
            return;	            
       }
```
and, for instance, `blockListener` will never be removed from the ethereum object.

I think a better solution would be to just check for not null before calling `refreshWorkTimer.cancel();` on the `stop()` method.

rskj-core/src/test/java/co/rsk/mine/MinerServerTest.java
pmprete: This is the error that i get if i don't use a mock to avoid the null value

> java.lang.IllegalArgumentException: Argument for @Nonnull parameter 'parentBlockHash' of co/rsk/mine/MinerWork.<init> must not be null

	at co.rsk.mine.MinerWork.$$$reportNull$$$0(MinerWork.java)
	at co.rsk.mine.MinerWork.<init>(MinerWork.java)
	at co.rsk.mine.MinerServerImpl.updateGetWork(MinerServerImpl.java:375)
	at co.rsk.mine.MinerServerImpl.buildBlockToMine(MinerServerImpl.java:412)
	at co.rsk.mine.MinerServerImpl.start(MinerServerImpl.java:175)
	at co.rsk.mine.MinerServerTest.gasUnitInDollarsIsInitializedOkAtConstructor(MinerServerTest.java:692)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)

rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
pmprete: Followed this advice, the changes are in the ammend

rskj-core/src/test/java/co/rsk/mine/MinerServerTest.java
martinmedina: Check if you have enabled the following setting 
Settings > Build, Execution, Deployment > Compiler > Add runtime assertions for not-null-annotated methods and parameters.
That is a possible reason to explain lain why it fails on your computer and not on CI

rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java
```diff
@@ -156,7 +156,9 @@ public void stop() {
         synchronized (lock) {
             started = false;
             ethereum.removeListener(blockListener);
-            refreshWorkTimer.cancel();
+            if (refreshWorkTimer != null) {
+                refreshWorkTimer.cancel();
+            }
             refreshWorkTimer = null;
```
martinmedina: This line could go inside the if clause

rskj-core/src/test/java/co/rsk/mine/MinerServerTest.java
pmprete: That was It! thanks @martinmedina

PR #883, opened 2019-06-17
Wasabi renaming

aeidelman:
Renaming everywhere "Wasabi" name usage is still pending


collheesi:
We should also update integration tests that manipulate the activation block to use the new name.

CC @rsksmart/rsk-core @rsksmart/rsk-qa @josedahlquist @pmprete @pamgonzalez

PR #885, opened 2019-06-18
Set BridgeSupport instanciation in Bridge initialization

josedahlquist:


rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -290,6 +290,8 @@ public void init(Transaction rskTx, Block rskExecutionBlock, Repository reposito
         this.rskExecutionBlock = rskExecutionBlock;
         this.repository = repository;
         this.logs = logs;
+
+        this.bridgeSupport = setup();
```
collheesi: The same `Factory` pattern we've been recently applying could be used with `BridgeSupport`. Many usages of `PowerMock` could also be changed with mocks of this factory object.

PR #887, opened 2019-06-18
1 preview configs

josedahlquist:


collheesi:
I left a TODO here: https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ConsensusRule.java#L23

I was hoping we would tackle it as part of the PR that created a new testnet.

Also, shouldn't we change the Genesis, the network id or _anything_?

PR #894, opened 2019-06-24
Checkstyle

nacho692:


collheesi:
This looks great, thanks a lot for the hard work! I verified the changes and they look good to me.

However, we need to sort the quality gate issue first. I think setting a new baseline after merging this would be the right approach.

CC @aeidelman @rsksmart/rsk-core

aeidelman:
@collheesi Agree.

collheesi:
To be clear: tomorrow we'll merge this PR (with a failing QG). Then we'll set `sonar.leak.period` to tomorrow's date (June 26 2019), making this version our baseline and turning the QG green again.

It is important to merge this PR first of all so everything works as expected. Per the docs:

> When specifying a number of days or a date, the snapshot selected as the baseline for comparison is the first one available inside the corresponding time range. **Specifically, the first analysis in the range is considered to be before the New Code Period**.

### Settings from Sonarcloud dashboard: 

image

I'll push a rebased version in a few minutes.

SergioDemianLerner:
MASsive Non-human-reviewed CHAnges (MANCHAs) like this only make life easier for an attacker to potentially introduce a backdoor. Even if this one seems harmless,  if we allow this MANCHA to go through, we're setting a bad precedent and open the possibility for future MANCHAs as we reduce our code review awareness levels, lower our guards, and force users to install and run more tools (sometimes in binary form) which we do not have control. 
Also MANCHAs will make harder to track important changes from before/after the change, so when really nasty bugs happen, we'll be unable to track the motives behind past changes. 
A codebase is not just snapshot of a system: it's the history of design choices and human desires to improve it, and this history must be preserved. Thanks to git most of this preservation is automatically performed, but it works as long as we keep the links intact. 
Finally, I haven't heard a single programmer complain on how the formatting of a comment affects code readability or security. On the other side, programmers do complain from awkward algorithmic implementations, which increases the probability bugs to go undetected in reviews.

PR #898, opened 2019-06-28
Get rid of MutableRepository#getSnapshotTo

collheesi:
The plan is to completely remove `Repository` from the context, and use the `RepositoryLocator` instead. This is a step towards that.

Next, we'll try to get rid of `Repository#syncToRoot`.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/NetworkStateExporter.java
+++ b/rskj-core/src/main/java/co/rsk/core/NetworkStateExporter.java
@@ -19,17 +19,19 @@
 package co.rsk.core;
 
 import co.rsk.core.bc.AccountInformationProvider;
+import co.rsk.db.RepositoryLocator;
 import co.rsk.panic.PanicProcessor;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectWriter;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.bouncycastle.util.encoders.Hex;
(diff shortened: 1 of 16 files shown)


rskj-core/src/main/java/co/rsk/core/NetworkStateExporter.java
```diff
@@ -44,18 +46,18 @@
...
 
-    public NetworkStateExporter(Repository repository) {
-        this.repository = repository;
+    public NetworkStateExporter(RepositoryLocator repositoryLocator, Blockchain blockchain) {
+        this.repositoryLocator = repositoryLocator;
+        this.blockchain = blockchain;
     }
 
     public boolean exportStatus(String outputFile) {
-        // This will only work if GlobalKeyMap.enabled = true when building the whole
```
diega: haha good 'ol comment

rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
```diff
@@ -327,18 +327,6 @@ public void syncTo(Trie root) {
         return rootHash.getBytes();
     }
 
-    // What's the difference between startTracking() and getSnapshotTo()?
```
diega: Until we have all the refactor in place, I would not lose the information in this comment. I would probably not keep it here, but maybe we can move the relevant parts into the javadoc of each method (I mean `co.rsk.db.RepositoryLocator#snapshotAt` and `org.ethereum.core.Repository#startTracking`)

collheesi: I don't see anything important in the comment, and also our next refactors (e.g remove `syncToRoot` will make this transparent 😄)

PR #904, opened 2019-07-02
Segregate the Repository reader interface

collheesi:
Introduce `RepositoryReader` and use it everywhere possible (mostly through `RepositoryLocator#snapshotAt`)

### DIFF
--- a/rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
@@ -18,8 +18,6 @@
 
 package co.rsk.rpc.modules.eth;
 
-import static org.ethereum.rpc.TypeConverter.toJsonHex;
-
 import co.rsk.bitcoinj.store.BlockStoreException;
 import co.rsk.config.BridgeConstants;
 import co.rsk.core.ReversibleTransactionExecutor;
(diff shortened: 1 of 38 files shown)


rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -86,7 +87,7 @@ public EthModule(
 
     public Map<String, Object> bridgeState() throws IOException, BlockStoreException {
         Block bestBlock = blockchain.getBestBlock();
-        Repository repository = repositoryLocator.snapshotAt(bestBlock.getHeader()).startTracking();
+        Repository repository = repositoryLocator.startTrackingAt(bestBlock.getHeader());
```
diega: I think that through this refactor an `up-to-grab` contribution may have rise. It makes no sense to need the whole repository for obtaining the `bridgeState`, a _reader_ should be safer and enough

collheesi: Yes, I tried but the `Bridge` components both read and write. That's the ISP principle right there.

collheesi:
@diega thanks, just fixed those issues! For the bridge one I'll let the `fed` guys know and let them add it to their backlog if they desire.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
diega: Absolutely, that's why I think it's a pretty safe place for an external contributor to dig in

collheesi:
Closed in favor of #907 because of the pipeline rejecting PRs from forks.

PR #908, opened 2019-07-03
Remove Repository getter from Blockchain

diega:


diega:
> The PR looks good in general, I would consider keeping the blockGenerator hidden since I haven't found any test where the construction outside from blockBuilder adds any value from the original version. Maybe adding another constructor would be enough.

I've moved, the `co.rsk.blockchain.utils.BlockGenerator` instantiation inside the only existing constructor of `co.rsk.test.builders.BlockBuilder`, thank you for the suggestion

PR #916, opened 2019-07-15
Convert some BlockValidatorTest tests to unit tests

collheesi:


rskj-core/src/test/java/co/rsk/validators/BlockTxsValidationRuleTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @Test
+    public void invalidTransactionsWithSameNonceAndSameSender() {
+        RskAddress sender = TestUtils.randomAddress();
+        initialNonce(sender, 42);
+
+        Block block = block(transaction(sender, 42), transaction(sender, 42));
+
+        Assert.assertFalse(rule.isValid(block, parent));
+    }
+
```
lsebrie: I would add a test where a list of txs are checked.
- (42, 43) should be valid
- (42, 44) should be invalid

rskj-core/src/test/java/co/rsk/validators/BlockParentGasLimitRuleTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @Test
+    public void invalidWhenGasIsOnRightLimit() {
+        whenGasLimitBoundDivisor(10);
+        whenGasLimit(parentHeader, 1000);
+        whenGasLimit(blockHeader, 1101);
+
+        Assert.assertFalse(rule.isValid(block, parent));
+    }
+
```
lsebrie: We should add a check to the rule to avoid accepting  values < 1 for boundDivisor.
We should test that check. We should check that it works with another value than 10

rskj-core/src/test/java/co/rsk/validators/BlockTxsValidationRuleTest.java
collheesi: Both added and passing :)

rskj-core/src/test/java/co/rsk/validators/BlockParentGasLimitRuleTest.java
collheesi: Done, but this is completely out of scope.

PR #920, opened 2019-07-16
Add the possibility to save block headers to the blockstore

nacho692:
Now both headers and blocks can be stored in the same block store. 

-Headers are stored as an RLP list with one element, the element being the encoded header.
-Blocks are stored the same way as before, as an RLP list.

It might be a good the idea to encapsulate the BlockStoreCache and the KeyValueDataSource management from the IndexedBlockStore.

This change is required to allow a future fast synchronization.


collheesi:
@nacho692 could you adjust your formatting settings so it doesn't change all our imports and other non-related stuff?

rskj-core/src/main/java/co/rsk/core/bc/PendingState.java
```diff
@@ -117,7 +117,7 @@ public BigInteger getNonce(RskAddress addr) {
         //Priority heap, and list of transactions are ordered by descending gas price.
         Comparator<Transaction> gasPriceComparator = reverseOrder(Comparator.comparing(Transaction::getGasPrice));
 
-        //First create a map to separate txs by each sender.
+        //First create a either to separate txs by each sender.
```
collheesi: outdated

rskj-core/src/main/java/co/rsk/net/BlockStoreCache.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param hash The look up key, cannot be null.
+     * @return An optional, empty if the header was not found in the cache.
+     */
+    public Optional<BlockHeader> getBlockHeaderByHash(Keccak256 hash) {
+        BlockHeader headerResult = headerMap.get(hash);
+        if (headerResult != null) {
+            return Optional.of(headerResult);
+        }
+
+        Block blockResult = blockMap.get(hash);
```
collheesi: Suggestion:

```java
return getBlockByHash(hash).map(Block::getHeader);
```

rskj-core/src/main/java/org/ethereum/db/BlockStoreEncoder.java
```diff
@@ -0,0 +1,95 @@
+package org.ethereum.db;
```
collheesi: Missed license header

rskj-core/src/main/java/org/ethereum/db/BlockStoreEncoder.java
```diff
@@ -0,0 +1,95 @@
...
+import org.ethereum.util.RLPList;
+
+/**
+ * Encodes and decodes both block headers and blocks allowing both value types to be stored the same block store.
+ */
+public class BlockStoreEncoder {
+
+    private final BlockFactory blockFactory;
+
+    /**
```
collheesi: No need to overdocument things like this constructor

rskj-core/src/main/java/org/ethereum/db/BlockStoreEncoder.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Encodes a header to store in the block store.
+     *
+     * @param blockHeader The block header to encode, cannot be null.
+     * @return The encoded header, never null.
+     */
+    public byte[] encodeBlockHeader(BlockHeader blockHeader) {
+        if (blockHeader == null) {
```
collheesi: Use `Objects.requireNonNull`

rskj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
```diff
@@ -276,37 +304,65 @@ public synchronized Block getChainBlockByNumber(long number){
...
 
-        return blockFactory.decodeBlock(blockRlp);
+    /**
+     * Wraps in an optional the retrieved bytes value from the key value data source.
+     * This method should be used instead of accessing the data source directly.
+     *
+     * @param hash The key to look up the stored values, cannot be null.
+     * @return The wrapped key value response.
+     */
+    private Optional<byte[]> getDSValue(byte[] hash) {
```
collheesi: Receive a `Keccak256` instance instead of a `byte[]`. Always work with high-level concepts when possible.

rskj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
```diff
@@ -276,37 +304,65 @@ public synchronized Block getChainBlockByNumber(long number){
...
+        return Optional.ofNullable(blocks.get(hash));
     }
 
     @Override
     public synchronized Map<Long, List<Sibling>> getSiblingsFromBlockByHash(Keccak256 hash) {
-        return this.remascCache.computeIfAbsent(hash, key -> getSiblingsFromBlock(getBlock(key.getBytes())));
+        return this.remascCache.computeIfAbsent(hash, key -> getSiblingsFromBlock(getBlock(key.getBytes()).get()));
```
collheesi: You should never call `.get` without an existence check first. Are you missing a case here? If you always know the value is present you should document that.

rskj-core/src/main/java/org/ethereum/db/BlockStore.java
```diff
@@ -41,11 +41,18 @@
...
 
     void removeBlock(Block block);
 
+    /**
+     * Retrieves a block header by its hash.
+     *
+     * @param hash The hash of the block to look up, cannot be null.
+     * @return The block header if found, null if not.
+     */
+    BlockHeader getBlockHeaderByHash(byte[] hash);
```
collheesi: New methods such as this one should use the correct API: hashes are represented by `Keccak256` instances.

It's OK to leave old method unchanged for now if that is not related with your PR.

rskj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
```diff
@@ -228,11 +254,13 @@ public synchronized void saveBlock(Block block, BlockDifficulty cummDifficulty,
         blockInfo.setHash(block.getHash().getBytes());
         blockInfo.setMainChain(mainChain);
 
-        if (blocks.get(block.getHash().getBytes()) == null) {
-            blocks.put(block.getHash().getBytes(), block.getEncoded());
+        Optional<byte[]> existingValue = getDSValue(block.getHash().getBytes());
```
collheesi: You should take advantage of chaining and handle errors at the end:

```java
Optional<Block> existingBlock = getDSValue(block.getHash().getBytes()).flatMap(blockStoreEncoder::decodeBlock);
if (!existingBlock.isPresent()) {
    // ...
```

rskj-core/src/main/java/org/ethereum/db/BlockStore.java
collheesi: Also: why return `null` and not an `Optional`?

rskj-core/src/test/java/co/rsk/db/IndexedBlockStoreTest.java
```diff
@@ -0,0 +1,239 @@
...
+import org.ethereum.datasource.KeyValueDataSource;
+import org.ethereum.db.BlockStoreEncoder;
+import org.ethereum.db.IndexedBlockStore;
+import org.ethereum.db.IndexedBlockStore.BlockInfo;
+import org.junit.Before;
+import org.junit.Test;
+import org.mapdb.DB;
+
+
+public class IndexedBlockStoreTest {
```
collheesi: Tests look really good! Finally we start having _unit_ tests. I would suggest leaving out the javadocs and making sure the test names are sufficiently descriptive (which I think they are)

rskj-core/src/main/java/org/ethereum/db/BlockStoreEncoder.java
nacho692: The thing about requireNonNull is that it throws a null pointer exception, I think of a null value in this situation as an IllegalArgument

rskj-core/src/test/java/co/rsk/db/IndexedBlockStoreTest.java
nacho692: I find it nice to understand tests such as "fails with this input" without internal knowledge of the class and cannot find a sufficiently descriptive name for that. I'll remove the ones I find redundant.

rskj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
nacho692: nice catch!

PR #922, opened 2019-07-16
Address issue where eth_call with 'pending' parameter would fail due to relying on an unupdated component

M-Picco:
As of now, said component relies on the miner server to be updated. When the miner server is
disabled the component is not updated on new blocks which causes that when queried it returns
incomplete or inconsistent data.


collheesi:
I think this should go to branch `release_1.0.0`

PR #941, opened 2019-07-30
Support unitrie migration after Wasabi activation

colltoaction:
This is intended to be released in v1.0.2 to support the case where a miner keeps producing Orchid blocks after Wasabi activation.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/db/migration/OrchidToUnitrieMigrator.java
+++ b/rskj-core/src/main/java/co/rsk/db/migration/OrchidToUnitrieMigrator.java
@@ -32,7 +32,9 @@ import org.ethereum.core.Repository;
 import org.ethereum.crypto.Keccak256Helper;
 import org.ethereum.datasource.HashMapDB;
 import org.ethereum.datasource.KeyValueDataSource;
+import org.ethereum.db.BlockStore;
 import org.ethereum.db.ByteArrayWrapper;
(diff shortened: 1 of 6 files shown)


rskj-core/src/main/java/co/rsk/db/migration/OrchidToUnitrieMigrator.java
```diff
@@ -116,10 +118,17 @@ public static void migrateStateToUnitrieIfNeeded(RskContext ctx) throws IOExcept
             return;
         }
 
+        BlockStore blockStore = ctx.getBlockStore();
+        long maximumBlockNumberToMigrate = ctx.getRskSystemProperties().getDatabaseMigrationMaximumHeight();
+        if (maximumBlockNumberToMigrate != 0 && blockStore.getMaxNumber() > maximumBlockNumberToMigrate) {
+            ((IndexedBlockStore) blockStore).rewind(maximumBlockNumberToMigrate);
```
nacho692: If we are keeping this class I would either receive an indexedblockstore or add the rewind functionality to the interface.

diega: The general idea is to remove this whole class in the next or the next+1 version, with that in mind we are not sure if this will be the final interface being used for when we implement the _rewind_ feature, that's why this wasn't promoted to the interface.

PR #956, opened 2019-08-15
Add support for non genesis starting block index

nacho692:
This change allows having an index with an arbitrary sequential interval of blocks.
The BlockInfo class cannot be moved from IndexedBlockstore due to the index encoding. An index  migration is required to move it.

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
+++ b/rskj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
@@ -497,10 +497,7 @@ public class IndexedBlockStore implements BlockStore {
     public void rewind(long blockNumber) {
         long maxNumber = getMaxNumber();
         for (long i = maxNumber; i > blockNumber; i--) {
-            List<BlockInfo> blockInfos = index.removeBlocksByNumber(i);
-            if (blockInfos == null) {
-                continue;
-            }
+            List<BlockInfo> blockInfos = index.removeLast();
 
             for (BlockInfo blockInfo : blockInfos) {
                 this.blocks.delete(blockInfo.getHash().getBytes());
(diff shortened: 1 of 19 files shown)


rskj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
```diff
@@ -127,7 +124,8 @@ public synchronized Block getBestBlock() {
     public byte[] getBlockHashByNumber(long blockNumber, byte[] branchBlockHash) {
         Block branchBlock = getBlockByHash(branchBlockHash);
         if (branchBlock.getNumber() < blockNumber) {
-            throw new IllegalArgumentException("Requested block number > branch hash number: " + blockNumber + " < " + branchBlock.getNumber());
+            throw new IllegalArgumentException("Requested block number > branch hash number: "
```
lsebrie: Add String format to avoid + concatenation please

rskj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
```diff
@@ -50,25 +50,22 @@
 
 public class IndexedBlockStore implements BlockStore {
 
-    private static final Logger logger = LoggerFactory.getLogger("general");
+    private static final Logger logger = LoggerFactory.getLogger(IndexedBlockStore.class);
```
lsebrie: We should improve the logback file for this change

rskj-core/src/main/java/org/ethereum/db/IndexedBlockStore.java
```diff
@@ -195,21 +193,13 @@ public boolean isBlockInMainChain(long blockNumber, Keccak256 blockHash){
     @Override
     public synchronized void flush() {
         Metric metric = profiler.start(Profiler.PROFILING_TYPE.DB_WRITE);
-        
-        //long t1 = System.nanoTime();
-        if (indexDB != null) {
```
lsebrie: Thanks :D

PR #968, opened 2019-08-20
Fix config parsing error messages

janhn:


colltoaction:
First: sorry about the Pipeline check, it's something we need to fix on our end.

With regards to your changes: thanks! The current message is awful, but you got something wrong: when the value is 20-bytes long, then the input must be a 40-character long hexadecimal string. Same with the other value.

Would you be so kind to amend this commit with this in mind? 

Thanks!

janhn:
> With regards to your changes: thanks! The current message is awful, but you got something wrong: when the value is 20-bytes long, then the input must be a 40-character long hexadecimal string. Same with the other value.

I am sorry, I don't follow. The change is deliberately so that the error message is not misleading about the length of the input. I.e. a configuration with 32-byte private key gives:
```
Exception in thread "main" java.lang.RuntimeException: The peer.privateKey needs to be Hex encoded and 32 byte length
```
until I checked the sources and realized that the input has to be 64-bytes, not 32.
What am I missing?

colltoaction:
32 bytes are encoded as 64 hex characters. The length **of the string** is 64, but the key is actually 32-bytes long.

janhn:
> 32 bytes are encoded as 64 hex characters. The length **of the string** is 64, but the key is actually 32-bytes long.

When giving an error message to a user that a string he provided has incorrect length, it should include the value that he can actually understand (64 char string), not some internal representation, right? (Are we talking Java internal string representation here?)

colltoaction:
Let's start again. 

1. The first message should point out that it is an invalid address. People configuring a miner MUST ABSOLUTELY KNOW what an address is, so the message could be something like `MINER_REWARD_ADDRESS_CONFIG + " is not a valid address"`.
2. The other is a little different, because there is no standard to specify a key. I think something like this would be good enough: `"The peer.privateKey must be 32-bytes long encoded in hexadecimal (two characters per byte). E.g: '010203...303132'"`

PR #969, opened 2019-08-21
Fee per kb validation

guidohernan93:
Added validation to check feePerKb is positive and unit tests for it

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -74,6 +74,7 @@ public class BridgeSupport {
     public static final Integer LOCK_WHITELIST_UNKNOWN_ERROR_CODE = 0;
     public static final Integer LOCK_WHITELIST_SUCCESS_CODE = 1;
     public static final Integer FEE_PER_KB_GENERIC_ERROR_CODE = -10;
+    public static final Integer NEGATIVE_FEE_PER_KB_ERROR_CODE = -1;
 
     public static final Integer BTC_TRANSACTION_CONFIRMATION_INEXISTENT_BLOCK_HASH_ERROR_CODE = -1;
     public static final Integer BTC_TRANSACTION_CONFIRMATION_BLOCK_NOT_IN_BEST_CHAIN_ERROR_CODE = -2;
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -74,6 +74,7 @@
     public static final Integer LOCK_WHITELIST_UNKNOWN_ERROR_CODE = 0;
     public static final Integer LOCK_WHITELIST_SUCCESS_CODE = 1;
     public static final Integer FEE_PER_KB_GENERIC_ERROR_CODE = -10;
+    public static final Integer NEGATIVE_FEE_PER_KB_ERROR_CODE = -10;
```
diega: Is it ok to use an existing error code?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1837,6 +1838,10 @@ public Integer voteFeePerKbChange(Transaction tx, Coin feePerKb) {
             return FEE_PER_KB_GENERIC_ERROR_CODE;
         }
 
+        if(!feePerKb.isPositive()){
```
diega: Shouldn't this be programmed as a network upgrade?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
guidohernan93: I thought I could use FEE_PER_KB_GENERIC_ERROR_CODE, but I was no sure. If it is right to use that, I can change it.

pmprete: i think its clearer if its a different error code

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
pmprete: Guido checked on mainnet and testnet and this function was never called with a negative number, this is not a public function. So we are ok going with this without a network upgrade

SergioDemianLerner: If isPositive does what it seems to do, then this line is preventing setting zero fees, which AFAIK it could be useful.

PR #934, opened 2019-07-19
Extend existing eth_getBlock methods to include merge mining fields

M-Picco:
This PR aims to include the merge mining related fields to the responses from rpc methods such as `eth_getBlockByHash` or `eth_getBlockByNumber`.

Said fields are:
 * `bitcoinMergeMiningHeader`: the bitcoin header which serves as proof of work for the rsk block
 * `bitcoinMergedMiningCoinbaseTransaction`: the bitcoin block coinbase transaction which links the btc block with the rsk block
 * `bitcoinMergedMiningMerkleProof`: the bitcoin block coinbase's SPV
 * `hashForMergedMining`: the rsk mining hash used to link the btc and rsk blocks through the coinbase

For more information about these fields refer to the [wiki](https://github.com/rsksmart/rskj/wiki/Merged-Mining)

PR #974, opened 2019-08-22
Added missing information to JsonRpc Block Result

pmprete:
The missing fields are: paidFees, bitcoinMergedMiningHeader, bitcoinMergedMiningMerkleProof and bitcoinMergedMiningCoinbaseTransaction
The extra information is needed to calculate the encoded block and verify that the hash of the block is correct.

Also we extended the web3 interface adding eth_getEncodedBlockByNumber and eth_getEncodedBlockByHash to directly obtain the encoded value without needing to encode it from the json response.
Added corresponding unit test


pmprete:
@diega  pointed out that the missing merge minning fields are added in this PR https://github.com/rsksmart/rskj/pull/934
I'll close the PR and until 934 is merged

PR #976, opened 2019-08-22
Rework internal services startup

colltoaction:
The node has many services that run for the whole duration of the program. We currently handle them in `FullNodeRunner`, but this is repetitive and we lose some possibilities, like only instantiating certain objects when the configuration requires it.

This PR reworks the startup logic with the help of the `InternalService` interface.

🤖, please run with *fed:services_startup*


smcsicardi:
Great work! This would be very useful for the light client implementation, in order to initialize only the specific classes needed

PR #990, opened 2019-08-27
Gas exactimation

SergioDemianLerner:
This is a complete gas exactimation for RSK without the EIP150  1/64th CALL pre-deposit. It doesn't require binary search. A simple computation test was added. 
The exactimation works as long as the code called is not malicious (do not use the GAS opcode to perform arbitrary actions based on its value)

This PR still needs some real-world test, however this exactimation CANNOT be worse than the previous implementation, because this one always returns a higher value than the previous one.


lsebrie:
Pipeline: run

Vovchyk:
@SergioDemianLerner this is already in master, isn't it? Done in scope of #1548. Can we close this PR?

Issue #901, opened 2019-07-01
Implement state garbage collector


collheesi:
RSKIP: https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP64.md.

### Sprint 100

- [x] implement POC: #897
- [ ] define next sprint tasks

### Tentative sprint 101 tasks

- [ ] create an abstraction with control over the flush behavior, and decouple it from `Blockchain`. It could be hooked to `onBlock` since this is the current behavior.
- [ ] decouple `Trie#save` so it can be used independently (or `TrieStore` if possible)

### Tentative backlog

- [ ] define behavior in case of a failed IO operation (e.g can't move the database paths around)
- [ ] flush databases in background
- [ ] handle missing state information gracefully (e.g `ExecutionBlockRetriever` for `web3`)
- [ ] review flushing interdependencies: keep states and blocks flushing in sync
- [ ] define the mapping between GC generations and disk storage layout (move folders vs keep track of the current generation in another database vs use the block number to keep track)

PR #996, opened 2019-08-29
Hook up BlockChainFlusher to onBestBlock event

colltoaction:
This is in preparation for the Garbage Collector (#901).


rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -715,6 +715,14 @@ protected NodeRunner buildNodeRunner() {
                 internalServices.add(getMinerClient());
             }
         }
+        if (getRskSystemProperties().isFlushEnabled()) {
```
diega: Have you evaluated to remove this flag? it seems to me that the full node is almost useless if it doesn't ever flush

PR #997, opened 2019-08-30
Import bootstrap data from external source

lsebrie:
First steps to download a bootstrap file to accelerate synchronization and have a faster operative node


rskj-core/src/main/java/co/rsk/db/importer/StateImporter.java
```diff
@@ -0,0 +1,79 @@
...
+    private void updateBlockStore(BlockStore bootstrapBlockStore, Block bestBlock) {
+        Block current = bestBlock;
+        while (current != null) {
+            blockStore.saveBlock(current, bootstrapBlockStore.getTotalDifficultyForHash(current.getHash().getBytes()), true);
+            current = bootstrapBlockStore.getBlockByHash(current.getParentHash().getBytes());
+        }
+    }
+
+    private void updateTrieStore(TrieStore bootstrapTrieStore, Block bestBlock) {
+        Trie node = bootstrapTrieStore.retrieve(bestBlock.getStateRoot());
```
colltoaction: You must take `StateRootHandler` into account

rskj-core/src/main/java/co/rsk/db/importer/StateImporter.java
```diff
@@ -0,0 +1,79 @@
...
+        BootstrapDataProvider bootstrapDataProvider = new BootstrapDataProvider(databaseDir, bootstrapURL, ctx);
+        importer.updateDatabaseIfNeeded(bootstrapDataProvider.getBootstrapBlockstore(), bootstrapDataProvider.getBootstrapTrieStore());
+    }
+
+    private static URL getBootstrapURL(RskContext ctx) {
+        String url = ctx.getRskSystemProperties().databaseImportUrl();
+        try {
+            return new URL(url);
+        } catch (MalformedURLException e) {
+            throw new IllegalArgumentException(e);
```
colltoaction: what's the plan for this exception? it's not caught as far as I can see

rskj-core/src/main/java/co/rsk/db/importer/StateImporter.java
lsebrie: Importing data from a previous height than wasabi won't be allowed. Thanks for appointing this :+1:

rskj-core/src/main/java/co/rsk/db/importer/StateImporter.java
lsebrie: It won't be catch. If the configuration is wrong the program will end. I'll improve the message for this :+1:

rskj-core/src/main/java/co/rsk/db/importer/provider/BootstrapFileHandler.java
```diff
@@ -0,0 +1,122 @@
...
+        Path zipPath = tempPath.resolve(BOOTSTRAP_DATA_NAME);
+
+        List<String> publicKeys = new ArrayList<>(selectedEntries.keySet());
+        String publicKey = publicKeys.get(new SecureRandom().nextInt(publicKeys.size()));
+        String expectedHash = selectedEntries.get(publicKey).getHash();
+
+        String dbSuffix = selectedEntries.get(publicKey).getDb();
+        URL dbURL = bootstrapUrlProvider.getFullURL(dbSuffix);
+
+        downloadFile(zipPath, dbURL);
```
nacho692: Given that the file is currently being loaded entirely on memory for decoding. Instead of storing the zip file on disk we could have it only on ram.
Or storing it in disk and in future iterations decode it linearly to avoid loading the full file on ram.

Just food for thought.

rskj-core/src/main/java/co/rsk/db/importer/provider/Unzipper.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+public class Unzipper {
+
+    public void unzip(InputStream sourceZipStream, Path destinationPath) throws IOException {
```
nacho692: I'd document exactly what unzip does. What exactly is the input? Is it a zip file? a path with zips?

rskj-core/src/main/java/co/rsk/db/importer/provider/index/BootstrapIndexRetriever.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class BootstrapIndexRetriever {
+
+    private static final String INDEX_NAME = "index.json";
+
+    private final List<String> publicKeys;
+    private final BootstrapURLProvider bootstrapUrlProvider;
+    private ObjectMapper objectMapper;
```
nacho692: make final please

rskj-core/src/main/java/co/rsk/db/importer/provider/index/BootstrapIndexMCH.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private final List<String> publicKeys;
+    private final int minimumAcceptableSources;
+
+    public BootstrapIndexMCH(List<String> publicKeys, int minimumAcceptableSources) {
+        this.publicKeys = new ArrayList<>(publicKeys);
+        this.minimumAcceptableSources = minimumAcceptableSources;
+    }
+
+    public MaximumCommonHeightData getMaximumCommonHeightData(List<BootstrapDataIndex> indexes) {
```
nacho692: I'd add some documentation as it took me some time to understand the state. I find the Map of Maps hard to follow.

Maybe this alternative implementation?
```
//Group by height
Map<Long, Long> entriesPerHeight = entries.stream().collect(
        Collectors.groupingBy(
                e -> e.height, Collectors.counting())
        );

//Maximum common height
int minimumAcceptableNumber = 2;
long maximumCommonHeight = entriesPerHeight.entrySet().stream()
        .filter(e -> e.getValue() >= minimumAcceptableNumber)
        .map(Map.Entry::getKey).max(Long::compare).orElseThrow(
        () -> new IllegalStateException("Downloaded files doesn't contain enough entries for a common height")
);

//Getting the maximum common height entries
Set<BootstrapDataEntry> maximumCommonEntries =
        entries.stream().filter(e -> e.height == maximumCommonHeight)
                .collect(Collectors.toSet());
```

rskj-core/src/main/java/co/rsk/net/BlockCache.java
```diff
@@ -46,4 +46,8 @@ public void addBlock(Block block) {
     public Block getBlockByHash(byte[] hash) {
         return blockMap.get(new Keccak256(hash));
     }
+
+    public void clear() {
```
nacho692: This is not being used i think

rskj-core/src/main/java/co/rsk/net/sync/DownloadingBackwardsBodiesSyncState.java
```diff
@@ -98,6 +102,8 @@ public void newBody(BodyResponseMessage body, MessageChannel peer) {
 
         if (child.getNumber() == 1) {
             connectGenesis(child);
+            blockStore.flush();
```
nacho692: Nice fix, i'd consider moving it to another commit

rskj-core/src/main/java/co/rsk/net/BlockCache.java
lsebrie: This method was called but not it's caller, so I'l remove all the chain of methods used in the `BlockStore::clear` interface

rskj-core/src/main/java/co/rsk/db/importer/BootstrapImporter.java
```diff
@@ -0,0 +1,107 @@
...
+        });
+        valuesData.forEach(i -> {
+            byte[] rlpData = i.getRLPData();
+            hashMapDB.put(Keccak256Helper.keccak256(rlpData), rlpData);
+        });
+
+        nodes.forEach(destinationTrieStore::save);
+    }
+
+    private static void insertBlocks(BlockStore blockStore,
```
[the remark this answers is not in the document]
lsebrie: Good catch, it would work after a couple of blocks are downloaded but it's better to don't rely on it

rskj-core/src/main/java/co/rsk/db/importer/BootstrapImporter.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private void updateDatabaseIfNeeded(long selectedHeight) {
+        if (!blockStore.isEmpty() &&  blockStore.getMaxNumber() >= selectedHeight) {
+            return;
+        }
+
+        byte[] encodedData = bootstrapDataProvider.getBootstrapData();
+        RLPList rlpElements = RLP.decodeList(encodedData);
+        insertBlocks(blockStore, blockFactory, rlpElements.get(0));
+        insertState(trieStore, rlpElements.get(1));
```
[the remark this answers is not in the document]
lsebrie: In the current implementation whenever you start an import it will reset db. I'll remove block numbers check that don't make sense anymore

rskj-core/src/main/java/co/rsk/db/importer/provider/BootstrapDataProvider.java
```diff
@@ -0,0 +1,97 @@
...
+            int minimumAcceptableSources) {
+
+        this.bootstrapIndexMCH = bootstrapIndexMCH;
+        this.bootstrapDataVerifier = bootstrapDataVerifier;
+        this.bootstrapFileHandler = bootstrapFileHandler;
+        this.bootstrapIndexRetriever = bootstrapIndexRetriever;
+        this.minimumAcceptableSources = minimumAcceptableSources;
+    }
+
+    public void retrieveData() {
```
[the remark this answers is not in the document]
lsebrie: I don't like it too, but for sake of testeability I chose this path :pray:

rskj-core/src/main/java/co/rsk/net/sync/DownloadingBackwardsBodiesSyncState.java
lsebrie: I'll move the changes on this class to another commit :+1:

rskj-core/src/main/java/co/rsk/db/importer/provider/BootstrapFileHandler.java
```diff
@@ -0,0 +1,122 @@
...
+    public BootstrapFileHandler(
+            BootstrapURLProvider bootstrapUrlProvider,
+            Unzipper unzipper) {
+        this.bootstrapUrlProvider = bootstrapUrlProvider;
+        this.unzipper = unzipper;
+    }
+
+    // for testing purposes these 3 methods are splitted into 3 public items
+    public void setTempDirectory() {
+        this.tempPath = getTempDirectory();
```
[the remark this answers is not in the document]
lsebrie: I tried to avoid since it would create a temp directory on every start of the node even when the import feature isn't running

rskj-core/src/main/java/co/rsk/db/importer/provider/index/BootstrapIndexMCH.java
lsebrie: I like the idea about collecting by height but it's important to validate that there aren't more than one entry for the same height and the same source (public key). I'll add some documentation to explain how data is organized.

rskj-core/src/main/java/co/rsk/db/importer/provider/BootstrapDataProvider.java
```diff
@@ -0,0 +1,97 @@
...
+            BootstrapFileHandler bootstrapFileHandler,
+            BootstrapIndexMCH bootstrapIndexMCH,
+            BootstrapIndexRetriever bootstrapIndexRetriever,
+            int minimumAcceptableSources) {
+
+        this.bootstrapIndexMCH = bootstrapIndexMCH;
+        this.bootstrapDataVerifier = bootstrapDataVerifier;
+        this.bootstrapFileHandler = bootstrapFileHandler;
+        this.bootstrapIndexRetriever = bootstrapIndexRetriever;
+        this.minimumAcceptableSources = minimumAcceptableSources;
```
joaquinlpereyra-iov: I think it would be interesting for `minimumAccetpableSources >= 3` to disallow misconfigurations which could lead to a corrupted database. 

Example scenario, not even an attack: a user messes with settings (not an impossible scenario for a settings file) and for some reason leaves only one public key on it. The BoostrapdataProvider would continue to function, but it would stop being a multisig scheme quite silently.

lsebrie: I'm not a fan of stepping over a user configuration value, the amount of public keys configured will determine this value. Nevertheless I can add a validation for this number with a warning in log messages whenever the value is lower than 2 (the needed amount of sources when there are 2 or more public keys used).

lsebrie:
@joaquinlpereyra-iov nice catch I'll be fixing this with a better filter when finding a height candidate.
Every entry that enters to `BootstrapDataVerifier` should have the same hash

rskj-core/src/main/java/co/rsk/db/importer/provider/BootstrapDataVerifier.java
```diff
@@ -0,0 +1,66 @@
...
+                throw new BootstrapImportException(String.format(
+                        "Error trying to verify different hashes: %s vs %s", verifiedHash, currentHash));
+            }
+            BootstrapDataSignature bds = bde.getSig();
+
+            // to use the public key we need to have an extra byte according to x9.62 declaring
+            // which format is using. The current format from signer is uncompressed
+            byte[] publicKey = Hex.decode(entry.getKey());
+
+            byte[] dbHash = Hex.decode(currentHash);
```
joaquinlpereyra-iov: We're still checking the signature against each of the hashes. The `if` condition catches the mistake, but a simpler and less prone to a `i refactored to delete an unnecessary if` is to just delete the `currentHash` variable and always compare the signature against the `verifiedHash`.

rskj-core/src/main/java/co/rsk/db/importer/provider/BootstrapDataVerifier.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Map;
+
+public class BootstrapDataVerifier {
+
+    public int verifyEntries(Map<String, BootstrapDataEntry> selectedEntries) {
+        int verifications = 0;
+        if (selectedEntries.isEmpty()) {
+            return 0;
+        }
+        String verifiedHash = selectedEntries.values().iterator().next().getHash();
```
joaquinlpereyra-iov: `verifiedHash` is quite the name. The hash still has not been verified by anyone, it is just `THE hash`. Like my other comment, a think a cleaner solution is just have `hash` and check the signatures against those. The if condition is nice, but should not be the main defense :)

rskj-core/src/main/java/co/rsk/db/importer/provider/index/BootstrapIndexCandidateSelector.java
```diff
@@ -0,0 +1,125 @@
...
+                continue;
+            }
+
+            Map<String, BootstrapDataEntry> entries = entry.getValue();
+            Map<String, Long> hashGroups = entries.values().stream()
+                    .collect(Collectors.groupingBy(BootstrapDataEntry::getHash, Collectors.counting()));
+            Optional<Map.Entry<String, Long>> bestHash = hashGroups.entrySet().stream()
+                    .max(Comparator.comparing(Map.Entry::getValue));
+
+            if (bestHash.isPresent() && bestHash.get().getValue() >= minimumRequiredSources) {
```
joaquinlpereyra-iov: Which condition should trigger the `bestHash.get().getValue > minimunRequiredSources`? I think we should change this to an equals and present a warning and skip the result if it is bigger than... 

I think bigger than means:
1. exporter ran twice on the same block
2. the signer did not care and signed something which was already in the index

those two things should not happen. if they do, something is broken.

rskj-core/src/main/java/co/rsk/db/importer/provider/BootstrapDataVerifier.java
lsebrie: Thx for the comment I improved this

rskj-core/src/main/java/co/rsk/db/importer/provider/BootstrapDataVerifier.java
lsebrie: Now I check only the starting hash, I am avoiding hex decode for each hash

rskj-core/src/main/java/co/rsk/db/importer/provider/BootstrapFileHandler.java
```diff
@@ -0,0 +1,129 @@
...
+    private Path getTempDirectory() {
+        try {
+            return Files.createTempDirectory("import");
+        } catch (IOException e) {
+            throw new BootstrapImportException(
+                    "Failed to create a temporary directory. Please start again the import process", e);
+        }
+    }
+
+    private void downloadFile(Path downloadPath, URL url) {
```
joaquinlpereyra-iov: we can implement a timeout of some kind here, taking advantage of the steam option. assuming someone has push access to s3, they could create arbitrary random signatures for non-existent data, and we would be able to check it just when the file has been fully downloaded. this could lead to a DoS. 

when we change to a more permission-less structure and ditch the authentication-to-upload part, we should look into better mitigations for this.

PR #998, opened 2019-09-02
State garbage collector v1

colltoaction:
This is a v1 implementation of the state garbage collection feature. The full definition of the expected feature is in [RSKIP64](https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP64.md). This v1 doesn't minimize the nodes stored in each epoch but when we decide to implement that it'll be retrocompatible with this implementation at storage level.

It's disabled by default, you have to opt-in through the `blockchain.gc.enabled` configuration flag.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/GarbageCollector.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/GarbageCollector.java
@@ -1,0 +1,91 @@
+/*
+ * This file is part of RskJ
+ * Copyright (C) 2019 RSK Labs Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package co.rsk.core.bc;
+
+import co.rsk.config.InternalService;
+import co.rsk.db.RepositoryLocator;
+import co.rsk.trie.MultiTrieStore;
+import org.ethereum.core.Block;
+import org.ethereum.core.BlockHeader;
+import org.ethereum.core.TransactionReceipt;
+import org.ethereum.db.BlockStore;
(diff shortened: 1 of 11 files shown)


rskj-core/src/main/java/co/rsk/trie/MultiTrieStore.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * @param forceSaveRoot allows saving the root node even if it's embeddable
+     */
+    private void save(Trie trie, boolean forceSaveRoot) {
+        // Invariant: a trie saved in epoch N is fully contained in d(N) and d(N - 1).
+        // This ensures all epoch N tries survive after discarding d(N - 2).
+        if (epochNSavedTries.contains(trie) || epochN1SavedTries.contains(trie)) {
+            // TODO(mc): can we return? could its children be in d(N - 2)?
```
nacho692: I think they cannot. d(N-1) should cover the d(N-2) nodes

rskj-core/src/main/java/co/rsk/trie/MultiTrieStore.java
```diff
@@ -0,0 +1,196 @@
...
+    @Override
+    public synchronized void collect(Trie lastFrontierTrie, long lastEpoch) {
+        // TODO(mc) could it be greater?
+        if (lastEpoch < currentEpoch) {
+            logger.warn("Skipping collection with epoch {} lower than current epoch {}", lastEpoch, currentEpoch);
+            return;
+        }
+
+        // TODO(mc) is this necessary with the save invariant?
+        ensureTrieForCollection(lastFrontierTrie, true);
```
nacho692: I think it is needed. d(N) and d(N-1) contain the full current state, a merge is required to generate the new d(N-1) storage.

rskj-core/src/main/java/co/rsk/core/bc/GarbageCollector.java
```diff
@@ -0,0 +1,91 @@
...
+        /**
+         * It'll collect all states older than <code>blocksPerEpoch * numberOfEpochs - 1</code> from
+         * the store (if there are enough blocks)
+         */
+        @Override
+        public void onBestBlock(Block block, List<TransactionReceipt> receipts) {
+            int statesToKeep = blocksPerEpoch * numberOfEpochs - 1;
+            long currentBlockNumber = block.getNumber();
+            if(currentBlockNumber % blocksPerEpoch == 0 && currentBlockNumber >= statesToKeep) {
+                collect(currentBlockNumber - statesToKeep);
```
nacho692: I find it strange that this logic is deciding statesToKeep and it has to match MultiTrieStore collect's save into the upcoming last epoch

diega: I'm not quite sure about this. I think passing the latest accesible hash after the `collect` takes place is independent on where that trie is stored

PR #1000, opened 2019-09-03
Get raw block header via JSON RPC

pmprete:
We needed to create json rpc methods to get the rlp encoded information needed to calculate the hash of the block, hash of the transaction receipt, and transaction receipt root.

This methods are:
rsk_getRawBlockHeaderByHash(String blockHash)
rsk_getRawBlockHeaderByNumber(String bnOrId)
rsk_getRawTransactionReceiptByHash(String transactionHash)
rsk_getTransactionReceiptNodesByHash(String blockHash, String transactionHash)

The extra information is needed to calculate the Merkle Tree and verify the Receipt Root
Created RskModule to add this new features from Web3 as they are not from ethereum
:robot: *fed:add-rsk-module*

### DIFF
--- a/rskj-core/src/main/java/co/rsk/trie/Trie.java
+++ b/rskj-core/src/main/java/co/rsk/trie/Trie.java
@@ -1370,4 +1370,56 @@ public class Trie {
             return TrieKeySlice.fromEncoded(encodedKey, 0, lshared, lencoded);
         }
     }
+
+    // Additional auxiliary methods for Merkle Proof
+
+    @Nullable
+    public List<Trie> getNodes(byte[] key) {
+        return findNodes(key);
+    }
+
+    @Nullable
+    public List<Trie> getNodes(String key) {
+        return this.getNodes(key.getBytes(StandardCharsets.UTF_8));
+    }
+
+    @Nullable
+    private List<Trie> findNodes(byte[] key) {
+        return findNodes(TrieKeySlice.fromKey(key));
+    }
+
+    @Nullable
+    private List<Trie> findNodes(TrieKeySlice key) {
+        if (sharedPath.length() > key.length()) {
+            return null;
+        }
+
+        int commonPathLength = key.commonPath(sharedPath).length();
+
+        if (commonPathLength < sharedPath.length()) {
+            return null;
+        }
+
+        if (commonPathLength == key.length()) {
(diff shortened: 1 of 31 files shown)


rskj-core/src/main/java/co/rsk/trie/Trie.java
```diff
@@ -1370,4 +1370,56 @@ private static TrieKeySlice deserialize(ByteBuffer message, boolean sharedPrefix
             return TrieKeySlice.fromEncoded(encodedKey, 0, lshared, lencoded);
         }
     }
+
+    // Additional auxiliary methods for Merkle Proof
+
+    @Nullable
+    public List<Trie> getNodes(byte[] key) {
```
diega: I think this whole list can be acquired through the use of the currently existing _iterators_. If that's the case, I would prefer to not add these methods to this class

rskj-core/src/main/java/co/rsk/rpc/Web3EthModule.java
```diff
@@ -105,6 +105,10 @@ default String eth_sendTransaction(Web3.CallArguments args) {
 
     BlockResultDTO eth_getBlockByNumber(String bnOrId, Boolean fullTransactionObjects) throws Exception;
 
+    String eth_getRawBlockHeaderByHash(String blockHash) throws Exception;
```
diega: I think all this shouldn't go under the `eth_` namespace as it may imply these are part of the standard. What about creating a new module for handling `ext_`?

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -303,6 +303,11 @@ public Keccak256 getHash() {
         return this.getEncoded(true, true);
     }
 
+    public byte[] getEncodedForBlockHash() {
```
diega: I think it's better to rename the current `getEncoded` as something like `getFullEncoded`, and then make this method to be the proper `getEncoded` one. This change should be done as a separate commit at the beginning of this PR

rskj-core/src/main/java/co/rsk/trie/Trie.java
diega: On a second thought, it's not easy to implement using the existing _iterators_

PR #1003, opened 2019-09-05
network upgrade wasabi 110

julianlen:


diega:
For any not yet planned to release RSKIP, please set it to `-1` in the `reference.conf`. When we decide what to include in a future network upgrade we'll make a PR like this but changing all the selected RSKIPs with `-1` to the name of the network upgrade

colltoaction:
E.g in your next PR you should leave `hardforkActivationHeights` unchanged and only add a new value to `consensusRules` like:

```
blockchain.config {
  consensusRules {
    ...
    rskip999 = -1
  }
}
```

PR #1019, opened 2019-10-01
Add Bridge locking cap - RSKIP-134

josedahlquist:
New Bridge methods:
- getLockingCap. Returns a long value representing the current locking cap
- increaseLockingCap. Increases the current locking cap. Authorized method. Only allows increasing the locking cap.
BridgeSupport refactor. BridgeSupport now receives an ActivationConfig.ForBlock instance to be used to detect network upgrades activations.

New logic in 2wp to reject locks that makes the Federation BTC balance bigger than the lockingCap configured in the Bridge.

Read the RSKIP [here](https://github.com/rsksmart/RSKIPs/blob/locking-cap/IPs/RSKIP134.md)

_fit: add-locking-cap-tests-finished_


rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -1033,6 +1038,24 @@ public long getFeePerKb(Object[] args)
         return bridgeSupport.getFeePerKb().getValue();
     }
 
+    public long getLockingCap(Object[] args) {
+        logger.trace("getLockingCap");
+
+        return bridgeSupport.getLockingCap().value;
```
pamgonzalez: Could you use getValue() instead of value?

rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
```diff
@@ -439,6 +439,28 @@
             (BridgeMethodExecutorTyped) Bridge::getStateForDebugging,
             true
     ),
+    GET_LOCKING_CAP(
+            CallTransaction.Function.fromSignature(
+                    "getLockingCap",
+                    new String[]{},
+                    new String[]{"int256"}
```
joaquinlpereyra-iov: Why is the argument to `INCREASE_LOCKING_CAP` an `int256` and not an `uint256`? it makes more sense to me :thinking:

rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -1033,6 +1038,24 @@ public long getFeePerKb(Object[] args)
...
+    public long getLockingCap(Object[] args) {
+        logger.trace("getLockingCap");
+
+        return bridgeSupport.getLockingCap().getValue();
+    }
+
+    public boolean increaseLockingCap(Object[] args) {
+        logger.trace("increaseLockingCap");
+
+        Coin newLockingCap = BridgeUtils.getCoinFromBigInteger((BigInteger)args[0]);
```
joaquinlpereyra-iov: It would be nice to see a test where we check that the overflow check is happening, in case by some of those laws which seem to govern software, the condition is forgotten, removed or changed by someone in the future :)

rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
joaquinlpereyra-iov: There's clearly an old bug in ethereumj which answers my question: https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/org/ethereum/solidity/SolidityType.java#L67

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -205,6 +206,17 @@ public static boolean isFromFederateMember(org.ethereum.core.Transaction rskTx,
...
     }
 
+    public static Coin getCoinFromBigInteger(BigInteger value) throws BridgeIllegalArgumentException {
+        if (value == null) {
+            throw new BridgeIllegalArgumentException("value cannot be null");
+        }
+        try {
+            return Coin.valueOf(value.longValueExact());
+        } catch(ArithmeticException e) {
+            throw new BridgeIllegalArgumentException(e.getMessage());
```
lsebrie: Why you pass only the message and leave the previous stack out of the exception?

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -1560,6 +1561,73 @@ public void getFeePerKbElection_withVotes() {
...
+
+        BridgeStorageProvider provider = new BridgeStorageProvider(
+                track, PrecompiledContracts.BRIDGE_ADDR,
+                config.getNetworkConstants().getBridgeConstants(), bridgeStorageConfiguration
+        );
+
+        // And then we get it back
+        assertThat(provider.getLockingCap(), is(expectedCoin));
+    }
+
```
lsebrie: Maybe you could add a test for getLockingCap after and before network upgrade activation without calling setLockingCap. I was lookging for a test where getInitialLockingCap was used

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1911,6 +1878,42 @@ public Integer setLockWhitelistDisableBlockDelay(Transaction tx, BigInteger disa
...
+            return false;
+        }
+        // new locking cap must be bigger than current locking cap
+        Coin currentLockingCap = this.getLockingCap();
+        if (newCap.compareTo(currentLockingCap) < 0) {
+            logger.warn("attempted value doesn't increase locking cap. Attempted: " + newCap.value);
+            return false;
+        }
+        Coin maxLockingCap = currentLockingCap.multiply(bridgeConstants.getLockingCapIncrementsMultiplier());
+        if (newCap.compareTo(maxLockingCap) > 0) {
```
joaquinlpereyra-iov: good to see this recommendation adopted :)

rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -1033,6 +1038,31 @@ public long getFeePerKb(Object[] args)
         return bridgeSupport.getFeePerKb().getValue();
     }
 
+    public long getLockingCap(Object[] args) {
```
SergioDemianLerner: Something is wrong. 
The code below suggests the value returned is in BTC (no satoshis)
Then the method should be called getLockingCapInBTC()

rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -1033,6 +1038,31 @@ public long getFeePerKb(Object[] args)
...
+        Coin lockingCap = bridgeSupport.getLockingCap();
+
+        // Parse value to BTC
+        return lockingCap.divide(Coin.COIN.getValue()).getValue();
+    }
+
+    public boolean increaseLockingCap(Object[] args) {
+        logger.trace("increaseLockingCap");
+
+        Coin newLockingCapInBTC = BridgeUtils.getCoinFromBigInteger((BigInteger) args[0]);
```
SergioDemianLerner: This is wrong. The Coin class stores the value in satoshis. You can't use Coin to store a value in BTC because the whole point of Coin is to protect from overflows and provide a standarized unit of account.

rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -1033,6 +1038,31 @@ public long getFeePerKb(Object[] args)
...
+        if (newLockingCapInBTC.getValue() <= 0) {
+            throw new BridgeIllegalArgumentException("Locking cap must be bigger than zero");
+        }
+        if (newLockingCapInBTC.compareTo(Coin.valueOf(21_000_000L)) > 0 ) {
```
SergioDemianLerner: again, mixining satoshis/BTC is not good

rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -1033,6 +1038,31 @@ public long getFeePerKb(Object[] args)
...
+    public long getLockingCap(Object[] args) {
+        logger.trace("getLockingCap");
+
+        Coin lockingCap = bridgeSupport.getLockingCap();
+
+        // Parse value to BTC
+        return lockingCap.divide(Coin.COIN.getValue()).getValue();
+    }
+
+    public boolean increaseLockingCap(Object[] args) {
```
SergioDemianLerner: Method should be increaseLockingCapInBTC()

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -2002,5 +2005,97 @@ private void ensureBtcBlockStore() throws IOException, BlockStoreException {
...
+        if (!activations.isActive(ConsensusRule.RSKIP134)) {
+            return true;
+        }
+
+        Coin fedCurrentFunds = getFederationCurrentFunds();
+        Coin lockingCap = this.getLockingCap();
+        logger.trace("Evaluating locking cap for: Sender {}. Value to lock {}. Current funds {}. Current locking cap {}", senderBtcAddress, totalAmount, fedCurrentFunds, lockingCap);
+        Coin fedUTXOsAfterThisLock = fedCurrentFunds.add(totalAmount);
+        // If the federation funds (including this new UTXO) are smaller than or equals to the current locking cap, we are fine.
+        if (fedUTXOsAfterThisLock.compareTo(lockingCap) <= 0) {
```
SergioDemianLerner: Here we're not accepting that the max locking cap is reached. Why?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -2002,5 +2005,97 @@ private void ensureBtcBlockStore() throws IOException, BlockStoreException {
...
+            panicProcessor.panic("locking-cap-exceeded-return-funds", String.format("locking cap exceeded! money return tx build for btc tx %s error. Return was to %s. Tx %s. Value %s", btcTx.getHash(), senderBtcAddress, rskTx, totalAmount));
+        }
+        return false;
+    }
+
+    private Optional<ReleaseTransactionBuilder.BuildResult> getRefundingTransaction(BtcTransaction btcTx, Address senderBtcAddress) throws IOException {
+
+        // Build the list of UTXOs in the BTC transaction sent to either the active
+        // or retiring federation
+        List<UTXO> utxosToUs = btcTx.getWalletOutputs(getNoSpendWalletForLiveFederations()).stream()
```
SergioDemianLerner: This is not our fault, but "getWalletOutputs" is a completely misleading name. It should be called "getOutputs" or "getTransactionOutputs". "Wallet" has nothing to do. I would prefer we don't use this method and we wrap it with an utility function getTransactionOutputs(btcTx,..) that uses getWalletOutputs with a big comment about this.

rskj-core/src/main/java/co/rsk/peg/Bridge.java
josedahlquist: changed it to use satoshis as it was originally intended

rskj-core/src/main/java/co/rsk/peg/Bridge.java
josedahlquist: changed it to use satoshis as it was originally intended

rskj-core/src/main/java/co/rsk/peg/Bridge.java
josedahlquist: changed it to use satoshis as it was originally intended

rskj-core/src/main/java/co/rsk/peg/Bridge.java
josedahlquist: changed it to use satoshis as it was originally intended

PR #1021, opened 2019-10-02
Altbn128 Precompiled Contracts

smcsicardi:
Hi, this is the PR for the precompiled contracts I've been working on

On gas costs:
Currently costs on the PR are the old ethereum ones, but this can be easily changed.

The implementation used is the one from EthereumJ, which is much faster than Pantheon's implementation. Still, it is considerably behind Parity's implementation and thus, the speed aim of the EIP changing the gas costs (is my tests it was between 5 and 7 times slower, on the same benchmark used for the gas costs correction).


rskj-core/src/main/resources/reference.conf
```diff
@@ -26,7 +26,8 @@ blockchain = {
             rskip125 = wasabi100,
             rskip126 = wasabi100,
             rskip132 = wasabi100,
-            rskip136 = bahamas
+            rskip136 = bahamas,
+            rskip137 = -1
```
diega: This should be assigned to `wasabiPlusOne`

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigTest.java
```diff
@@ -60,6 +60,7 @@
             "    rskip126: wasabi100",
             "    rskip132: wasabi100",
             "    rskip136: bahamas",
+            "    rskip137: bahamas",
```
diega: I'm not quite sure this change is needed, anyway, _rskip137_ shouldn't belong to **bahamas**

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
```diff
@@ -422,4 +453,209 @@ private BigInteger parseArg(byte[] data, int offset, int len) {
...
+     * input data[]:<br/>
+     * two points encoded as (x, y), where x and y are 32-byte left-padded integers,<br/>
+     * if input is shorter than expected, it's assumed to be right-padded with zero bytes<br/>
+     * <br/>
+     *
+     * output:<br/>
+     * resulting point (x', y'), where x and y encoded as 32-byte left-padded integers<br/>
+     *
+     */
+    public static class BN128Addition extends PrecompiledContract {
```
diega: I think it would be better if you extract the `BN128Addition`, `BN128Multiplication` and `BN128Pairing` contracts to their own files

rskj-core/src/test/java/co/rsk/pcc/AltBN128Test.java
```diff
@@ -0,0 +1,718 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Sebastian Sicardi
+ * @since 10.09.2019
+ */
+
+public class AltBN128Test {
+    private final TestSystemProperties config = new TestSystemProperties();
```
diega: I think we prefer to initialize fields in the constructor

rskj-core/src/test/java/co/rsk/pcc/AltBN128Test.java
```diff
@@ -0,0 +1,718 @@
...
+
+        PrecompiledContracts.PrecompiledContract altBN128Add = precompiledContracts.getContractForAddress(activations, altBN128AddAddr);
+        PrecompiledContracts.PrecompiledContract altBN128Mul = precompiledContracts.getContractForAddress(activations, altBN128MulAddr);
+        PrecompiledContracts.PrecompiledContract altBN128Pair = precompiledContracts.getContractForAddress(activations, altBN128PairAddr);
+
+        Assert.assertNotNull(altBN128Add);
+        Assert.assertNotNull(altBN128Mul);
+        Assert.assertNotNull(altBN128Pair);
+
+        Assert.assertEquals(PrecompiledContracts.BN128Addition.class, altBN128Add.getClass());
```
diega: We are trying to use _assertEquals_ which is more JUnit idiomatic

colltoaction: You mean `assertThat`?

diega: Yes, I'm sorry, that's what I meant

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
smcsicardi: Do you want me to changed the implementation to the new NativeContract interface? That way they can be on its own class also

rskj-core/src/test/java/co/rsk/pcc/AltBN128Test.java
smcsicardi: It's a testing class, they usually don't have a constructor, do you want me to do one anyway?

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
diega: I'm happy with just moving it outside the `org.ethereum.vm.PrecompiledContracts`, if you want to give a try to the `NativeContract` interface go ahead :)

smcsicardi:
> Hey @smcsicardi, mostly looks good, a couple of requests.
> 
> 1. Please use safe gas operations, you can find the static methodos on `GasCost.java` to add, multiply and subtraction.
> 2. If we don't use `isGroupMember`, why is it even written? Delete it?
> 3. Where's the switch to the native or java implementation? :thinks:

1. Will update the code.
2. The implementation was taken from EthereumJ and I didnt want to touch anything from it. I can remove but I would rather not to for consistency.
3. https://github.com/rsksmart/rskj/blob/altbn128/rskj-core/src/main/java/co/rsk/pcc/altBN128/AltBN128.java#L28

horca17:
ok, will try later now ZoKrates on master.

PR #1047, opened 2019-10-24
Improving remasc blocks retrieval; removing unused BlockStore method

ajlopezrsk:
Remasc precompiled contract usually  retrieved 4000 blocks in mainchain, each time. Now, the retrieve of the execution block (current block - maturity aprox) was replaced by an algorithm using in other uses cases. The old method used was removed.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/remasc/Remasc.java
+++ b/rskj-core/src/main/java/co/rsk/remasc/Remasc.java
@@ -115,11 +115,11 @@ public class Remasc {
         int uncleGenerationLimit = constants.getUncleGenerationLimit();
         Deque<Map<Long, List<Sibling>>> descendantsBlocks = new LinkedList<>();
 
-        // this search can be optimized if have certainty that the execution block is not in a fork
-        // larger than depth
-        Block currentBlock = blockStore.getBlockByHashAndDepth(
-                executionBlock.getParentHash().getBytes(),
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/co/rsk/remasc/Remasc.java
```diff
@@ -115,11 +115,11 @@ void processMinersFees() {
         int uncleGenerationLimit = constants.getUncleGenerationLimit();
         Deque<Map<Long, List<Sibling>>> descendantsBlocks = new LinkedList<>();
 
-        // this search can be optimized if have certainty that the execution block is not in a fork
-        // larger than depth
-        Block currentBlock = blockStore.getBlockByHashAndDepth(
-                executionBlock.getParentHash().getBytes(),
-                remascConstants.getMaturity() - 1 - uncleGenerationLimit
+        // this search is noe optimized if have certainty that the execution block is not in a fork
```
lsebrie: Is this change on the comment intended?

ajlopez: Yes, the algorithm supports the existence of a fork, but it could be degraded to the original algorithm performance if the fork involves > blocks than the remasc maturity parameter

diegomasini:
LGTM! I just found a minor typo, @ajlopez please fix it and I will approve the PR.

Issue #1043, opened 2019-10-23
EIP-152:  Add BLAKE2 compression function `F` precompile


SergioDemianLerner:
### Prize Bounty
The Prize Bounty is 5.25 ETH.

### Challenge Description

 Add BLAKE2 compression function `F` precompile as specified by EIP-152 (https://github.com/ethereum/EIPs/blob/master/EIPS/eip-152.md).

### Submission Requirements

A valid submission should modify the RSKj node to implement the aforementioned functionality.

The code should respect the contribution standards specified in https://github.com/rsksmart/rskj/blob/master/CONTRIBUTING.md

### Submission Deadline
The deadline for submissions is November 11, 2019.

### Judging Criteria
The prize will be awarded to the first valid submission.

### Winner Announcement Date
The submission will be reviewed between November 11 and November 18. Winners will be privately notified before a public announcement of all hackathon prices given by IOV Labs. The announcement will occur before November 31 and together with prizes payouts.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 weeks from now.
Please review their action plans below:


**1) [bakaoh](https://gitcoin.co/bakaoh) has started work.**

I have done add blake2f precompile in this pr https://github.com/rsksmart/rskj/pull/1049

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/rsksmart/rskj/1043/3576).

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Work for 5.25 ETH (945.65 USD @ $180.12/ETH) has been submitted by__:

 1. [@bakaoh](https://gitcoin.co/bakaoh)



@SergioDemianLerner please take a look at the submitted work:
* [PR](https://github.com/rsksmart/rskj/pull/1049) by @bakaoh

<hr>

* Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/rsksmart/rskj/1043/3576)
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/rsksmart/rskj/1043/3576).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
[...]

PR #1049, opened 2019-10-25
EIP-152: Add BLAKE2 Compression Function F Precompile

bakaoh:
Add BLAKE2 compression function F precompile as specified by [EIP-152](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-152.md).

Fixed: #1043


rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
```diff
@@ -91,6 +97,7 @@
             RIPEMPD160_ADDR,
             IDENTITY_ADDR,
             BIG_INT_MODEXP_ADDR,
+            BLAKE2F_ADDR,
```
nacho692: Precompiled contracts created after the network creation shouldn't be added here because it forks the network. In this case, pre existing number 1 blocks will have an incorrect parent hash.

Check this example as to how to implement it:
https://github.com/rsksmart/rskj/commit/e0e2c8652af0930b12ad7a31a0d329821674ae12

They are activated based on an RSKIP which is tied to a NetworkUpgrade which is tied to a block number.
So nodes update their code and when a certain block number is reached the new configuration loads.

As for what RSKIP to use, just define a non existing one, we'll handle the change when we merge it into master, for the NetworkUpgrade use `wasabiPlusOne`.

bakaoh: Thanks for clarifying, i updated and defined `RSKIP153`

Issue #1038, opened 2019-10-22
Native keccak256


SergioDemianLerner:
### Prize Bounty
The Prize Bounty is XX ETH

### Challenge Description
RSKj implements a Java version of the Keccack cryptographic hash function. This function is required for computation of hash digests in many parts of the RSK protocol.

The task is to replace the current implementation with an compliant implementation written in C/C++ and compiled into native code that can be linked through JINI.

Submission should link to a small external library. If this is not the case, the library should be  pruned to contain only the required functions to compute the hash function.

### Submission Requirements
A valid submission should modify the RSKj node to pass all tests with the new Keccack implementation. The C/C+ implementation should not be obscured in any way and must be easily verified against the specification, or be part of a popular standardized library.

Submissions should not require large external library dependencies unless this is proven necessary.

The code should respect the contribution standards specified in https://github.com/rsksmart/rskj/blob/master/CONTRIBUTING.md

### Submission Deadline
The deadline for submissions is November 11, 2019.

### Judging Criteria
The prize will be awarded to the first valid submission.

### Winner Announcement Date
The submission will be reviewed between November 11 and November 18. Winners will be privately notified before a public announcement of all hackaton prices given by IOV Labs. The announcement will occur before November 31 and together with prizes payouts.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 weeks, 1 day from now.
Please review their action plans below:


**1) [gpas](https://gitcoin.co/gpas) has started work.**

I will create Wrapper Classes for the following Keccak Hash Implementation: https://github.com/XKCP/XKCP/blob/master/lib/high/Keccak/FIPS202/KeccakHash.h

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/rsksmart/rskj/1038/3582).

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Work for 3.5 ETH (630.43 USD @ $180.12/ETH) has been submitted by__:

 1. [@bakaoh](https://gitcoin.co/bakaoh)



@SergioDemianLerner please take a look at the submitted work:
* [PR](https://github.com/rsksmart/rskj/pull/1051) by @bakaoh

<hr>

[...]

PR #1051, opened 2019-10-27
Native Keccak256 Implementation

bakaoh:
Fix: #1038 
Keccak jni using https://github.com/aidansteele/sphlib

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ jobs:
       - checkout
       - run:
           name: Setup
-          command: apt update -y && apt install -y gnupg2
+          command: apt update -y && apt install -y gnupg2 gcc g++
       - run:
           name: Verify files
           command: |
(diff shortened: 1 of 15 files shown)


nacho692:
Hey bakaoh, thanks for the contribution!, this seems to be working fine.

One thing I'll ask of you is to trim the files and keep only the code required for `keccak256`, as every type of keccak is included in the PR.
Also, there are a couple of files which do not seem to be required:

```
rskj-core/src/main/c/keccak/api.h
rskj-core/src/main/c/keccak/crypto_hash.h
rskj-core/src/main/c/keccak/export.h
rskj-core/src/main/c/keccak/hash.c
rskj-core/src/main/c/keccak/sha3_test.c
```

What's up with that?

bakaoh:
thanks @nacho692, I removed all unnecessary files and kept only the required code

PR #1059, opened 2019-10-31
WIP: Issue 1048

zyfrank:


zyfrank:
interesting, I think my added test should cover most new added code, but sonar said no

PR #1071, opened 2019-11-04
blockstore and locatorRepository added in lightprocessor

julianlen:


rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
```diff
@@ -42,13 +43,16 @@
 public class LightProcessor {
     private static final Logger logger = LoggerFactory.getLogger("lightprocessor");
     // keep tabs on which nodes know which blocks.
-    private final BlockSyncService blockSyncService;
+    private final BlockStore store;
```
smcsicardi: Change variable name to blockStore to be in line with other dependencies

PR #1072, opened 2019-11-04
[Gitcoin] Replace The Multilevel Cache System With A Journaling System

### LINKED ISSUE #1040: Replace the multilevel cache system to a journaling system (Gitcoin hackaton project sponsored by IOV Labs)
### Prize Bounty
The Prize Bounty is XX ETH

### Challenge Description

Currently the RSKj nodes employs a multi-level cache system to track changes to the world state.
This allows to revert the state in case of REVERT opcode or OOG.

The main class involved is MutableTrieCache.java. Every new contract called gets a new MutableTrieCache on top of the previous one.

This implementation is sub-optimal in terms of efficiency. Put/get are O(N) operations, where N is the level of the cache while revert is O(1).
It is preferable to use a journaling cache "JournalTrieCache" that:
-  Stores the latest values for key/values that have changed, but it is a pass-through for keys that have not. 
- Past values to changed keys are stored in a log, with entries of type (key,oldValue). Reversal implies going over the log in reverse order and reversing the changes made. Recursive deletion may require special functionality, as the current system marks accounts to imply recursive deletion.

The task is to change the structure of the node such that: for each block a MutableTrieCache is used, but for each transaction a JournalTrieCache is used (and the same object is is used for all internal CALL frames).

### Submission Requirements
A valid submission should modify the RSKj node to implement JournalTrieCache integrated to transaction processing, pass all unit tests and synchronize with the RSK mainnet.

The code should respect the contribution standards specified in https://github.com/rsksmart/rskj/blob/master/CONTRIBUTING.md

### Submission Deadline
The deadline for submissions is November 11, 2019.

### Judging Criteria
The prize will be awarded to the best valid submission.

### Winner Announcement Date
The submission will be reviewed between November 11 and November 18. Winners will be privately notified before a public announcement of all hackaton prices given by IOV Labs. The announcement will occur before November 31 and together with prizes payouts.

levalleux-ludo:
resolves #1040


lsebrie:
@levalleux-ludo please do a rebase so all the checks run again

levalleux-ludo:
Hi @lsebrie, 
I think that my PR is stuck on the last step "Pipeline Expected — Waiting for status to be reported".
I've just rebased on the last master to avoid any conflicts, but it has been stuck in this state for the last 2 days, after all checks have passed successfully.
Any thoughts ?
Thanks

levalleux-ludo:
> Hi @lsebrie,
> I think that my PR is stuck on the last step "Pipeline Expected — Waiting for status to be reported".
> I've just rebased on the last master to avoid any conflicts, but it has been stuck in this state for the last 2 days, after all checks have passed successfully.
> Any thoughts ?
> Thanks

@lsebrie  - PR is still stuck and I don't know what to do. Please could you give me some help ?

lsebrie:
Hi @levalleux-ludo we still have some tests that are ran by manual activation for external contributors. Sorry for the delay, I'll start the Pipeline step.

lsebrie:
Pipeline: run

levalleux-ludo:
OK, many thanks. I thought I have missed something

lsebrie:
We'll need a reviewer first, I'll see if there is anybody available for this topic

lsebrie:
Pipeline: run

lsebrie:
Pipeline: run

lsebrie:
Pipeline: run

lsebrie:
Pipeline: run

levalleux-ludo:
Pipeline: run

levalleux-ludo:
Pipeline: run

rskj-core/src/main/java/co/rsk/db/JournalTrieCache.java
```diff
@@ -0,0 +1,763 @@
...
+                Map<Integer, byte[]> valuesPerLevel = valuesPerLevelPerKey.get(key);
+                if (valuesPerLevel != null) {
+                    return valuesPerLevel.get(cacheLevel);
+                }
+            }
+            // TODO Prefer Optional<T> over null
+            return null;
+        }
+
+        byte[] getNewestValue(ByteArrayWrapper key, int maxLevel) {
```
SergioDemianLerner: This is not O(1). This is O(L) where L is the number of levels the value has been modified. This is no different with the current multi-level cache approach.
The spec required a cache based on single last state which could be checked in O(1). All changes should go to a log.

levalleux-ludo: OK. I've been focused on having the same structure (instance) to store data across the different cache levels, and I miss the point regarding the complexity class of the get method.
I look into it to upgrade my PR

levalleux-ludo: @SergioDemianLerner 
I've just upgraded my PR to take into account your comment.
Now the get accesses to the cache are O(1) as requested

lazaridiscom:
@levalleux-ludo , looked at your PR purely out of curiosity. Reviewing this is very(!) difficult, mainly because `JournalTrieCache.java` is introduced in an non-OO way.

The workaround of directly comparing `JournalTrieCache.java` to `MutableTrieCache.java` is not working well, due to the high fragmentation of changes (VSC / VS2017 are "loosing it").

I think you would speed-up the review-process if you:

* either inherit `MutableTrieCache.java`, thus your changes become nicely visible.
* create a common based-class (e.g. `TrieCacheBase`), where you keep the common functionality of Journal/Mutable implementations.

I don't understand why you chose to modify `MutableTrieCache` directly, I sense that I should have stayed "as-is", to allow the system to switch between "mutable" and "journal".

But then, I've overflown this just, quite abstractly, and I could be 100% wrong here with everything I wrote.

levalleux-ludo:
Hi @lazaridiscom ,
Thanks for your feedback, I find it very interesting.
Let me give some answers to your questions:
1.	Modifying the class MutableTrieCache was not functionally required and I wouldn’t do it, if I didn’t need to reuse the class StorageKeysIterator (returned by getStorageKeys()). Then I had to externalize it (before, it was a static nested class of MutableTrieCache) and this is only a structural change, in order to avoid code duplication.
2.	Going further, it’s true that there are some commonalities between JournalTrieCache and MutableTrieCache. However, they both are implementing the same interface (MutableTrie), they could also inherit from a new abstract call in order to gather the common code. But doing that I would do change the original MutableTrieCache class and that was not my initial choice (refer to point 1).
3.	By the way, I was only trying to gather as much as possible my changes to limit my ‘intrusion’ in the existing code, just because it’s my first contribution in that project and I didn’t want to impact too much. That’s why I’ve implemented several classes in the same file (MultiLevelCache, MultiLevelCacheSnapshot, ICache, JournalTrieCache) and this appears to be a bad choice for readability.
Then, I propose to improve the readability and the global design of the PR by this refactoring:
-	Implement an abstract class, say AbstractTrieCache, gathering the common parts of JournalTrieCache and MutableTrieCache.
-	Externalize MultiLevelCache, MultiLevelCacheSnapshot, and ICache from JournalTrieCache

Please let me know if you would find it useful

lazaridiscom:
Sounds fine. Whatever makes modifications/addition clearly visible is fine.

----

Note that personally, I'd processed this like this (which is a possible alternative) :

```java
class MutableTrieCacheAlt extends MutableTrieCache {};

// possibly absolute minimal modification of MutableTrieCache, e.g. change visibility of "StorageKeysIterator"

class JournalTrieCache extends MutableTrieCacheAlt {};

// any newly introduced constructs into separate file
// (MultiLevelCache, MultiLevelCacheSnapshot, ICache, JournalTrieCache)
```

This would allow for an easy review of the modifications/additions, and additionally, `MutableTrieCache` continues working EXACTLY as it did before.

Within thew new files, you'd keep notes on what should be further done (e.g. reduce duplication by altering the original class), thus the original author can react to it.

lazaridiscom:
small addition re:

// any newly introduced constructs into separate file
// (MultiLevelCache, MultiLevelCacheSnapshot, ICache)

or at least at the end of the file, thus any diff stays clean.

lazaridiscom:
looks much better, but I still think that a `MutableTrieCachAlt` class (as mentioned in https://github.com/rsksmart/rskj/pull/1072#issuecomment-563980777) would make things much cleaner (and allow for A/B comparisons). 

(sidenote: you need to do the usual commit-squashing, too, in order to reduce commit-count)

levalleux-ludo:
Hi @SergioDemianLerner ! I've tried to contact you on Discord Gitcoin channel regarding my submission, without success.
I've seen lots of great news regarding IOV Labs recently. Congrats. I guess you're still very busy but I would really appreciate if you could take time to reconsider my submission of the journallized trie cache as you've proposed in our last discussion on 25th of Dec.
The bounty is closed now, but I didn't get the payout I was expected.
As a reminder, you gave me a first feedback on the 28th of November, noticing that my new algorithm was not O(1), that was right, and I've fixed that immediately to resubmit one day later, but stayed without an updated feedback from yours.
Thanks in advance

SergioDemianLerner:
Hi levalleux-ludo,
 I will evaluate your out-of-contest submission this week to see if we can give you a special prize. Thanks for waiting.

levalleux-ludo:
> Hi levalleux-ludo,
> I will evaluate your out-of-contest submission this week to see if we can give you a special prize. Thanks for waiting.

Oh great ! That's nice. Thanks @SergioDemianLerner

levalleux-ludo:
> Hi levalleux-ludo,
> I will evaluate your out-of-contest submission this week to see if we can give you a special prize. Thanks for waiting.

Hi @SergioDemianLerner ! Would you have had time for evaluating, then ?
Thanks

ajlopezrsk:
@levalleux-ludo thanks for your contribution. These are my comments:

In the issue description, I read

Past values to changed keys are stored in a log, with entries of type (key,oldValue). Reversal implies going over the log in reverse order and reversing the changes made. Recursive deletion may require special functionality, as the current system marks accounts to imply recursive deletion.

I didn't find anything that points to this kind of implementation. I read a multi level cache system, but when a level is rollbacked, no reversal of the changes is executed. Only a call to MultiLevelCache.recomputeLatestValue.

And that recomputeLatestValue (that could fill the level with the values of all the previous levels, I'm not sure), it is only invoked when rollback. When a deeper level is created, the values of the other levels ARE STILL IN THOSE LEVELS, so, that is the reason for my second comment, below:

Second comment:
There is a new MultiLevelCache class. It's getNewestValue method returns the latest value of a key. If the value is not present at the current level, it queries the previous level, and so on... Then, if a value is found, NO PROVISION is made to at least put the found value recorded at the current level.

So, a smart contract, running at level 100, could query the balance of an account, 100 times, and if that value is at level 20, every query visits ALL the cache from level 100 to level 20, EACH TIME.
Is it right?

levalleux-ludo:
Hi @ajlopezrsk 
Many thanks for having taken time to review my submission.
I wish we would have this share and feedback during the hackathon in November, but better late than never ;-) 
Please find my answer to your comments below, and I'm trying to explain better my implementation

On you first remark, "Past values to changed keys are stored in a log, with entries of type (key,oldValue)."
While implementing this issue in the way as described, I found that it was absolutely necessary to keep at which level the oldValue is referring to.
There was many usecases in the non-regression tests suite that was failing otherwise.
(Just excuse me I don't remember exactly which ones, 6 months have passed. Just think, at least this would be something with the ‘collectKeys()’ method)
Hence, it was not possible to store entries exactly as described (key, oldValue) and I had to keep storing oldValue(s) par key AND per level.
That could explain why my design is slightly different than expected.
Here after a very simple graphical overview of the design:
image

The special case of recursive deletion of account is of course taken into account in my implementation: see MultiLevelCache::deleteAccount().
To get confidence in my code, I was referring a lot on existing tests usecases, If you compare JournalTrieCacheTest.java with MutableTrieCacheTest.java: you can see that I’ve used the same unit tests cases for JournalTrieCache than the ones existing for MutableTrieCache, in order to verify that the new implementation works the same as the old one. However, after upgrading MutableRepository to use JournalTrieCache instead of MutableTrieCache, I’ve also checked that all the project’s tests passed.
I’ve even added a specific test for JournalTrieCache (testMultiPutsIncludingAccountDeletion).that focuses on deletion and rollbacks at different levels.

Going further in your comments: When a rollback is done at a given level, there IS a reversal of the changes done at that level: the ‘clear()’ method is called to erase the values stored at that level and the ‘recomputeLatestValue()’ is refreshing the “Latest Values Map”. However, when rollbacking the deepest level of the MultiLevelCache (and only in that case), the cache max deepest level is decremented.
You can see that ‘recomputeLatestValue()’ is not only called when ‘rollback()’ but also when ‘deleteAccount()’. However the underlying “Latest Values Map” is also updated while putting a new value for the key at the deepest level (‘put()’).

Rephrasing your second comment with the enlightening done before:
There is a new MultiLevelCache class (note: only one instance exists for one MutableTrie, look at the multiplicity in the schema above). Its (getNewestValue()’ method returns the latest value of a key WITHOUT recursively looking at previous level, thanks to the ‘Latest Values Map’
 So, a smart contract, running at level 100, could query the balance of an account, 100 times, and if that value is at level 20, every query will return the value without parsing deeper levels, just getting the value from the “Latest Values Map”.

Once again, keeping the values ‘per level’ in the map table structure might have differed from the expected design, but IMO remains necessary to handle usecases such as deletions, rollbacks, and keys collections when they occur at an intermediary level (I mean not the deepest one). While running the complete test suite, I’ve seen these cases occurring so that my implementation.

levalleux-ludo:
Hi @ajlopezrsk ,
Thanks for your feedback.
I understand that you want me to remove the for loops as well as possible, replacing them with complete key/value maps at each level, meaning having to manage lots of copies of the same data.
I think there are some tricky usecases, as when the values are changed at an intermediate level (not always the deepest one), that require to be analyzed to check feasibility.
However, I wonder if copying data at each level and keeping them consistent with all usecases would not lead to a significant performance overhead (performances are often a compromise: what you win in some usecases is lost from others).
I’m very busy at the moment but I propose to look forward and complete this issue by the end of July.
Please let me know whether you’re OK on not with that deadline.

ajlopezrsk:
Hi @levalleux-ludo° Sorry for the delay.

Yes, the use of `for` to visit many levels in some use case is a concern. 

But you don't need to copy a lot of key values between levels. You only use a key value (I'm simplifying, there key values for accounts, address key value for storage, etc) along the levels.

Let's call this key value the UniqueKeyValue. Then, when at level 1, you call a contract, a level 2 is created, and the SAME UniqueKeyValue is pass to level 2. At level 2, you can get new key values from trie, and the put them in the UniqueKeyValue (to avoid go to the trie again each time you query them). When at level 2, you modify the value of a key, it is saved in the UniqueKeyValue: same key, new value replace the old value. When finish the level 2, the UniqueKeyValue goes back to level 1, and so on.

The only thing to keep BY LEVEL, is another key map: key + original value.

When at level 2 you modify a key value, replace the value associated to that key within UniqueKeyValue. But if it is the FIRST time you modify that key in this level, then add key - original value to the key original value associated to level 2.

If at level 2 you must rollback, you restore the key original values to the UniqueValueKey, before returning to level 1.

If at level 2 you commit, then you add the key original values of level 2 to the associated key original values of level 1 (with care: only the keys that were modified at level 2 BUT NOT at level 1). In this way, if level 1 should be rollbacked, its associated kjey original value map has the info to be restored (accumulating all the changes during the level 1 life, including level 2, level 3s and more)

In short: one key value map along all the process; key original value map per level. No `for` is needed to visit MANY levels. When a key is absent in uniquekeyvalue, then go directly to the storage/trie

I hope I could explain the proposed algorithm. Just write here if you need more info or have a use case that cannot be resolved in this way.

levalleux-ludo:
Hi @ajlopezrsk !
I've now implemented the JournalTrieCache in the way you've proposed in your last comment.

Please have a look at the updated PR, and let me know if it matches your expectations.

Just please note that there are some specific usecases, for instance when setting a new value for a key at a given cache level L (for instance level 1) **when the same key has already been set another value at a deeper level L'** (for instance level 7), and **only** in that cases, where we still need to loop over the cache levels (from 2 to 7 in the example) to update the internal maps in a correct way.
See calls to the method findOriginalValuesPerKeyAtDeeperLevel() that deal with that particular cases.
Thanks

levalleux-ludo:
Hi @ajlopez @ajlopezrsk. Any feedback? Thanks

PR #1073, opened 2019-11-05
Add UPnP support for peer discovery.  Resolves #1036

### LINKED ISSUE #1036: Add UPnP Support (Gitcoin hackaton project sponsored by IOV Labs)
### Prize Bounty
The Prize Bounty is XX ETH

### Challenge Description

The peer discovery component has a problem when announcing the local node to another peers: it doesn't traverse through NAT routers.
A way to address this would be implementing UPnP protocol.

https://en.wikipedia.org/wiki/Universal_Plug_and_Play

An sample solution applied to Pegasys Phanteon is this: https://github.com/PegaSysEng/pantheon/pull/1334

### Submission Requirements

A valid submission should modify the RSKj node to allow seamless discovery and communication through NAT routers using uPnP. 
Submissions should not require large external library dependencies unless this is proven necessary. 

The code should respect the contribution standards specified in https://github.com/rsksmart/rskj/blob/master/CONTRIBUTING.md

### Submission Deadline

The deadline for submissions is November 11st, 2019.

### Judging Criteria

The prize will be awarded to the first valid submission.

### Winner Announcement Date

The submission will be reviewed between November 11st and November 18th. Winners will be privately notified before a public announcement of all hackaton prices given by IOV Labs. The announcement will occur before November 31st and together with prizes payouts.

CoeJoder:
Feature description:

- UPnP is toggleable by the config property `peer.discovery.upnp_enabled` and is enabled by default.

- If automatic port mapping fails, a log message notifies the user to enable UPnP on their router, or disable UPnP in the node conf and forward router ports manually.

- Any automatic port mapping created via UPnP on the router is deleted upon node shutdown.

Implementation notes:

- The `UpnpService` is injected into `UDPServer` as an `Optional` which is present if the service was enabled in the conf.  This is the only service using UPnP right now.

- There are three TCP services running: `Web3WebSocketServer`, `Web3HttpServer` , `PeerServerImpl`.  All three of these services are configurable with different network interfaces/local addresses, each of which could be connected to unique WAN gateways.  If `UpnpService` was injected into these server objects, it could handle the port mapping easily.  But that is out of scope for this ticket.

- Code coverage is 95.6% because two unit tests were deleted which required PowerMockito, which is not supported by SonarQube.  Neither unit test was important (extraneous Exception handling unrelated to functionality of the code).

TODO:
There is a new dependency on the weUPnP library, see rsksmart/reproducible-builds#19.  The `build.gradle` dependency is currently using the SHA hash of the public repo, but should be switched to that of the reproducible build when it is ready.

### DIFF
--- a/rskj-core/build.gradle
+++ b/rskj-core/build.gradle
@@ -68,6 +68,7 @@ ext {
     logbackVersion = '1.2.2'
     bitcoinjVersion = '0.14.4-rsk-8'
     nettyVersion = '4.0.56.Final'
+    weupnpVersion = '0.1.4'
 }
 
 dependencies {
@@ -84,6 +85,7 @@ dependencies {
     compile "org.mapdb:mapdb:2.0-beta13"
     compile "co.rsk.bitcoinj:bitcoinj-thin:${bitcoinjVersion}"
     compile 'com.github.briandilley.jsonrpc4j:jsonrpc4j:1.5.1'
+    compile "org.bitlet:weupnp:${weupnpVersion}"
 
     runtime "org.slf4j:log4j-over-slf4j:${slf4jVersion}"
     runtime "ch.qos.logback:logback-classic:${logbackVersion}"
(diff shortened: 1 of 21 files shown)


CoeJoder:
I'm going to resubmit this PR with a cleaner commit history.

CoeJoder:
There is a new PR here:  #1088

Issue #1065, opened 2019-11-02
Incoming Block is usually validated TWICE


ajlopez:
When a `BlockMessage` arrives, the `MessageVisitor` validates the incoming block. BUT the `BlockchainImpl` also validates it AGAIN, apparently using the same rules.

The validation of uncles could be expensive, so the first advice is to validate the block ONLY in the most sensitive process, inside `BlockchainImpl` and removing the validation check from ``MessageVisitor`.

PR #1080, opened 2019-11-06
Remove block validation from MessageVisitor

ajlopezrsk:
The block validation is already included in `BlockchainImpl`

Solving #1065


lucasvuotto:
pipeline: run

PR #1082, opened 2019-11-06
Native bn128

mraof:
I replaced the classes in org.ethereum.crypto.altbn128 with ones using the native implementation that Parity uses, which should work and retain its speed on any platform Rust supports.

The source code for the native library is located at https://github.com/mraof/rskj-bn128


nacho692:
Hey @mraof, sorry to keep you waiting. Thanks a lot for your contribution, nevertheless, after running some internal benchmarks we decided to go with the Cloudflare solution

PR #1083, opened 2019-11-07
Add support for disabling GC if it was ever enabled

### LINKED ISSUE #1018: Support disabling garbage collector once enabled

diega:
Fixes #1018

### DIFF
--- a/rskj-core/src/main/java/co/rsk/RskContext.java
+++ b/rskj-core/src/main/java/co/rsk/RskContext.java
@@ -130,6 +130,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Clock;
 import java.util.*;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.HashMap;
 import java.util.List;
@@ -814,14 +815,32 @@ public class RskContext implements NodeBootstrapper {
     private TrieStore buildAbstractTrieStore(Path databasePath) {
         TrieStore newTrieStore;
         GarbageCollectorConfig gcConfig = getRskSystemProperties().garbageCollectorConfig();
+        final String multiTrieStoreNamePrefix = "unitrie_";
         if (gcConfig.enabled()) {
             try {
-                newTrieStore = buildMultiTrieStore(databasePath, "unitrie_", gcConfig.numberOfEpochs());
+                newTrieStore = buildMultiTrieStore(databasePath, multiTrieStoreNamePrefix, gcConfig.numberOfEpochs());
             } catch (IOException e) {
                 throw new IllegalStateException("Unable to build multi trie store", e);
             }
         } else {
-            newTrieStore = buildTrieStore(databasePath.resolve("unitrie"));
+            Path trieStorePath = databasePath.resolve("unitrie");
+            try (Stream<Path> databasePathFilesStream = Files.list(databasePath)) {
+                List<Path> multiTrieStorePaths = databasePathFilesStream
+                        .filter(p -> p.getFileName().toString().startsWith(multiTrieStoreNamePrefix))
+                        .collect(Collectors.toList());
+
+                boolean gcWasEnabled = !multiTrieStorePaths.isEmpty();
+                if (gcWasEnabled) {
+                    LevelDbDataSource.mergeDataSources(trieStorePath, multiTrieStorePaths);
+                    // cleanup MultiTrieStore data sources
+                    multiTrieStorePaths.stream()
+                            .map(Path::toString)
+                            .forEach(FileUtil::recursiveDelete);
+                }
+            } catch (IOException e) {
+                logger.error("Unable to check if GC was ever enabled", e);
+            }
+            newTrieStore = buildTrieStore(trieStorePath);
         }
         return newTrieStore;
     }
@@ -934,7 +953,6 @@ public class RskContext implements NodeBootstrapper {
                 disposedEpoch -> FileUtil.recursiveDelete(databasePath.resolve(namePrefix + disposedEpoch).toString())
         );
     }
-
 
     protected RepositoryLocator buildRepositoryLocator() {
         return new RepositoryLocator(getTrieStore(), getStateRootHandler());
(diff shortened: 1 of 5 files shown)


nacho692:
run tests

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -811,17 +815,36 @@ protected SolidityCompiler buildSolidityCompiler() {
         return new SolidityCompiler(getRskSystemProperties());
     }
 
-    private TrieStore buildAbstractTrieStore(Path databasePath) {
+    @VisibleForTesting
+    protected TrieStore buildAbstractTrieStore(Path databasePath) {
```
nacho692: I'd rather you used `getTrieStore()` which is public already instead of exposing internals

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -935,6 +958,20 @@ private TrieStore buildMultiTrieStore(Path databasePath, String namePrefix, int
         );
     }
 
+    @VisibleForTesting
+    public static void mergeMultiTrieStoreDBs(Path trieStorePath, List<Path> multiTrieStorePaths) {
```
nacho692: same here

rskj-core/src/test/java/co/rsk/RskContextTest.java
```diff
@@ -21,47 +21,16 @@
...
-import org.powermock.modules.junit4.PowerMockRunner;
-
-import java.io.ByteArrayInputStream;
-import java.io.InputStream;
 
 import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.assertThat;
-import static org.powermock.api.mockito.PowerMockito.*;
 
-@RunWith(PowerMockRunner.class)
```
nacho692: Great on removing this

rskj-core/src/main/java/co/rsk/RskContext.java
diega: You were right, it was possible to test the `RskContext#getTrieStore` directly by mocking the _databaseDir_ property

rskj-core/src/main/java/co/rsk/RskContext.java
diega: I rather keep it this way. Maybe a different component should take care of of the whole `TrieStore` lifecycle, but doing that goes well beyond the scope of this PR. So far, being this method _static_, it should be pretty harmless and easy to refactor in the future.

diega:
@nacho692 I agree on changing the default from `3` to `2` but I'm not quite sure about the integration testing procedures for validating this. I would suggest to keep it as `3` and then change it once it has been properly tested.
Also, could you please indicate the internal pipeline to run again?

nacho692:
pipeline: run

nacho692:
Pipeline: run

rskj-core/src/test/java/co/rsk/RskContextTest.java
```diff
@@ -43,25 +60,150 @@ public void getCliArgsSmokeTest() {
...
+        }
+
+        Assert.assertThat(Files.list(testDatabasesDirectory).count(), is(preExistingEpochs));
+        TrieStore trieStore = rskContext.getTrieStore();
+        Assert.assertThat(trieStore, is(instanceOf(TrieStoreImpl.class)));
+        Assert.assertThat(Files.list(testDatabasesDirectory).count(), is(1L));
+    }
+
+    @Test
+    public void shouldBuildMultiTrieStore() throws IOException {
```
lsebrie: Please add one test case when multiple folders already exist and the method getTrieStore should build a multi trie store. The other cases are covered :+1:

rskj-core/src/test/java/co/rsk/RskContextTest.java
```diff
@@ -43,25 +60,150 @@ public void getCliArgsSmokeTest() {
...
+            sourcePaths.add(sourcePath);
+            KeyValueDataSource originDataSource = RskContext.makeDataSource(sourcePath);
+            byte[] currentElement = {(byte) i};
+            sourceKeys.add(currentElement);
+            originDataSource.put(currentElement, currentElement);
+            originDataSource.close();
+        }
+
+        Path destination = testDatabasesDirectory.resolve("destination");
+        RskContext.mergeMultiTrieStoreDBs(destination, sourcePaths);
```
lsebrie: Please move this method out of RskContext to MultiTrieStore or TrieStoreImpl

rskj-core/src/test/java/co/rsk/RskContextTest.java
```diff
@@ -43,25 +60,150 @@ public void getCliArgsSmokeTest() {
...
+        KeyValueDataSource destinationDataSource = RskContext.makeDataSource(destination);
+        try {
+            Set<byte[]> destinationKeys = destinationDataSource.keys();
```
lsebrie: I would try to build two maps and probe that both contains the same elements. In the first part of the test you are only checking one side of inclusion all destination items in sum(sources). You also need to probe that items in sum(sources) are included in destination.

rskj-core/src/test/java/co/rsk/RskContextTest.java
diega: Given that `MultiTrieStore` or `TrieStoreImpl` doesn't have visibility of any kind of `java.util.Path`, I moved it to `LevelDbDataSource#mergeDataSources`

rskj-core/src/test/java/co/rsk/RskContextTest.java
diega: The test is built inserting a single element into each of the _origin_ datasources, I added an extra assertion for the keys after merge to be exactly the same count as _origin_ datasources. Then I check for the value to be the expected one (the same as key, this was made just to simplify the test). I think the whole PR is already pretty covered, but if you think it's really important to improve this particular test, could you suggest a more explicit code change (a diff would be optimal)? I may have not understood your request really well.

lsebrie:
Pipeline: run

Issue #1045, opened 2019-10-23
Parallel transaction executions based on RSKIP144


SergioDemianLerner:
### Prize Bounty
The Prize Bounty is XX ETH

### Challenge Description
The RSKIP144 (https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP144.md) specifies a method to execute transactions in parallel based on a schedule proposed by the miner.
The execution part requires a modification of the consensus of RSK while the transaction scheduling algorithm can be any and it is decided by the miner.

The task is to write modification of the RSK node and consensus protocol that implements this RSKIP.

### Submission Requirements
A valid submission should modify the RSKj node to implement RSKIP144. The submitter is allowed to improve the scheduling protocol. The submission must also include the required documentation to fully understand the scheduler proposed.

Submissions should not require external libraries dependencies unless this is proven necessary.

The code should respect the contribution standards specified in https://github.com/rsksmart/rskj/blob/master/CONTRIBUTING.md

### Submission Deadline
The deadline for submissions is November 11, 2019.

### Judging Criteria
The prize will be split 80%/ 20% between the best valid submission and the second best valid submission. The submission is valid if it implements the protocol specified in RSKIP144, optionally providing proven improvements. 

The judging criteria will be based on (in prioritized order):

1. The efficiency of the scheduler, measured by executing predefined set of blocks that contain assorted transactions including standard RBTC and ERC-20 token transfers.
2. The quality and formalization of the scheduler, if it was improved.
3. The quality of the code
4. The quality of the associated documentation

If there is only one submission, then the second prize will stay vacant.

### Winner Announcement Date
The submission will be reviewed between November 11 and November 18. Winners will be privately notified before a public announcement of all hackaton prices given by IOV Labs. The announcement and prize payouts will occur before November 31, 2019.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 2 weeks, 4 days from now.
Please review their action plans below:


**1) [levalleux-ludo](https://gitcoin.co/levalleux-ludo) has started work.**

I'm ready to start implementing RSKIP144 as specified.
Not sure if can complete this task before the end of Web3 World hackathon on Nov 11th, but at least I propose to start it and I'll give feedback on my progress before this deadline.
I will be able to complete later on, but not before November 22nd (because I'm not available on next week).
[...]

PR #1090, opened 2019-11-09
Parallel transaction rskip144 (partial)

levalleux-ludo:
This PR starts to implement #1045 
This work is still in progress but the PR is here as a milestone to give a progress status before Gitcoin hackathon is completed, as discussed on Gitcoin's Discord.
Any feedback is welcome.

RSKIP144 - Progress status (9th November):

Achievements 

- BlockExecutor is able to run some TransactionExecutor in concurrent threads, according to a new field 'partitionEnds' added in the block header.
However, as there is no consensus rule added yet for this new header structure, this field can only be added in block header for test purpose.
By default only one thread executes all transactions (hence, serially), because no 'partitionEnds' field is specified in block header.

Implementation notes : 
We keep the TransactionExecutor class as is, as it is used in different context, at runtime or for test, assuming that we still execute transactions without parallelism in these cases.
For the specific usecases where we want a parallel execution (in BlockExecutor), we use a derivated class TransactionConcurrentExecutor, that, unlike the normal TransactionExecutor,
doesn't transfer the fees to the miner on the fly (after each transaction) otherwise it would create a conflict (different threads write to the same state key -> the miner account).
Then we differ the transfer fees in BlockExecutor after all partitions are completed.

- BlockExecutor is able to detect conflicts between concurrent transactions, that would mean 2 transactions executed in different threads are accessing (read, write or delete) to the same key in the cached repository.
(It is for now a simplified version of the conflict detection to be implemented wrt RSKIP144).
In case of such a conflict is detected, the block validation immediately stops, all threads are stopped and an INVALID_BLOCK result is returned.
Some new tests have been added in BlockExecutorTest to verify that : 
* if transactions are executed in parallel (thanks to a partitionEnds specified in block header) and these transactions access the same key (read or write), then the block is declared invalid and its validation stops immediately.
* if transactions are executed in parallel (thanks to a partitionEnds specified in block header) and these transactions does not access the same keys (read or write), then the block is validated and the execution results are equivalent to the non-concurrent execution

Implementation notes : 
Marking of the read, written and deleted keys is realized by a 'tracking' feature implemented in the MutableTrieCache.
Unlike specified in the RSKIP144, marking is not done per thread (ie using the threadId as discriminant), because the execution of the transaction 
does not involve only one thread (the partition thread) : actually another thread (at least ?) is used during the transaction execution to run contract code
 (the thread is called 'vwExecutor'). Then we must consider a kind of delegation by the TxExecutor thread ( = the partition thread ) to the vmExecutor thread.
Hence, when the vmExecutor thread writes or reads keys, it shall be considered as it would come from the TxExecutor thread, but the threadIds are different.
To fix this issue, we're using the threadGroup as a discriminant, instead of the threadId, assuming that each partition thread is created in a dedicated group, 
and this group is inherited by every vmExecutor threads created for each Tx in the context of the partition.

- When executing different transactions in parallel, the 'executedTransactions' and 'transactionReceipts' lists in the BlockResult, being populated on the fly, receive the execution results in an undeterminist order.
So, after all execution have completed, they are reordered to match the "submission order" (as ordered in the block to be validated).

- All the unit tests have also been passed to verify that the implemented changes do not introduce a regression.

Next steps :

- Implement and test the conflict detection as specified in RSKIP144, including :
	- the case where some transactions only read the same key but none of them write this key. So there shall be no conflict in that case and such transaction can effectively been parallelised

- Implement and test the case of block to be mined, that executes the transactions serially, looking for potential access conflicts and fix these conflicts by adjusting the partitioning of transactions across threads.
This will result in defining the content of the 'partitionEnds' that is to be written in the block header.

- Define a new consensus rule to insure comparability between old and new block header structure, specifying a block number when the new consensus will apply.

Proposed deadline (as discussed on Gitcoin Discord : 22nd Nov)


levalleux-ludo:
Although very close to the RSKIP144 proposed design, my implementation is slightly different for the 3 following points :
1.	The tracking of state key access during the transaction execution is not realized ‘per thread’ but ‘per threadGroup’ instead This is because the execution of one transaction can involve more than one thread (for instance when vmExecutor is called). Using a dedicated threadGroup per partition and ensuring that all the involved threads are created in that group give us the discriminant we need to track accesses per partition (1 partition = 1 group)
2.	During a block validation, there is no need to execute all transactions, before analyzing/comparing the maps looking for conflicts. Actually a conflict is detected ‘on the fly’ during the execution of transactions. In case one or more conflicts have been detected, the block validation fails the sooner as possible and we don’t need to wait for all the transactions/threads are completed.
3.	During the creation of an (unresolved) block, all candidate transactions are run a first time as ‘in a sandbox’: we don’t take into account the state changes but only tracking conflicts between txs in order to sort them across partitions (with a maximum of 16 partitions). After this partitioning is finished, we run the transactions again, but this time in parallel according to the partitioning result, to confirm the block is valid (this also confirms the partitioning). Compared to the proposed design, this implementation looks, IMHO, more reliable, because:
•	the final world state resulting from the block validation is computed by the miner, in the second run, with the exact same context as the validators will do.
•	we don’t need to ‘join the threads’ in a way that don’t change the resulting world state. The proposed design had specific issue with existence of loop cases. Here it doesn’t matter and we don’t need to exclude transactions from the block without fearing potential tx censorship.
However, this simplification results in a simpler code, easier to maintain. Nevertheless, I must admit that, because of the double-run of transactions when building a new block, my implementation consumes probably more time for this particular usecase. But I think it worth the benefit brought by the concurrent txs execution at block validation (synchronization). Running all the tests with the new BlockExecutor implementation, I did not notice any significant increase of the execution time compared to before.

levalleux-ludo:
A few refactoring compared to my previous comment : I've replaced the tracking 'per ThreadGroup' with a tracking 'per partitionId' because usage of ThreadGroup is denied by quality gates (Sonar).
For that, I've implemented a derivated Thread class to include the partitionId at the thread level and propagate it across children threads.

SergioDemianLerner:
Hi @levalleux-ludo, thank you very much for coding your flavor of RSKIP144. 
Today I'm formally beginning the evaluation of submissions.  
While you've implemented a variation of RSKIP144, I think that, if the evaluation result is that your submission is close to the spec, you can be awarded all or almost all of the established price (this I need to check with the other judges to be fair).

levalleux-ludo:
Hi @SergioDemianLerner 
Thanks, I understand.
Please refer to my explanation in comments above for justifications of the slight 'variation'

levalleux-ludo:
I've just added the Consensus Rule for RSKIP144 and the encoding/decoding of the 'partitionEnds' field in BlockHeader (only when the new consensus rule is activated).
Waiting for your feedback :-)

rskj-core/src/main/java/co/rsk/core/bc/TransactionConflictDetector.java
```diff
@@ -0,0 +1,367 @@
...
+        return true;
+    }
+
+    private boolean checkKeyAlreadyRead(ByteArrayWrapper key, int partitionId) {
+        Collection<Integer> otherPartIds = partitionReadersPerKey.get(key);
+        if (otherPartIds == null) {
+            return false;
+        }
+        // there can not be a conflict with itself
+        otherPartIds.remove(partitionId);
```
SergioDemianLerner: This remove() actually modifies the underlying data structure so has side effects. It should either re-insert the partitionId or don't remove it in the first place and check for size()==1 and the first element.

rskj-core/src/main/java/co/rsk/core/bc/TransactionConflictDetector.java
```diff
@@ -0,0 +1,367 @@
...
+        return true;
+    }
+
+    private boolean checkAccountAccessed(ByteArrayWrapper account, int partitionId) {
+        Collection<Integer> otherPartIds = accessedAccounts.get(account);
+        if (otherPartIds == null) {
+            return false;
+        }
+        // there can not be a conflict with itself
+        otherPartIds.remove(partitionId);
```
SergioDemianLerner: side effects here too

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -265,97 +447,147 @@ private BlockResult executeInternal(
...
-        Coin totalPaidFees = Coin.ZERO;
-        List<TransactionReceipt> receipts = new ArrayList<>();
-        List<Transaction> executedTransactions = new ArrayList<>();
-        Set<DataWord> deletedAccounts = new HashSet<>();
+        BlockSharedData blockSharedData = new BlockSharedData(track, programTraceProcessor);
 
-        int txindex = 0;
+        if (!block.isSealed()
+                && block.useParallelTxExecution()
+                && !computeTxPartitioning(block, track, programTraceProcessor, acceptInvalidTransactions, discardInvalidTxs)) {
```
SergioDemianLerner: The pre-run of transactions sequentially only occurs during mining, and this is correctly detected by block.isSealed(). Nevertheless, a comment would help the reader understand it.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -239,13 +247,187 @@ public void traceBlock(
...
+     */
+    private boolean computeTxPartitioning(Block block, RepositorySnapshot track, ProgramTraceProcessor programTraceProcessor,
+                                          boolean acceptInvalidTransactions, boolean discardInvalidTxs) {
+        //
+        // Work with a cached repository that will be rolled back in the end of transaction execution
+        TransactionsPartitioner partitioner = new TransactionsPartitioner();
+        Repository dummyRepo = track.startTracking();
+        BlockSharedData dummySharedData = new BlockSharedData(dummyRepo, programTraceProcessor);
+        TransactionConflictDetectorWithCache transactionConflictDetector = new TransactionConflictDetectorWithCache(partitioner);
+        if (dummyRepo.isCached()) {
```
SergioDemianLerner: I don't have enough clarity why the TransactionConflictDetectorWithCache is used here and not the TransactionConflictDetector.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -265,97 +447,147 @@ private BlockResult executeInternal(
...
 
-            long gasUsed = txExecutor.getGasUsed();
-            totalGasUsed += gasUsed;
-            Coin paidFees = txExecutor.getPaidFees();
-            if (paidFees != null) {
-                totalPaidFees = totalPaidFees.add(paidFees);
+        // Wait for all thread terminate
+        try {
+            partitioner2.waitForAllThreadTermination(
+                    10000,
```
SergioDemianLerner: Magic number should be a pre-defined constant

rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
```diff
@@ -189,24 +221,31 @@ private BlockHeader decodeHeader(RLPList rlpHeader, boolean sealed) {
         int r = 15;
 
         int uncleCount = 0;
-        if (rlpHeader.size() == RLP_HEADER_SIZE || rlpHeader.size() == RLP_HEADER_SIZE_WITH_MERGED_MINING) {
-            byte[] ucBytes = rlpHeader.get(r++).getRLPData();
-            uncleCount = parseBigInteger(ucBytes).intValueExact();
-        }
+        byte[] ucBytes = rlpHeader.get(r++).getRLPData();
+        uncleCount = parseBigInteger(ucBytes).intValueExact();
```
SergioDemianLerner: I must check if it ever happened in Mainnet, but this doesn't seem to catch the case where rlpHeader.size() == 15.

SergioDemianLerner:
@levalleux-ludo: I have a question regarding your comment:

> During the creation of an (unresolved) block, all candidate transactions are run a first time as ‘in a sandbox’: we don’t take into account the state changes but only tracking conflicts between txs in order to sort them across partitions (with a maximum of 16 partitions). After this partitioning is finished, we run the transactions again, but this time in parallel according to the partitioning result, to confirm the block is valid (this also confirms the partitioning). 

As far as I understood it, your algorithm won't solve correctly the following problem:
If a transaction A conflicts with a transaction B, you will put B in the same partition as A, and then re-execute the block under this new setting assuming no other problem can arise. However, the second execution the transaction B could produce different changes in the read/write key sets because transaction A may have modified the environment. Your solution only works for simple transfers, that do not change the key access pattern based on previous executions.
Your algorithm could be modified to add the conflicting tx to the conflicting partition, and then loop again and execute it to the test if there is any new conflict. This could lead to a cascade of partition merges, and multiple executions for a single tx.
Another possible patch would be to differentiate between contract executions and simple transfers.
For contract executions, move conflicting transactions to a special "last serial thread" which executes after all parallel processing has ended. For simple transfers, re-run in the same partition to see if the source balance is enough to covert the cost.
This will parallelize fully disjoint contract executions, but will serialize any execution having a conflict.

rskj-core/src/main/java/co/rsk/core/bc/TransactionConflictDetector.java
levalleux-ludo: @SergioDemianLerner  Well done !
You’re right, there is a bug here.
Following your comment, I’ve tried to find a usecase which reveals it, and add a test case and check the correction fix it effectively.
But I couldn’t find this case: this bug has actually no real impact because the only case when you remove the current partition from the readers list of the key is when the same partition is writing this key too.
So, in the end there will be a conflict with any other partition that would attempt to access the same key, even if the readers list has been purged (unintentionally I confess).
Anyway, I make the fix. This should have a positive impact on performances, however.

rskj-core/src/main/java/co/rsk/core/bc/TransactionConflictDetector.java
levalleux-ludo: Here again, the only case when you remove the current partition from the accessors list of the account is when the same partition is deleting this account.
So, in the end there will be a conflict with any other partition that would attempt to access the same account, even if the accessors list has been purged (unintentionally I confess).
Anyway, I fix the bug here too, and I take this opportunity to factorize this code to reduce duplication.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
levalleux-ludo: I've added some comments to explain better

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
levalleux-ludo: During block mining, in case a transaction is invalid, we actually have started to execute it. All the state changes will be rollbacked thanks to the cache system at the repository level. Nevertheless, the conflict detector has recorded the accesses to the state keys from this transaction execution. We also need to purge the access when discarding the tx from the block. Then, we have 2 options :
1.	We can record accesses to state keys from the partition executing the transaction and, if it is discarded, we purge the entries from that partition in internal maps. But this reverse mechanism is not really efficient because the accesses are mapped ‘per key’, not ‘per partition’ (for performance reason in the nominal cases).
2	Or, we can cache the accesses to state keys from the partition executing the transaction and, if it is NOT discarded, we merge the entries from that partition in internal maps.

I’ve started with implementing the 1st option, successfully, but I was not happy with the resulting purging code, not really efficient and a too tricky for maintenance.

Then, I’ve decided to refactor towards the second option using a derivated class of TransactionConflictDetector that caches the entries until the commitPartition() method is called (merge implementation). Calling commitPartition() is only called if the transaction is valid, otherwise, we just clear the cache to ignore all accesses from the associated partition. 

Ultimately, TransactionConflictDetector is still used for block validation (executing txs in parallel regarding the partitioning) to detect conflicts and immediately stop block execution if such a case happens, whereas TransactionConflictDetectorWithCache is used for block mining.

I’ve added a comment to describe the TransactionConflictDetectorWithCache class and explain what this class is used for.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
levalleux-ludo: Sure !
Done.

rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
levalleux-ludo: The original file (in master branch) specifies only 2 valid sizes : 16 and 19.
Look at ; https://github.com/rsksmart/rskj/blob/f747100c327e72e8ab29f0ff9d53bfd5c7e95a1d/rskj-core/src/main/java/org/ethereum/core/BlockFactory.java#L143-L148

As I've added the partitionEnds fields, now 17 and 20 are allowed too.

Also I've removed the test on line 192 because I though it was redondant with the one at the beginning of the method (line 143 on master, see above),
https://github.com/rsksmart/rskj/blob/f747100c327e72e8ab29f0ff9d53bfd5c7e95a1d/rskj-core/src/main/java/org/ethereum/core/BlockFactory.java#L192

levalleux-ludo:
First, let me try to explain better:
During the pre-run, each executed transaction effectively changes the world state in a cached repository, that is kept all over the loop.
Then, when executing transaction B after transaction A, all effects from A are taken into account by B
So, I think there is no reason why the transaction B could produce different changes in the read/write keys because it’s already taking into account the world state after transaction A.
Saying that, we assume we take care of the order the transactions are executed, as soon as they are accessing the same keys in world state. That’s why for a given partition, the order of transaction is given by their respective order in the input list of the block being mined, and this order is maintained by the partitioning mechanism.
Only after all transactions have been pre-run, the cached repository used for this phase is clear (not committed) because we expect the world state resulting from this pre-run may differ from the one obtained after executing the transactions in parallel regarding to the computed partitioning. (I assume that, per precaution, but may be wrong?).
Then, we run the transaction list again, but this time using parallelization and regarding the computed partitioning (exactly in the same way the validators will do it after the block is broadcasted). This time, we consider the world state resulting from this execution as authentic and we publish the BlockResult with it.

I wish I would create some additional test with contract transactions but I didn’t find pretty simple example to inspire me, then it appears to me quite tricky to do it. Then, based on your remark, I’m trying to find some usecases that cause problem.

- Say Alice account balance is 150
- Say a transaction TA transferring 100 from Alice account to another account.
- Say a transaction TB is calling the following method of a contract :
`IF balance of account Alice >100 
THEN transfer 100 from account Bob to account Carol 
ELSE transfer 100 from account Denise to account Elise.
`
Executing TA then TB gives a very different result than executing TB then TA.
But, let say during the pre-run TA is executed BEFORE TB, then TA writes Alice key, TB is reading Alice key and writing Denise and Elise keys. 
Because the 2 transactions are in conflict (read Alice / write Alice), they will be added to the same partition, in the same order (TA then TB).
Then, during the concurrent execution, TA will effectively be executed BEFORE TB, again. It should give exactly the same result.
Am I wrong?
Do you see some other usecases that would really cause problems?

SergioDemianLerner:
@levalleux-ludo:
Now I understand the algorithm. I'll take a bit more time to think about it.
For the next hackathon that requires designing and writing complex code I will add to the requirements a textual description of the algorithms proposed, to facilitate evaluation.

PR #1096, opened 2019-11-12
Improved DataSourceWithCache: auto flush, sync locks; TrieStore avoid to write known state

ajlopez:
The implementation of `DataSourceWithCache` is improved with:

- Automanage of write flush
- Synchronization locks to support multithreading

And `TrieStoreImpl` avoid the write of a known trie world state

It includes and improve #1079 #1078 #1077

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
+++ b/rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
@@ -23,6 +23,7 @@ import org.ethereum.db.ByteArrayWrapper;
 import org.ethereum.util.ByteUtil;
 
 import java.util.*;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.BiFunction;
 import java.util.function.BinaryOperator;
 import java.util.function.Function;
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
```diff
@@ -21,92 +21,152 @@
...
 
+    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
+
     public DataSourceWithCache(KeyValueDataSource base, int cacheSize) {
+        this.cacheSize = cacheSize;
         this.base = base;
-        this.uncommittedCache = new HashMap<>();
-        this.committedCache = new MaxSizeHashMap<>(cacheSize, true);
+        this.uncommittedCache = new LinkedHashMap<>(cacheSize / 8, (float)0.75, false);
+        this.committedCache = Collections.synchronizedMap(new MaxSizeHashMap<>(cacheSize, true));
```
nacho692: Having a synchronized map here will needlessly lock every get to the `comittedCache` and make the multiple concurrency of `DataSourceWithCache#get()` useless.

I'd rather use the `WriteLock`, a `ConcurrentMap` or another `Lock`, the problem with `concurrentMap` is that we are losing the `MaxSize` property. 
We could roll our own `concurrentMasSizeHashMap` but that might be out of the scope of this PR, so i'd just use the `WriteLock`.

ajlopez: Add write lock, done!

nacho692:
pipeline: run

ajlopez:
Replaced by #1151

PR #1100, opened 2019-11-14
Refactor BridgeStorageProvider to use Activations class to check rskips

josedahlquist:
- Removed BridgeStorageConfiguration class
- Added ActivationConfig.ForBlock parameter in BridgeStorageProvider
- Replaced checks for activations in BridgeStorageProvider
- Refactored configurations in tests to use ActivationConfigForTest where possible


rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -27,15 +27,19 @@
...
 import java.io.IOException;
 import java.util.*;
 
+import static org.ethereum.config.blockchain.upgrades.ConsensusRule.RSKIP123;
+import static org.ethereum.config.blockchain.upgrades.ConsensusRule.RSKIP87;
+
 /**
  * Provides an object oriented facade of the bridge contract memory.
- * @see co.rsk.remasc.RemascStorageProvider
+ * @see co.rsk.peg.BridgeStorageProvider
```
marcos-iov: References itself?

josedahlquist: good catch, I'll resolve this in a separated PR though.

PR #1108, opened 2019-11-19
Change log data and address format

nacho692:
Changing log values based on ethereum specifications. Empty `data` should resolve to `0x` instead of `0x00`, same with `address`
*fed:version120* *fit:120*


lsebrie:
Pipeline: run

lsebrie:
Pipeline: run

lsebrie:
Pipeline: run

PR #1110, opened 2019-11-20
Chainid

smcsicardi:
PR for the ChainID opcode


smcsicardi:
> This looks great!
> I'd add a test verifying that `0x46` maps to `CHAINID` operation. This is to avoid a future possible scenario in which either the number is changed involuntarily or another operation is also mapped to `0x46`
> 
> It should be easy with `OpCode.code(byte code)`

Great, I've added a test that check both places where the opcode number 0x46 is created (Opcode and Opcodes classes).

PR #1113, opened 2019-11-25
Improving remasc blocks retrieval; removing unused BlockStore method

ajlopez:
Remasc precompiled contract usually retrieved 4000 blocks in mainchain, each time. Now, the retrieve of the execution block (current block - maturity aprox) was replaced by an algorithm using in other uses cases. The old method used was removed.


rskj-core/src/main/java/co/rsk/remasc/Remasc.java
```diff
@@ -115,11 +115,11 @@ void processMinersFees() {
         int uncleGenerationLimit = constants.getUncleGenerationLimit();
         Deque<Map<Long, List<Sibling>>> descendantsBlocks = new LinkedList<>();
 
-        // this search can be optimized if have certainty that the execution block is not in a fork
-        // larger than depth
-        Block currentBlock = blockStore.getBlockByHashAndDepth(
-                executionBlock.getParentHash().getBytes(),
-                remascConstants.getMaturity() - 1 - uncleGenerationLimit
+        // this search is now optimized if have certainty that the execution block is not in a fork
```
nacho692: This comment should be removed

nacho692 (dismissed):
Nice

PR #1114, opened 2019-11-25
Return 0x instead of 0x00 in empty byte array results via JSON RPC

ajlopez:
Some formatting code was changed from `TypeConverter.toJsonValue` to `TypeConverter.toUnformattedJsonHex`.

With this code, getting code from a contract that IS NOT a contract, returns `0x`, like other implementations (ganache-cli)

And if a call to a contract returns nothing, it returns 0x instead of `0x00`, again, like other implementations.

Byte arrays in logs were also changed.

ADVICE: when this code was stabilized, detect the few use cases were a `0x00` return value is needed. Apparently, there is none associated to a byte array, only to other numeric values (maybe represented in byte arrays). Then, change the names of the `TypeConverter` methods to more descriptive ones.

*fed:version120* *fit:120*


lsebrie:
Pipeline: run

lsebrie:
Pipeline: run

PR #1116, opened 2019-11-26
Synchronized savedTries

nacho692:
Now trieStore is thread-safe iif the underlying store is thread-safe

*fed:version120* *fit:120*


nacho692:
Pipeline: run

PR #1117, opened 2019-11-27
Selfbalance opcode

smcsicardi:
Simple implementation for the SELFBALANCE opcode, given that it is asking for the balance of the owner of the contract, it is assumed that the balance is already in cache, and thus is cheaper than a call to BALANCE.

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/vm/OpCode.java
+++ b/rskj-core/src/main/java/org/ethereum/vm/OpCode.java
@@ -262,6 +262,10 @@ public enum OpCode {
      * (0x45) Get the block’s gas limit
      */
     GASLIMIT(0x45, 0, 1, BASE_TIER),
+    /**
+     * (0x47) Get the senders balance
+     */
+    SELFBALANCE(0x47, 0, 1, LOW_TIER),
 
     /**
      * (0x46) Get the chain id
@@ -283,7 +287,7 @@ public enum OpCode {
      */
     MSTORE(0x52, 2, 0, VERY_LOW_TIER),
     /**
-     * (0x53) Save byte to memory
+     * `(0x53) Save byte to memory
      */
     MSTORE8(0x53, 2, 0, VERY_LOW_TIER),
     /**
--- a/rskj-core/src/main/java/org/ethereum/vm/VM.java
+++ b/rskj-core/src/main/java/org/ethereum/vm/VM.java
@@ -1071,6 +1071,19 @@ public class VM {
         program.step();
     }
 
+    protected void doSELFBALANCE(){
+        spendOpCodeGas();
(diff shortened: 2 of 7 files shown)


rskj-core/src/main/java/org/ethereum/vm/OpCode.java
```diff
@@ -262,6 +262,10 @@
      * (0x45) Get the block’s gas limit
      */
     GASLIMIT(0x45, 0, 1, BASE_TIER),
+    /**
+     * (0x47) Get the senders balance
+     */
+    SELFBALANCE(0x47, 0, 1, LOW_TIER),
```
fedejinich: verify that selfbalance gas cost matches cost from [this](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1884.md) EIP 

> SELFBALANCE is priced as GasFastStep, at 5 gas.

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1067,6 +1067,19 @@ protected void doCHAINID() {
         program.step();
     }
 
+    protected void doSELFBALANCE(){
+        spendOpCodeGas();
+        // EXECUTION PHASE
+        DataWord balance = program.getBalance(program.getOwnerAddress());
+
+        if (isLogEnabled) {
+            hint = "gaslimit: " + balance;
```
lsebrie: Why the hint says "gaslimit"?

rskj-core/src/test/java/co/rsk/vm/VMExecutionTest.java
```diff
@@ -710,14 +712,73 @@ private void executeCHAINIDWithActivations(byte chainIDExpected, ActivationConfi
...
+        int balanceValue = 100;
+
+        RskAddress testAddress = new RskAddress(invoke.getCallerAddress());
+        invoke.setOwnerAddress(testAddress);
+        invoke.getRepository().addBalance(testAddress, Coin.valueOf(balanceValue));
+
+        Program program = executeCodeWithActivationConfig("SELFBALANCE", 1, activations);
+        Stack stack = program.getStack();
+
+        long gasSpent = 5;
```
lsebrie: Add a constant instead of a var for this value

rskj-core/src/main/java/org/ethereum/vm/VM.java
smcsicardi: Copy paste typo, corrected

rskj-core/src/test/java/co/rsk/vm/VMExecutionTest.java
smcsicardi: Added

PR #1123, opened 2019-12-14
Decoupling Trie from Repository

ajlopez:
TBD


SergioDemianLerner:
I disagree with this change for several reasons:
-It will force us to throw to the trash the existent solutions for storage rent and parallel transaction executions. It will delay our scaling roadmap for 2020. ajlopez has a different parallel transaction execution code, which may be compatible with this change, and therefore should be discussed in parallel to this change.
- It seems to prevent direct access to trie nodes for retrieving old node timestamps to compute storage rent. Therefore it only allow per-account storage rent, which was specified in RSKI61 and was found to be problematic for the prallelization of ERC20 tokens. RSKIP113 requires access to individual node's timestamps.
- It prevents uniform detection key/value collisions in parallel execution. It forces the execution engine to add thread-id information for every kind of object that is monitored for concurrent access (AccountState / code / storage-cells / other future objects such as LPTCP presets). The mutableTrie shows uniformity in concurrent access detection in all types of objects.

AFAIK a good improvement to mutableTrie is decoupling the AccountState prefix check, and moving that to Repository.

ajlopez:
Well, this proposal has no relation with my parallel transaction proposal.

> * It will force us to throw to the trash the existent solutions for storage rent and parallel transaction executions. It will delay our scaling roadmap for 2020. ajlopez has a different parallel transaction execution proposal (without code yet), which may be compatible with this change, and therefore should be discussed in parallel to this change.

I think that the existent solutions could be easily adapted to this code. Where is the code? Any difficulty in adapting the code could be a code smell: the repository should be decoupled from trie, and any solution should be follow that.

Ie, any retrieval of old storage could be detected in the `TopRepository` and auxiliary code. Or directly in trie wrapper.

> * It seems to prevent direct access to trie nodes for retrieving old node timestamps to compute storage rent. Therefore it only allow per-account storage rent, which was specified in RSKI61 and was found to be problematic for the prallelization of ERC20 tokens. RSKIP113 requires access to individual node's timestamps.

I don't see so. That query could be easy added to this proposal.

> * It prevents uniform detection key/value collisions in parallel execution. It forces the execution engine to add thread-id information for every kind of object that is monitored for concurrent access (AccountState / code / storage-cells / other future objects such as LPTCP presets). The mutableTrie shows uniformity in concurrent access detection in all types of objects.

Again, a trie wrapper could be added at any moment to `TopRepository`.
> 
> AFAIK a good improvement to mutableTrie is decoupling the AccountState prefix check, and moving that to Repository.

Yes, it really use logical access instead of trie calisthenics. I will add a comment with some performance comparison.

ajlopez:
In a particular machine (the results could vary in different machines, but the relations are similar), I just tested adjacent mainnet nodes in long synchronization.

The result:
Using master code:
no blocks 200
process time (ns) 13429801594

ie 13s

Using this proposal
no blocks 200
process time (ns) 3815684598

ie 3 seconds

A boost in performance of 4x

We should test this performance using more complete blocks (usually mainnet has few transactions per block) and transactions with more smart contract logic invoked

SergioDemianLerner:
The cause of the current trie code being slow may have nothing to do with the MutableTrie cache class, but with some other bottleneck introduced by core. For example, heavy use of slow keccak hash in key-mapping. The original code had a small keccak cache to avoid repeated hashing. I suggest we repeat the performance test with @raullaprida test suite that enables us to detect where the bottlenecks are.

SergioDemianLerner:
To be clear, is not that I'm against improving the code. I just need to know for sure that this new code supports parallel transaction execution the way it was designed and improved over the years. Implementing this and then reverting to the current class system 4 months later would be a pita.
If you really really want to push this at this time, I suggest you implement parallel transaction processing together with this proposal.
We have the option to delay this change to after parallel transaction processing is merged.

SergioDemianLerner:
Regarding the slow keccak mapping: we already have a native implementation that we can use for this purpose, so we should compare apples-to-apples using this implementation

PR #1126, opened 2019-12-19
Set wasabi plus one activation height in zero for regtest

smcsicardi:


aeidelman:
@smcsicardi can you please resolve conflicts so we can merge?

Issue #1130, opened 2019-12-20
Transaction queued for ever (until timeout discard) and not relayed in TransactionPool


ajlopez:
This tests (add it to `TransactionPoolImplTest`) fails:

```
    @Test
    public void addTwoTransactionsOutOfOrderInNonceUsingAddTransaction() {
        Coin balance = Coin.valueOf(1000000);
        createTestAccounts(2, balance);

        Transaction tx1 = createSampleTransactionWithGasPrice(1, 0, 1000, 0, 1);
        Transaction tx2 = createSampleTransactionWithGasPrice(1, 0, 2000, 1, 2);

        Assert.assertTrue(transactionPool.addTransaction(tx2).transactionWasAdded());
        Assert.assertTrue(transactionPool.addTransaction(tx1).transactionWasAdded());

        Assert.assertTrue(transactionPool.getPendingTransactions().stream().anyMatch(tx -> tx.getHash().equals(tx2.getHash())));
        Assert.assertTrue(transactionPool.getPendingTransactions().stream().anyMatch(tx -> tx.getHash().equals(tx1.getHash())));
    }
```

because it uses `TransactionPool.addTransaction` directly. I don't know any case to use this method from OUTSIDE. The problem is that the out of order transaction is queued when it is added, but it is NOT removed from queued transactions and added to pending transactions when the missing companion transaction is added.

Compare with this tests that runs OK:

```
    @Test
    public void addTwoTransactionsOutOfOrderInNonceUsingAddTransactions() {
        Coin balance = Coin.valueOf(1000000);
        createTestAccounts(2, balance);

        Transaction tx1 = createSampleTransactionWithGasPrice(1, 0, 1000, 0, 1);
        Transaction tx2 = createSampleTransactionWithGasPrice(1, 0, 2000, 1, 2);

        List<Transaction> txs = new ArrayList<>();

        txs.add(tx2);
        txs.add(tx1);

        transactionPool.addTransactions(txs);

        Assert.assertTrue(transactionPool.getPendingTransactions().stream().anyMatch(tx -> tx.getHash().equals(tx2.getHash())));
        Assert.assertTrue(transactionPool.getPendingTransactions().stream().anyMatch(tx -> tx.getHash().equals(tx1.getHash())));
    }
```

This second test uses `TransactionPoolImpl.addTransactions`. This is the expected method to use to keep the queued and pending transaction state OK.

[...]

PR #1134, opened 2019-12-21
Fixing transaction pool add transaction behaviour

### LINKED ISSUE #1132: Transaction events triggered twice in transaction process
The method `TransactionPoolImpl.addTransactions` invokes `TransactionPoolImpl.addTransaction`. If the transactions are accepted as new pending transactions, both methods trigger the events `onPendingTransactionsReceived` and `onTransactionPoolChanged` TWICE.

It's not the expected behaviour, and it could lead to additional spurious execution.

ajlopez:
The method `TransactionPoolImpl.addTransaction` was modified to consider the adding of queued successors.

This modificacion fixes #1132 and #1130  The tests mentioned in #1130 were added and run without failure.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
@@ -158,28 +158,36 @@ public class TransactionPoolImpl implements TransactionPool {
         return repositoryLocator.snapshotAt(getBestBlock().getHeader());
     }
 
+    private List<Transaction> addSuccesors(Transaction tx) {
+        List<Transaction> added = new ArrayList<>();
+        Optional<Transaction> succesor = this.getQueuedSuccesor(tx);
+
+        while (succesor.isPresent()) {
+            Transaction found = succesor.get();
+            queuedTransactions.removeTransactionByHash(found.getHash());
+
+            if (!this.internalAddTransaction(found).transactionWasAdded()) {
+                break;
+            }
+
+            added.add(found);
+
+            succesor = this.getQueuedSuccesor(found);
+        }
+
+        return added;
+    }
+
     @Override
     public synchronized List<Transaction> addTransactions(final List<Transaction> txs) {
         List<Transaction> added = new ArrayList<>();
 
         for (Transaction tx : txs) {
-            if (this.addTransaction(tx).transactionWasAdded()) {
+            TransactionPoolAddResult result = this.internalAddTransaction(tx);
+
+            if (result.transactionWasAdded()) {
                 added.add(tx);
-
-                Optional<Transaction> succesor = this.getQueuedSuccesor(tx);
-
-                while (succesor.isPresent()) {
-                    Transaction found = succesor.get();
-                    queuedTransactions.removeTransactionByHash(found.getHash());
-
-                    if (!this.addTransaction(found).transactionWasAdded()) {
-                        break;
-                    }
-
-                    added.add(found);
-
-                    succesor = this.getQueuedSuccesor(found);
-                }
+                added.addAll(this.addSuccesors(tx));
             }
         }
 
@@ -208,18 +216,10 @@ public class TransactionPoolImpl implements TransactionPool {
                 .findFirst();
     }
 
-    @Override
-    public synchronized TransactionPoolAddResult addTransaction(final Transaction tx) {
-        if (pendingTransactions.hasTransaction(tx)) {
-            return TransactionPoolAddResult.withError("pending transaction with same hash already exists");
-        }
-
-        if (queuedTransactions.hasTransaction(tx)) {
-            return TransactionPoolAddResult.withError("queued transaction with same hash already exists");
-        }
-
+    private TransactionPoolAddResult internalAddTransaction(final Transaction tx) {
         RepositorySnapshot currentRepository = getCurrentRepository();
         TransactionValidationResult validationResult = shouldAcceptTx(tx, currentRepository);
+
         if (!validationResult.transactionIsValid()) {
             return TransactionPoolAddResult.withError(validationResult.getErrorMessage());
         }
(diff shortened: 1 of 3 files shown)


rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -158,37 +158,49 @@ private RepositorySnapshot getCurrentRepository() {
         return repositoryLocator.snapshotAt(getBestBlock().getHeader());
     }
 
-    @Override
-    public synchronized List<Transaction> addTransactions(final List<Transaction> txs) {
+    private List<Transaction> addSuccesors(Transaction tx) {
```
Vovchyk, line 161: @ajlopez, It seems to me that `addSuccesors` and `getQueuedSuccesor` methods could be merged into single one? So we'd reduce a number of paths through the `queuedTransactions` collection.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -158,37 +158,49 @@ private RepositorySnapshot getCurrentRepository() {
...
-                    queuedTransactions.removeTransactionByHash(found.getHash());
+        while (succesor.isPresent()) {
+            Transaction found = succesor.get();
+            queuedTransactions.removeTransactionByHash(found.getHash());
 
-                    if (!this.addTransaction(found).transactionWasAdded()) {
-                        break;
-                    }
+            if (!this.internalAddTransaction(found).transactionWasAdded()) {
+                break;
```
Vovchyk, line 170: should we `break` here? `getQueuedSuccesor` method returns the first tx from the collection of transactions with the same nonce. There might be some other tx with the same nonce which potentially could be added? Another question: should the `getQueuedSuccesor` method return the first tx or the oldest one (from the list of txs with the same nonce)?

nicops: re: _There might be some other tx with the same nonce which potentially could be added?_, NO, but because the queuedTransactions collection is a TransactionSet, which is a particular class that doesn't allow more than one tx with the same nonce, see: https://github.com/rsksmart/rskj/blob/918e14623851fc58aaf3b7a39d2ef680adbc686f/rskj-core/src/main/java/org/ethereum/core/TransactionSet.java#L44

The way the method works is that the new tx always removes any preexisting tx with the same nonce in the TransactionSet.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -252,14 +264,25 @@ public synchronized TransactionPoolAddResult addTransaction(final Transaction tx
...
-        if (listener != null) {
-            EventDispatchThread.invokeLater(() -> {
-                listener.onPendingTransactionsReceived(Collections.singletonList(tx));
-                listener.onTransactionPoolChanged(TransactionPoolImpl.this);
-            });
+        return TransactionPoolAddResult.ok();
+    }
+
+    @Override
+    public synchronized TransactionPoolAddResult addTransaction(final Transaction tx) {
```
nicops, line 271: I would rewrite this to call addTransactions with a list of one to avoid duplicated logic, but it's a minor detail

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
ajlopez: but `getQueuedSuccesor` is called TWICE from `addSuccesors` in two different logic

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
ajlopez: Originally, transaction pool allowed many transactions from the same sender with the same nonce (with different values, and then, different hash). The last implementors decided to not allow such case. I'm not sure it is the right way. Maybe the client/user changed their intention, and the put another competing transaction, to replace an old one. Maybe we should review other implementations (geth, parity). 

But now, there are no transactions with same sender, same nonce

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
Vovchyk: I guess the current version is pretty sustainable and it's more about refactoring, reducing the amount of code and personal preference )

something like:

```java
        Map<BigInteger, Transaction> txMap = queuedTransactions.getTransactionsWithSender(tx.getSender()).stream()
                .collect(Collectors.toMap(Transaction::getNonceAsInteger, t -> t));

        Transaction nextSuccessor;
        BigInteger nextNonce = tx.getNonceAsInteger().add(BigInteger.ONE);
        while ((nextSuccessor = txMap.get(nextNonce)) != null && internalAddTransaction(nextSuccessor).transactionWasAdded()) {
            added.add(nextSuccessor);
            nextNonce = nextNonce.add(BigInteger.ONE);
        }
```

I also noticed a type in the `getQueuedSuccesor` method name and that `getTransactionsWithSender` cannot return `null`, so no need to check for `null` inside the `getQueuedSuccesor` method.


Inside the `getTransactionsWithSender` method I'd replace `return new ArrayList<>();` onto `Collections.emptyList()`, even though it's not related to the pull request.

patogallaiovlabs:
pipeline:run

patogallaiovlabs:
pipeline:run

patogallaiovlabs:
pipeline:run

josedahlquist:
pipeline: run

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
ajlopezrsk: Well, the `addTransactions` method returns a list of transaction, that it was the original intention of the method. Maybe we could use this results in the (possible) removal of `TransactionGateway`, to return to the original behavior: `TransactionPool` returns the added transactions, to be relayed to the other nodes by the caller object. So, we would visit this refactor in that upcoming issue

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
ajlopezrsk: OK, now using `Collections.emptyList()` and typos fixed, thanks!

ajlopezrsk:
run: pipeline

patogallaiovlabs:
pipeline:run

patogallaiovlabs:
pipeline:run

ajlopezrsk:
pipeline:run

ajlopezrsk:
pipeline:run

PR #1166, opened 2020-01-30
Difficulty divisor drop

smcsicardi:
Difficulty divisor is increased from 50 to 400

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
+++ b/rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
@@ -45,19 +45,18 @@ public class DifficultyCalculator {
             }
         }
 
-        return getBlockDifficulty(header, parentHeader, constants);
+        return getBlockDifficulty(header, parentHeader);
     }
 
-    private static BlockDifficulty getBlockDifficulty(
+    private BlockDifficulty getBlockDifficulty(
             BlockHeader curBlockHeader,
-            BlockHeader parent,
-            Constants constants) {
+            BlockHeader parent) {
         BlockDifficulty pd = parent.getDifficulty();
         long parentBlockTS = parent.getTimestamp();
         int uncleCount = curBlockHeader.getUncleCount();
         long curBlockTS = curBlockHeader.getTimestamp();
         int duration = constants.getDurationLimit();
-        BigInteger difDivisor = constants.getDifficultyBoundDivisor();
+        BigInteger difDivisor = constants.getDifficultyBoundDivisor(activationConfig.forBlock(curBlockHeader.getNumber()));
         BlockDifficulty minDif = constants.getMinimumDifficulty();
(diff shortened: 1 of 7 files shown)


rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -48,7 +48,7 @@ public BlockDifficulty calcDifficulty(BlockHeader header, BlockHeader parentHead
         return getBlockDifficulty(header, parentHeader, constants);
     }
 
-    private static BlockDifficulty getBlockDifficulty(
+    private BlockDifficulty getBlockDifficulty(
             BlockHeader curBlockHeader,
             BlockHeader parent,
             Constants constants) {
```
M-Picco: Since the method is no longer static I think the `constants` parameter could be dropped

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -58,6 +58,11 @@ private static BlockDifficulty getBlockDifficulty(
         long curBlockTS = curBlockHeader.getTimestamp();
         int duration = constants.getDurationLimit();
         BigInteger difDivisor = constants.getDifficultyBoundDivisor();
+        if (activationConfig.isActive(ConsensusRule.RSKIP156, curBlockHeader.getNumber())
+            && constants.getChainId() != Constants.REGTEST_CHAIN_ID) {
+            // Unless we are in regtest, this RSKIP increments the difficulty divisor from 50 to 400
+            difDivisor = BigInteger.valueOf(400);
```
M-Picco: Could a constant be introduced for the value `400`?

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -58,6 +58,11 @@ private static BlockDifficulty getBlockDifficulty(
         long curBlockTS = curBlockHeader.getTimestamp();
         int duration = constants.getDurationLimit();
         BigInteger difDivisor = constants.getDifficultyBoundDivisor();
+        if (activationConfig.isActive(ConsensusRule.RSKIP156, curBlockHeader.getNumber())
```
M-Picco: This new block could be moved to a private method which returns the final divisor in order to improve legibility.

rskj-core/src/main/java/org/ethereum/config/Constants.java
```diff
@@ -89,6 +89,7 @@ public BlockDifficulty getFallbackMiningDifficulty() {
     }
 
     public BigInteger getDifficultyBoundDivisor() {
```
M-Picco: Since this "constant" is no longer constant maybe we should find a way to remove it from here

smcsicardi:
> The code changes seem OK, what I would like to document here is the rationale for this decision?

I talked with @SergioDemianLerner about it. Will document this on the RSKIP

PR #1167, opened 2020-01-31
Signature cache v2

julianlen:


rskj-core/src/main/java/org/ethereum/core/ReceivedTxSignatureCache.java
```diff
@@ -23,10 +42,27 @@ public RskAddress getSender(Transaction transaction) {
             return RemascTransaction.REMASC_ADDRESS;
         }
 
-        return addressesCache.computeIfAbsent(transaction, Transaction::getSender);
```
joaquinlpereyra-iov: Why change this for the if? Looks like this is a better option.

diega:
#1050 😢

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
```diff
@@ -710,6 +711,52 @@ public void checkTxWhichCanNotBePaidIsRejected() {
...
+        Coin balance = Coin.valueOf(1000000);
+        createTestAccounts(2, balance);
+        Account account1 = createAccount(1);
+        Transaction tx = createSampleTransaction(1, 2, 1000, 0);
+        transactionPool.addTransaction(tx);
+
+        Assert.assertTrue(signatureCache.containsTx(tx));
+        Assert.assertArrayEquals(signatureCache.getSender(tx).getBytes(), account1.getAddress().getBytes());
+    }
+
```
pmprete: could you add a spy to signatureCache = rskTestContext.getReceivedTxSignatureCache();
And create a test where you add two times the same transaction, then check that the cached called only once the storeSender for that transaction and the getSender was called at least twice

rskj-core/src/test/java/co/rsk/test/builders/BlockChainBuilder.java
```diff
@@ -206,19 +206,24 @@ public BlockChainImpl build() {
...
+                blockTxSignatureCache
         );
         repositoryLocator = new RepositoryLocator(trieStore, stateRootHandler);
+
         transactionPool = new TransactionPoolImpl(
                 config, repositoryLocator, this.blockStore, blockFactory, new TestCompositeEthereumListener(),
-                transactionExecutorFactory, 10, 100
-        );
+                transactionExecutorFactory, receivedTxSignatureCache, 10,
+                100);
```
pmprete: put the 100 on the same line as the rest

rskj-core/src/main/java/org/ethereum/core/BlockTxSignatureCache.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public void storeSender(Transaction transaction) {
+
+        if (transaction instanceof RemascTransaction) {
+            return;
+        }
+
+        if (cacheN2.containsTx(transaction)) {
+            RskAddress sender = cacheN2.getSender(transaction);
```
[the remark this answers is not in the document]
julianlen: The problem with that solution is `getSender` method has two chances
1. If sender is in cache, it returns the queried sender
2. If it is not, the cache computes sender from the signature and then returns it. 

Then in the case which BlockTxSignatureCache queries a transaction sender to cacheN2, it first needs to verify if cacheN2 containsTx and in affirmative case, compute cacheN2.getSender. Because in the opposite case, cacheN2 will extract sender from the signature (which is not desired)

rskj-core/src/main/java/org/ethereum/core/BlockTxSignatureCache.java
```diff
@@ -0,0 +1,78 @@
...
+            addressesCache.put(transaction, sender);
+            return sender;
+        }
+
+        return transaction.getSender();
+    }
+
+    @Override
+    public boolean containsTx(Transaction transaction) {
```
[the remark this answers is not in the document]
julianlen: GetSender is responsible for returns sender if it cache has the transaction or compute the sender, store in the cache and then returns it.
Null is never returned.

rskj-core/src/main/java/org/ethereum/core/BlockTxSignatureCache.java
```diff
@@ -0,0 +1,78 @@
...
+        if (transaction instanceof RemascTransaction) {
+            return RemascTransaction.REMASC_ADDRESS;
+        }
+
+        if (containsTx(transaction)) {
+            return addressesCache.get(transaction);
+        }
+
+        if (cacheN2.containsTx(transaction)) {
+            RskAddress sender = cacheN2.getSender(transaction);
```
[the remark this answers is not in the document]
julianlen: You repeat the comment here https://github.com/rsksmart/rskj/pull/1167#discussion_r375812188

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
julianlen: i will do this in "integration" tests

rskj-core/src/main/java/org/ethereum/core/BlockTxSignatureCache.java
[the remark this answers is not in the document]
pmprete: What i mean is instead of 
if (cacheN2.containsTx(transaction)) {
            RskAddress sender = cacheN2.getSender(transaction);
            addressesCache.put(transaction, sender);
}
Use
RskAddress sender = cacheN2.getSender(transaction);
if (sender != null) {
            addressesCache.put(transaction, sender);
}

So you have only 1 access

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
pmprete: ok

julianlen:
> I'm not a fan of `ReceivedTxSignatureCache` and `BlockTxSignatureCache` being so similar, I feel like there's an abstraction there waiting to happen where one class can fulfill both roles easily.
> 
> `ìnternalCache` could be an optinal, or even a stack.
> `MAX_CACHE_SIZE` could be decided in the initialization and injected by the creator.
> 
> But I don't like bikeshedding, so I'm just commenting, if that makes sense...
> 
> Other questions:
> 
> 1. Why in one implementation the `MaxSizeHashMap` is ordered by accessOrder and in the other one it does not? I don't see us iterating over the map, this would only mean that the map performs extra work when `get`ting from the hashamp, am I wrong?
> 2. Are we sure access and writing is done by at most one thread at the time? The structure is not synced by default.

1. FYI the access order determines if the hashmap is LRU (true) or FIFO (false). The block signature cache has a FIFO hashmap because after three blocks, there is a really low probability that a received tx appear in a new received block. Furthermore, the tx received sign cache, is an LRU cache because if a transaction is broadcasted it will be high probability of receive the same tx from the network. What are you talking about when said "map performs extra work when `get`ting from the hashamp" ?

2. If the cache is acceded by many threads, in the worst case the tx sender wont be stored in cache and the signature is going to be computed many times (as we have today).

jjtechuy:
QA: Approved based on benchmarks (profiler) results and functional tests on these changes.

PR #1170, opened 2020-02-13
First check if the transaction is already added in TransactionPoolAddResult

pmprete:
Validating a transaction is more expensive than verifying if it's already added in a Hash map.
Performing first the check if the Tx hash exists in pendingTransactions or queuedTransactions improves performance


pmprete:
> It makes sense. Is there a settled variable when tx is verified? (may be there is stored some evidence that the transaction is invalid)

The value that is stored is only for informing the RPC caller https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/org/ethereum/core/TransactionPoolAddResult.java#L39

PR #1172, opened 2020-02-14
initial queue commit

joaquinlpereyra-iov:


joaquinlpereyra-iov:
pipeline: run

joaquinlpereyra-iov:
pipeline: run

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -123,7 +124,10 @@ private void tryAddMessage(Peer sender, Message message) {
                 }
                 this.receivedMessages.add(encodedMessage);
             }
-            if (!this.queue.offer(new MessageTask(sender, message))){
+
+            double score = sender.score(System.currentTimeMillis(), message.getMessageType());
+
+            if (score >= 0 && !this.queue.offer(new MessageTask(sender, message, score))){
```
joaquinlpereyra-iov: This condition is never true, as `this.queue.offer` never returns false: 
```
     * Inserts the specified element into this priority queue.
     * As the queue is unbounded, this method will never return {@code false}.
```

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -236,6 +247,31 @@ public String toString() {
                     ", message=" + message +
                     '}';
         }
+
+        @Override
+        public int compareTo(MessageTask messageTask) {
```
joaquinlpereyra-iov: I think it's weird to have `MessageTask` be naturally ordered by their score. It certainly helps in this implementation, but it is weird to read `taskA > taskB` and assign it any meaning. Maybe use the other constructor to the queue which allows to define a `Comparator<MessageTask>`?

Maybe I'm being too picky here, take just a note/discussion.

joaquinlpereyra-iov:
pipeline: run

PR #1175, opened 2020-02-18
Improve transaction execution with trace in BlockExecutor

ajlopezrsk:
Avoid saving end state when reexecute a transaction


jjtechuy:
pipeline: run

PR #1180, opened 2020-03-02
Update network upgrade name from wasabiPlusOne to papyrus200

josedahlquist:
*fit: rename-network-upgrade*


josedahlquist:
pipeline:run

rskj-core/src/main/resources/reference.conf
```diff
@@ -1,5 +1,5 @@
 # disable wasabi+1 consensus rules by default for every network
```
marcos-iov: Change this comment to # disable papyrus?

PR #1171, opened 2020-02-14
UMM network upgrade

SergioDemianLerner:
Plus BlockHeaderBuilder refactor

PR #1182, opened 2020-03-03
UMM Upgrade

M-Picco:
Introduce consensus changes to pave down the road for future UMM changes.

This PR decouples the UMM changes from https://github.com/rsksmart/rskj/pull/1171

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
+++ b/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
@@ -515,7 +515,7 @@ public class BlockHeader {
     }
 
     private byte[] getHashRootForMergedMining(byte[] leftHash) {
-        if ((ummRoot.length != UMM_LEAVES_LENGTH) && (ummRoot.length != 0)){
+        if (ummRoot.length != UMM_LEAVES_LENGTH){
             throw new IllegalStateException(
                     String.format("UMM Root length must be either 0 or 20. Found: %d", ummRoot.length)
             );
--- a/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
+++ b/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
@@ -47,6 +47,7 @@ public class BlockHeader {
 
     private static final int HASH_FOR_MERGED_MINING_PREFIX_LENGTH = 20;
     private static final int FORK_DETECTION_DATA_LENGTH = 12;
+    private static final int UMM_ROOT_LENGTH = 20;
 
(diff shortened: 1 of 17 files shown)


M-Picco:
pipeline:run

M-Picco:
pipeline:run

M-Picco:
pipeline:run

M-Picco:
pipeline:run

M-Picco:
pipeline:run

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -496,6 +514,20 @@ public String getShortHashForMergedMining() {
         return hashForMergedMining;
     }
 
+    private byte[] getHashRootForMergedMining(byte[] leftHash) {
+        if ((ummRoot.length != UMM_LEAVES_LENGTH) && (ummRoot.length != 0)){
```
SergioDemianLerner: This method does not work when ummRoot.length == 0. The exception should be thrown on ummRoot.length == 0 also

SergioDemianLerner: This method should not be called when ummRoot.length ==0. It's better to throw an exception in that case.

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -349,6 +353,10 @@ public Coin getMinimumGasPrice() {
         byte[] uncleCount = RLP.encodeBigInteger(BigInteger.valueOf(this.uncleCount));
         fieldToEncodeList.add(uncleCount);
 
+        if (this.ummRoot != null) {
```
joaquinlpereyra-iov: Should this be the condition an not `isActive(umm)`? For example, `getPendingState` always create a `null` `ummRoot`, even after activation... If coded this way, we will have two kind of blocks inside the node, some which encode to something with ummroots, and some which dont... is this intended? Sounds like it will complicate things at some point.

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -477,9 +485,19 @@ public String getShortHashForMergedMining() {
         return HashUtil.shortHash(getHashForMergedMining());
     }
 
+    public boolean isUMMBlock() {
```
joaquinlpereyra-iov: Same comments as above, shouldn't all blocks be UMM Blocks after the activation?

PR #1183, opened 2020-03-04
Light client capability handshake

julianlen:


rskj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
```diff
@@ -375,18 +376,23 @@ private void recordEvent(ChannelHandlerContext ctx, EventType event) {
     private void processHelloMessage(ChannelHandlerContext ctx, HelloMessage helloMessage) {
         List<Capability> capInCommon = configCapabilities.getSupportedCapabilities(helloMessage);
         channel.initMessageCodes(capInCommon);
+        List<Capability> capsToActivate = new LinkedList();
         for (Capability capability : capInCommon) {
             // It seems that the only supported capability is RSK, and everything else is ignored.
```
smcsicardi: we should change this comment to reflect new behaviour

PR #1184, opened 2020-03-05
Light client capabilities

julianlen:
In the current PR, the following features were added:

- Light Client as a capability

- LightClientMessageCode and LightClientMessageFactory: When a message arrives at the node, the MessageCodec class catches it and then creates a light client message.

- Channel: This class accepts Light Client as a capability. In this class, the handshake at an RLPx level starts. Once the handshake is finished, a second handshake starts for each protocol (In the Light client protocol, a status message is sent). 

- Light Client peer: Represents the channel between our node and a light client peer, it wraps the Channel class.

- LightClientMessageHandler: It starts in the Channel class and waits for a new message (performed in the netty ChannelRead0 method). Then, using a visitor pattern design, the new message is settled as a light client message and finally is sent to some processor.

- LightProcessor and SyncProcessor: They process the messages. The first one will be used for messages On-Demand Request (ODR) related. At the moment, the processing for ODR messages hasn't been implemented yet. The second one processes messages that were involved in the sync (e.g. status message, getBlockHeader, blockHeader).

- Finally each light client message.


rskj-core/src/main/java/co/rsk/net/light/LightPeer.java
```diff
@@ -0,0 +1,61 @@
...
+    public void sendMessage(Message message) {
+        msgQueue.sendMessage(message);
+    }
+
+    public void disconnect(ReasonCode reasonCode) {
+        msgQueue.disconnect(reasonCode);
+    }
+
+    public void receivedBlock(BlockHeader blockHeader) {
+        blockHeaders.add(blockHeader);
```
nicops, line 55: This seems to be for testing only, since the getBlocks() method is only called from test code, maybe it merits a comment or a TODO.

rskj-core/src/main/java/co/rsk/net/light/MessageVisitor.java
```diff
@@ -0,0 +1,125 @@
...
+import co.rsk.net.eth.LightClientHandler;
+import co.rsk.net.light.message.*;
+import io.netty.channel.ChannelHandlerContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Created by Julian Len and Sebastian Sicardi on 21/04/20.
+ */
+public class MessageVisitor {
```
nicops, line 30: Since there is another MessageVisitor, even if it is on a different namespace, I'm tempted to change the name, (MessageVisitor isn't on itself very meaningful either), maybe to something like LightClientMessageVisitor?

rskj-core/src/main/java/co/rsk/core/bc/BlockChainStatus.java
```diff
@@ -50,4 +51,8 @@ public BlockDifficulty getTotalDifficulty() {
     public boolean hasLowerTotalDifficultyThan(Status status) {
         return this.totalDifficulty.compareTo(status.getTotalDifficulty()) < 0;
     }
+
+    public boolean hasLowerDifficultyThan(LightStatus status) {
```
nicops, line 55: Since the calling code that uses this method says "this works only for testing purposes", should this method have a similar comment? Or will it make sense later?

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.LinkedList;
+import java.util.List;
+
+
+/**
+ * Created by Julian Len and Sebastian Sicardi on 21/10/19.
+ */
+public class LightProcessor {
+    private static final Logger logger = LoggerFactory.getLogger("lightprocessor");
+    // keep tabs on which nodes know which blocks.
```
nicops, line 45: This comment seems off. That's not what the BlockStore does, or is it?

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
```diff
@@ -0,0 +1,222 @@
...
+     * @param requestId the id of the request
+     * @param blockHash   the requested block hash.
+     * @param lightPeer
+     */
+    public void processGetBlockReceiptsMessage(long requestId, byte[] blockHash, LightPeer lightPeer) {
+        String blockHashLog = Hex.toHexString(blockHash);
+        logger.trace("Processing block receipts request {} block {}", requestId, blockHashLog);
+        final Block block = blockStore.getBlockByHash(blockHash);
+
+        if (block == null) {
```
nicops, line 68: For this and all other methods, will this be the long term solution, to respond nothing? In https://openethereum.github.io/wiki/The-Parity-Light-Protocol-%28PIP%29 there is a mention of exclusion proofs, would that apply here? Alternatively, should we account for spam in some way?

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
```diff
@@ -0,0 +1,222 @@
...
+            // Don't waste time sending an empty response.
+            return;
+        }
+
+        List<TransactionReceipt> receipts = new LinkedList<>();
+
+        for (Transaction tx :  block.getTransactionsList()) {
+            TransactionInfo txInfo = blockchain.getTransactionInfo(tx.getHash().getBytes());
```
nicops, line 76: Could this possibly return null?

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
```diff
@@ -0,0 +1,222 @@
...
+            receipts.add(txInfo.getReceipt());
+        }
+
+        Message responseMessage = new BlockReceiptsMessage(requestId, receipts);
+        lightPeer.sendMessage(responseMessage);
+    }
+
+    public void processBlockReceiptsMessage(long id, List<TransactionReceipt> blockReceipts, LightPeer lightPeer) {
```
nicops, line 84: I would delete this and all the other methods that aren't supported. They don't belong together at all: One thing is to reply to LC queries as a full node, something very different is to consume replies to those queries as a light node. I would also rename the class to reflect this, maybe something like LightClientQueriesResolver (I didn't think this through).

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
```diff
@@ -0,0 +1,222 @@
...
+        if (block == null) {
+            // Don't waste time sending an empty response.
+            return;
+        }
+
+        RepositorySnapshot repositorySnapshot = repositoryLocator.snapshotAt(block.getHeader());
+        RskAddress addr = new RskAddress(address);
+        Keccak256 codeHash = repositorySnapshot.getCodeHash(addr);
+
+        CodeMessage response = new CodeMessage(requestId, codeHash.getBytes());
```
nicops, line 126: If this is modelled after https://openethereum.github.io/wiki/The-Parity-Light-Protocol-%28PIP%29 then shouldn't this be the bytecode rather than the code hash?

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
```diff
@@ -0,0 +1,222 @@
...
+        if (block == null) {
+            // Don't waste time sending an empty response.
+            return;
+        }
+
+        RepositorySnapshot repositorySnapshot = repositoryLocator.snapshotAt(block.getHeader());
+        RskAddress address = new RskAddress(addressHash);
+        AccountState state = repositorySnapshot.getAccountState(address);
+
+        AccountsMessage response = new AccountsMessage(id, new byte[] {0x00}, state.getNonce().longValue(),
```
nicops, line 148: I would add a TODO comment to mark that the merkle proof is yet to be implemented, and the same below

rskj-core/src/main/java/co/rsk/net/light/message/BlockBodyMessage.java
```diff
@@ -0,0 +1,120 @@
...
+
+        this.code = BLOCK_BODY.asByte();
+    }
+
+    @Override
+    public byte[] getEncoded() {
+        byte[] rlpId = encodeBigInteger(BigInteger.valueOf(getId()));
+
+        byte[][] rlpTransactions = getTransactions().stream()
+                .map(Transaction::getEncoded)
```
nicops, line 84: This method encodes only the signature, which doesn't seem correct.

rskj-core/src/main/java/co/rsk/net/light/message/BlockBodyMessage.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Override
+    public Class<?> getAnswerMessage() {
+        return null;
+    }
+
+    @Override
+    public String toString() {
+        return "";
```
nicops, line 101: Why empty toString on this and the other messages?

rskj-core/src/main/java/co/rsk/net/light/message/BlockBodyMessage.java
```diff
@@ -0,0 +1,120 @@
...
+        List<BlockHeader> uncleList = new LinkedList<>();
+        for (int k = 0; k < rlpUncles.size(); k++) {
+            byte[] rlpData = rlpUncles.get(k).getRLPData();
+            BlockHeader uncle = blockFactory.decodeHeader(rlpData);
+            uncleList.add(uncle);
+        }
+
+        this.transactions = transactionList;
+        this.uncles = uncleList;
+        this.code = BLOCK_BODY.asByte();
```
nicops, line 67: In this and other places, the usage of import static for enum can be a little hard to read. It is of course, a matter of personal preference.

rskj-core/src/main/java/co/rsk/net/light/message/BlockBodyMessage.java
```diff
@@ -0,0 +1,120 @@
...
+    private final List<BlockHeader> uncles;
+
+    public BlockBodyMessage(byte[] encoded, BlockFactory blockFactory) {
+        RLPList list = (RLPList) decode2(encoded).get(0);
+        byte[] rlpId = list.get(0).getRLPData();
+        this.id = rlpId == null ? 0 : BigIntegers.fromUnsignedByteArray(rlpId).longValue();
+
+        RLPList rlpTransactions = (RLPList) decode2(list.get(1).getRLPData()).get(0);
+
+        List<Transaction> transactionList = new LinkedList<>();
```
nicops, line 49: In this and other places, there is a repeated usage of an structure that could be abstracted away in a lambda, maybe as a static method on a utils package. Something like this:


    public static <T> List<T> mapListRLP(RLPList inputList, Function<byte[], T> lambda) {
        List<T> output = new LinkedList<>();
        for (int k = 0; k < inputList.size(); k++) {
            byte[] rlpData = inputList.get(k).getRLPData();
            T outputElement = lambda.apply(rlpData);
            output.add(outputElement);
        }
        return output;
    }

And then you can replace this with simply        

`List<Transaction> transactionList = mapListRLP(rlpTransactions, rlpData -> new ImmutableTransaction(rlpData));`

rskj-core/src/main/java/co/rsk/net/light/LightSyncProcessor.java
```diff
@@ -0,0 +1,190 @@
...
+                return;
+            }
+
+            Keccak256 genesisHash = genesis.getHash();
+            Keccak256 msgGenesisHash = new Keccak256(status.getGenesisHash());
+            if (!msgGenesisHash.equals(genesisHash)) {
+                loggerNet.info("Removing LCHandler for {} due to unexpected genesis", ctx.channel().remoteAddress());
+                loggerNet.info("Config genesis hash {} - message genesis hash {}",
+                        genesisHash, msgGenesisHash);
+                lightPeer.disconnect(ReasonCode.UNEXPECTED_GENESIS);
```
nicops, line 115: This repeated code makes me think that you should have an auxiliary method that returns a disconnection code or null, and then you disconnect with that code.

rskj-core/src/main/java/co/rsk/net/light/LightSyncProcessor.java
```diff
@@ -0,0 +1,190 @@
...
+                ctx.pipeline().remove(lightClientHandler);
+                return;
+            }
+        } catch (NoSuchElementException e) {
+            loggerNet.debug("LCHandler already removed - exception: {}", e.getMessage());
+        }
+
+        if (peerStatuses.size() >= MAX_PEER_CONNECTIONS) {
```
nicops, line 123: I don't understand this line nor the usage of the Map. Why use a Map for what essentially could be an int? why you don't disconnect in this case? Shouldn't there be a mechanism to account for disconnections?

rskj-core/src/main/java/co/rsk/net/light/LightSyncProcessor.java
```diff
@@ -0,0 +1,190 @@
...
+            return;
+        }
+
+        if (!hasLowerDifficulty(status)) {
+            return;
+        }
+
+        peerStatuses.put(lightPeer, status);
+
+        if (msg.isTxRelay()) {
```
nicops, line 133: This is always false, except on the unit tests. Right?

rskj-core/src/main/java/co/rsk/net/light/LightSyncProcessor.java
```diff
@@ -0,0 +1,190 @@
...
+    public void processBlockHeaderMessage(long id, BlockHeader blockHeader, LightPeer lightPeer) {
+        if (!isPending(id, BLOCK_HEADER)) {
+            return;
+        }
+
+        pendingMessages.remove(id, BLOCK_HEADER);
+        lightPeer.receivedBlock(blockHeader);
+    }
+
+    public boolean hasTxRelay(LightPeer peer) {
```
nicops, line 163: hasTxRelay should have a @VisibleForTesting annotation, or similar. Because it is a method ONLY for testing

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -318,7 +318,7 @@ public String helloPhrase() {
     }
 
     public List<String> peerCapabilities() {
-        return configFromFiles.hasPath("peer.capabilities") ?  configFromFiles.getStringList("peer.capabilities") : new ArrayList<>(Arrays.asList("rsk"));
+        return configFromFiles.hasPath("peer.capabilities") ?  configFromFiles.getStringList("peer.capabilities") : new ArrayList<>(Arrays.asList("rsk","lc"));
```
nicops, line 321: To ask for the presence is a pattern that reoccurs through the code, but it is bad pattern: this should be set in `reference.conf`, that assures that the value can never be null. Besides that, `lc` should **not** be a default

rskj-core/src/test/java/co/rsk/net/LightProcessorTest.java
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    @Test
+    public void processGetBlockReceiptMessageAndShouldReturnsReceiptsCorrectly() {
+        List<Transaction> txs = new LinkedList<>();
+        long requestId = 0;
+        List<TransactionReceipt> receipts = new LinkedList<>();
+        TransactionReceipt receipt = createReceipt();
+        receipts.add(receipt);
+        final Block block = mock(Block.class);
```
nicops, line 85: Like ajl says on his blog https://blogs.msmvps.com/lopez/2012/08/09/tdd-unit-tests-and-mocks/ it's undesirable to mock so much when testing. Ideally, we should have factories to build these objects for us, it is a strong smell to resort to mock all these classes, especially domain classes like Block and Transaction, since it's not exactly the external, unpredictable world the one we're mocking.

There is a BlockBuilder class for this, it might be a bit unwieldy for the purposes of this test, but I think it merits trying to make efforts in order to make non mocking tests easy to code.

rskj-core/src/main/java/co/rsk/net/light/LightPeer.java
julianlen: Received block headers will be stored in a header chain. These methods together with the block header list in each Light peer will disappear.

rskj-core/src/main/java/co/rsk/core/bc/BlockChainStatus.java
julianlen: That comment points to the "hasTotalDifficulty" variable. Not the whole method.

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
julianlen: In the long term, a light client will have available an amount of "credits". Once these credits are finished, the light peer cannot request anything to the full node.

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
julianlen: This is weird, all transactions return a receipt, it doesn't matter if the transaction was successful or not.

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
julianlen: @smcsicardi

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
smcsicardi: You're right, this seems to be an old comment, we should remove it, it makes no sense there

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
smcsicardi: Totally right! Clear mistake on my part, will change it now.

rskj-core/src/main/java/co/rsk/net/light/message/BlockBodyMessage.java
smcsicardi: Actually the method uses another method which, given the signature, encodes the whole transaction

Line it calls encoding method: https://github.com/rsksmart/rskj/blob/9feb5a08b0c4d1996760b43f69d293367388073f/rskj-core/src/main/java/org/ethereum/core/Transaction.java#L482

Method itself: https://github.com/rsksmart/rskj/blob/9feb5a08b0c4d1996760b43f69d293367388073f/rskj-core/src/main/java/org/ethereum/core/Transaction.java#L488

rskj-core/src/main/java/co/rsk/net/light/message/BlockBodyMessage.java
smcsicardi: I really like it, the RLP encoding is always cumbersome, we just didn't took the time to refactor it. We will organize with @julianlen for this and the other fixes.

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
smcsicardi: We will check that (the pattern).

On the other hand, the Light Client _capability_ **should** be a default. By default any full node should serve Light Clients, removing it will cause that no Full Node will serve them.

I think the root of this problem is the bad way that defaults and config settings are handled in general in the code.

rskj-core/src/main/java/co/rsk/net/light/message/BlockBodyMessage.java
nicops: you're right!

rskj-core/src/main/java/co/rsk/net/light/message/BlockBodyMessage.java
julianlen: Ready

rskj-core/src/main/java/co/rsk/net/light/LightSyncProcessor.java
julianlen: This is another case of "dummy things". The map is the basic approach to handle peers. For now, there is no something like "PeerHandlers". In the sync stage, peerStatuses is used to get the best peer, I mean, the peer with high total difficult and with the best block.
Then, each node accepts a max quantity of peers.

rskj-core/src/main/java/co/rsk/net/light/message/BlockBodyMessage.java
julianlen: Nice

rskj-core/src/main/java/co/rsk/net/light/LightSyncProcessor.java
julianlen: It's false by default

rskj-core/src/main/java/co/rsk/net/light/LightSyncProcessor.java
julianlen: At this moment, it should be visible for testing, right.

PR #1185, opened 2020-03-06
New RLP implementation

ajlopezrsk:
In this RLP implementation the decode of the data is lazy, only the first level is decoded, and then, the rest of the levels (nested lists) are decoded on demand.

`RLPList` was simplified to not be a full list, only having `size` and `get` methods.

`RLPElementView` was removed.

`BridgeSerializationUtilsTest` was refactored to use RLP directly. In the previous implementation, it used a custom ad-hoc serialization format, injected via mocking RLP coding and decoding methods. Not, RLP format is used in the test.

*fed: newrlp-fixes*


josedahlquist:
pipeline:run

aeidelman:
pipeline:run

PR #1188, opened 2020-03-12
Switch logger levels in `org.ethereum.core.TransactionExecutor#execEr…

### LINKED ISSUE #987: Expected VM errors shouldn't be logged as errors
E.g the `org.ethereum.core.TransactionExecutor#execError(java.lang.String)` logs errors such as `Not enough gas for transaction execution`, which isn't a program error but a user error. I suggest making this a `logger.trace`.

On the other hand I think `org.ethereum.core.TransactionExecutor#execError(java.lang.Throwable)` should be upgraded to `logger.error` because we should never be reaching the catch-all exception.

kiminuo:
…ror` methods

Fixes #987


fedejinich:
pipeline: run

fedejinich:
pipeline: run

PR #1190, opened 2020-03-12
Use `ImmutableList.copyOf(list)` instead of `Collections.unmodifiable…

### LINKED ISSUE #1158: Transaction and Uncles Lists in Block are not Immutable
This code is from one of the `Block` constructors:

```
        this.transactionsList = Collections.unmodifiableList(transactionsList);
        this.uncleList = Collections.unmodifiableList(uncleList);

```

Altough this code prevents the modification of `this.transactionList` directly, the original list could still be modified externally, affecting the integrity of the block.

To prevent such behavior, maybe a list copy with unmodifiable view could be used.

kiminuo:
…List(list)`

Fixes #1158

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/Block.java
+++ b/rskj-core/src/main/java/org/ethereum/core/Block.java
@@ -26,6 +26,7 @@ import co.rsk.core.bc.BlockHashesHelper;
 import co.rsk.crypto.Keccak256;
 import co.rsk.panic.PanicProcessor;
 import org.bouncycastle.util.Arrays;
+import com.google.common.collect.ImmutableList;
 import org.bouncycastle.util.BigIntegers;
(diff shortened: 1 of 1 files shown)


fedejinich:
why should we use google collection instead of java native implementation?

kiminuo:
@fedejinich Have you read the original issue?

> Altough this code prevents the modification of this.transactionList directly, the original list could still be modified externally, affecting the integrity of the block.

And that's why using a 3rd party library may make sense in this case. Because Google's immutable list won't change. The items are still mutable though. In the current version and even in the proposed PR version.

This PR targets Java 8. Java 9+ would allow to use `List.of()` which would be ideal.

fedejinich:
pipeline: run

fedejinich:
pipeline: run

Issue #1192, opened 2020-03-16
Bail out on configuration error


lucasvuotto:
Currently, the node will go completely silent about errors in configuration file, which can lead to several hours wasted trying to debug a problem that turned to be a typo.

Most daemons out there will be noisy about such errors, and ideally will refuse to start in such situations. It would be great if rskj behaved the same way.

PR #1201, opened 2020-03-30
Added verification of node consolidated config property names

Vovchyk:
This verification process is turned off by default. The way to turn it on is using cli flag `--verify-config` or configuration property `blockchain.config.verify`.


rskj-core/src/main/resources/reference.conf
```diff
@@ -3,6 +3,20 @@ blockchain.config.hardforkActivationHeights.papyrus200 = -1
```
patogalla: reference.conf is used by convention :
https://github.com/lightbend/config#standard-behavior

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
```diff
@@ -128,4 +150,52 @@ private Config getNetworkDefaultConfig(Config userConfig) {
...
     }
+
+    private static void verify(String key, @Nullable ConfigValue expectedValue, ConfigValue actualValue) {
+        if (expectedValue == null) {
+            throw unexpectedKeyException(key, actualValue);
+        }
+
+        switch (actualValue.valueType()) {
+            case OBJECT:
+                if (!expectedValue.valueType().equals(ConfigValueType.OBJECT)) {
```
patogalla: I would suggest doing like : "Enum.VALUE.equals(value)", 
good practice

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
```diff
@@ -128,4 +150,52 @@ private Config getNetworkDefaultConfig(Config userConfig) {
         logger.info("Network not set, using mainnet by default");
         return ConfigFactory.load(MAINNET_RESOURCE_PATH);
     }
+
+    private static void verify(String key, @Nullable ConfigValue expectedValue, ConfigValue actualValue) {
+        if (expectedValue == null) {
+            throw unexpectedKeyException(key, actualValue);
+        }
+
```
patogalla: Isnt a way to compare types before going to the switch, it seems redundant to me, asking in each switch statement for the equal type. Whats the logic there?

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
```diff
@@ -128,4 +150,52 @@ private Config getNetworkDefaultConfig(Config userConfig) {
...
+        switch (actualValue.valueType()) {
+            case OBJECT:
+                if (!expectedValue.valueType().equals(ConfigValueType.OBJECT)) {
+                    throw typeMismatchException(key, expectedValue, actualValue);
+                }
+                for (Map.Entry<String, ConfigValue> actualEntry : ((ConfigObject) actualValue).entrySet()) {
+                    ConfigValue expectedEntryValue = ((ConfigObject) expectedValue).get(actualEntry.getKey());
+                    verify((key.isEmpty() ? "" : key + ".") + actualEntry.getKey(), expectedEntryValue, actualEntry.getValue());
```
patogalla: "(key.isEmpty() ? "" : key + ".")" logic should be outside "for", no need to ask in each iteration.

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
```diff
@@ -128,4 +150,52 @@ private Config getNetworkDefaultConfig(Config userConfig) {
...
+
+    private static void verify(String key, @Nullable ConfigValue expectedValue, ConfigValue actualValue) {
+        if (expectedValue == null) {
+            throw unexpectedKeyException(key, actualValue);
+        }
+
+        switch (actualValue.valueType()) {
+            case OBJECT:
+                if (!expectedValue.valueType().equals(ConfigValueType.OBJECT)) {
+                    throw typeMismatchException(key, expectedValue, actualValue);
```
patogalla: instead of throwing, it would be nice to collect them, and throw/log them all at once.

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
Vovchyk: I've updated the code a bit.

rskj-core/src/main/resources/reference.conf
Vovchyk: Thanks @patogalla

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
```diff
@@ -56,13 +59,38 @@ public ConfigLoader(CliArgs<NodeCliOptions, NodeCliFlags> cliArgs) {
...
+                .withFallback(systemEnvConfig)
+                .withFallback(userCustomConfig)
+                .withFallback(installerConfig);
         Config networkBaseConfig = getNetworkDefaultConfig(userConfig);
-        return userConfig.withFallback(networkBaseConfig);
+        Config unifiedConfig = userConfig.withFallback(networkBaseConfig);
+
+        if (unifiedConfig.getBoolean(SystemProperties.PROPERTY_BC_VERIFY)) {
+            Config expectedConfig = ConfigFactory.load(EXPECTED_RESOURCE_PATH);
+            Config actualConfig = ConfigFactory.empty()
```
Vovchyk, line 79: Config without `systemPropsConfig` and `systemEnvConfig` which are out of our interest

rskj-core/src/main/resources/expected.conf
```diff
@@ -0,0 +1,458 @@
+# disable wasabi+1 consensus rules by default for every network
```
Vovchyk, line 1: The idea is that the expected.conf file should contain all possible config settings which are expected by the node

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
nicops: can you add this as a comment on the code directly?

rskj-core/src/main/resources/expected.conf
nicops: same as before, can you add this as a header comment on the expected.conf file, to explain what the file is about?

rskj-core/src/main/resources/expected.conf
```diff
@@ -0,0 +1,458 @@
...
+
+peer {
+    # Boot node list
+    # Use to connect to specific nodes
+    active = [ ]
+
+    # list of trusted peers the incoming connections is always accepted from. Even if the max amount of connections is reached
+    # This is used to create a filter of Trusted peers
+    trusted = [
+        # Sample entries:
```
nicops, line 70: maybe we can flesh out these entries so it gets validated?

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
```diff
@@ -128,4 +156,66 @@ private Config getNetworkDefaultConfig(Config userConfig) {
         logger.info("Network not set, using mainnet by default");
         return ConfigFactory.load(MAINNET_RESOURCE_PATH);
     }
+
+    private static void verify(String keyPath, @Nullable ConfigValue expectedValue, ConfigValue actualValue, List<String> problems) {
+        Objects.requireNonNull(keyPath);
+        Objects.requireNonNull(actualValue);
+        ConfigValueType actualValueType = Objects.requireNonNull(actualValue.valueType());
+
+        if (expectedValue == null) {
```
patogalla, line 165: This is the original case? When a misspelled property will be warned... am i right?

rskj-core/src/main/resources/expected.conf
```diff
@@ -0,0 +1,458 @@
...
+        # the peer window will show
+        # only what retrieved by active
+        # peer [true/false]
+        enabled = true
+
+        # List of the peers to start
+        # the search of the online peers
+        # values: [ip:port]
+        ip.list = [
+            "bootstrap01.rsk.co:5050",
```
nicops, line 126: for clarity I would replace this with dummy data

rskj-core/src/main/resources/expected.conf
```diff
@@ -0,0 +1,458 @@
...
+
+    # Port for server to listen for incoming connections
+    port = 5050
+
+    # Network id
+    networkId = 775
+}
+
+# the folder resources/genesis contains several versions of genesis configuration according to the network the peer will run on
+genesis = rsk-mainnet.json
```
nicops, line 153: same here, and everything else related to mainnet, I would put dummy data.

rskj-core/src/main/resources/reference.conf
```diff
@@ -3,6 +3,10 @@ blockchain.config.hardforkActivationHeights.papyrus200 = -1
 
 blockchain = {
     config = {
+        # Flag that forces to verify whether a node is started with unknown configuration property names.
+        # Whole list of possible property names is defined in the reference.conf file.
```
nicops, line 7: replace reference.conf in comment with expected.conf

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
```diff
@@ -55,14 +51,59 @@ public ConfigLoader(CliArgs<NodeCliOptions, NodeCliFlags> cliArgs) {
...
+        if (unifiedConfig.getBoolean(SystemProperties.PROPERTY_BC_VERIFY)) {
+            Config expectedConfig = ConfigFactory.parseResourcesAnySyntax(EXPECTED_RESOURCE_PATH)
+                    .withFallback(systemPropsConfig)
+                    .withFallback(systemEnvConfig);
+
+            ArrayList<String> problems = new ArrayList<>();
+            verify("", expectedConfig.root(), unifiedConfig.root(), problems);
+            if (!problems.isEmpty()) {
+                throw new RskConfigurationException("Verification of node configs has failed. The following problems were found: "
+                        + String.join("; ", problems));
```
lucasvuotto, line 102: I can see this getting overly verbose and quite difficult to read and search for problems, should multiple configuration problems be present. I suggest making `verify` return a boolean instead, signaling if the verification was successful, and logging each of the misconfigurations instead of pushing them into `problems`.

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
```diff
@@ -55,14 +51,59 @@ public ConfigLoader(CliArgs<NodeCliOptions, NodeCliFlags> cliArgs) {
...
+                .withFallback(cliConfig)
+                .withFallback(systemPropsConfig)
+                .withFallback(systemEnvConfig)
+                .withFallback(userCustomConfig)
+                .withFallback(installerConfig);
         Config networkBaseConfig = getNetworkDefaultConfig(userConfig);
-        return userConfig.withFallback(networkBaseConfig);
+        Config unifiedConfig = userConfig.withFallback(networkBaseConfig);
+
+        if (unifiedConfig.getBoolean(SystemProperties.PROPERTY_BC_VERIFY)) {
```
lucasvuotto, line 93: Can't we run the config verification always and just abort execution only when this flag is set?

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
lucasvuotto: This seems to be my original case presented in #1192. Maybe a comment explaining how we got here can help.

rskj-core/src/main/resources/expected.conf
```diff
@@ -0,0 +1,160 @@
...
+                bind_address = <bind_address>
+                port = <port>
+            }
+        }
+    }
+    modules = [
+        {
+            name = <name>
+            version = <version>
+            enabled = <enabled>
```
lucasvuotto, line 135: You're missing a key in here:

```
            methods = {
                enabled = []
                disabled = []
            }
```

(cf https://github.com/rsksmart/rskj/blob/1e842d8a9728d4a791dbeef04753e207c68533d6/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java#L238)

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
nicops: that would be interesting. In which case we catch the exception and only log it?

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
nicops: agree, I would add a comment here.

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
lucasvuotto: About what exception are we talking? I only see `verify` throwing an exception if there is a malformed entry in `expected.conf`.

nicops: sorry, I did not read the code carefully between all the changes. Right now this method throws the exception, it could skip throwing the exception if the flag is not set.

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
nicops: I agree and disagree.
**Agree:** I would iterate over problems and log each of them as a separate line.
**Disagree:** I think the method works fine as it is, right now we don't unit test it directly, but it could be directly unit tested, and it's a richer interface to have it return a problems array instead of a boolean. What I would add is an overload:
```
 private static List<String> verify(ConfigValue expectedRoot, ConfigValue actualRoot) {
    ArrayList<String> problems = new ArrayList<>();
    verify("", expectedConfig.root(), unifiedConfig.root(), problems);
    return problems;
}
``` 

IMHO it's nice to add an overload in such cases to make the calling code more oblivious to the internal details of the recursive method.

Vovchyk: I just don't want to overarchitect this solution. Good point about doing verification all the time and only throwing an exception if the flag is present.

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
Vovchyk: Yes, the `verify` method is recursive one and `expectedValue` could be `null` during recursive iterations when an appropriate matching key path is to found.

rskj-core/src/main/resources/expected.conf
Vovchyk: thanks

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
Vovchyk: well.. there're pros and cons.. people from sonarsource suggest using `==` - https://rules.sonarsource.com/java/RSPEC-4551.

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
Vovchyk: done

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
Vovchyk: done

rskj-core/src/test/java/co/rsk/config/ConfigLoaderTest.java
```diff
@@ -181,13 +168,7 @@ public void detectExpectedScalarValueProblemInList() {
...
         mockConfigFactory(defaultConfig, expectedConfig);
 
-        try {
-            loader.getConfig();
-        } catch (RskConfigurationException e) {
-            assertThat(e.getMessage(), containsString(ConfigProblems.expectedScalarValueProblem("expectedKey[0]", EMPTY_LIST_VALUE)));
-
-            throw e;
-        }
+        loader.getConfig();
```
nicops, line 171: with the latest change, this test and the one above it is not making any assertion

aeidelman:
pipeline: run

aeidelman:
pipeline: run

aeidelman:
@Vovchyk can you rebase please?

Vovchyk:
> @Vovchyk can you rebase please?

@aeidelman done

ajlopezrsk:
pipeline: run

jjtechuy:
pipeline: run

PR #1202, opened 2020-04-02
Fix minor debug logging error in Program class

nicops:
The changed logging lines were mistakenly using a `{number}` positional syntax that's not actually used in the logging methods, instead `{}` simply replaces the corresponding parameter that has been passed. Add further confusion, if the last parameter is an exception, it should not be referenced in the format string, instead the stack trace is printed below the string. 

So to make the fix I removed the 0 from `{0}` and also I removed the format reference to the exception parameter.


jjtechuy:
pipeline: run

PR #1204, opened 2020-04-07
GetBlockHeader and BlockHeader messages

julianlen:
Added GetBlockHeader and BlockHeader messages.


rskj-core/src/test/java/co/rsk/net/light/message/BlockHeaderMessageTest.java
```diff
@@ -0,0 +1,57 @@
+package co.rsk.net.light.message;
```
smcsicardi, line 1: Please insert header

PR #1208, opened 2020-04-08
Persistent store for logs bloom

ajlopezrsk:


rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomStore.java
```diff
@@ -29,22 +32,58 @@
...
+
+        if (blocksBloom != null) {
+            return blocksBloom;
+        }
+
+        if (this.dataSource == null) {
+            return null;
+        }
+
+        byte[] data = this.dataSource.get(longToKey(key));
```
nicops, line 68: wouldn't you want to load everything in memory when it starts?

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomEncoder.java
```diff
@@ -0,0 +1,62 @@
...
+    private BlocksBloomEncoder() {
+
+    }
+
+    public static byte[] encode(BlocksBloom blocksBloom) {
+        byte[] rlpFrom = encodeLong(blocksBloom.fromBlock());
+        byte[] rlpTo = encodeLong(blocksBloom.toBlock());
+        byte[] rlpData = RLP.encodeElement(blocksBloom.getBloom().getData());
+
+        return RLP.encodeList(rlpFrom, rlpTo, rlpData);
```
nicops, line 42: should we encode something extra like noBlocks / noConfirmations or a maybe just a version number to ease a future migration if the structure of this changes?

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomStore.java
ajlopezrsk: We don't have the information at start (the log blooms are built on demand). The node maybe never would be asked for that log range, etc... So the memory cache is only to working nodes that have a behavior: someone is requesting the logs, from JSON RPC. The store is the new piece in this pull request: then, when a node restart and is queried for logs, it could respond faster.

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomEncoder.java
ajlopezrsk: Not needed. Only the confirmed blocks are processed in this logic, so each accepted block is processed only once. Check configuration, there is a number of blocks to take as confirmation blocks.

nicops: Ok, basically if we happen to have to migrate this structure we would have to delete or rename the whole datasource. So we don't need to version this.

jjtechuy:
pipeline: run

PR #1215, opened 2020-04-16
Recursive contracts to test 400 recursive levels in VM, VM Refactor 

ajlopezrsk:
The new tests are based in two DSL (Domain Specific Language) files

The class `VM` was refactored to have a better stack usage

Recommendation: add this kind of test also in integration tests


rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1553,25 +1549,38 @@ protected void doCALL(){
...
+        return new MessageCall(
                 MsgType.fromOpcode(op),
                 DataWord.valueOf(calleeGas), codeAddress, value, inDataOffs, inDataSize,
                 outDataOffs, outDataSize);
+    }
 
-        callToAddress(codeAddress, msg);
-
-        program.step();
+    private DataWord calculateCallValue(ActivationConfig.ForBlock activations) {
```
nicops, line 1558: I would call this something different from calculate. calculate makes me think this is a transparently referential function, but because this can do a program.stackPop(), it's actually imperative. In fact, it almost always does a stackPop, it seems, except in two exceptional op codes. So it makes sense to me to call it something like popCallValue(), or even maybePopCallValue().

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1553,25 +1549,38 @@ protected void doCALL(){
...
+            // value is always zero in a DELEGATECALL operation
+            value = op == OpCode.DELEGATECALL ? DataWord.ZERO : program.stackPop();
+        }
+        return value;
     }
 
-    private void callToAddress(DataWord codeAddress, MessageCall msg) {
-        ActivationConfig.ForBlock activations = program.getActivations();
-        PrecompiledContracts.PrecompiledContract contract = precompiledContracts.getContractForAddress(activations, codeAddress);
+    private long calculateGetMinimumTransferGas(DataWord value, long remainingGas) {
```
nicops, line 1570: This function is not cohesive and the naming is unclear. It's doing two things, and one of them is validating and maybe throwing an exception. At the very least is should be called calculateAndValidateMinimumTransferGas

PR #1217, opened 2020-04-17
GetStorage and Storage Messages

smcsicardi:
// Request for a proof of contract storage.
Request::Storage {
    ID: 6
    Inputs:
        Loose(H256) // block hash
        Loose(H256) // address hash
        Loose(H256) // storage key hash
    Outputs:
        [U8](U8) // merkle inclusion proof from storage trie
        H256 reusable_as(0) // storage value
}

### DIFF
--- a/rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
+++ b/rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
@@ -28,6 +28,7 @@ import org.ethereum.net.message.Message;
 import org.bouncycastle.util.encoders.Hex;
 import org.ethereum.db.BlockStore;
 import org.ethereum.db.TransactionInfo;
+import org.ethereum.vm.DataWord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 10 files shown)


rskj-core/src/test/java/co/rsk/net/LightProcessorTest.java
```diff
@@ -329,6 +330,45 @@ public void processBlockBodyMessageAndShouldThrowAnException() {
...
+                new byte[] {0x00}, lightPeer);
+
+        verify(msgQueue, times(0)).sendMessage(any());
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void processStorageMessageAndShouldThrowAnException() {
+        lightProcessor.processStorageMessage(0, new byte[] {0x00}, new byte[] {0x00}, lightPeer);
+    }
+
```
julianlen, line 371: You should test the case in which the address has null storage

rskj-core/src/main/java/co/rsk/net/light/LightProcessor.java
```diff
@@ -185,4 +186,30 @@ public void processGetBlockBodyMessage(long id, byte[] blockHash, LightPeer ligh
     public void processBlockBodyMessage(long id, List<BlockHeader> uncles, List<Transaction> transactions, LightPeer lightPeer) {
         throw new UnsupportedOperationException("Not supported BlockBody processing");
     }
+
+    public void processGetStorageMessage(long id, byte[] blockHash, byte[] addressHash, byte[] storageKeyHash, LightPeer lightPeer) {
+        logger.trace("Processing storage request {} block {} address {} storage key {}", id,
```
julianlen, line 191: To fix the SonarCloud advice "Invoke method only conditionally", we will start to log these messages in the next manner:

```
String logBlockHash = Hex.toHexString(blockHash);
        logger.trace("Processing block header request {} block {}", id, logBlockHash);
```

PR #1220, opened 2020-04-17
Fix and enhance testing

julianlen:


smcsicardi:
Update: Its was shown this check corectly checks what is needed so @julianlen please make another push so the pipeline runs

Issue #1145, opened 2020-01-03
Block/Transaction Node Information: Harcoded max blocks and peers (50) values


juli:
Maximum number of peers and blocks must not be hardcoded and should match the node's configuration. This harcoded value (50) could cause problems to bootstrap nodes that have more than 100 peers.

`BlockNodeInformation` tracks information about which blocks are known by which peers, and provides functions to retrieve all the block hashes known by a node, and which nodes know a certain block.  
``` 
public BlockNodeInformation() {
        this(1000, 50);
    }
```
https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/net/BlockNodeInformation.java#L62

https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/net/TransactionNodeInformation.java#L53

https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/RskContext.java#L1276

*edit* check also TransactionNodeInformation.java

PR #1222, opened 2020-04-20
Remove unused code from BlockNodeInformation & TransactionNodeInformation

nicops:
# Rationale

See https://github.com/rsksmart/rskj/issues/1145

## BlockNodeInformation

maxPeers here is not actually used for anything. maxPeers is hardcoded to 50 and it limits the total size of a HashMap, the attribute blocksByNode of the BlockNodeInformation class. But, that structure is not actually queried at all in the main code, it should be queried by the getBlocksByNode methods (there are two overrides) but it is only queried on the test code. 

For clarity, this PR removes the unused attribute and changes the test code to make the same assertions in a different way.

## TransactionNodeInformation

In this class, maxPeers here is not actually used for anything, literally the attribute is unused. So we are removing the maxPeers from the class and the constructor.

## maxBlocks & maxTransactions

These two attributes simply limit the amount of blocks/transactions remembered by removing old entries after the 1000th. This means forgetting which node knows about a block/transaction when the block/transaction is old enough. Currently there are no config values for this and in principle it should not be an issue for bootstrapping a node.

### DIFF
--- a/rskj-core/src/test/java/co/rsk/net/BlockNodeInformationTest.java
+++ b/rskj-core/src/test/java/co/rsk/net/BlockNodeInformationTest.java
@@ -38,38 +38,6 @@ public class BlockNodeInformationTest {
     }
 
     @Test
-    public void nodeEvictionPolicy() {
-        final BlockNodeInformation nodeInformation = new BlockNodeInformation();
-        final Keccak256 block = createBlockHash(10);
-
-        // Add a few nodes, not exceeding the node limit.
-        // These nodes should contain the block when we call getBlocksByNode
-        for (int i = 0; i < 30; i++) {
-            final NodeID node = createNodeID(i);
-            nodeInformation.addBlockToNode(block, node);
-        }
-
-        Assert.assertTrue(nodeInformation.getBlocksByNode(createNodeID(10)).contains(block));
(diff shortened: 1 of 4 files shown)


rskj-core/src/test/java/co/rsk/net/BlockNodeInformationTest.java
```diff
@@ -37,38 +37,6 @@ private NodeID createNodeID(int i) {
         return new NodeID(ByteBuffer.allocate(4).putInt(i).array());
     }
 
-    @Test
-    public void nodeEvictionPolicy() {
```
nicops, line 41: I removed this test because it checked the behaviour of the getBlocksByNode method and the underlying map, which is now gone

aeidelman:
pipeline: run

patogallaiovlabs:
pipeline: run

aeidelman:
pipeline: run

nicops:
Added a new commit to make the two classes thread safe by adding synchronized to the methods and returning a copy of the set in the read methods.

ajlopezrsk:
*pipeline: run*

ajlopezrsk:
pipeline: run

PR #1223, opened 2020-04-21
Message Visitor for the light client handler

smcsicardi:
Created a visitor pattern to replace the ugly switch we had.

The only concern is the creation of the Visitor each time a new message is created. This is because a message need the ChannelHandlerContext passed to the channelRead0 function, so the context is passed in the constructor.

Other possible solutions:

1. setCtx (not that pretty)
2. pass the ctx to the accept method (I really dont like this because ctx is only needed for the Status message, and this would mean we have to change the whole signature of both accept AND apply in EVERY message for just only one)


smcsicardi:
We could also separate the STATUS message and treat it differently (as it does), because almost all dependencies on the visitor are for this message. What do you think?

PR #1226, opened 2020-04-27
CLI Tools

ajlopezrsk:
Command line tools:

- Export blocks
- Export state
- Show state info
- Import blocks
- Connect blocks
- Execute blocks
- Import State


rskj-core/src/main/java/co/rsk/ConnectBlocks.java
```diff
@@ -0,0 +1,72 @@
...
+    public static void main(String[] args) throws IOException {
+        RskContext ctx = new RskContext(args);
+
+        BlockFactory blockFactory = ctx.getBlockFactory();
+        Blockchain blockchain = ctx.getBlockchain();
+        TrieStore trieStore = ctx.getTrieStore();
+        BlockStore blockStore = ctx.getBlockStore();
+        ReceiptStore receiptStore = ctx.getReceiptStore();
+
+        String filename = args[0];
```
nicops, line 45: should we check the presence of the filename? same in the other classes

rskj-core/src/main/java/co/rsk/ConnectBlocks.java
```diff
@@ -0,0 +1,72 @@
...
+            if (parts.length < 4) {
+                continue;
+            }
+
+            byte[] encoded = Hex.decode(parts[3]);
+
+            Block block = blockFactory.decodeBlock(encoded);
+            block.seal();
+
+            blockchain.tryToConnect(block);
```
nicops, line 65: this returns a value that explains if the block has been connected or not. what happens if we don't check it?

rskj-core/src/main/java/co/rsk/ExecuteBlocks.java
```diff
@@ -0,0 +1,53 @@
...
+        long toBlockNumber = Long.parseLong(args[1]);
+
+        for (long n = fromBlockNumber; n <= toBlockNumber; n++) {
+            Block block = blockStore.getChainBlockByNumber(n);
+            Block parent = blockStore.getBlockByHash(block.getParentHash().getBytes());
+
+            blockExecutor.execute(block, parent.getHeader(), false, false);
+        }
+
+        trieStore.flush();
```
nicops, line 50: why do we need to flush? since we're re executing blocks, we shouldn't be having new data here. or could we?

rskj-core/src/main/java/co/rsk/ImportBlocks.java
```diff
@@ -0,0 +1,66 @@
...
+            execute(blockFactory, blockStore, reader);
+        }
+    }
+
+    public static void execute(BlockFactory blockFactory, BlockStore blockStore, BufferedReader reader) throws IOException {
+        for (String line = reader.readLine(); line != null; line = reader.readLine()) {
+            String[] parts = line.split(",");
+
+            if (parts.length < 4) {
+                continue;
```
nicops, line 52: shouldn't we raise an error here instead of continue? why silently swallow the error?

nicops:
I would add a println that says "this is an experimental/unsupported tool" in each of these main classes just to make clear that these tools are not meant for regular node use.

rskj-core/src/main/java/co/rsk/ConnectBlocks.java
ajlopezrsk: The current code is oriented to be simple. We could improve the user experience if these tools are really used

rskj-core/src/main/java/co/rsk/ConnectBlocks.java
ajlopezrsk: Nothing. The supplied blocks could be connected best ones, or only connect, or even invalid. One of the use case of this tool is to run the external blocks, maybe to check the execution logs

rskj-core/src/main/java/co/rsk/ExecuteBlocks.java
ajlopezrsk: Well, it's related with the "missing trie" issue, just in case a trie that should be in the database, was not before the run of this command

rskj-core/src/main/java/co/rsk/ImportBlocks.java
ajlopezrsk: Only for simplicity, until these tools were really used

ajlopezrsk:
> I would add a println that says "this is an experimental/unsupported tool" in each of these main classes just to make clear that these tools are not meant for regular node use.

Well, I would prefer not pollute output with message, keep the first implementation simple, so we could improve them when we have more use pattern info

ajlopezrsk:
pipeline:run

ajlopezrsk:
pipeline: run

ajlopezrsk:
pipeline: run

patogallaiovlabs:
pipeline:run

patogallaiovlabs:
pipeline:run

PR #1227, opened 2020-04-28
Adding a Cache to Transaction Receipts Store

ajlopezrsk:


rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -835,8 +835,8 @@ protected NodeRunner buildNodeRunner() {
                 getRskSystemProperties().flushNumberOfBlocks(),
                 getCompositeEthereumListener(),
                 getTrieStore(),
-                getBlockStore()
-        ));
+                getBlockStore(),
+                null));
```
patogallaiovlabs, line 839: Shouldn't call to "getReceiptStore()" instead of null?

patogallaiovlabs:
As i see, it also fix the issue of flushing before stoping the node. The one that someone complained, that it was not flushing because of never reaching the minimum flushing amount of blocks before stoping it. Am i right?

rskj-core/src/main/java/co/rsk/RskContext.java
ajlopezrsk: Good catch! Fixing it

ajlopezrsk:
> As i see, it also fix the issue of flushing before stoping the node. The one that someone complained, that it was not flushing because of never reaching the minimum flushing amount of blocks before stoping it. Am i right?

Yes, we could fix now, or later. Maybe add a QA explicit test

PR #1230, opened 2020-04-30
Refactor of ECKey signature related methods: signatureToKey / recover…

patogallaiovlabs:
The idea is to refactor ECKey class in order to prepare functionality for Native Signature implementation with a C Native library.
What i did is a Singleton service class, that defines an interface of methods that should be implemented in Bouncy Castle / Native implementations.
The refactor includes only the methods (and where its used):

1. **signatureToKey**()
- PeerDiscoveryMessage.getKey()
- Transaction.getSender()
- PrecompiledContract.ERecover.execute()

2. **recoverFromSignature**()
 - ProofOfWorkRule.validFallbackBlockSignature()
 - Transaction.getKey()
 - ECKey.sign() → (for making it recoverable)
 - EncryptionHandshake.makeAuthInitiate()

3. -**verify**()
 - BootstrapDataVerifier.verifyEntries()


rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -835,6 +835,7 @@ protected NodeRunner buildNodeRunner() {
                     getRepositoryLocator()
             ));
         }
+        SignatureService.initialize(getRskSystemProperties());
```
nicops, line 838: something to ponder: command line and all that jazz, do they start the nodeRunner? if not, maybe there is a better place to initialize a singleton.

patogallaiovlabs: Yep, i will need some help with that. Dont know where it should be the "best" place. 
Maybe @ajlopez could help me?

rskj-core/src/main/java/org/ethereum/crypto/signature/SignatureService.java
```diff
@@ -0,0 +1,135 @@
...
+            INSTANCE = new SignatureServiceBC();
+        }
+    }
+    /**
+     * As a singleton this should be the only entry point for creating instances of SignatureService classes.
+     *
+     * @return either {@link SignatureServiceBC} or Native Signature (future) implementation.
+     */
+    public synchronized static final SignatureService getInstance() {
+        if (INSTANCE == null) {
```
Vovchyk, line 65: just an idea. Alternatively instead of creating the default implementation we may throw an exception if the service is not initialised (when INSTANCE == null). In this case I guess this method may not be `synchronized` which would have a little positive impact on the performance. Also I'd make `initialize` method `synchronized`.

rskj-core/src/main/java/org/ethereum/crypto/signature/SignatureService.java
```diff
@@ -0,0 +1,135 @@
...
+     * <p> It should be called only once in Node Startup.</p>
+     *
+     * <p> It reads "crypto.library" property to decide which impl to initialize.</p>
+     *
+     * <p> By default it initialize Bouncy Castle impl.</p>
+     *
+     * @param rskSystemProperties
+     */
+    public static void initialize(RskSystemProperties rskSystemProperties) {
+        if (NATIVE_LIB.equals(rskSystemProperties.cryptoLibrary())) {//TODO: Check if Native library is loaded.
```
Vovchyk, line 53: according to the description - `It should be called only once in Node Startup.` - I guess we should throw an exception if `INSTANCE != null`

rskj-core/src/main/java/co/rsk/RskContext.java
ajlopezrsk: Check federator code. I'm not sure NodeRunner is unique, maybe they have another Start code

rskj-core/src/test/java/org/ethereum/crypto/signature/Secp256k1Test.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    public void testInitialization_Native() {
+        // Test Native init
+        Secp256k1.reset();
+        RskSystemProperties properties = Mockito.mock(RskSystemProperties.class);
+        Mockito.when(properties.cryptoLibrary()).thenReturn("native");
+        Secp256k1.initialize(properties);
+        assertTrue(Secp256k1.getInstance() instanceof Secp256K1ServiceNative);
```
ajlopezrsk, line 53: Are we sure that this test will pass in ANY development environment? ie Windows 10

rskj-core/src/test/java/org/ethereum/crypto/signature/Secp256K1ServiceNativeTest.java
```diff
@@ -0,0 +1,8 @@
+package org.ethereum.crypto.signature;
+
+public class Secp256K1ServiceNativeTest extends Secp256k1ServiceTest {
+
+    public Secp256K1ServiceNativeTest() {
+        super(new Secp256K1ServiceNative());
```
ajlopezrsk, line 6: Are we sure that this test will pass in any development environment? ie Windows 10

rskj-core/src/test/java/org/ethereum/crypto/signature/Secp256k1Test.java
patogallaiovlabs: Thats something that we have to evaluate when implementing the Native part of the app. In this case we should "mock" that part of the logic. Considering the case when the OS is not supported. I leave a TODO in the instance initialization.

rskj-core/src/test/java/org/ethereum/crypto/signature/Secp256K1ServiceNativeTest.java
patogallaiovlabs: For now yes. But when we actually implement this, we should consider OS. So, its a really good question for the next PR.

rskj-core/src/main/java/org/ethereum/crypto/signature/Secp256k1.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public static final Secp256k1Service getInstance() {
+        if (instance == null) {
+            setDefault();
+        }
+        return instance;
+    }
+
+    private static synchronized void setDefault() {
+        instance = new Secp256k1ServiceBC();
```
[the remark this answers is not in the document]
patogallaiovlabs: I change the implementation again. Check what you think of that.

patogallaiovlabs:
pipeline: run

rskj-core/src/main/java/org/ethereum/crypto/signature/SignatureService.java
patogallaiovlabs: I change the implementation again. Tell me what you think of it.

PR #1231, opened 2020-05-05
Refactor registerBtcTransaction

guidohernan93:
Create method validationsForRegisterBtcTransaction

Add validation methods on BridgeUtils

Add unit tests on BridgeSupportTest and BridgeUtilTest

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -119,7 +119,7 @@ public class BridgeSupport {
     private final FederationSupport federationSupport;
 
     private final Context btcContext;
-    private BtcBlockStoreWithCache.Factory btcBlockStoreFactory;
+    private final BtcBlockStoreWithCache.Factory btcBlockStoreFactory;
     private BtcBlockStoreWithCache btcBlockStore;
     private BtcBlockChain btcBlockChain;
     private final org.ethereum.core.Block rskExecutionBlock;
@@ -263,7 +263,7 @@ public class BridgeSupport {
         Sha256Hash btcTxHash = BtcTransactionFormatUtils.calculateBtcTxHash(btcTxSerialized);
 
         // Check the tx was not already processed
-        if (isAlreadyBtcTxHashProcessedHeight(btcTxHash)) {
+        if (isAlreadyBtcTxHashProcessed(btcTxHash)) {
             return;
         }
 
@@ -276,7 +276,7 @@ public class BridgeSupport {
         btcTx.verify();
 
         // Check again that the tx was not already processed but making sure to use the txid (no witness)
-        if (isAlreadyBtcTxHashProcessedHeight(btcTx.getHash(false))) {
+        if (isAlreadyBtcTxHashProcessed(btcTx.getHash(false))) {
             return;
         }
 
@@ -1163,7 +1163,7 @@ public class BridgeSupport {
      * @return true or false according
      * @throws  IOException
      * */
-    public boolean isAlreadyBtcTxHashProcessedHeight(Sha256Hash btcTxHash) throws IOException {
+    public boolean isAlreadyBtcTxHashProcessed(Sha256Hash btcTxHash) throws IOException {
         if (getBtcTxHashProcessedHeight(btcTxHash) > -1L) {
             logger.warn("Supplied Btc Tx {} was already processed", btcTxHash);
             return true;
@@ -2273,7 +2273,8 @@ public class BridgeSupport {
         }
 
         // Validates inputs count
-        BridgeUtils.validateInputsCount(btcTxSerialized, activations.isActive(ConsensusRule.RSKIP143), btcTxHash);
+        logger.info("Going to validate inputs for btc tx {}", btcTxHash);
+        BridgeUtils.validateInputsCount(btcTxSerialized, activations.isActive(ConsensusRule.RSKIP143));
 
         // Check the the merkle root equals merkle root of btc block at specified height in the btc best chain
         // BTC blockstore is available since we've already queried the best chain height
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -260,77 +260,15 @@ public class BridgeSupport {
      */
     public void registerBtcTransaction(Transaction rskTx, byte[] btcTxSerialized, int height, byte[] pmtSerialized) throws IOException, BlockStoreException {
         Context.propagate(btcContext);
-
         Sha256Hash btcTxHash = BtcTransactionFormatUtils.calculateBtcTxHash(btcTxSerialized);
+
         // Check the tx was not already processed
-        if (getBtcTxHashProcessedHeight(btcTxHash) > -1L) {
-            logger.warn("Supplied Btc Tx {} was already processed", btcTxHash);
+        if (isAlreadyBtcTxHashProcessedHeight(btcTxHash)) {
             return;
         }
 
-        if (height < 0) {
-            String panicMessage = String.format("Btc Tx %s Supplied Height is %d but should be greater than 0", btcTxHash, height);
-            logger.warn(panicMessage);
-            panicProcessor.panic("btclock", panicMessage);
-            return;
-        }
-
-        // Check there are at least N blocks on top of the supplied height
-        int btcBestChainHeight = getBtcBlockchainBestChainHeight();
-        int confirmations = btcBestChainHeight - height + 1;
-        if (confirmations < bridgeConstants.getBtc2RskMinimumAcceptableConfirmations()) {
-            logger.warn(
-                    "Btc Tx {} at least {} confirmations are required, but there are only {} confirmations",
-                    btcTxHash,
-                    bridgeConstants.getBtc2RskMinimumAcceptableConfirmations(),
-                    confirmations
-            );
-            return;
-        }
-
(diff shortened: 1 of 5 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -243,4 +241,49 @@ private static boolean isFromFeePerKbChangeAuthorizedSender(org.ethereum.core.Tr
         AddressBasedAuthorizer authorizer = bridgeConfiguration.getFeePerKbChangeAuthorizer();
         return authorizer.isAuthorized(rskTx);
     }
+
+    public static boolean validateHeightAndConfirmations(int height, int btcBestChainHeight, int acceptableConfirmationsAmount, Sha256Hash btcTxHash) throws Exception {
+        // Check there are at least N blocks on top of the supplied height
+        if (height < 0) {
+            throw new Exception();
```
marcos-iov, line 248: Maybe it would be better to throw a custom exception

guidohernan93: It could be a VerificationException maybe

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1219,6 +1157,21 @@ public Long getBtcTxHashProcessedHeight(Sha256Hash btcTxHash) throws IOException
         return provider.getHeightIfBtcTxhashIsAlreadyProcessed(btcTxHash).orElse(-1L);
     }
 
+    /**
+     * Returns if tx was already processed by the bridge
+     * @param btcTxHash the btc tx hash for which to retrieve the height.
+     * @return true or false according
+     * @throws  IOException
+     * */
+    public boolean isAlreadyBtcTxHashProcessedHeight(Sha256Hash btcTxHash) throws IOException {
```
josedahlquist, line 1166: Funny name
```suggestion
    public boolean isAlreadyBtcTxHashProcessed(Sha256Hash btcTxHash) throws IOException {
```

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
josedahlquist: Yes, we should throw meaningful exceptions. Typed and/or with a message to help the log reader grasp what happened.

Although in this case you could just log and return false. No need to throw an exception I think.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -4147,6 +4143,303 @@ public void hasBtcCoinbaseTransaction_fails_with_null_coinbase_information_after
...
+        BridgeSupport bridgeSupport = getBridgeSupport(
+                bridgeConstants,
+                mock(BridgeStorageProvider.class),
+                mock(Repository.class),
+                mock(BridgeEventLogger.class),
+                null,
+                mockFactory
+        );
+
+        Assert.assertFalse(bridgeSupport.validationsForRegisterBtcTransaction(btcTx.getHash(), 0, pmt.bitcoinSerialize(), btcTx.bitcoinSerialize()));
```
josedahlquist, line 4261: the argument for the block height doesn't match the height you are setting in the stored block

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -4147,6 +4143,303 @@ public void hasBtcCoinbaseTransaction_fails_with_null_coinbase_information_after
...
+    }
+
+    @Test(expected = VerificationException.class)
+    public void validationsForRegisterBtcTransaction_tx_without_inputs_before_rskip() throws IOException, BlockStoreException {
```
josedahlquist, line 4265: I'm a bit confused. Regardles of the rskip, we would never accept tx without inputs.

josedahlquist: and please specify the rskip in the test name

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -4147,6 +4143,303 @@ public void hasBtcCoinbaseTransaction_fails_with_null_coinbase_information_after
...
+                null,
+                mockFactory,
+                activations
+        );
+
+        bridgeSupport.validationsForRegisterBtcTransaction(btcTx.getHash(), 0, pmt.bitcoinSerialize(), btcTx.bitcoinSerialize());
+    }
+
+    @Test(expected = VerificationException.class)
+    public void validationsForRegisterBtcTransaction_tx_without_inputs_after_rskip() throws IOException, BlockStoreException {
```
josedahlquist, line 4307: Same comment

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -4147,6 +4143,303 @@ public void hasBtcCoinbaseTransaction_fails_with_null_coinbase_information_after
...
+        BtcECKey srcKey = new BtcECKey();
+        tx.addInput(PegTestUtils.createHash(1), 0, ScriptBuilder.createInputScript(null, srcKey));
+
+        // Create header and PMT
+        byte[] bits = new byte[1];
+        bits[0] = 0x3f;
+        List<Sha256Hash> hashes = new ArrayList<>();
+        hashes.add(tx.getHash());
+        PartialMerkleTree pmt = new PartialMerkleTree(bridgeConstants.getBtcParams(), bits, hashes, 1);
+        co.rsk.bitcoinj.core.BtcBlock btcBlock =
```
josedahlquist, line 4369: A comment here could help understand the test.
I understand we are creating a tx and a PMT on one hand, and a block which has no relation with said tx and PMT on the other hand; therefore the transaction will be rejected because the merkle block won't match.
Is that right?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -4147,6 +4143,303 @@ public void hasBtcCoinbaseTransaction_fails_with_null_coinbase_information_after
...
+        tx.addInput(PegTestUtils.createHash(1), 0, ScriptBuilder.createInputScript(null, srcKey));
+
+        // Create header and PMT
+        byte[] bits = new byte[1];
+        bits[0] = 0x3f;
+        List<Sha256Hash> hashes = new ArrayList<>();
+        hashes.add(tx.getHash());
+        PartialMerkleTree pmt = new PartialMerkleTree(bridgeConstants.getBtcParams(), bits, hashes, 1);
+        Sha256Hash merkleRoot = pmt.getTxnHashAndMerkleRoot(new ArrayList<>());
+        co.rsk.bitcoinj.core.BtcBlock btcBlock =
```
josedahlquist, line 4419: A comment in the test would help reviewers.
In this test, the block includes a valid merkleRoot that you calculate from the PMT you craft.
Good test!

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -48,8 +50,10 @@
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import sun.nio.ch.Net;
```
josedahlquist, line 53: This dependency is truly used?!

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -511,6 +515,79 @@ public void getCoinFromBigInteger() {
...
+    }
+
+    @Test
+    public void validateHeightAndConfirmation_enough_confirmations() throws Exception {
+        Assert.assertTrue(BridgeUtils.validateHeightAndConfirmations(2, 5, 3, Sha256Hash.of(Hex.decode("ab"))));
+    }
+
+    @Test(expected = Exception.class)
+    public void calculateMerkleRoot_invalid_pmt() {
+        BridgeUtils.calculateMerkleRoot(mock(NetworkParameters.class), Hex.decode("ab"), null);
```
josedahlquist, line 535: Be careful sending a mock in this way, plus expecting `Exception`. The method could be failing with an NPE if internally depends on any value from the mocked NetworkParameters.

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -243,4 +241,49 @@ private static boolean isFromFeePerKbChangeAuthorizedSender(org.ethereum.core.Tr
...
+        List<Sha256Hash> hashesInPmt = new ArrayList<>();
+        Sha256Hash merkleRoot = pmt.getTxnHashAndMerkleRoot(hashesInPmt);
+        if (!hashesInPmt.contains(btcTxHash)) {
+            logger.warn("Supplied Btc Tx {} is not in the supplied partial merkle tree", btcTxHash);
+            return null;
+        }
+        return merkleRoot;
+    }
+
+    public static void validateInputsCount(byte[] btcTxSerialized, boolean isActiveRskip, Sha256Hash btcTxHash) throws VerificationException.EmptyInputsOrOutputs {
```
josedahlquist, line 274: You could probably remove hash parameter as it doesn't have an actual usage internally, and instead, logging in the callers directly.

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
guidohernan93: Throws exception with message now, as it was throwing exception originally. In case we prefer to log and return false, we should remove the try/catch block on the call.

PR #1235, opened 2020-05-11
Refactor of ECKey signature related methods: signatureToKey / recover…

patogallaiovlabs:
The idea is to refactor ECKey class in order to prepare functionality for Native Signature implementation with a C Native library. What i did is a Singleton service class, that defines an interface of methods that should be implemented in Bouncy Castle / Native implementations. The refactor includes only the methods (and where its used):
1. signatureToKey()
* PeerDiscoveryMessage.getKey()
* Transaction.getSender()
* PrecompiledContract.ERecover.execute()
2. recoverFromSignature()
* ProofOfWorkRule.validFallbackBlockSignature()
* Transaction.getKey()
* ECKey.sign() → (for making it recoverable)
* EncryptionHandshake.makeAuthInitiate()
3. -verify()
* BootstrapDataVerifier.verifyEntries()


josedahlquist:
pipeline:run

patogallaiovlabs:
pipeline:run

patogallaiovlabs:
pipeline:run

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -848,6 +849,7 @@ protected NodeRunner buildNodeRunner() {
                     getRepositoryLocator()
             ));
         }
+        Secp256k1.initialize(getRskSystemProperties());
```
nicops, line 852: I would move this somewhere else because CLI tools don't build the node runner and thus they never execute this line. We would want ideally to have this run in those tools as well.

rskj-core/src/test/java/org/ethereum/crypto/signature/ECDSASignatureTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+
+    /**
+     * @param r    -
+     * @param s    -
+     * @param hash - the hash used to compute this signature
+     * @param pub  - public key bytes, used to calculate the recovery byte 'v'
+     * @return -
+     */
+    public static ECDSASignature fromComponentsWithRecoveryCalculation(byte[] r, byte[] s, byte[] hash, byte[] pub) {
```
nicops, line 137: remove these two methods and point to the actual methods in ECDSASignature

rskj-core/src/main/java/co/rsk/RskContext.java
patogallaiovlabs: I've been thinking about that... Where else would you suggest to put it?
Because if you dont initialize, it would take default Bouncy Castle. So CLI tools not taking the config... shouldn’t be actually a problem, plus CLI are still experimental.
And couldn't find any other possible way of running (besides CLI) without building a NodeRunner, Fed Node extends from this, so its safe.

rskj-core/src/test/java/org/ethereum/crypto/signature/ECDSASignatureTest.java
patogallaiovlabs: Done

patogallaiovlabs:
pipeline:run

patogallaiovlabs:
pipeline:run

fedejinich:
> I recommend catching exceptions from Secp256k1Service methods always and returning a failure value. But that can wait until a native library is implemented.

in that case, shouldn't you stop the node?

PR #1251, opened 2020-06-04
CI: migrate SonarQube docker image to JDK 11

lucasvuotto:
While there, rename jvm-executor to rskj-executor to avoid confusion.


patogallaiovlabs (approved):
LGTM!

PR #1252, opened 2020-06-08
Add support of Java11; Migrate to Gradle v6.5

Vovchyk:
With this change it is now possible to run the rskj node under jre11.

Changes introduced in this PR:
- Added support of java11, but source compatibility is still java8 (`sourceCompatibility = 1.8`)
- Migrated to `Gradle v6.5`
- Removed `witness` plugin in favor of Gradle built-in deps verification capability
- Minor unit tests fixes

Note 1: It was added a feature to print system info (e.g. `java.version`, `os.name` etc.) with help of `--print-system-info` cli flag.

Note 2: `gradle/verification-metadata.xml` file with deps hashes is generated by this command:
```bash
./gradlew --write-verification-metadata sha256 assemble
```
... so if new dependencies are added later, this command can be used to update `verification-metadata.xml`.

*fed: j11-support*


Vovchyk:
Resolves #1255 and #1256

aeidelman:
@josedahlquist please review changes in Federation code

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -54,7 +55,7 @@ public Federation(List<FederationMember> members, Instant creationTime, long cre
         // effectively immutable
         this.members = Collections.unmodifiableList(members.stream().sorted(FederationMember.BTC_RSK_MST_PUBKEYS_COMPARATOR).collect(Collectors.toList()));
 
-        this.creationTime = creationTime;
+        this.creationTime = creationTime.truncatedTo(ChronoUnit.MILLIS);
```
josedahlquist, line 58: I'm a bit confused by this change. Why can't we directly assign the `creationTime` as parameterised?

Vovchyk: I did leave a comment here about this change, but seems it's disappeared after I've update the PR.

The issue here is that in java11 nanoseconds part of a `Instant` object is not zero. In the `BridgeSerializationUtils` class we have two helper methods `serializeFederationWithSerializer` and `deserializeFederationWithDesserializer`. And they do not take into account nanoseconds while serializing / deserializing `Federation` objects (e.g the `serializeFederationWithSerializer` method uses this approach `federation.getCreationTime().toEpochMilli()`, so only milliseconds are being taken and nanos are ignored). Without this change unit tests fail, as serialized `Federation` object after deserializing does not equal to original one.

Also not sure if we can serialize nanos, as older nodes and the ones run by jvn8 would not understand it

Vovchyk:
pipeline: run

rskj-core/src/main/java/co/rsk/peg/Federation.java
josedahlquist: thanks for the clarifications

josedahlquist:
pipeline:run

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

bcodesido (changes requested):
Nice work!
Please modify the `configure.sh` in order to stop downloading the gradle-witness plugin because isn't necessary anymore.
Once modified, we can push the new signature of that file.

bcodesido (dismissed):

aeidelman (approved):

PR #1254, opened 2020-06-10
Lc sync common ancestor

julianlen:


julianlen:
This is the first step of the sync. The searching of the common ancestor is useful to start the syncing between two peers (Light client peer or LC and Full client peer or FN) from a known (in the network) block. Once the light peer sends the status message and the full peer answers, the ancestor search begins.   

To implement this, we use the state pattern. The lightSyncProcessor in LC starts in IdleSyncState. When a Status message arrives, the sync starts with DecidingLightSyncState. The latter decides the transition depends on the best block stored in the blockchain. If the best block is different from zero, then LC transits to the CommonAncestorSearchState state. 

The LC in CommonAncestorSearchState state requests to FN (starting from its best block number and going backward) all FN's blocks. LC repeats this process until an FN's block is known by LC.
Finally, it produces the starting block for sync.

PR #1258, opened 2020-06-18
Implementation of Recovery Signature with Native Library

patogallaiovlabs:
- Implementation of **Secp256k1ServiceNative.recoverFromSignature()** method.
It can be configured by property, but also if the OS is not compatible, we use the BouncyCastle implementation and we dont run the tests related with this implementation.

- Also added a check of Point at Infinity when recovering a Pub Key that its point is infinity. If thats the case, the recovery returns "null". The most used case is the one that Priv Key = 0 (used in unit tests).

- Fixed some init issues of Singleton service class Secp256k1.
- Originally we were going to implement "verify()" method too, but it was harder than we thought and worthless so far.

- **Note** : It only supports Linux OS.

*fed: signature_native*


patogallaiovlabs:
pipeline:run

rskj-core/build.gradle
```diff
@@ -217,6 +219,7 @@ task jar(type:Jar, overwrite: true, dependsOn: generateResources) {
     }
     from sourceSets.main.output.classesDirs
     from sourceSets.main.output.resourcesDir
+    from sourceSets.test.output.classesDirs //Just for TestSystemProperties
```
Vovchyk, line 222: could you explain a bit more why we need this? Does it mean that the final jar will contain test classes?

patogallaiovlabs: Nice catch! Forgot to remove it, i added it when playing with benchmarking tool.

rskj-core/src/main/java/co/rsk/remasc/RemascTransaction.java
```diff
@@ -73,7 +74,7 @@ public long transactionCost(Constants constants, ActivationConfig.ForBlock activ
     }
 
     @Override
-    public RskAddress getSender() {
+    public synchronized RskAddress getSender(ActivationConfig.ForBlock activations, SignatureCache signatureCache) {
```
nicops, line 77: synchronized seems off here

donequis: It's definitely off for me too. In both remasc and normal transactions

rskj-core/src/main/java/org/ethereum/crypto/signature/Secp256k1ServiceNative.java
```diff
@@ -19,10 +19,88 @@
...
+
+    /**
+     * Returns a (r.length + s.length) bytes array long
+     *
+     * Note: we take 32 bytes from "r" and 32 bytes from "s".
+     *
+     * @param sig {r,s}
+     * @return r + s (64 length byte array)
+     */
+    byte[] concatenate(ECDSASignature sig) {
```
donequis, line 74: This method is ugly when r and s are longer than 32 bytes. If we prepend garbage before the real r and s values, those bytes will be ignored. This does not happen with the BC implementation.

There's no immediate issue with that because the Transaction class does an extra validation on the r and s components, but those are not the only places where the recover from signature is used, causing some ugly cases there.

I think that we should fail quickly here. If r or s are too big.

patogallaiovlabs:
pipeline:run

patogallaiovlabs:
pipeline:run

donequis (dismissed):

patogallaiovlabs:
pipeline:run

rskj-core/src/main/java/org/ethereum/crypto/signature/Secp256k1ServiceNative.java
patogallaiovlabs: fixed.

ajlopezrsk (approved):

PR #1259, opened 2020-06-23
Create a new function calculateReceiptsTrieRoot to calculate the rece…

pamgonzalez:
…ipt merkle proof of the transaction.

Given the hash of a transaction, the block and receiptStore, the receipt merkle proof is returned. For the construction of this function, the getTransactionReceiptNodesByHash function of RskModule was considered.


josedahlquist:
pipeline:run

PR #1262, opened 2020-06-26
Use udp packet port in PeerExplorer

nicops:
This makes the PeerExplorer reply to ping and pongs using the ip and port of the udp packet instead of only the ip plus the port inside the message. 

Without this change, nodes that are behind routers that change the udp port cannot connect to the network succesfully.


patogallaiovlabs:
pipeline:run

Issue #1127, opened 2019-12-19
Transaction is not being forwarded to all the peers after (what we think a reorganization).


ignaciopulicedonatto:
**Overview**: 
We have a script that sends a transaction every certain time (time doesn’t matter except you can reproduce it quicker if time is short between requests).

**Debugging steps**: 
1) Transaction is send to node
2) Transaction is accepted, everything looks good with the transaction
3) Transaction is send to the peers 
4) Transaction comes in one block mined.
5) There are several reorganization in the network that affect the mined block.
transaction should be added again into txpool.
6) From here this transaction is never mined again.
7) We send a new transaction with the same hash (same nonce, same price, same everything) so hash is the same as the previous one.
8) Up to here nothing happens and the new transaction will never get mined.

**Things we believe is happening**:
1) Transaction with the same hash was previously sent and our node doesn't resend it because it has been sent to all the peers. 
(TransactionNodeInFormation.java method addTransactionNode saves a map key-pair for every transaction it has, a set of node friends, this starts to clean after reaching 1000 transactions).
Peers friends may have the same problem when they receive transactions with the same hash, they discard it because they already received it.
2) Reorganization is discarding/losing original transaction. For some reason, that we don’t know, this transaction is never added in one block to mine after reorganizations. This reorganization at certain heights happens 3 times as we could see ( IMPORTED_BEST blocks). And then it never gets mined again.

**Up to here, we think there is more than one bug**. 
1) The transaction with the same hash should be able to resend it to the network after a period of time.
2) In reorganizations, transactions are being discarded or lost.

**Things we had tried to workaround the bug**:
Workaround 1. We tried to restart the node, nothing happened almost every time. 
Workaround 2. Every time we build the transaction we change the price a little bit (adding a random value) to change the value of the hash, and that works.

I uploaded all the logs (script.log,  rsk.log) in a [drive](https://drive.google.com/open?id=1mm5ma2x6D0nsgAVkr79vXux-e0pkACpz). With this you can see that the first time we get an ERROR in ourscript.log there is a reorganization at that time. When a reorganization happens for block X you will see in rsk.log that there are 3 IMPORTED_BEST  for that block.
The raw transaction is printed in script.log, so you can start from that point.

ajlopez:
The transaction is forwarded to all the peers of the node, when it is received and validated. There is no need to forward the transaction again.

[...]

PR #1266, opened 2020-07-07
Resend transactions fix and refactor

fedejinich:
# Resend Transactions

## Goals
- reBroadcast transactions
- Remove TransactionNodeInfo which was storing a lot of transactions (increasing memory use) 

## Context

There are some cases where you want to resend transactions to activePeers for example when, they have expired, or mining nodes have been restarted, or if there were a bug where tx were lost like the [previous bug](https://github.com/rsksmart/rskj/issues/1127).

When @nicops  [fixed](https://github.com/rsksmart/rskj/pull/1253) [this issue](https://github.com/rsksmart/rskj/issues/1127), he discover that it wasn't posible to rebroadcast transactions to activePeers because `TransactionNodeInformation` was blocking it. Then we decided to remove that class because that blocking was no longer needed.

This is the actual flow (before this PR) when a transaction arrives to the node:
- RPC flows:
	- `PersonalModuleWalletEnabled.sendTransaction` -> `EthereumImpl.submitTransaction` -> `TransactionPoolImpl.addTransaction`
	 - `Web3EthModule.eth_sendRawTransaction` -> `EthModuleTransactionBase.sendRawTransaction` -> `TransactionPoolImpl.addTransaction`
	 - `Web3EthModule.eth_sendTransaction` -> `EthModuleTransactionBase.sendTransaction` -> `TransactionPoolImpl.addTransaction`
- Peer node flow:
    - `MessageVisitor.apply(TransactionsMessage)` -> `TransactionGateway.receiveTransactionsFrom` -> `TransactionPool.addTransactions` -> `TransactionPool.addTransaction`

## Implemented

- Removed `TransactionNodeInfo` class (and tests)
- Refactored `TransactionGateway` class
    - `receiveTransactionsFrom` now broadcasts transactions after adding them to `TransactionPool`
    - Removed a nonsense listener: `OnPendingTransactionsReceivedListener`
- Incoming transactions should only be added to TxPool via TxGateway: replaced where we used to call `transactionPool.addTransaction` with `transactionGateway.receiveTransaction/sFrom` (because invoking `addTransaction` was also producing a `onPendingTransactionsReceived` & `onTransactionPoolChanged` event )
- Removed `@Ignore` annotation from ChanelManagerImplTests 

Note: `onPendingTransactionsReceived` was being consumed from `TransactionGateway` which then used to call `channelManager.broadcastTransaction`. Apart from this there is another consumer from this event which is not related with this PR.


fedejinich:
added some comments at `TransactionPool`

rskj-core/src/main/java/co/rsk/net/TransactionGateway.java
```diff
@@ -18,66 +18,41 @@
...
-        @Override
-        public void onPendingTransactionsReceived(List<Transaction> txs) {
-            for (Transaction tx : txs) {
-                Keccak256 txHash = tx.getHash();
-                Set<NodeID> nodesToSkip = new HashSet<>(transactionNodeInformation.getNodesByTransaction(txHash));
-                Set<NodeID> newNodes = channelManager.broadcastTransaction(tx, nodesToSkip);
-
-                newNodes.forEach(nodeID -> transactionNodeInformation.addTransactionToNode(txHash, nodeID));
-            }
+    public TransactionPoolAddResult receiveTransactionFrom(Transaction transaction) {
```
patogallaiovlabs, line 51: why the suffix "From"?

rskj-core/src/main/java/org/ethereum/core/TransactionPoolAddResult.java
```diff
@@ -42,11 +48,15 @@ public void ifTransactionWasNotAdded(Consumer<String> errorConsumer) {
...
 
-    public static TransactionPoolAddResult ok() {
-        return new TransactionPoolAddResult(true, null);
+    public static TransactionPoolAddResult ok(Transaction transaction) {
+        return new TransactionPoolAddResult(true, null, Collections.singletonList(transaction));
     }
 
     public static TransactionPoolAddResult withError(String errorMessage) {
-        return new TransactionPoolAddResult(false, errorMessage);
+        return new TransactionPoolAddResult(false, errorMessage, null);
```
patogallaiovlabs, line 56: This could cause a NPE in the check "if(!result.getTransactionsAdded().isEmpty()) {" L53 TransactionGateway.

rskj-core/src/main/java/org/ethereum/net/server/ChannelManager.java
```diff
@@ -78,4 +78,6 @@
     Collection<Peer> getActivePeers();
 
     boolean isAddressBlockAvailable(InetAddress address);
+
+    Set<NodeID> broadcastTransactions(@Nonnull List<Transaction> transactions, @Nonnull Set<NodeID> nodeID);
```
patogallaiovlabs, line 82: Be careful with the use of NodeID... most of the impl returns a copy of those. So, i wonder if it is actually cheaper to return the Node instead of NodeID.

rskj-core/src/main/java/org/ethereum/core/TransactionPoolAddResult.java
fedejinich: fixed

rskj-core/src/main/java/org/ethereum/net/server/ChannelManager.java
fedejinich: didn't get that. do you think it's worth it? this is just a reuse of broadcastTransaction which has been there before this PR

fedejinich:
the weird thing is that there was an `@Ignore` for ChannelManagerImplTests. I've removed that

fedejinich:
pipeline:run

rskj-core/src/main/java/co/rsk/net/TransactionGateway.java
fedejinich: removed "from" suffix

fedejinich:
pipeline:run

fedejinich:
squashed commits

nicops:
> When @nicops fixed issue, he discover that there was another step to complete that fix: When a reorganisation was produced the node wasn’t resending those transactions to active peers, because we used to kept track of transactions that were already sent to other nodes.

Just to clarify, the fix I did for reorganization is more or less independent of this. Yes, when a reorganization is added, the node isn't resending those transactions. But neither is on this PR, because it doesn't go through the tx gateway. And that's fine because you don't need to resend them, it is expected other nodes already have them.

Instead, this change would apply for *any* scenario you want to resend a tx if, for example, they have expired, or mining nodes have been restarted, or if there were a bug where tx were lost like the previous bug.

fedejinich:
Last updates:
- Only broadcasting pending transactions instead of broadcasting queued and pending transactions
- Avoid caching Message.Transaction at NodeMessageHandler.receivedMessage -> this allows resending expired txs

fedejinich:
Now It's also posible to share the same logic at `addTransaction/addTransactions` at `TransactionPool`

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
```diff
@@ -512,8 +514,10 @@ public void addTwiceAndGetQueuedTransaction() {
         transactionPool.addTransaction(tx);
 
         TransactionPoolAddResult result = transactionPool.addTransaction(tx);
-        Assert.assertFalse(result.transactionWasAdded());
-        result.ifTransactionWasNotAdded(msg -> Assert.assertEquals("queued transaction with same hash already exists", msg));
+        Assert.assertFalse(result.transactionsWereAdded());
+        if(!result.transactionsWereAdded()) {
```
nicops, line 518: You should remove this if, it is confusing, you already asserted it was false the line above

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
```diff
@@ -557,8 +561,10 @@ public void checkTxWithSameNonceIsRejected() {
 
         TransactionPoolAddResult result = transactionPool.addTransaction(tx2);
 
-        Assert.assertFalse(result.transactionWasAdded());
-        result.ifTransactionWasNotAdded(msg -> Assert.assertEquals("gas price not enough to bump transaction", msg));
+        Assert.assertFalse(result.transactionsWereAdded());
+        if(!result.transactionsWereAdded()) {
```
nicops, line 565: ditto

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
```diff
@@ -583,8 +589,10 @@ public void checkTxWithHighGasLimitIsRejected() {
 
         TransactionPoolAddResult result = transactionPool.addTransaction(tx);
 
-        Assert.assertFalse(result.transactionWasAdded());
-        result.ifTransactionWasNotAdded(msg -> Assert.assertEquals("transaction's gas limit of 3000001 is higher than the block's gas limit of 3000000", msg));
+        Assert.assertFalse(result.transactionsWereAdded());
+        if(!result.transactionsWereAdded()) {
```
nicops, line 593: ditto

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
```diff
@@ -802,8 +825,8 @@ public void addTwoTransactionsOutOfOrderInNonceUsingAddTransaction() {
         Transaction tx1 = createSampleTransactionWithGasPrice(1, 0, 1000, 0, 1);
         Transaction tx2 = createSampleTransactionWithGasPrice(1, 0, 2000, 1, 2);
 
-        Assert.assertTrue(transactionPool.addTransaction(tx2).transactionWasAdded());
-        Assert.assertTrue(transactionPool.addTransaction(tx1).transactionWasAdded());
+        Assert.assertTrue(transactionPool.addTransaction(tx2).transactionsWereAdded());
+        Assert.assertTrue(transactionPool.addTransaction(tx1).transactionsWereAdded());
```
nicops, line 829: I would make more assertions here about the return value of each addTransaction. Check that the first one returns ok queued with tx2,  the latter one should add ok pending with the list of the two tx in the right order (first tx1, then tx2).

nicops: the test below (addTwoTransactionsOutOfOrderInNonceUsingAddTransactions), you didn't modify it on this PR, but should have assertions done to the return value as well

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModuleTransactionBase.java
```diff
@@ -64,12 +67,15 @@ public synchronized String sendTransaction(Web3.CallArguments args) {
                 args.data = args.data.substring(2);
             }
 
-            synchronized (transactionPool) {
+            synchronized (transactionGateway) {
```
nicops, line 70: I don't think this is correct. This method derives the account nonce based on the line right below this one. To do that consistently it needs to synchronized over the `transactionPool` because it is accessing it as you can see in the line right below this one.

fedejinich: got it

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
fedejinich: yeah it sounds weird but in this way i'm asserting error message

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
fedejinich: also this `addTransaction` only retrieves an internal `TransactionPoolAddResult` instead of returning internal result+successors (successors are also in pending state now)

fedejinich: I'll raise another issue to refactor those two methods into one

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
fedejinich: removed unnecesary if

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
fedejinich: removed unnecessary if

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
fedejinich: removed unnecessary if

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModuleTransactionBase.java
fedejinich: now it synchronises with txPool instead of txGateway

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
fedejinich: #1286 https://github.com/rsksmart/rskj/issues/1286

fedejinich:
In [last commit](https://github.com/rsksmart/rskj/commit/f6216e6df378c957096921d8450c505c0ef4819c) 
- I've added more detailed test cases for `TransactionPool` 
- Fixed a bug at `addTransaction`: it was only returning one pending transaction instead of returning that pending transaction and its successors

fedejinich:
Also spotted a posible refactor #1286

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
nicops: :ok_hand: :ok_hand: :ok_hand:

fedejinich:
In [last commit](https://github.com/rsksmart/rskj/commit/12839722581405c1d3446298a32e880c172b0481) 
- I've added more detailed test cases for `TransactionPool` 
- Synchronised `transactionPool` instead of `transactionGateway` at `EthModuleTransactionBase.sendTransaction`
- Fixed a bug at `addTransaction`: it was only returning one pending transaction instead of returning that pending transaction and its successors

fedejinich:
pipeline:run

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -118,7 +119,7 @@ public void postMessage(Peer sender, Message message) {
     private void tryAddMessage(Peer sender, Message message) {
         Keccak256 encodedMessage = new Keccak256(HashUtil.keccak256(message.getEncoded()));
         if (!receivedMessages.contains(encodedMessage)) {
-            if (message.getMessageType() == MessageType.BLOCK_MESSAGE || message.getMessageType() == MessageType.TRANSACTIONS) {
+            if (message.getMessageType() == MessageType.BLOCK_MESSAGE) {
```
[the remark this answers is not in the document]
fedejinich: well this was a team decision, the idea was based on the case "when a transaction expires". in the old way we couldn't even process that resented transaction (despite expiration time)

rskj-core/src/main/java/co/rsk/net/TransactionGateway.java
```diff
@@ -18,66 +18,41 @@
...
-import java.util.Set;
+import javax.annotation.Nonnull;
+import java.util.*;
 
 /**
  * Centralizes receiving and relaying transactions, so we can only distribute information to nodes that don't already
  * have it.
  */
-public class TransactionGateway implements InternalService {
+public class TransactionGateway {
```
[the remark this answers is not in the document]
fedejinich: 💯

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -159,23 +159,23 @@ private RepositorySnapshot getCurrentRepository() {
     }
 
     private List<Transaction> addSuccessors(Transaction tx) {
-        List<Transaction> added = new ArrayList<>();
+        List<Transaction> pendingTransactionsToBeAdded = new ArrayList<>();
```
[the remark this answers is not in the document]
fedejinich: yes you're right!

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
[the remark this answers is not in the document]
fedejinich:
```
But now, this logic helps a lot to discard incoming message that came many times (ie, now it's easy to see the transaction message come from many peers, even before the first transaction message could be processed)
```
what do you think about this @juli @donequis ?

fedejinich:
In [this commit](https://github.com/rsksmart/rskj/commit/bd15017245dad912d1c58281d395f459af9bed0d) 
- I've added more detailed test cases for `TransactionPool` 
- Synchronised `transactionPool` instead of `transactionGateway` at `EthModuleTransactionBase.sendTransaction`
- Fixed a bug at `addTransaction`: it was only returning one pending transaction instead of returning that pending transaction and its successors

NOTE: I've added one test case to `NodeMessageHandlerTest` in [this](https://github.com/rsksmart/rskj/pull/1266/commits/a615c435adfce21b68a9fa05f881b9b196fbc9b8) commit

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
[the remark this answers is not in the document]
donequis: Does it still help? These are the first lines of the `internalAddMethod`

```
        if (pendingTransactions.hasTransaction(tx)) {
            return TransactionPoolAddResult.withError("pending transaction with same hash already exists");
        }

        if (queuedTransactions.hasTransaction(tx)) {
            return TransactionPoolAddResult.withError("queued transaction with same hash already exists");
        }
```

fedejinich:
pipeline:run

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -127,7 +126,8 @@ private void tryAddMessage(Peer sender, Message message) {
 
             double score = sender.score(System.currentTimeMillis(), message.getMessageType());
 
-            if (score >= 0 && !this.queue.offer(new MessageTask(sender, message, score))) {
+            boolean notAdded = !this.queue.offer(new MessageTask(sender, message, score));
```
donequis, line 129: Messages should be offered only on non negative scores

fedejinich: extracted that part into `addMessage` method

PR #1268, opened 2020-07-07
Start Round Sync State

julianlen:
On the following pr. the next feature were added:

- In GetBlockHeader message were renamed variables. This message is going to ask for a number of headers from a starting number instead of from a best block. Because this, the variable bestBlockHeader where renamed as startBlockHeader

- Because the light client works with headers, syncprocessor doesn't receive blocks.

- Two peers (light and full) are being sync. The full peer starts sending to the light peer blocks until the latter receives a known block. The known block is the common ancestor between peers. Then, the light peer transits to StartRoundSync state where starts downloading block headers, not all blockheaders from the common ancestor to the best block known by the light peer, but some of them which are called pivots.

- Also in testing, two mains things were made. The first one, change all testing where the state's transition ends in SearchCommonAncestor to ends in StartSyncRound or, to an error state MoreBlocksThanAllowed or MessageNotRequested and also add tests for the StartRoundSync state. The second one, extracted methods in tests to made these more readable. This is a step to move all auxiliar methods to TestUtils. 

It's important to know that the error states will disappear and will be replaced for a error handler.


rskj-core/src/main/java/co/rsk/net/light/LightSyncProcessor.java
```diff
@@ -120,44 +123,77 @@ public void sendStatusMessage(LightPeer lightPeer) {
                 block.getNumber(), lightPeer.getPeerIdShort());
     }
 
-    public void sendBlockHeadersMessage(LightPeer lightPeer, byte[] bestBlock, int max, int skip, boolean reverse) {
-        GetBlockHeadersByHashMessage blockHeaderMessage = new GetBlockHeadersByHashMessage(++lastRequestedId, bestBlock, max, skip, reverse);
+    public void sendBlockHeadersMessage(LightPeer lightPeer, byte[] startBlockHash, int maxAmountOfHeaders, int skip, boolean reverse) {
```
smcsicardi, line 126: the method naming is a little confusing, I know they do the same thing, but with different messages, you should rename it reflecting that

PR #1270, opened 2020-07-08
Fix race condition around newPeers in ChannelManagerImpl

nicops:
This fixes the bug where a MessageQueue was witnessed accumulating messages that are never sent, creating a memory leak.

The root cause of the bug is due to a race condition that happens when there is a disconnection of a channel at the same time the channel is being promoted to active in the ChannelManagerImpl class. 

How it happens

1. There is a disconnection in the channel. Two things happen here. There is a listener set up on EthereumChannelInitializer for disconnection that ends up calling ChannelManagerImpl#notifyDisconnect with the Channel. This ends up removing the channel from activePeers and/or newPeers collection. There is also the P2pHandler channelInactive method which is also called on the disconnection by Netty which ends up calling MessageQueue#close, which stops the MessageQueue timer that sends the messages and empties the queue. 
2. At the same time this is happening, ChannelManagerImpl#handleNewPeersAndDisconnections is being called, since it runs automatically once per second with an ScheduledExecutorService. So the channel got removed from newPeers at the same time it was being promoted to active. This is the sequence being called: ChannelManagerImpl#handleNewPeersAndDisconnections -> ChannelManagerImpl#tryProcessNewPeers -> ChannelManagerImpl#processNewPeers -> ChannelManagerImpl#processNewPeer -> ChannelManagerImpl#addToActives -> SyncPool#addToActives

So after that, the MessageQueue gets deactivated (logged as "Cancelling Message Queue") but because it's still referenced as active by ChannelManagerImpl, it still sends messages. The messages AREN'T actually being sent, the deactivated MessageQueue simply keeps them piling up, thus creating a memory leak.

To reproduce it, I added this at the end of EthereumChannelInitializer

            new Thread(() -> {
                try {
                    int milis = new Random().nextInt(3000);
                    logger.info("sleeping {}", milis);
                    Thread.sleep(milis);
                    logger.info("finished sleeping {}", milis);
                } catch (InterruptedException e) {
                    e.printStackTrace();
                }
                ch.close();
            }).start();

To kill every new connection after a few seconds. Also I added a Thread Sleep at the beginning of ChannelManagerImpl#processNewPeer. This forces the race condition to happen, which can be verified by the new logs I added to MessageQueue.

To fix the race condition I added synchronization blocks around the usage of the newPeers attribute in ChannelManagerImpl. I've verified that, with the code I added to force the race condition, the problem does not appear. If disconnection happens at the same time a promotion to active were to happen, promotion OR disconnection can't run at the same time. So it either gets disconnected so it can't be promoted, or if it happened to be promoted then it's disconnected.

I decided to include the new logs in the MessageQueue in this PR in case we happen to need them again, but by default, those logs are not set up in the logback.xml

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/net/server/ChannelManagerImpl.java
+++ b/rskj-core/src/main/java/org/ethereum/net/server/ChannelManagerImpl.java
@@ -292,10 +292,12 @@ public class ChannelManagerImpl implements ChannelManager {
         channel.onDisconnect();
         synchronized (newPeers) {
             if(newPeers.remove(channel)) {
-                logger.info("Peer removed from active peers: {}", channel.getPeerId());
+                logger.info("Peer removed from new peers list: {}", channel.getPeerId());
             }
             synchronized (activePeersLock){
-                activePeers.values().remove(channel);
+                if(activePeers.values().remove(channel)) {
+                    logger.info("Peer removed from active peers list: {}", channel.getPeerId());
+                }
             }
             syncPool.onDisconnect(channel);
         }
--- a/rskj-core/src/main/java/org/ethereum/net/server/ChannelManagerImpl.java
+++ b/rskj-core/src/main/java/org/ethereum/net/server/ChannelManagerImpl.java
@@ -120,9 +120,11 @@ public class ChannelManagerImpl implements ChannelManager {
     }
 
     private void processNewPeers() {
-        List<Channel> processedChannels = new ArrayList<>();
-        newPeers.stream().filter(Channel::isProtocolsInitialized).forEach(c -> processNewPeer(c, processedChannels));
-        newPeers.removeAll(processedChannels);
+        synchronized (newPeers) {
+            List<Channel> processedChannels = new ArrayList<>();
+            newPeers.stream().filter(Channel::isProtocolsInitialized).forEach(c -> processNewPeer(c, processedChannels));
+            newPeers.removeAll(processedChannels);
+        }
     }
 
     private void processNewPeer(Channel channel, List<Channel> processedChannels) {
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/org/ethereum/net/server/ChannelManagerImpl.java
```diff
@@ -288,12 +290,14 @@ public void add(Channel peer) {
...
         channel.onDisconnect();
-        syncPool.onDisconnect(channel);
-        synchronized (activePeersLock){
-            activePeers.values().remove(channel);
-        }
-        if(newPeers.remove(channel)) {
-            logger.info("Peer removed from active peers: {}", channel.getPeerId());
+        synchronized (newPeers) {
+            if(newPeers.remove(channel)) {
+                logger.info("Peer removed from active peers: {}", channel.getPeerId());
```
patogallaiovlabs, line 295: could we log both cases? "new" and "active"?

rskj-core/src/main/java/org/ethereum/net/server/ChannelManagerImpl.java
```diff
@@ -288,12 +290,14 @@ public void add(Channel peer) {
     public void notifyDisconnect(Channel channel) {
         logger.debug("Peer {}: notifies about disconnect", channel.getPeerIdShort());
         channel.onDisconnect();
-        syncPool.onDisconnect(channel);
-        synchronized (activePeersLock){
-            activePeers.values().remove(channel);
-        }
-        if(newPeers.remove(channel)) {
-            logger.info("Peer removed from active peers: {}", channel.getPeerId());
+        synchronized (newPeers) {
```
ajlopezrsk, line 293: Why not add synchronization to add(Channel peer)?

nicops: It wouldn't make a difference. The newPeers array itself is thread safe. EthereumChannelInitializer both calls this add method and registers the listener that will call notifyDisconnect, and it does the add first, so there is no possible race condition there.

See https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/org/ethereum/net/server/EthereumChannelInitializer.java#L119

donequis:
I don't see any problem from a security standpoint.

I'd argue that an issue here is the message queue silently fails when it is used after being closed. There's an interface for closeable stuff, an alternative would be to implement it, and throw an exception if called after being closed. Just an opinion, it can be merged as it is now.

patogallaiovlabs:
pipeline:run

nicops:
I just did a rebase

patogallaiovlabs:
pipeline:run

PR #1273, opened 2020-07-13
TransactionAddResult removal and TransactionPool refactor for addTransaction/s

fedejinich:
This PR is an extension of [previous pr](https://github.com/rsksmart/rskj/pull/1266).

Aims to remove `TransactionAddResult` class which wasn't adding any significant value and was making things more complex.

Also I added a small refactor to `TransactionPool` `addTransaction/s` methods, now `addTransactions` relies on `addTransaction` method. This should have been the first approach of that method but it didn't, now after some refactors we could achieve that approach. 

It's important to mark that when a transaction fails at `addTransactions` for loop, it just logs a warning and continues with the next transaction (since we don't care the status of each one and we don't want to stop the execution)

Plus now an `RskJsonRpcRequestException` has a property for a transaction that produced error


fedejinich:
we've decided another approach which is gonna be implemented in the future

spoiler: make `addTransactions` return `TransactionAddResult` instead of `List<Transaction>`

PR #1276, opened 2020-07-20
Light Client Abort System for the Sync

smcsicardi:
Each time a check is not met, the Abort System removes the Light Peer from the Sync list


rskj-core/src/main/java/co/rsk/net/light/LightSyncProcessor.java
```diff
@@ -305,27 +293,46 @@ public void startFetchRound() {
         //Starting fetch sub chains
     }
 
-    public void differentFirstBlocks() {
-
+    public void differentFirstBlocks(LightPeer lightPeer) {
```
julianlen, line 296: we should sort publics methods on the top and then the private ones

rskj-core/src/test/java/co/rsk/net/light/LightSyncStateTest.java
```diff
@@ -283,8 +292,9 @@ public void peerInStartRoundGetLesserHeadersThanAsked() {
 
         newBlockHeadersInStartRoundSyncState(startBlockHeader, blockHeaders, longLightPeer);
 
-        verify(lightSyncProcessor, times(1)).failedAttempt();
+        verify(lightSyncProcessor, times(1)).failedAttempt(longLightPeer);
         assertEquals(lightSyncProcessor.getSyncState().getClass(), StartRoundSyncState.class);
+        assertThat(lightPeersInformation.getConnectedPeersSize(), is(1));
     }
```
julianlen, line 299: What's happen if a peer connects to the lightpeer, then it has a bad behavior, the process aborts, and then the peer tries to connect again ?

rskj-core/src/main/java/co/rsk/net/light/LightSyncProcessor.java
smcsicardi: Methods sorted better

rskj-core/src/test/java/co/rsk/net/light/LightSyncStateTest.java
smcsicardi: This should be taken into account in the Timeouts system, mainly because the peer should be banned for a period of time at least after some tries

PR #1277, opened 2020-07-20
Check if an account exists before add it.

jonathansmirnoff:
Added logic to check if an account exists before add it to the Wallet.

Now it is possible to added several time the same account using the method `personal_newAccountWithSeed`.
This PR fix this issue adding a logic that checks before add.

Please @ajlopez and @fedejinich take a look of this.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/Wallet.java
+++ b/rskj-core/src/main/java/co/rsk/core/Wallet.java
@@ -178,7 +178,9 @@ public class Wallet {
         Account account = new Account(ECKey.fromPrivate(privateKeyBytes));
         synchronized (accessLock) {
             RskAddress addr = addAccount(account);
-            this.initialAccounts.add(addr);
+            if (!this.initialAccounts.contains(addr)) {
+                this.initialAccounts.add(addr);
+            }
             return addr.getBytes();
         }
     }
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/core/Wallet.java
```diff
@@ -193,7 +195,9 @@ public RskAddress addAccountWithPrivateKey(byte[] privateKeyBytes, String passph
 
     private void saveAccount(Account account) {
         synchronized (accessLock) {
-            accounts.put(account.getAddress(), account.getEcKey().getPrivKeyBytes());
+            if (!accounts.containsKey(account.getAddress())) {
```
fedejinich, line 198: this seems unnecessary, why don't just put the value on the same address? wouldn't it always be the same deriving same priv?

fedejinich: @ajlopez I just tried removing this `if` and seems fine

jonathansmirnoff: It is possible to avoid this validation and just have the other. I don not know so far the code so I prefer to wait until @ajlopez comments about this.

jonathansmirnoff:
@fedejinich good catch. I review this PR with @ajlopez and you were right. I updated the PR removing the unnecessary if.

fedejinich:
everything works as expected @jonathansmirnoff, nice work!

fedejinich:
agreed to wait for QA and security review

dieguito9000:
> Wallet is insecure and must be deprecated.

Let's make the appropriate disclaimers, in the future we can deactivate it by default.

patogallaiovlabs:
pipeline:run

PR #1278, opened 2020-07-22
Logging improvements

Vovchyk:
Changes that are done if scope of this pull request:

1. Removed truncation of ids, hashes etc. while logging them.
2. Encapsulated `Hex.toHexString()` method inside the `ByteUtil` class.
3. Improved `Transaction.toString()` method. Now inside the `Transaction.toString()` method the `gasPrice`, `receiveAddress` and `value` fields are printed directly without calling `toString()` on them (which could cause `NullPointerException` when, for instance, `gasPrice` or the other fields are `null`).
4. Added logging of tx hash in Program class

Note: there are a lot of files with changes, but the changes themselves are pretty straightforward. Mostly replacements in logging methods.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/rpc/modules/personal/PersonalModuleWalletEnabled.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/modules/personal/PersonalModuleWalletEnabled.java
@@ -29,6 +29,7 @@ import org.ethereum.core.TransactionPool;
 import org.ethereum.facade.Ethereum;
 import org.ethereum.rpc.TypeConverter;
 import org.ethereum.rpc.Web3;
+import org.ethereum.util.ByteUtil;
 import org.ethereum.vm.GasCost;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -172,7 +173,7 @@ public class PersonalModuleWalletEnabled implements PersonalModule {
                 throw new Exception("Address private key is locked or could not be found in this node");
             }
 
-            return s = TypeConverter.toJsonHex(Hex.toHexString(account.getEcKey().getPrivKeyBytes()));
+            return s = TypeConverter.toJsonHex(ByteUtil.toHexString(account.getEcKey().getPrivKeyBytes()));
         } finally {
             LOGGER.debug("personal_dumpRawKey(*****): {}", s);
         }
@@ -187,7 +188,7 @@ public class PersonalModuleWalletEnabled implements PersonalModule {
             throw new Exception("From address private key could not be found in this node");
         }
 
-        String toAddress = args.to != null ? Hex.toHexString(TypeConverter.stringHexToByteArray(args.to)) : null;
+        String toAddress = args.to != null ? ByteUtil.toHexString(TypeConverter.stringHexToByteArray(args.to)) : null;
 
         BigInteger accountNonce = args.nonce != null ? TypeConverter.stringNumberAsBigInt(args.nonce) : transactionPool.getPendingState().getNonce(account.getAddress());
         BigInteger value = args.value != null ? TypeConverter.stringNumberAsBigInt(args.value) : BigInteger.ZERO;
--- a/rskj-core/src/main/java/org/ethereum/core/TransactionReceipt.java
+++ b/rskj-core/src/main/java/org/ethereum/core/TransactionReceipt.java
@@ -19,21 +19,17 @@
 
 package org.ethereum.core;
 
-import org.bouncycastle.util.BigIntegers;
-import org.bouncycastle.util.encoders.Hex;
-import org.ethereum.util.RLP;
-import org.ethereum.util.RLPElement;
-import org.ethereum.util.RLPItem;
-import org.ethereum.util.RLPList;
-import org.ethereum.vm.LogInfo;
+    import org.bouncycastle.util.BigIntegers;
+    import org.ethereum.util.*;
+    import org.ethereum.vm.LogInfo;
 
-import java.math.BigInteger;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
+    import java.math.BigInteger;
+    import java.util.ArrayList;
+    import java.util.Arrays;
+    import java.util.List;
 
-import static co.rsk.util.ListArrayUtil.nullToEmpty;
-import static org.ethereum.util.ByteUtil.EMPTY_BYTE_ARRAY;
+    import static co.rsk.util.ListArrayUtil.nullToEmpty;
+    import static org.ethereum.util.ByteUtil.EMPTY_BYTE_ARRAY;
 
 /**
  * The transaction receipt is a tuple of three items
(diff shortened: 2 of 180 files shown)


donequis:
https://github.com/rsksmart/rskj/pull/1278/files#diff-647a4aa5e0133fbd56feac559152d4e2R115
https://github.com/rsksmart/rskj/pull/1278/files#diff-00ca2825626bea19c3501e29d04ffe19R87
https://github.com/rsksmart/rskj/pull/1278/files#diff-00ca2825626bea19c3501e29d04ffe19R187

Shouldn't these kind of examples be calling ByteUtil.toHexString()? I'm trying to understand the change

Vovchyk:
> https://github.com/rsksmart/rskj/pull/1278/files#diff-647a4aa5e0133fbd56feac559152d4e2R115
> https://github.com/rsksmart/rskj/pull/1278/files#diff-00ca2825626bea19c3501e29d04ffe19R87
> https://github.com/rsksmart/rskj/pull/1278/files#diff-00ca2825626bea19c3501e29d04ffe19R187
> 
> Shouldn't these kind of examples be calling ByteUtil.toHexString()? I'm trying to understand the change

Historically in the code there was a mix of calls to the following methods: `ByteUtil.toHexString()` (wrapper around `Hex.toHexString()` which could properly handle `null` input) and `Hex.toHexString()` (method from the third-party lib which does not allow `null` as an input value).

Now `ByteUtil.toHexString()` is a version which does not allow `null`'s and the `ByteUtil.toHexStringOrEmpty()` method is the one which simply returns empty string for the `null` input. In other words, third-party `Hex.toHexString()` method was incapsulated in the `ByteUtil.toHexString()`.

Vovchyk:
I guess we cannot use `ByteUtil.toHexString()` everywhere, as in some places `null` may come and throw an exception

donequis:
In some classes, like Program.java and VM.java, preventing a NullPointerException may create a hardfork. Still throwing the exception is the right behavior in some places

Vovchyk:
> In some classes, like Program.java and VM.java, preventing a NullPointerException may create a hardfork. Still throwing the exception is the right behavior in some places

Yes, the idea was to keep the logic the same and not use unified version in all the places.
- `ByteUtil.toHexString()` is older `Hex.toHexString()` that does not support nulls
- `ByteUtil.toHexStringOrEmpty()` is older `ByteUtil.toHexString()` that supports nulls

donequis:
Oh, I see.  My bad

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

rskj-core/src/test/java/org/ethereum/core/LogInfoTest.java
```diff
@@ -19,16 +19,14 @@
 
 package org.ethereum.core;
 
+import org.bouncycastle.util.encoders.Hex;
+import org.ethereum.util.ByteUtil;
 import org.ethereum.vm.LogInfo;
-
+import org.junit.Ignore;
```
patogallaiovlabs, line 25: Are ignored tests?

rskj-core/src/main/java/co/rsk/rpc/modules/personal/PersonalModuleWalletEnabled.java
```diff
@@ -172,7 +173,7 @@ public String dumpRawKey(String address) throws Exception {
                 throw new Exception("Address private key is locked or could not be found in this node");
             }
 
-            return s = TypeConverter.toJsonHex(Hex.toHexString(account.getEcKey().getPrivKeyBytes()));
+            return s = TypeConverter.toJsonHex(ByteUtil.toHexString(account.getEcKey().getPrivKeyBytes()));
```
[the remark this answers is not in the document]
Vovchyk: `s` is being used below in the `final` block for logging. This style of coding with the `s` and `final` block is used in multiple places in the code, not only here. Not sure we should address this sort of refactoring in scope of this pull request (as this pr is already pretty big).

rskj-core/src/test/java/org/ethereum/core/LogInfoTest.java
Vovchyk: yes, correct. One test is ignored in this class.

Vovchyk:
pipeline: run

rskj-core/src/test/java/org/ethereum/core/TransactionTest.java
```diff
@@ -712,6 +712,32 @@ function doIt(address invokedAddress) {
...
+        BigInteger value = new BigInteger("1000000000000000000000");
+        byte[] senderPrivKey = HashUtil.keccak256("cow".getBytes());
+        byte[] privKey = HashUtil.keccak256("cat".getBytes());
+        ECKey ecKey = ECKey.fromPrivate(privKey);
+        byte[] gasPrice = Hex.decode("09184e72a000");
+        byte[] gas = Hex.decode("4255");
+
+        Transaction tx = new Transaction(new byte[0], gasPrice, gas, ecKey.getAddress(), value.toByteArray(),null);
+        tx.sign(senderPrivKey);
+        tx.verify();
```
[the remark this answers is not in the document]
Vovchyk: done

Vovchyk:
> Did you test the case when there could be a NPE in toString of TX?
> I couldnt find it.

Added a test for that

rskj-core/src/main/java/org/ethereum/core/TransactionReceipt.java
```diff
@@ -19,21 +19,17 @@
...
+    import org.ethereum.util.*;
+    import org.ethereum.vm.LogInfo;
+
+    import java.math.BigInteger;
+    import java.util.ArrayList;
+    import java.util.Arrays;
+    import java.util.List;
+
+    import static co.rsk.util.ListArrayUtil.nullToEmpty;
+    import static org.ethereum.util.ByteUtil.EMPTY_BYTE_ARRAY;
```
nicops, line 32: fix indentation?

Vovchyk: Not sure I've got the question. This was done automatically by an IDE (`Optimize Imports` command)

rskj-core/src/main/java/org/ethereum/vm/trace/OpActions.java
```diff
@@ -135,7 +135,7 @@ public Action addMemoryExtend(long delta) {
     public Action addMemoryWrite(int address, byte[] data, int size) {
         return addAction(memory, Action.Name.WRITE)
                 .addParam("address", address)
-                .addParam("data", toHexString(data).substring(0, size));
+                .addParam("data", toHexStringOrEmpty(data).substring(0, size));
```
nicops, line 138: My main concern with this PR is the replacement of toHexString with toHexStringOrEmpty, like for example in this line... In many cases it is not replaced, but in some cases it is replaced, even though it clearly cannot be null. Why is that?

PR #1285, opened 2020-08-04
ForkDetectionData retrieval performance

M-Picco:
We found out that the conversion of native byte arrays (`byte[]`) to object lists (`List<Byte>`) was adding a bit of overhead when performing merge mining tag lookups in the block's btc coinbase data in `ProofOfWorkRule` as well as in the fork detection data retrieval in `BlockHeader`. So we decided to temporarily use an implementation of the lookup algorithm that uses native types instead of objects. Altough the change is small, we felt it was a good first step before jumping to a faster or complex implementation.

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
+++ b/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
@@ -18,13 +18,14 @@
  */
 package org.ethereum.core;
 
+import co.rsk.config.RskMiningConstants;
 import co.rsk.core.BlockDifficulty;
 import co.rsk.core.Coin;
 import co.rsk.core.RskAddress;
 import co.rsk.crypto.Keccak256;
+import co.rsk.util.ListArrayUtil;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
-import org.apache.commons.lang3.ArrayUtils;
 import org.ethereum.crypto.HashUtil;
 import org.ethereum.util.RLP;
 import org.ethereum.util.Utils;
@@ -32,7 +33,6 @@ import org.ethereum.util.Utils;
 import javax.annotation.Nullable;
 import java.math.BigInteger;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 
 import static java.lang.System.arraycopy;
@@ -490,13 +490,7 @@ public class BlockHeader {
     }
 
     public byte[] getHashForMergedMining() {
-        byte[] encodedBlock = getEncoded(false, false);
-        byte[] hashForMergedMining = HashUtil.keccak256(encodedBlock);
-
-        if (isUMMBlock()) {
-            byte[] leftHash = Arrays.copyOf(hashForMergedMining, UMM_LEAVES_LENGTH);
-            hashForMergedMining = this.getHashRootForMergedMining(leftHash);
-        }
+        byte[] hashForMergedMining = this.getBaseHashForMergedMining();
 
         if (includeForkDetectionData) {
             byte[] mergedMiningForkDetectionData = hasMiningFields() ?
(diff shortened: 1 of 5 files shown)


rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -570,6 +560,24 @@ public String getParentShortHash() {
...
+     * This base hash is later modified to include the forkdetectiondata in its last 12 bytes
+     *
+     * @return The computed hash for merged mining
+     */
+    private byte[] getBaseHashForMergedMining() {
+        byte[] encodedBlock = getEncoded(false, false);
+        byte[] hashForMergedMining = HashUtil.keccak256(encodedBlock);
+
+        if (isUMMBlock()) {
+            byte[] leftHash = Arrays.copyOfRange(hashForMergedMining, 0, 20);
```
ajlopezrsk, line 574: Could we use UMM_LEAVES_LENGTH instead of magic number 20?

M-Picco: Done

M-Picco:
Fixed some issues pointed out by Sonar on `ListArrayUtil#lastIndexOfSubList`

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -539,28 +533,24 @@ public String getParentShortHash() {
...
                     throw new IllegalStateException(
                             String.format("Mining fork detection data could not be found. Header: %s", getShortHash())
                     );
                 }
 
-                int from = position + HASH_FOR_MERGED_MINING_PREFIX_LENGTH;
+                int from = position + RskMiningConstants.RSK_TAG.length + HASH_FOR_MERGED_MINING_PREFIX_LENGTH;
                 int to = from + FORK_DETECTION_DATA_LENGTH;
+
+                // should the fork detection data in the coinbase be under 12 bytes, copyOfRange pads with zeros
```
amendelzon, line 553: I think we should definitely validate that the fork detection data in the coinbase transaction is always 12 bytes and avoid these potentially consensus threatening blocks.

M-Picco: Done

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -539,28 +533,29 @@ public String getParentPrintableHash() {
...
                     );
                 }
 
-                int from = position + HASH_FOR_MERGED_MINING_PREFIX_LENGTH;
+                int from = position + RskMiningConstants.RSK_TAG.length + HASH_FOR_MERGED_MINING_PREFIX_LENGTH;
                 int to = from + FORK_DETECTION_DATA_LENGTH;
+
+                if (coinbaseTransaction.length < to) {
+                    throw new IllegalStateException(
+                            String.format("Invalid fork detection data length. Expected: %d. Got: %d. Header: %s", FORK_DETECTION_DATA_LENGTH, FORK_DETECTION_DATA_LENGTH - (to - coinbaseTransaction.length), getPrintableHash())
```
amendelzon, line 555: Shouldn't `FORK_DETECTION_DATA_LENGTH - (to - coinbaseTransaction.length)` (the `Got: %d` bit) be `coinbaseTransaction.length - from - 1`?

amendelzon (changes requested):
Looks good. Very minor comment.

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
M-Picco: Results are the same. However, the other way (`length - from`) is way clearer, so I'll change it to that.

amendelzon: Thanks

PR #1288, opened 2020-08-05
Add thread names & mdc to log

nicops:
Added mdc & thread to log output.
Added mdc keys to BlockChainImpl (blockHash & blockHeight)
Added names to various threads
Added logging around BlockChainImpl calls to transactionPool

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
@@ -144,6 +144,9 @@ public class BlockChainImpl implements Blockchain {
             }
 
             try {
+                org.slf4j.MDC.put("blockHash", block.getHash().toHexString());
+                org.slf4j.MDC.put("blockHeight", Long.toString(block.getNumber()));
+
                 logger.trace("Try connect block hash: {}, number: {}",
                              block.getShortHash(),
                              block.getNumber());
@@ -160,10 +163,16 @@ public class BlockChainImpl implements Blockchain {
                 logger.error("Unexpected error: ", t);
                 return ImportResult.INVALID_BLOCK;
             }
+            finally {
+                org.slf4j.MDC.remove("blockHash");
+                org.slf4j.MDC.remove("blockHeight");
+
+            }
         }
         finally {
             this.lock.readLock().unlock();
         }
+
     }
 
     private ImportResult internalTryToConnect(Block block) {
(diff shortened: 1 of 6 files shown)


rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
```diff
@@ -457,8 +457,11 @@ private void saveReceipts(Block block, BlockResult result) {
     }
 
     private void processBest(final Block block) {
+        logger.debug("Starting to run transactionPool.processBest(block)");
```
ajlopezrsk, line 460: Could we add the block number and short hash to the log message? It could help when grep over the log lines. I'm not sure about the current format (there is a pull request that standardize the format)

nicops: Added it directly as MDC keys to the whole block processing

rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
```diff
@@ -144,6 +144,9 @@ public ImportResult tryToConnect(Block block) {
             }
 
             try {
+                org.slf4j.MDC.put("blockHash", block.getHash().toHexString());
+                org.slf4j.MDC.put("blockHeight", Long.toString(block.getNumber()));
```
patogallaiovlabs, line 148: Wouldn't be better if we wrap-up org.slf4j.MDC with an own implementation? Kind of helper or sthg?

rskj-core/src/main/java/co/rsk/mine/MinerClientImpl.java
```diff
@@ -77,7 +77,7 @@ public RefreshWork createRefreshWork() {
     }
 
     public Thread createDoWorkThread() {
-        return new Thread() {
+        return new Thread("miner client") {
```
ajlopezrsk, line 80: AFAIR, this is an auxiliary class, which its only real use is in regtest. Maybe we could add also the name of the thread to MinerServer, it has a thread

nicops: Yes, but still, it's nice to have the logs in regtest. I didn't add the name to the MinerSever thread... because the name is already set: https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/mine/MinerServerImpl.java#L194

patogallaiovlabs:
pipeline:run

patogallaiovlabs:
pipeline:run

PR #1292, opened 2020-08-12
Fix index DB shutdown NPE

patogallaiovlabs:
- Issue

When we a halt of the JVM (on purpose or not), we have a NPE trying to commit the changes on the indexDB. So last blocks were lost. 
The problem was that the DB was initialised with a property of closeOnShutdown, so when we try to flush it's already closed (shouldn't throw a NPE, but thats because we use an older version of mapdb).

- Solution

What i did was just remove the option of automatically closing the db when shutdown, and do it "manually" after flushing.

- Test

Dont know how to properly test it, i just run a node -> stop it by halting the jvm -> start it again (checking that the last block number its the same as when starts).

- Exception thrown

```
Exception in thread "Thread-3" java.lang.NullPointerException
	at org.mapdb.StoreDirect.indexValGet(StoreDirect.java:1575)
	at org.mapdb.StoreWAL.indexValGet(StoreWAL.java:307)
	at org.mapdb.StoreDirect.update2(StoreDirect.java:378)
	at org.mapdb.StoreCached.flushWriteCacheSegment(StoreCached.java:454)
	at org.mapdb.StoreCached.flushWriteCache(StoreCached.java:429)
	at org.mapdb.StoreWAL.commit(StoreWAL.java:613)
	at org.mapdb.Engine$CloseOnJVMShutdown.commit(Engine.java:489)
	at org.mapdb.DB.commit(DB.java:2638)
	at co.rsk.db.MapDBBlocksIndex.flush(MapDBBlocksIndex.java:136)
	at org.ethereum.db.IndexedBlockStore.flush(IndexedBlockStore.java:182)
	at co.rsk.core.bc.BlockChainFlusher.flushAll(BlockChainFlusher.java:87)
	at co.rsk.core.bc.BlockChainFlusher.stop(BlockChainFlusher.java:69)
	at co.rsk.FullNodeRunner.stop(FullNodeRunner.java:76)
	at java.lang.Thread.run(Thread.java:748)
```

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockChainFlusher.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockChainFlusher.java
@@ -72,6 +72,13 @@ public class BlockChainFlusher implements InternalService {
     public void stop() {
         emitter.removeListener(listener);
         flushAll();
+        closeAll();
+    }
+
+    private void closeAll() {
+        logger.trace("closing blockStore.");
+        blockStore.close();
+        logger.trace("blockStore closed.");
     }
 
     private void flush() {
--- a/rskj-core/src/main/java/co/rsk/RskContext.java
+++ b/rskj-core/src/main/java/co/rsk/RskContext.java
@@ -1674,7 +1674,6 @@ public class RskContext implements NodeBootstrapper {
         }
 
         DB indexDB = DBMaker.fileDB(dbFile)
-                .closeOnJvmShutdown()
                 .make();
 
         KeyValueDataSource blocksDB = LevelDbDataSource.makeDataSource(Paths.get(databaseDir, "blocks"));
--- a/rskj-core/src/main/java/co/rsk/db/BlocksIndex.java
+++ b/rskj-core/src/main/java/co/rsk/db/BlocksIndex.java
@@ -84,4 +84,6 @@ public interface BlocksIndex {
      * Commits the changes to the underlying permanent storage.
      */
     void flush();
+
+    void close();
 }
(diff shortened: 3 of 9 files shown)


patogallaiovlabs:
pipeline:run

rskj-core/src/main/java/co/rsk/core/bc/BlockChainFlusher.java
```diff
@@ -67,6 +67,13 @@ public void start() {
     public void stop() {
         emitter.removeListener(listener);
         flushAll();
+        closeAll();
+    }
+
+    private void closeAll() {
+        logger.trace("closing blockStore.");
+        blockStore.close();
+        logger.trace("blockStore closed.");
```
ajlopezrsk, line 76: Should we add  `trieStore.dispose()`?

patogallaiovlabs: Well... DataSourceWithCache does that job. I wouldn’t analyse that in the scope of this PR.

Vovchyk: Seems that `DataSource.close()` is not called on shutdown, and, as a result, `LevelDB` instances are not getting disposed gracefully.

In the examples [here](https://github.com/fusesource/leveldbjni) they do close db.

I guess we should investigate this more deeply in scope of this or another task.

patogallaiovlabs: Yep, totally agreed. I will add another task to make a deep analysis.

ajlopezrsk (dismissed):

ajlopezrsk (approved):

PR #1303, opened 2020-09-07
QG in red fixes

patogallaiovlabs:
Fixing sonar issues:
https://sonarcloud.io/dashboard?id=rskj

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockChainFlusher.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockChainFlusher.java
@@ -23,6 +23,7 @@ import co.rsk.metrics.profilers.Profiler;
 import co.rsk.metrics.profilers.ProfilerFactory;
 import co.rsk.trie.TrieStore;
 import co.rsk.util.FormatUtils;
+import com.google.common.annotations.VisibleForTesting;
 import org.ethereum.core.Block;
 import org.ethereum.core.TransactionReceipt;
 import org.ethereum.db.BlockStore;
@@ -77,9 +78,9 @@ public class BlockChainFlusher implements InternalService {
     }
 
     private void closeAll() {
-        logger.trace("closing blockStore.");
+        getLogger().trace("closing blockStore.");
         blockStore.close();
-        logger.trace("blockStore closed.");
+        getLogger().trace("blockStore closed.");
     }
 
     private void flush() {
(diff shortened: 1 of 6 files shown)


rskj-core/src/main/java/org/ethereum/util/ByteUtil.java
```diff
@@ -366,14 +366,14 @@ public static String oneByteToHexString(byte value) {
     public static int numBytes(String val) {
 
         BigInteger bInt = new BigInteger(val);
-        int bytes = 0;
+        int bytes = 1;
```
Vovchyk, line 369: how about this way?
```
    public static int numBytes(String val) {
        return new BigInteger(val).bitLength() / 8 + 1;
    }
```

rskj-core/src/main/java/co/rsk/core/bc/BlockChainFlusher.java
```diff
@@ -77,9 +78,9 @@ public void stop() {
     }
 
     private void closeAll() {
-        logger.trace("closing blockStore.");
+        getLogger().trace("closing blockStore.");
```
nicops, line 81: I don't think this is the right way to tackle this. Instead, I would change the logging level of the test logback.xml

rskj-core/src/main/java/org/ethereum/util/ByteUtil.java
```diff
@@ -366,14 +366,14 @@ public static String oneByteToHexString(byte value) {
...
 
         while (!bInt.equals(BigInteger.ZERO)) {
             bInt = bInt.shiftRight(8);
             ++bytes;
         }
-        if (bytes == 0) {
```
nicops, line 375: SonarQube says this expression is always true, but that seems off. Surely it is not guaranteed it is always true, given the while above? bad static analysis?

rskj-core/src/main/java/org/ethereum/util/ByteUtil.java
patogallaiovlabs: worked like a charm!

rskj-core/src/main/java/co/rsk/core/bc/BlockChainFlusher.java
patogallaiovlabs: changed as suggested, thanks!

patogallaiovlabs:
pipeline:run

patogallaiovlabs:
pipeline:run

Vovchyk (approved):

patogallaiovlabs:
pipeline:run

ajlopezrsk (approved):

PR #1311, opened 2020-09-11
Fixed issue when contract deployment burns previous balance

Vovchyk:
Contract addresses are deterministic. So it's possible to calculate an address of a contract before actual deployment (for some particular Externally Owned Account and nonce).

This pull request addresses the issue when some funds are transferred in advance to the contract address which is not deployed yet. And then after deployment the balance becomes zero.

Note: The issue is only relevant to the case when a contract is created directly, not via `CREATE` or `CREATE2` opcodes.


Vovchyk:
pipeline: run

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -369,7 +369,15 @@ private void call() {
...
-        cacheTrack.createAccount(newContractAddress); // pre-created
+
+        // Keep prior balance of a contract address (if any)
+        if (cacheTrack.isExist(newContractAddress)) {
+            Coin oldBalance = cacheTrack.getBalance(newContractAddress);
+            cacheTrack.createAccount(newContractAddress);
+            cacheTrack.addBalance(newContractAddress, oldBalance);
+        } else {
+            cacheTrack.createAccount(newContractAddress);
+        }
```
ajlopezrsk, line 380: Why not simply NOT create account?

Vovchyk: I just implemented it the same way as it's done for `create` and `create2` opcodes - https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/org/ethereum/vm/program/Program.java#L545

`Repository.createAccount` fully recreates an account from scratch. Do you thing we should simplify this logic and do nothing for existing accounts?

patogallaiovlabs (dismissed):
LGTM

patogallaiovlabs (approved):

patogallaiovlabs (approved):

patogallaiovlabs (approved):
LGTM

PR #1317, opened 2020-09-25
Added logic of new RskLab address activation

Vovchyk:
Added logic for HF activation of a new RskLab address

Note: exact RSKIP number and actual account address will be provided later.


patogallaiovlabs (changes requested):
You could add some simple unit test for making sure you get one address or the other before and after activation block.
The rest LGTM.

Vovchyk:
already in master. Done in scope of another PR

PR #1318, opened 2020-09-28
Testing remaining bytes after merged mining hash.

ignaciopulicedonatto:
## Description
New test validations for remaining bytes after merged mining hash.

## Motivation and Context
More than 2 outputs placed after merged mining hash (36 bytes each) make a solution invalid. 
A test with 2 outputs is needed in order to be alert that the condition for remaining bytes after merged mining tag should trigger some test alarm, otherwise, if is changed MAX_BYTES_AFTER_MERGED_MINING_HASH  for a value smaller than 128 could make miner solutions invalid.
Nowadays, existing miners have 1 or 2 outputs.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/mine/MinerUtils.java
+++ b/rskj-core/src/main/java/co/rsk/mine/MinerUtils.java
@@ -55,15 +55,22 @@ public class MinerUtils {
     }
 
     public static co.rsk.bitcoinj.core.BtcTransaction getBitcoinMergedMiningCoinbaseTransaction(co.rsk.bitcoinj.core.NetworkParameters params, byte[] blockHashForMergedMining) {
-        co.rsk.bitcoinj.core.BtcTransaction coinbaseTransaction = new co.rsk.bitcoinj.core.BtcTransaction(params);
-        //Add a random number of random bytes before the RSK tag
         SecureRandom random = new SecureRandom();
         byte[] prefix = new byte[random.nextInt(1000)];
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/co/rsk/mine/MinerUtils.java
```diff
@@ -55,15 +55,22 @@
...
+
+        return getBitcoinCoinbaseTransaction(params, bytes);
+    }
+
+    public static co.rsk.bitcoinj.core.BtcTransaction getBitcoinCoinbaseTransaction(co.rsk.bitcoinj.core.NetworkParameters params, byte[] additionalData) {
+
+        co.rsk.bitcoinj.core.BtcTransaction coinbaseTransaction = new co.rsk.bitcoinj.core.BtcTransaction(params);
+
+        // Add the data to the scriptSig of first input
+        co.rsk.bitcoinj.core.TransactionInput ti = new co.rsk.bitcoinj.core.TransactionInput(params, coinbaseTransaction, new byte[0]);
```
donequis, line 71: I'm missing why it doesn't use the  additionalData param

ignaciopulicedonatto: What we are passing frond that parameter is output data, not input data.

donequis (approved):

ignaciopulicedonatto:
Opening a new branch for this change

PR #1319, opened 2020-09-28
Create scafold in bridge methods for the new registerBtcTransfer func…

pamgonzalez:
…tion.

Create the function in BridgeMethods, in Bridge and it is BridgeSupport (without logic).
Test the external call to the method.


josedahlquist (commented):
LGTM
But I think it's worthy to add tests to check the data parsing.

guidohernan93 (approved):

josedahlquist (approved):

PR #1320, opened 2020-09-29
Bls12 draft (EIP 2537)  - WIP

nicops:
## Description

Implementation of EIP-2537 (rskip yet to be created).

This is a black-box implementation of the EIP-2537 based on besu's implementation https://github.com/hyperledger/besu/commit/29e668a2a0877c4c7180b48c571e40cd97ba2692 which in turn uses the native reference implementation https://github.com/matter-labs/eip1962

## Motivation and Context
Set up compatibility with soon to be released ethereum hardfork.

## How Has This Been Tested?
I tested this by unit testing the contract classes directly against the same csv with all the examples as besu, and also did two dsl tests for success and failure. I also ran tx against regtest using the same solidity contract as the dsl tests.

Currently there is a breaking test, see more on that below.

## Types of changes
  (this might include a breaking change, see below)

## Checklist:

* **Other information**:

This PR is incomplete:
* In order to run it, you must compile manually besu library and push it into the maven local repository. This works as a proof of concept, but a rsk native version is actually required, among other things because besu doesn't support windows.
* when starting rsk context, we should probably check somehow that if the native library couldn't be loaded AND the rskip has activation != -1, the node should *not* start. 
* Following the EIP it seems that a CALL to these precompiles with error should put a zero in the stack. Nevertheless the current code seems to work differently: exceptions in precompiles are not captured but instead they bubble through the whole execution. This seems different than what happens with CALLs to other precompiles. So I changed the precompile execution code to capture exceptions and return them (I'm very unsure that the code is correct). This change is needed in order to make the dsl test for the error case to pass. This might be a breaking change, and might require a HF, in fact, it breaks a specific bridge test.


nicops:
Another thing I didn't do but could be nice to have would be to check if there is any kind of truffle testing with some sollidity contracts on this. the solidity contract used in the dsl was written by me based on some examples on the internet.

PR #1327, opened 2020-10-05
Build Block Blooms Database

ajlopezrsk:
## Description
Using a new configuration option, the build and save of block blooms data could be executed within the normal operation of the node. This data is used to retrieve log information efficiently.

## Motivation and Context

Usually, this block blooms information is used only in nodes that has JSON RPC open, and log query operation enabled. And with the first query, the data is build and saved. But even with this feature, the operation could take many minutes and resources from the node.

So, this new code allow the build of the same data along the blockchain node normal processing, ie, during long synchronization or normal synchronization. To do this, a new internal service was added, and it is activated by configuration (the default value is not activated),

## How Has This Been Tested?

It has code tests, and also was tested running a long sync against testnet network,

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomProcessor.java
+++ b/rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomProcessor.java
@@ -25,13 +25,13 @@ import org.ethereum.db.BlockStore;
 /**
  * Created by ajlopez on 29/09/2020.
  */
-public class BlocksBloomBuilder {
+public class BlocksBloomProcessor {
     private final BlocksBloomStore blocksBloomStore;
     private final BlockStore blockStore;
 
     private BlocksBloom blocksBloomInProcess = null;
 
-    public BlocksBloomBuilder(BlocksBloomStore blocksBloomStore, BlockStore blockStore) {
+    public BlocksBloomProcessor(BlocksBloomStore blocksBloomStore, BlockStore blockStore) {
         this.blocksBloomStore = blocksBloomStore;
         this.blockStore = blockStore;
     }
@@ -55,7 +55,8 @@ public class BlocksBloomBuilder {
 
         if (this.blocksBloomInProcess == null) {
             this.blocksBloomInProcess = new BlocksBloom();
-            fromBlock = this.blocksBloomStore.firstNumberInRange(blockNumber);
+            fromBlock = this.blocksBloomStore.
+                    firstNumberInRange(blockNumber);
         }
         else {
             fromBlock = this.blocksBloomInProcess.toBlock() + 1;
@@ -87,7 +88,7 @@ public class BlocksBloomBuilder {
         this.blocksBloomInProcess.addBlockBloom(blockNumber, bloom);
 
         if (blockNumber == this.blocksBloomStore.lastNumberInRange(blockNumber)) {
-            this.blocksBloomStore.setBlocksBloom(this.blocksBloomInProcess);
+            this.blocksBloomStore.addBlocksBloomCache(this.blocksBloomInProcess);
             this.blocksBloomInProcess = null;
         }
     }
--- a/rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomProcessor.java
+++ b/rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomProcessor.java
@@ -88,7 +88,7 @@ public class BlocksBloomProcessor {
         this.blocksBloomInProcess.addBlockBloom(blockNumber, bloom);
 
         if (blockNumber == this.blocksBloomStore.lastNumberInRange(blockNumber)) {
-            this.blocksBloomStore.addBlocksBloomCache(this.blocksBloomInProcess);
+            this.blocksBloomStore.addBlocksBloom(this.blocksBloomInProcess);
             this.blocksBloomInProcess = null;
         }
     }
--- a/rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomProcessor.java
+++ b/rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomProcessor.java
@@ -19,6 +19,7 @@
 
 package co.rsk.logfilter;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.ethereum.core.Bloom;
 import org.ethereum.db.BlockStore;
 
@@ -36,11 +37,16 @@ public class BlocksBloomProcessor {
         this.blockStore = blockStore;
     }
 
+    @VisibleForTesting
     public BlocksBloom getBlocksBloomInProcess() {
         return this.blocksBloomInProcess;
     }
 
-    public void processNewBlockNumber(long newBlockNumber) {
+    public synchronized void processNewBlockNumber(long newBlockNumber) {
+        if (newBlockNumber < this.blocksBloomStore.getNoConfirmations()) {
+            return;
+        }
+
         long blockNumber = newBlockNumber - this.blocksBloomStore.getNoConfirmations();
 
         addBlocksUpToNumber(blockNumber);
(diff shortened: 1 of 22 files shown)


rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomService.java
```diff
@@ -0,0 +1,73 @@
...
+        emitter.removeListener(listener);
+    }
+
+    public void processNewBlock(long blockNumber) {
+        this.blocksBloomProcessor.processNewBlockNumber(blockNumber);
+    }
+
+    private class OnBlockListener extends EthereumListenerAdapter {
+        @Override
+        public void onBlock(Block block, List<TransactionReceipt> receipts) {
```
Vovchyk, line 69: shouldn't we listen to `onBestBlock` event instead?

Vovchyk (changes requested):

rskj-core/src/main/resources/reference.conf
```diff
@@ -403,3 +403,10 @@ crypto {
     # - 'native' (Native C lib) <- not yet implemented
     library: 'bc'
 }
+
+blooms = {
+    blocks = 64
+    service = false
```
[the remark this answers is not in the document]
ajlopezrsk: OK. Done. I'm a bit confused due that equal sign is used in expected.conf

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomService.java
ajlopezrsk: We use onBlock only to take advantage of no of confirmations at the earliest time possible, in an independent way of being processing the next bestchain or not

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomStore.java
```diff
@@ -40,22 +44,25 @@ public BlocksBloomStore(int noBlocks, int noConfirmations, KeyValueDataSource da
...
 
-        if (this.dataSource != null && this.dataSource.get(longToKey(blockNumber)) != null) {
-            return true;
-        }
+        return hasBlockNumberInCache(key)
+            || hasBlockNumberInStore(key);
+    }
+
+    private boolean hasBlockNumberInStore(long key) {
+        return this.dataSource != null && this.dataSource.get(longToKey(key)) != null;
```
Vovchyk, line 55: Seems `dataSource` field cannot be `null` (`non-nullable`)? It's being checked for `null` in a few places in this class. But in other places (e.g. in `flush` method) it is not.

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomStore.java
```diff
@@ -29,9 +31,11 @@
  * Created by ajlopez on 05/02/2019.
  */
 public class BlocksBloomStore {
+    private static final Logger logger = LoggerFactory.getLogger("blooms");
+
     private final int noBlocks;
     private final int noConfirmations;
-    private final Map<Long, BlocksBloom> blocksBloom = new ConcurrentHashMap<>();
+    private final Map<Long, BlocksBloom> blocksBloomCache = new ConcurrentHashMap<>();
```
Vovchyk, line 38: I guess here we could use `HashMap` instead of `ConcurrentHashMap`. Access to this filed is synchronized in this class.

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+    public BlocksBloomProcessor(BlocksBloomStore blocksBloomStore, BlockStore blockStore) {
+        this.blocksBloomStore = blocksBloomStore;
+        this.blockStore = blockStore;
+    }
+
+    public BlocksBloom getBlocksBloomInProcess() {
+        return this.blocksBloomInProcess;
+    }
+
+    public void processNewBlockNumber(long newBlockNumber) {
```
Vovchyk, line 43: Looks like this method is not thread-safe. Shouldn't it be `synchronized`?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private BlocksBloom blocksBloomInProcess = null;
+
+    public BlocksBloomProcessor(BlocksBloomStore blocksBloomStore, BlockStore blockStore) {
+        this.blocksBloomStore = blocksBloomStore;
+        this.blockStore = blockStore;
+    }
+
+    public BlocksBloom getBlocksBloomInProcess() {
+        return this.blocksBloomInProcess;
```
Vovchyk, line 40: `blocksBloomInProcess` field is not synchronized. And it looks like it's only needed for unit testing. Shouldn't the method be marked as `synchronized` or `VisibleForTesting`?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    public void processNewBlockNumber(long newBlockNumber) {
+        long blockNumber = newBlockNumber - this.blocksBloomStore.getNoConfirmations();
```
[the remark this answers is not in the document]
ajlopezrsk: nice catch! Done

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomStore.java
ajlopezrsk: Done!

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomProcessor.java
ajlopezrsk: Currently only one thread at a time access this method, but yes, it could be fragile if new changes are added to the block processing logic (like processing blocks in multiple threads). So, done!

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomProcessor.java
ajlopezrsk: Done

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomStore.java
ajlopezrsk: Done, I just added additional checks

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloomStore.java
```diff
@@ -21,17 +21,31 @@
...
+ *
  * Created by ajlopez on 05/02/2019.
  */
 public class BlocksBloomStore {
+    private static final Logger logger = LoggerFactory.getLogger("blooms");
+
     private final int noBlocks;
     private final int noConfirmations;
-    private final Map<Long, BlocksBloom> blocksBloom = new ConcurrentHashMap<>();
+    private final Map<Long, BlocksBloom> blocksBloomCache = new HashMap<>();
```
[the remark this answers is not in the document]
ajlopezrsk: Good. But I though it should be to refine in another tasks. The idea is to not be disruptive with the current behavior the main consumers of the logs have now (like Lumino and token bridge). And all this work (the new blooms) are optional, only to be adopted by power users.

Even with 2 millions of blocks, the used memory should be few megabytes.

Vovchyk (dismissed):

fedejinich (approved):
lgtm

fedejinich (dismissed):
reviewed

patogallaiovlabs (approved):

patogallaiovlabs (approved):

PR #1339, opened 2020-10-20
Fixed length of root field in tx receipt response

Vovchyk:
This sets the length of the `root` field that's being returned by `eth_getTransactionReceipt` JSON-RPC call to 32 bytes.

## Description
Right now rskj returns 0x00,0x01 instead of 0x0...0 and 0x0...1 (32 bytes) for the `root` field. This makes some js libraries incompatible with rskj.

## Motivation and Context
This change is required for the rskj node to be compatible with client libraries at the JSON-RPC level.

## How Has This Been Tested?
JSON-RPC response of `eth_getTransactionReceipt` has been checked.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/test/java/org/ethereum/util/ByteUtilTest.java
+++ b/rskj-core/src/test/java/org/ethereum/util/ByteUtilTest.java
@@ -500,13 +500,18 @@ public class ByteUtilTest {
 
     @Test
     public void testToBytesWithLeadingZeros_WithLeadingZeros() {
-        byte[] src = new byte[]{1};
+        byte[] src = new byte[]{1, 2};
 
         byte[] actualResult = ByteUtil.toBytesWithLeadingZeros(src, 10);
 
         assertEquals(10, actualResult.length);
-        for (int i = 0; i < actualResult.length - src.length; i++) {
+
+        int srcStart = actualResult.length - src.length;
+        for (int i = 0; i < srcStart; i++) {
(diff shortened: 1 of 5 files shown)


rskj-core/src/test/java/org/ethereum/util/ByteUtilTest.java
```diff
@@ -466,7 +466,47 @@ public void testParseBytes() {
...
+    @Test
+    public void testToBytesWithLeadingZeros_WithLeadingZeros() {
+        byte[] src = new byte[]{1};
+
+        byte[] actualResult = ByteUtil.toBytesWithLeadingZeros(src, 10);
+
+        assertEquals(10, actualResult.length);
+        for (int i = 0; i < actualResult.length - src.length; i++) {
+            assertEquals(0, actualResult[i]);
+        }
```
patogallaiovlabs, line 510: I would add an assert of the last byte, matching the src.

patogallaiovlabs (changes requested):
We could add a test for the final DTO, that we currently dont have

fedejinich:
is it posible to add a test case for `testToBytesWithLeadingZeros_TxReceiptRoot`?

patogallaiovlabs (approved):

patogallaiovlabs (approved):

ajlopezrsk (approved):

PR #1349, opened 2020-10-27
Update registerBtcTransfer signature

guidohernan93:
- Update BridgeMethods and Bridge classes to cast BTC addresses as bytes
- Update tests

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
@@ -580,7 +580,7 @@ public enum BridgeMethods {
     REGISTER_BTC_TRANSFER(
             CallTransaction.Function.fromSignature(
                     "registerBtcTransfer",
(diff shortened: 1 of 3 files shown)


josedahlquist (approved):

rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
```diff
@@ -580,7 +580,7 @@
     REGISTER_BTC_TRANSFER(
             CallTransaction.Function.fromSignature(
                     "registerBtcTransfer",
-                    new String[]{"bytes", "uint256", "bytes", "bytes32", "string", "string", "string", "bool"},
+                    new String[]{"bytes", "uint256", "bytes", "bytes32", "bytes", "string", "bytes", "bool"},
```
marcos-iov, line 583: Not sure if it makes a difference, but you could use `bytes20` as the data type for btc addresses since they are always 20 bytes long

marcos-iov (approved):

PR #1360, opened 2020-11-11
Make BtcBlockStore cache configurable

patogallaiovlabs:
## Description

Make BtcBlockStore cache configurable, with 2 options: depth and size.
Add more logs, to identify possible issues in long sync.

## Motivation and Context

When making a long-sync in testnet, I found some range of blocks with a lot of BTC blocks, which makes it really hard to process. Increasing the cache, would make it faster and would take more advantage of this cache. As the configuration of this cache are now static fields, I made this change so we could configure it depending on the chain we are. 
For the purpose of this PR, we wanted to change only testnet params.

## How Has This Been Tested?

Syncing with Boton instance on testnet.

## Types of changes

## Checklist:

* **Other information**:

Properties:
- cache.btcBlockStore.size: default to 10k
- cache.btcBlockStore.depth: default to 5k (depth is calculated by the height of the block)

### DIFF
--- a/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
+++ b/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
@@ -350,6 +350,14 @@ public class RskSystemProperties extends SystemProperties {
         return configFromFiles.getInt("cache.receipts.max-elements");
     }
 
+    public int getBtcBlockStoreCacheSize() {
+        return configFromFiles.getInt("cache.btcBlockStore.size");
+    }
+
+    public int getBtcBlockStoreCacheDepth() {
+        return configFromFiles.getInt("cache.btcBlockStore.depth");
+    }
+
     public long getVmExecutionStackSize() {
         return configFromFiles.getBytes("vm.executionStackSize");
     }
--- a/rskj-core/src/main/resources/config/testnet.conf
+++ b/rskj-core/src/main/resources/config/testnet.conf
@@ -71,3 +71,10 @@ wallet {
     enabled = false
     accounts = []
 }
+
+cache {
+    btcBlockStore {
+        depth: 50000,
+        size: 100000
+    }
+}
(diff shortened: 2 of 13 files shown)


rskj-core/src/main/resources/config/testnet.conf
```diff
@@ -71,3 +71,10 @@ wallet {
     enabled = false
     accounts = []
 }
+
+cache {
+    btcBlockStore {
+        depth: 50000,
```
donequis, line 77: This is different than the default value, is it ok?

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
```diff
@@ -338,6 +338,14 @@ public int getReceiptsCacheSize() {
         return configFromFiles.getInt("cache.receipts.max-elements");
     }
 
+    public int getBtcBlockStoreCacheSize() {
+        return configFromFiles.getInt("cache.btcBlockStore.size");
+    }
+
+    public int getBtcBlockStoreCacheDepth() {
+        return configFromFiles.getInt("cache.btcBlockStore.depth");
```
donequis, line 346: There are two things that you can check here (or in the caller)

1. That `depth <= size`

2. That `depth >= 4000` or some other constant. You can add a warning log in case that it fails, there's no reason to fail. But a lower value sounds problematic

rskj-core/src/main/resources/config/testnet.conf
patogallaiovlabs: No, I added an extra 0, good catch!

patogallaiovlabs: Yes, this is why I made it configurable, to change it in testnet. I'm testing syncing fully in testnet, and its performing much better. I will try to do more testing, but we can discuss this if you want.

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
patogallaiovlabs: Ok, I will add some warning.

patogallaiovlabs:
pipeline:run

fedejinich (approved):
lgtm

PR #1362, opened 2020-11-16
Implemented timestamp linking validation

Vovchyk:
This implementation checks if BTC and RSK block timestamps are close enough.

## Description
This extends `BlockTimeStampValidationRule` class with the timestamps validation logic and its activation in the next HF.

## Motivation and Context
This aims to prevent a merge-miner attacker from mining old RSK blocks with the current Bitcoin hashrate. 

## How Has This Been Tested?
Using unit tests and running the node locally.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/util/TimeProvider.java
+++ b/rskj-core/src/main/java/co/rsk/util/TimeProvider.java
@@ -1,0 +1,25 @@
+/*
+ * This file is part of RskJ
+ * Copyright (C) 2017 RSK Labs Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package co.rsk.util;
+
+public interface TimeProvider {
+
+    long currentTimeMillis();
+
+}
--- a/rskj-core/src/main/java/co/rsk/util/TimeProvider.java
+++ b/rskj-core/src/main/java/co/rsk/util/TimeProvider.java
@@ -1,6 +1,6 @@
 /*
  * This file is part of RskJ
- * Copyright (C) 2017 RSK Labs Ltd.
+ * Copyright (C) 2020 RSK Labs Ltd.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
--- a/rskj-core/src/main/java/co/rsk/validators/BlockTimeStampValidationRule.java
+++ b/rskj-core/src/main/java/co/rsk/validators/BlockTimeStampValidationRule.java
@@ -18,10 +18,21 @@
 
 package co.rsk.validators;
 
+import co.rsk.bitcoinj.core.BtcBlock;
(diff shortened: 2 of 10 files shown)


Vovchyk:
pipeline: run

donequis:
Did you consider that the activation could happen based on block timestamp instead of block number?

Even there I don't have a clear idea of the benefit of this change. What kind of attack is it preventing? A miner will need the same amount of hashing power and time for a double spend or any similar attack. I'm missing something here.

Vovchyk:
pipeline: run

Vovchyk:
> Did you consider that the activation could happen based on block timestamp instead of block number?
> 
> Even there I don't have a clear idea of the benefit of this change. What kind of attack is it preventing? A miner will need the same amount of hashing power and time for a double spend or any similar attack. I'm missing something here.

@donequis the idea and motivation behind this is described in this RSKIP - https://github.com/rsksmart/RSKIPs/blob/timestamplink/IPs/RSKIP179.md

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ConsensusRule.java
```diff
@@ -54,6 +54,7 @@
     RSKIP171("rskip171"),
     RSKIP172("rskip172"),
     RSKIP174("rskip174"),
+    RSKIP179("rskip179"),
```
patogallaiovlabs, line 57: Could you leave a comment here, telling what #179 is all about, short comment (title?)?
We are trying to identify faster whats the main reason behind the code.

rskj-core/src/main/java/co/rsk/validators/BlockTimeStampValidationRule.java
```diff
@@ -18,22 +18,51 @@
...
+                                        TimeProvider timeProvider, NetworkParameters bitcoinNetworkParameters) {
         this.validPeriodLength = validPeriodLength;
+        this.activationConfig = Objects.requireNonNull(activationConfig);
+        this.timeProvider = Objects.requireNonNull(timeProvider);
+        this.bitcoinNetworkParameters = Objects.requireNonNull(bitcoinNetworkParameters);
+    }
+
+    public BlockTimeStampValidationRule(int validPeriodLength, ActivationConfig activationConfig,
+                                        TimeProvider timeProvider) {
+        this(validPeriodLength, activationConfig, timeProvider, RegTestParams.get());
```
patogallaiovlabs, line 61: Its the same to use RegTestParams than MainNetParams in this logic class? Because I dont see any other use than this default constructor, so no matter which network we are, it will use regtest params.

patogallaiovlabs (changes requested):

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ConsensusRule.java
Vovchyk: you can find some details of this RSKIP [here](https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP179.md)

rskj-core/src/main/java/co/rsk/validators/BlockTimeStampValidationRule.java
Vovchyk: Yes, this is bitcoin network params and only this `RegTestParams` is being used across our code base.

rskj-core/src/main/java/co/rsk/util/TimeProvider.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * This file is part of RskJ
+ * Copyright (C) 2017 RSK Labs Ltd.
```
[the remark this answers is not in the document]
Vovchyk: done

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ConsensusRule.java
Vovchyk: done

patogallaiovlabs (dismissed):
LGTM

patogallaiovlabs (approved):

PR #1367, opened 2020-11-18
Removing mocked transaction executor from Web3ImplTest

ajlopez:
## Description

A mocked `ReversibleTransactionExecutor` was removed from `Web3ImplTest`, being replaced by a concrete instance.

## Motivation and Context

The motivation is to improve those tests. Before this change, many test could run in green, even without running all the preconditions.

Now, writing `eth_call` with real contracts is feasible. Before this change, you could call `eth_call` with wrong data, and the result was OK.

## Types of changes

## Checklist:

* **Other information**:
Some tests were rewritten, to use a real `Greeter` contract. Previous tests used the runtime bytes to deploy the contract, instead of the full bytecode.

Gas provided was also wrong, it was not enough for some tests.

I recommend to review `TestContract.greeter` because its use in other tests could also be polluted by mocks, that give a false sense of security


Vovchyk (approved):

ajlopezrsk:
run:pipeline

patogallaiovlabs:
pipeline:run

PR #1378, opened 2020-11-26
Pegin to any address

marcos-iov:
*fed:pegin-to-any-address-integration*
*fit:peg-in-to-any-address*


marcos-iov:
pipeline:run

josedahlquist (approved):

rskj-core/src/main/java/co/rsk/peg/PeginInformation.java
```diff
@@ -0,0 +1,131 @@
...
+            default:
+                String message = String.format("Invalid protocol version: %d", protocolVersion);
+                logger.warn("[parseFromPeginInstructions] {}", message);
+                throw new PeginInstructionsException(message);
+        }
+    }
+
+    private void parseFromPeginInstructionsVersion1(PeginInstructionsVersion1 peginInstructions) {
+        Optional<Address> btcRefundAddressOptional = peginInstructions.getBtcRefundAddress();
+        if (btcRefundAddressOptional.isPresent()) {
```
donequis, line 126: We might want to prevent addresses like the federation here. Not a big deal, but would just waste resources

rskj-core/src/main/java/co/rsk/peg/PeginInformation.java
```diff
@@ -0,0 +1,131 @@
...
+        if (activations.isActive(ConsensusRule.RSKIP170)) {
+            peginInstructionsOptional = peginInstructionsProvider.buildPeginInstructions(btcTx);
+            if (peginInstructionsOptional.isPresent()) {
+                PeginInstructions peginInstructions = peginInstructionsOptional.get();
+                parseFromPeginInstructions(peginInstructions);
+            }
+        }
+
+        // If BtcLockSender could not be parsed and peg-in instructions were not provided, then this tx can't be processed
+        if(!btcLockSenderOptional.isPresent() && !peginInstructionsOptional.isPresent()) {
```
donequis, line 85: I could be wrong on this, but I feel that we have an unexpected change in consensus here.

If we send a transaction from a wp2pkh without presenting the witness, before the activation of the rskip170, it used to reject the transaction and send the money back because with no sender the transaction wouldn't be lockeable, but now it does not seem to send the money back. Or, in general, non lockeable transaction because there's no sending, before rskip170 used to generate a rejection release while it does not happen anymore.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -328,52 +338,148 @@ protected TxType getTransactionType(BtcTransaction btcTx) {
...
+            btcLockSenderProvider,
+            peginInstructionsProvider,
+            activations
+        );
+        try {
+            peginInformation.parse(btcTx);
+        } catch (PeginInstructionsException e) {
+            if (activations.isActive(ConsensusRule.RSKIP170)) {
+                // If possible to get the sender address, refund
+                refundTxSender(btcTx, rskTx, peginInformation, totalAmount);
```
donequis, line 361: Here: refund is only created when there's a pegininstructionexception only after the rskip170 activation

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

donequis (dismissed):

guidohernan93 (dismissed):

pamgonzalez (approved):

PR #1395, opened 2020-12-22
Added Peg-in rejection events

Vovchyk:
## Description

This adds changes to the Bridge peg-in protocol in order to register evidence on-chain when a peg-in was rejected and why it happened.

## Motivation and Context

These changes will improve the user experience of the peg-in.

## Types of changes

## Checklist:

* **Other information**:

*fit:add-pegin-rejection-events*

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -356,7 +356,9 @@ public class BridgeSupport {
             peginInformation.parse(btcTx);
         } catch (PeginInstructionsException e) {
             if (activations.isActive(ConsensusRule.RSKIP170)) {
-                eventLogger.logRejectedPegin(btcTx, RejectedPeginReason.PEGIN_V1_INVALID_PAYLOAD);
+                if (activations.isActive(ConsensusRule.RSKIP181)) {
+                    eventLogger.logRejectedPegin(btcTx, RejectedPeginReason.PEGIN_V1_INVALID_PAYLOAD);
+                }
 
                 // If possible to get the sender address, refund
                 refundTxSender(btcTx, rskTx, peginInformation, totalAmount);
@@ -406,7 +408,9 @@ public class BridgeSupport {
         if (!BridgeUtils.txIsProcessableInLegacyVersion(senderBtcAddressType, activations)) {
             logger.warn("[processPeginVersionLegacy] [btcTx:{}] Could not get BtcLockSender from Btc tx", btcTx.getHash());
 
-            eventLogger.logRejectedPegin(btcTx, RejectedPeginReason.LEGACY_PEGIN_UNDETERMINED_SENDER);
+            if (activations.isActive(ConsensusRule.RSKIP181)) {
+                eventLogger.logRejectedPegin(btcTx, RejectedPeginReason.LEGACY_PEGIN_UNDETERMINED_SENDER);
+            }
 
             throw new RegisterBtcTransactionException("Could not get BtcLockSender from Btc tx");
         }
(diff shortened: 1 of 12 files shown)


marcos-iov (approved):

rskj-core/src/test/java/co/rsk/peg/utils/BridgeEventLoggerImplTest.java
```diff
@@ -625,6 +625,76 @@ public void logReleaseBtcRequested() {
...
+        CallTransaction.Function event = BridgeEvents.REJECTED_PEGIN.getEvent();
+
+        byte[][] topics = event.encodeEventTopics(btcTx.getHash().getBytes());
+
+        for (int i=0; i<topics.length; i++) {
+            Assert.assertArrayEquals(topics[i], result.getTopics().get(i).getData());
+        }
+
+        // Assert log data
+        Assert.assertArrayEquals(event.encodeEventData(RejectedPeginReason.PEGIN_CAP_SURPASSED.getValue()), result.getData());
```
ajlopezrsk, line 660: Could we have an explicit test of event having the signature expected by external tools? ie, an assert equals against a calculated hash. So, we could be sure the event is consumible and queryable by external tools, using a code test. My concert is that, if we apply fuzzy testing to the event function code (and then, changing the resulting signature), all these tests could pass in green

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -475,6 +495,14 @@ private void refundTxSender(
             generateRejectionRelease(btcTx, btcRefundAddress, rskTx, amount);
         } else {
             logger.debug("[refundTxSender] No btc refund address provided, couldn't get sender address either. Can't refund");
+
+            if (activations.isActive(ConsensusRule.RSKIP181)) {
+                if (activations.isActive(ConsensusRule.RSKIP170)) {
+                    eventLogger.logUnrefundablePegin(btcTx, UnrefundablePeginReason.PEGIN_V1_REFUND_ADDRESS_NOT_SET);
+                } else {
+                    eventLogger.logUnrefundablePegin(btcTx, UnrefundablePeginReason.LEGACY_PEGIN_UNDETERMINED_SENDER);
+                }
```
josedahlquist, line 504: This is not correct.
The status of RSKIP170 doesn't ensure the pegin was a pegin-v1. pegin v1 and pegin legacy will co exist after Iris activation.
You can check the type of pegin using the peginInformation object (see https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/PeginInformation.java#L44 )
```suggestion
                if (peginInformation.getProtocolVersion() == 1) {
                    eventLogger.logUnrefundablePegin(btcTx, UnrefundablePeginReason.PEGIN_V1_REFUND_ADDRESS_NOT_SET);
                } else {
                    eventLogger.logUnrefundablePegin(btcTx, UnrefundablePeginReason.LEGACY_PEGIN_UNDETERMINED_SENDER);
                }
```
This could be improved by at least encoding the `1` in an enum. What do you think @marcos-iov ?

rskj-core/src/test/java/co/rsk/peg/utils/BridgeEventLoggerImplTest.java
josedahlquist: @ajlopezrsk has a good point here. We are doing this though in the integration tests already, so we can leave this comment as tech debt to be solved by the fed team afterwards

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
marcos-iov: Good catch! Yes, an enum with the different peg-in protocol versions (0 and 1 for the moment) should do it

Vovchyk: thanks @josedahlquist. Applied your suggestion.

josedahlquist (dismissed):

marcos-iov (dismissed):

donequis (approved):
Seems good

marcos-iov (approved):

PR #1403, opened 2021-01-07
New bridge constants: irisActivationBtcHeight and maxDepthToSearchBlo…

marcos-iov:


rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
```diff
@@ -60,6 +60,9 @@
 
     protected int lockingCapIncrementsMultiplier;
 
+    protected int irisActivationBtcHeight;
```
josedahlquist, line 63:
```suggestion
    protected int btcHeightWhenBlockIndexActivates;
```
Maybe a name that doesn't necessary ties to Iris is better, what do you think?

rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
```diff
@@ -60,6 +60,9 @@
...
+    protected int maxDepthToSearchBlocksBelowIrisActivation;
```
josedahlquist, line 64:
```suggestion
    protected int maxDepthToSearchBlocksBelowIndexActivation;
```
Same?

josedahlquist (approved):

josedahlquist (approved):

PR #1410, opened 2021-01-11
Program changes

josedahlquist:
### Add originHash to InternalTransaction 
This field is only to be used in memory, doesn't change the encoded internal Transaction.
- originHash: will be the tx hash of the first transaction that initiated a chain of internal transactions.

### Change Program for precompiled contracts 
Add log support for precompiled contracts when executing internal transaction

## Motivation and Context
Allow execution internal transactions in precompiled contracts

## How Has This Been Tested?
Unit tests provided

## Types of changes

## Checklist:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/vm/program/InternalTransaction.java
+++ b/rskj-core/src/main/java/org/ethereum/vm/program/InternalTransaction.java
@@ -29,22 +29,29 @@ import static co.rsk.util.ListArrayUtil.*;
 
 public class InternalTransaction extends Transaction {
 
-    private byte[] parentHash;
-    private int deep;
-    private int index;
+    private final byte[] originHash;
+    private final byte[] parentHash;
+    private final int deep;
+    private final int index;
+    private final String note;
     private boolean rejected = false;
-    private String note;
 
-    public InternalTransaction(byte[] parentHash, int deep, int index, byte[] nonce, DataWord gasPrice, DataWord gasLimit,
-                               byte[] sendAddress, byte[] receiveAddress, byte[] value, byte[] data, String note) {
+    public InternalTransaction(byte[] originHash, byte[] parentHash, int deep, int index, byte[] nonce, DataWord gasPrice, DataWord gasLimit,
+        byte[] sendAddress, byte[] receiveAddress, byte[] value, byte[] data, String note) {
 
         super(nonce, getData(gasPrice), getData(gasLimit), receiveAddress, nullToEmpty(value), nullToEmpty(data));
 
+        this.originHash = originHash.clone();
         this.parentHash = parentHash;
         this.deep = deep;
         this.index = index;
         this.sender = RLP.parseRskAddress(sendAddress);
         this.note = note;
+    }
+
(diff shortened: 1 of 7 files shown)


guidohernan93 (approved):

rskj-core/src/main/java/org/ethereum/vm/program/InternalTransaction.java
```diff
@@ -73,7 +79,11 @@ public String getNote() {
     }
 
     public byte[] getParentHash() {
-        return parentHash;
+        return parentHash.clone();
+    }
+
+    public byte[] getOriginHash() {
+        return originHash.clone();
```
ajlopezrsk, line 86: I don't see any use of this new information in current logic, only in test. Could you elaborate the rationale to add this info now? BTW, the concept of internal transaction is flawed in our implementation: there is no internal transaction as transacton, only message. The original implementors chose to inherits from Transaction, but really, in Ethereum world there is no internal transaction AS transaction, only as a message. That's the reason why Solidity has msg.sender, not tx.sender

josedahlquist: Thanks for taking a look at this Angel!
The originHash is not used in this PR. It will be used by the Bridge though.
The new fast-bridge logic will be consumed by contracts, so the executing transaction will be an InternalTransaction. The bridge needs the origin transaction hash to store it internally.
This information is used by the powpeg node to extract the release_requested log from said tx. (This is required by the peg-out process)

pamgonzalez (approved):

rskj-core/src/main/java/org/ethereum/vm/program/InternalTransaction.java
ajlopezrsk: Thanks for the description and information

I'm a bit afraid to use something (initial transaction hash) that the rest of Ethereum world does not use.

To me, the original sin was to inject a Transaction into Bridge init. In other circunstances (ie the add of TxIndex, now removed) we used EthereumJ ProgramInvoke.

But even if the tx hash is needed (ie in the bridge process (I don't see any other use in OTHER precompiled contracts), as in current code (tx hash is used in the process you mentioned), adding this feature to InternalTransaction maybe DON'T cover all the use case. IE, I could imagine a transaction calling a DeFi contract A, that internally makes MANY CALLS and at the end, produces MANY calls to bridge release process, using the SAME tx hash (and even the same repeated amount). That is one reason to not use transaction hash in Ethereum world: it does not identify practically anything, you need something more to refer to an event or work

I will contact you to discuess directly (in Spanish I hope :-))

pamgonzalez (dismissed):

ajlopezrsk (dismissed):

rskj-core/src/main/java/org/ethereum/vm/program/InternalTransaction.java
donequis: I'm with aj on this. It's hard to imagine reasons to use the origin tx hash and being something error-prone is a reason to not use it

donequis (dismissed):

guidohernan93 (approved):

marcos-iov (approved):

PR #1412, opened 2021-01-15
[RSKIP186] Active Federation creation block height registration

Vovchyk:
## Description
When the Federation change voting process finishes, the Bridge emmits a commit_federation event. This event can be used to provide a blockchain proof of the Federation address when requested. Nowadays doing this would imply navigating the blockchain to determine when it was emmited, or constantly monitoring said events.

This RSKIP proposes adding on-chain information to easily fetch the block where the Federation was changed for the last time. With the block, a system could provide proof that it belongs to the mainchain.

Additionally, once a Federation migration process finishes, the retiring Federation is removed. This RSKIP proposes to keep the last retiring Federation address in storage, to be used in case there are outstanding funds to migrate after the retiring age.

## Motivation and Context

When a user gets the Federation address they usually interact with a node/service and have to trust this information. This exposes a risk where a man-in-the-middle kind of attack could lead users to send their funds to an invalid address thus losing them.

## Types of changes

## Checklist:

*fed:active-block-reg*

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/Bridge.java
+++ b/rskj-core/src/main/java/co/rsk/peg/Bridge.java
@@ -1066,6 +1066,12 @@ public class Bridge extends PrecompiledContracts.PrecompiledContract {
         return bridgeSupport.hasBtcBlockCoinbaseTransactionInformation(blockHash);
     }
 
+    public long getActiveFederationCreationBlockHeight(Object[] args) {
+        logger.trace("getActiveFederationCreationBlockHeight");
+
+        return bridgeSupport.getActiveFederationCreationBlockHeight();
+    }
+
     public static BridgeMethods.BridgeMethodExecutor activeAndRetiringFederationOnly(BridgeMethods.BridgeMethodExecutor decoratee, String funcName) {
         return (self, args) -> {
             Federation retiringFederation = self.bridgeSupport.getRetiringFederation();
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
@@ -450,6 +450,17 @@ public enum BridgeMethods {
             activations -> activations.isActive(RSKIP134),
             true
     ),
+    GET_ACTIVE_FEDERATION_CREATION_BLOCK_HEIGHT(
+            CallTransaction.Function.fromSignature(
+                    "getActiveFederationCreationBlockHeight",
+                    new String[]{},
+                    new String[]{"uint256"}
+            ),
+            fixedCost(3_000L),
+            (BridgeMethodExecutorTyped) Bridge::getActiveFederationCreationBlockHeight,
+            activations -> activations.isActive(RSKIP186),
+            true
+    ),
     INCREASE_LOCKING_CAP(
             CallTransaction.Function.fromSignature(
                     "increaseLockingCap",
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
@@ -459,7 +459,7 @@ public enum BridgeMethods {
             fixedCost(3_000L),
             (BridgeMethodExecutorTyped) Bridge::getActiveFederationCreationBlockHeight,
             activations -> activations.isActive(RSKIP186),
-            true
+            false
     ),
     INCREASE_LOCKING_CAP(
             CallTransaction.Function.fromSignature(
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
@@ -19,6 +19,7 @@
 package co.rsk.peg;
 
 import co.rsk.bitcoinj.core.*;
+import co.rsk.bitcoinj.script.Script;
 import co.rsk.config.BridgeConstants;
 import co.rsk.core.RskAddress;
 import co.rsk.crypto.Keccak256;
@@ -61,6 +62,11 @@ public class BridgeStorageProvider {
     private static final DataWord LOCKING_CAP_KEY = DataWord.fromString("lockingCap");
     private static final DataWord RELEASE_REQUEST_QUEUE_WITH_TXHASH = DataWord.fromString("releaseRequestQueueWithTxHash");
     private static final DataWord RELEASE_TX_SET_WITH_TXHASH = DataWord.fromString("releaseTransactionSetWithTxHash");
+
+    // Federation creation keys
+    private static final DataWord ACTIVE_FEDERATION_CREATION_BLOCK_HEIGHT_KEY = DataWord.fromString("activeFedCreationBlockHeight");
+    private static final DataWord NEXT_FEDERATION_CREATION_BLOCK_HEIGHT_KEY = DataWord.fromString("nextFedCreationBlockHeight");
+    private static final DataWord LAST_RETIRED_FEDERATION_P2SH_SCRIPT_KEY = DataWord.fromString("lastRetiredFedP2SHScript");
 
     // Version keys and versions
     private static final DataWord NEW_FEDERATION_FORMAT_VERSION = DataWord.fromString("newFederationFormatVersion");
@@ -108,6 +114,10 @@ public class BridgeStorageProvider {
     private HashMap<Sha256Hash, Long> btcTxHashesToSave;
 
     private Map<Sha256Hash, CoinbaseInformation> coinbaseInformationMap;
+
+    private Long activeFederationCreationBlockHeight;
+    private Long nextFederationCreationBlockHeight; // if -1, then clear value
+    private Script lastRetiredFederationP2SHScript;
 
     public BridgeStorageProvider(Repository repository, RskAddress contractAddress, BridgeConstants bridgeConstants, ActivationConfig.ForBlock activations) {
         this.repository = repository;
@@ -592,6 +602,89 @@ public class BridgeStorageProvider {
             safeSaveToRepository(getStorageKeyForCoinbaseInformation(blockHash), data, BridgeSerializationUtils::serializeCoinbaseInformation));
     }
 
+    public Long getActiveFederationCreationBlockHeight() {
+        if (!activations.isActive(RSKIP186)) {
+            return null;
+        }
+
+        if (activeFederationCreationBlockHeight != null) {
+            return activeFederationCreationBlockHeight;
+        }
+
+        activeFederationCreationBlockHeight = safeGetFromRepository(ACTIVE_FEDERATION_CREATION_BLOCK_HEIGHT_KEY, BridgeSerializationUtils::deserializeOptionalLong).orElse(null);
+        return activeFederationCreationBlockHeight;
+    }
+
+    public void setActiveFederationCreationBlockHeight(long activeFederationCreationBlockHeight) {
+        this.activeFederationCreationBlockHeight = activeFederationCreationBlockHeight;
+    }
+
+    protected void saveActiveFederationCreationBlockHeight() {
+        if (activeFederationCreationBlockHeight == null || !activations.isActive(RSKIP186)) {
+            return;
+        }
+
+        safeSaveToRepository(ACTIVE_FEDERATION_CREATION_BLOCK_HEIGHT_KEY, activeFederationCreationBlockHeight, BridgeSerializationUtils::serializeLong);
+    }
+
+    public Long getNextFederationCreationBlockHeight() {
+        if (!activations.isActive(RSKIP186)) {
+            return null;
+        }
+
+        if (nextFederationCreationBlockHeight != null) {
+            return nextFederationCreationBlockHeight;
+        }
+
+        nextFederationCreationBlockHeight = safeGetFromRepository(NEXT_FEDERATION_CREATION_BLOCK_HEIGHT_KEY, BridgeSerializationUtils::deserializeOptionalLong).orElse(null);
+        return nextFederationCreationBlockHeight;
+    }
+
+    public void setNextFederationCreationBlockHeight(long nextFederationCreationBlockHeight) {
+        this.nextFederationCreationBlockHeight = nextFederationCreationBlockHeight;
+    }
+
+    public void clearNextFederationCreationBlockHeight() {
+        this.nextFederationCreationBlockHeight = -1L;
+    }
+
+    protected void saveNextFederationCreationBlockHeight() {
+        if (nextFederationCreationBlockHeight == null || !activations.isActive(RSKIP186)) {
+            return;
+        }
+
+        if (nextFederationCreationBlockHeight == -1L) {
+            safeSaveToRepository(NEXT_FEDERATION_CREATION_BLOCK_HEIGHT_KEY, null, BridgeSerializationUtils::serializeLong);
+        } else {
+            safeSaveToRepository(NEXT_FEDERATION_CREATION_BLOCK_HEIGHT_KEY, nextFederationCreationBlockHeight, BridgeSerializationUtils::serializeLong);
+        }
+    }
+
+    public Script getLastRetiredFederationP2SHScript() {
+        if (!activations.isActive(RSKIP186)) {
+            return null;
+        }
+
+        if (lastRetiredFederationP2SHScript != null) {
+            return lastRetiredFederationP2SHScript;
+        }
+
+        lastRetiredFederationP2SHScript = safeGetFromRepository(LAST_RETIRED_FEDERATION_P2SH_SCRIPT_KEY, BridgeSerializationUtils::deserializeScript);
+        return lastRetiredFederationP2SHScript;
+    }
+
+    public void setLastRetiredFederationP2SHScript(Script lastRetiredFederationP2SHScript) {
+        this.lastRetiredFederationP2SHScript = lastRetiredFederationP2SHScript;
+    }
+
+    protected void saveLastRetiredFederationP2SHScript() {
+        if (lastRetiredFederationP2SHScript == null || !activations.isActive(RSKIP186)) {
+            return;
+        }
+
+        safeSaveToRepository(LAST_RETIRED_FEDERATION_P2SH_SCRIPT_KEY, lastRetiredFederationP2SHScript, BridgeSerializationUtils::serializeScript);
+    }
+
     public void save() throws IOException {
         saveBtcTxHashesAlreadyProcessed();
 
@@ -619,6 +712,10 @@ public class BridgeStorageProvider {
         saveHeightBtcTxHashAlreadyProcessed();
 
         saveCoinbaseInformations();
+
+        saveActiveFederationCreationBlockHeight();
+        saveNextFederationCreationBlockHeight();
+        saveLastRetiredFederationP2SHScript();
     }
 
     private DataWord getStorageKeyForBtcTxHashAlreadyProcessed(Sha256Hash btcTxHash) {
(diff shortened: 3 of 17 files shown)


Vovchyk:
pipeline: run

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -2149,6 +2150,138 @@ public void saveCoinBaseInformation_after_RSKIP143() throws IOException {
         );
     }
```
marcos-iov, line 2152: Add tests for the new `get..` methods

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -114,22 +112,28 @@ public static boolean isValidLockTx(BtcTransaction tx) {
...
      * @param tx the BTC transaction to check
-     * @param federations the active federations
+     * @param activeFederations the active federations
+     * @param retiredFederationP2SHScript the retired federation P2SHScript. Could be {@code null}.
      * @param btcContext the BTC Context
      * @param bridgeConstants the Bridge constants
      * @return true if this is a valid lock transaction
      */
-    public static boolean isLockTx(BtcTransaction tx, List<Federation> federations, Context btcContext, BridgeConstants bridgeConstants) {
+    public static boolean isLockTx(BtcTransaction tx, List<Federation> activeFederations, Script retiredFederationP2SHScript,
```
guidohernan93, line 121: It could be useful to rename this method to `isPegInTx` so that expression `lock` is no longer used

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -5096,11 +5096,15 @@ public void getTransactionType_migration_tx() {
         }
         migrationTxInput.setScriptSig(inputScript);
         Assert.assertEquals(BridgeSupport.TxType.MIGRATION, bridgeSupport.getTransactionType(migrationTx));
+
+        when(mockFederationSupport.getRetiringFederation()).thenReturn(null);
+        when(provider.getLastRetiredFederationP2SHScript()).thenReturn(retiringFederation.getP2SHScript());
+        Assert.assertEquals(BridgeSupport.TxType.MIGRATION, bridgeSupport.getTransactionType(migrationTx));
     }
```
marcos-iov, line 5104: Add tests for updateCollections now with `updateFederationCreationBlockHeights`. See `callProcessFundsMigration_is_migrating_after_rskip_146_activation` for example

rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -1066,6 +1066,12 @@ public boolean hasBtcBlockCoinbaseTransactionInformation(Object[] args) {
         return bridgeSupport.hasBtcBlockCoinbaseTransactionInformation(blockHash);
     }
 
+    public long getActiveFederationCreationBlockHeight(Object[] args) {
+        logger.trace("getActiveFederationCreationBlockHeight");
```
ajlopezrsk, line 1070: Maybe, we could add the result into the log message

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -592,6 +602,89 @@ private void saveCoinbaseInformations() {
...
+        nextFederationCreationBlockHeight = safeGetFromRepository(NEXT_FEDERATION_CREATION_BLOCK_HEIGHT_KEY, BridgeSerializationUtils::deserializeOptionalLong).orElse(null);
+        return nextFederationCreationBlockHeight;
+    }
+
+    public void setNextFederationCreationBlockHeight(long nextFederationCreationBlockHeight) {
+        this.nextFederationCreationBlockHeight = nextFederationCreationBlockHeight;
+    }
+
+    public void clearNextFederationCreationBlockHeight() {
+        this.nextFederationCreationBlockHeight = -1L;
```
ajlopezrsk, line 648: Maybe we should check if the activation is on

ajlopezrsk, line 648: It's not clear that the `BridgeSerializationUtils` used for this variable could manage a negative value (my first look: it uses `BigInteger.fromUnsignedByteArray`). But then it uses `BigInteger.longValue` ;  do we have this test?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -811,6 +817,18 @@ private void processReleaseTransactions(Transaction rskTx) {
         }
     }
 
+    private void updateFederationCreationBlockHeights() {
+        if (!activations.isActive(RSKIP186)) {
+            return;
+        }
+
+        Long nextFederationCreationBlockHeight = provider.getNextFederationCreationBlockHeight();
+        if (nextFederationCreationBlockHeight != null) {
```
ajlopezrsk, line 826: I'm not sure about using null as an "not present" value. Usually, an smart contract using storage finds the unused cell already initialized in zero.

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
```diff
@@ -1009,6 +1012,27 @@ public void deserializeInteger() {
...
+
+        Assert.assertEquals(expectedScript, new Script(((RLPList) RLP.decode2(actualData).get(0)).get(0).getRLPData()));
+    }
+
+    @Test
+    public void deserializeScript() {
+        Script expectedScript = ScriptBuilder.createP2SHOutputScript(2, Lists.newArrayList(new BtcECKey(), new BtcECKey(), new BtcECKey()));
+        byte[][] rlpElements = new byte[1][];
+        rlpElements[0] = RLP.encodeElement(expectedScript.getProgram());
+        byte[] data = RLP.encodeList(rlpElements);
```
ajlopezrsk, line 1029: Minor comment: we could use  `byte[] data = RLP.encodeList(RLP.encodeElement(expectedScript.getProgram()));` directly instead of declaring explicitly an array of arrays

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
```diff
@@ -1009,6 +1012,27 @@ public void deserializeInteger() {
         Assert.assertEquals(1200, BridgeSerializationUtils.deserializeInteger(RLP.encodeBigInteger(BigInteger.valueOf(1200))).intValue());
     }
 
+    @Test
+    public void serializeScript() {
+        Script expectedScript = ScriptBuilder.createP2SHOutputScript(2, Lists.newArrayList(new BtcECKey(), new BtcECKey(), new BtcECKey()));
```
ajlopezrsk, line 1017: Maybe it is a good time to add some direct explicit test exercising `ScriptBuilder`

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -811,6 +817,18 @@ private void processReleaseTransactions(Transaction rskTx) {
...
+            provider.setActiveFederationCreationBlockHeight(nextFederationCreationBlockHeight);
```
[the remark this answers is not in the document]
Vovchyk: hmm.. interesting question. @josedahlquist what is your thought?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -2070,6 +2097,27 @@ public boolean hasBtcBlockCoinbaseTransactionInformation(Sha256Hash blockHash) {
...
+                return nextFederationCreationBlockHeight;
+            }
+        }
+
+        Long activeFederationCreationBlockHeight = provider.getActiveFederationCreationBlockHeight();
+        if (activeFederationCreationBlockHeight != null) {
+            return activeFederationCreationBlockHeight;
+        }
+
+        return 0L;
```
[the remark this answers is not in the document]
Vovchyk: @josedahlquist your suggestion was to return zero if neither active nor next block height is set yet, wasn't it? Do you think we should return `federationSupport.getActiveFederation().getCreationBlockNumber()` instead?

rskj-core/src/main/java/co/rsk/peg/Bridge.java
Vovchyk: well, we could, but then it won't be consistent with other trace messages in this class.

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
Vovchyk: not sure this is the responsibility of this class. It's data level entity - it simply stores and retrieves data.

Vovchyk: `-1L` value is not being serialized. `null` is being stored instead. It's just some sort of flag that indicates that the value should be cleared during the `save` stage.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
Vovchyk: `null` is only being used at the data level - to distinguish if it was set. Precompiled contract returns `zero`, if not set.

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
Vovchyk: @ajlopezrsk  not sure I got it correctly. Do you mean adding some tests for `ScriptBuilder` class? It's from the third-party library.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -811,6 +817,16 @@ private void processReleaseTransactions(Transaction rskTx) {
         }
     }
 
+    private void updateFederationCreationBlockHeights() {
+        if (!activations.isActive(RSKIP186)) {
+            return;
+        }
+
+        Long nextFederationCreationBlockHeight = provider.getNextFederationCreationBlockHeight();
+        provider.setActiveFederationCreationBlockHeight(nextFederationCreationBlockHeight);
```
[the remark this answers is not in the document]
Vovchyk: looks like an outdated change. Posted a comment on the similar question below.

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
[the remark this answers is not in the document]
Vovchyk: done

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -159,24 +163,31 @@ private static boolean isReleaseTx(BtcTransaction tx, Federation federation) {
     }
 
     public static boolean isReleaseTx(BtcTransaction tx, List<Federation> federations) {
```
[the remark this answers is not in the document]
Vovchyk: done

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
Vovchyk: done

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
Vovchyk: done

Vovchyk:
pipeline: run

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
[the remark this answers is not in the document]
josedahlquist: Yes, we should be checking if the active Federation became active.... The same should be done in the method `getActiveFederationCreationBlockHeight` as there might be a gap between the last call to the update_collections and the actual activation of the new Federation.
I mean that we should also check this in the `getActiveFederationCreationBlockHeight` method, and adjust accordingly (we cannot do it in that method only because that method will most likely not be mined).

The validation should be equivalent to what co.rsk.peg.FederationSupport#shouldFederationBeActive does:
```
    private boolean shouldFederationBeActive(Federation federation) {
        long federationAge = executionBlock.getNumber() - federation.getCreationBlockNumber();
        return federationAge >= bridgeConstants.getFederationActivationAge();
    }
```

If the new Federation should be active, then we have to replace the `activeFederationCreationBlockHeight` with `nextFederationCreationBlockHeight` and clear `nextFederationCreationBlockHeight`.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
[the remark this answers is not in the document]
josedahlquist: correct. This is to ensure we won't return invalid data. A zero indicates we don't have the right data, yet.

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -2149,6 +2150,435 @@ public void saveCoinBaseInformation_after_RSKIP143() throws IOException {
...
+        // Once the network upgrade is active, we will store it in the repository
+        verify(repository, times(1)).addStorageBytes(
+                PrecompiledContracts.BRIDGE_ADDR,
+                DataWord.fromString("activeFedCreationBlockHeight"),
+                BridgeSerializationUtils.serializeLong(10L)
+        );
+    }
+
+    @Test
+    public void saveActiveFederationCreationBlockHeight_before_RSKIP186() {
```
josedahlquist, line 2270:
```suggestion
    public void getActiveFederationCreationBlockHeight_before_RSKIP186() {
```

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -2149,6 +2150,435 @@ public void saveCoinBaseInformation_after_RSKIP143() throws IOException {
...
+        // Once the network upgrade is active, we will store the value in the repository
+        verify(repository, times(1)).addStorageBytes(
+                PrecompiledContracts.BRIDGE_ADDR,
+                DataWord.fromString("nextFedCreationBlockHeight"),
+                BridgeSerializationUtils.serializeLong(1L)
+        );
+    }
+
+    @Test
+    public void getNextFederationCreationBlockHeight_before_fork() {
```
josedahlquist, line 2323: duplicated?

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -2149,6 +2150,435 @@ public void saveCoinBaseInformation_after_RSKIP143() throws IOException {
...
+        // Once the network upgrade is active, we will store it in the repository
+        verify(repository2, times(1)).addStorageBytes(
+                PrecompiledContracts.BRIDGE_ADDR,
+                DataWord.fromString("nextFedCreationBlockHeight"),
+                null
+        );
+    }
+
+    @Test
+    public void saveNextFederationCreationBlockHeight_before_RSKIP186() {
```
josedahlquist, line 2420: you are testing getActiveFederationCreationBlockHeight again (?)

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -2149,6 +2150,435 @@ public void saveCoinBaseInformation_after_RSKIP143() throws IOException {
...
+        // Once the network upgrade is active, we will store it in the repository
+        verify(repository, times(1)).addStorageBytes(
+                PrecompiledContracts.BRIDGE_ADDR,
+                DataWord.fromString("lastRetiredFedP2SHScript"),
+                BridgeSerializationUtils.serializeScript(script)
+        );
+    }
+
+    @Test
+    public void saveLastRetiredFederationP2SHScript_before_RSKIP186() {
```
josedahlquist, line 2566: set and save tests are testing the same, you can keep one of them and the coverage will remain the same :)

josedahlquist (commented):
The changes look good, I would change the returns in the storageProvider to use Optional.of instead of returning null.
And we have to make the changes Marcos mentioned regarding checking the activation of the new Fed before setting its creation block height

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
ajlopezrsk: But, why to have a -1L. Which is the use case? A normal contract would retrieve 0 and nothing more (and manage the zero at code level). All the logic BEFORE the activation should live with a 0 in that data slot

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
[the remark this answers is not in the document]
Vovchyk: done

rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
```diff
@@ -450,6 +450,17 @@
...
+    GET_ACTIVE_FEDERATION_CREATION_BLOCK_HEIGHT(
+            CallTransaction.Function.fromSignature(
+                    "getActiveFederationCreationBlockHeight",
+                    new String[]{},
+                    new String[]{"uint256"}
+            ),
+            fixedCost(3_000L),
+            (BridgeMethodExecutorTyped) Bridge::getActiveFederationCreationBlockHeight,
+            activations -> activations.isActive(RSKIP186),
+            true
```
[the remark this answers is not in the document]
Vovchyk: done

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -592,6 +602,89 @@ private void saveCoinbaseInformations() {
             safeSaveToRepository(getStorageKeyForCoinbaseInformation(blockHash), data, BridgeSerializationUtils::serializeCoinbaseInformation));
     }
 
+    public Long getActiveFederationCreationBlockHeight() {
+        if (!activations.isActive(RSKIP186)) {
+            return null;
+        }
```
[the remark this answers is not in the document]
Vovchyk: done

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
Vovchyk: done. Removed `set*` tests.

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
Vovchyk: as discussed zero value doesn't clear it

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
josedahlquist: @Vovchyk in this test you are not saving, just getting. Seems like a duplicate from a previous test or am I wrong?

josedahlquist (dismissed):
a minor detail is still missing to me with a test, but other than that LGTM, great work !

marcos-iov (dismissed):

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
Vovchyk: thanks @josedahlquist. Good catch. Actually in this class we have two `getXXX` tests (before n after hf), two `saveXXX` (before n after hf) and one `saveXXXAndGetXXX` (which tests saving to and getting from non-mocked repository). I've fixed `saveXXX_before_RSKIP186` test that you're refering to above in here ce43150

donequis (dismissed):

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

marcos-iov (approved):

PR #1419, opened 2021-01-21
Depth limit for Receive Header.

pamgonzalez:
Check the depth of the block, taking into account the depth in the storage of the previous block.
If the depth of the block is too old, then it returns error. If it can't find the previous block in storage, then it also returns error.
The new errors that are introduced here are:

-2 -> RECEIVE_HEADER_BLOCK_TOO_OLD
-3 -> RECEIVE_HEADER_CANT_FOUND_PREVIUOS_BLOCK

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
@@ -5166,7 +5166,6 @@ public class BridgeSupportTest {
 
         bridgeSupport.receiveHeader(btcBlock2);
 
-        // Calls put when is adding the block header. (Saves his storedBlock)
         verify(btcBlockStore, times(1)).put(storedBlock2);
         verify(provider, times(1)).getReceiveHeadersLastTimestamp();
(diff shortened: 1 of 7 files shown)


rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -5134,6 +5134,244 @@ public void processPegIn_tx_no_lockable_by_surpassing_locking_cap() throws IOExc
...
+                activationsAfterForks
+        );
+;
+        when(storedBlock.getHeight()).thenReturn(10, 5000, 10);
+
+        int result = bridgeSupport.receiveHeader(btcBlock2);
+
+        StoredBlock storedBlock2 = storedBlock.build(btcBlock2);
+
+        // Calls put when is adding the block header. (Saves his storedBlock)
```
marcos-iov, line 5368: This comment is repeated in all tests, check when it's valid and when it should say something else

josedahlquist (approved):

PR #1433, opened 2021-02-02
Bridge performance improvement

marcos-iov:
*fed:bridge-performance-improvement-integration*


marcos-iov:
pipeline:run

rskj-core/src/main/java/co/rsk/peg/RepositoryBtcBlockStoreWithCache.java
```diff
@@ -132,7 +197,48 @@ public StoredBlock getStoredBlockAtMainChainHeight(int height) throws BlockStore
         StoredBlock chainHead = getChainHead();
         int depth = chainHead.getHeight() - height;
         logger.trace("Getting btc block at depth: {}", depth);
-        return getStoredBlockAtMainChainDepth(depth);
+
+        if (depth < 0) {
```
donequis, line 201: If this case is possible, what happens now? It is not throwing an exception but returning null. Does the caller handle it the same way?

donequis: I'd probably do it slightly different. Just add before this a `if (!activations.isActive(...)) { return getStoredBlockAtMainChainDepth(depth))` and then check for `depth < 0` and all the following code with no more activations guards.

rskj-core/src/main/java/co/rsk/peg/RepositoryBtcBlockStoreWithCache.java
```diff
@@ -132,7 +197,48 @@ public StoredBlock getStoredBlockAtMainChainHeight(int height) throws BlockStore
...
+            int limit;
+            if (chainHead.getHeight() - btcHeightWhenBlockIndexActivates > maxDepthToSearch) {
+                limit = btcHeightWhenBlockIndexActivates;
+            } else {
+                limit = chainHead.getHeight() - maxDepthToSearch;
+            }
+            logger.trace("[getStoredBlockAtMainChainHeight] Chain head height is {} and the depth limit {}", chainHead.getHeight(), limit);
+
+            if (height < limit) {
+                String message = String.format(
```
donequis, line 223: If `height < limit` this method is returning early instead of calling `getStoredBlockAtMainChainDepth`. Is this desired? I have no access to the rskip199 for checking this

donequis: In fact there's this bug here where there will be an unwanted limit imposed by `getBtcHeightWhenBlockIndexActivates`. This is likely not what we want

rskj-core/src/main/java/co/rsk/peg/RepositoryBtcBlockStoreWithCache.java
```diff
@@ -108,6 +147,31 @@ public synchronized void setChainHead(StoredBlock newChainHead) {
         if (cacheBlocks != null) {
             populateCache(newChainHead);
         }
+        setMainChainBlock(newChainHead.getHeight(), newChainHead.getHeader().getHash());
```
donequis, line 150: setChainHead seems to set the chainhead only. This is, this method is not telling you anything about forks. Probably I'm missing it, but is there a test of btc forks with depth > 1? I feel like this will not work correctly

rskj-core/src/main/java/co/rsk/peg/RepositoryBtcBlockStoreWithCache.java
```diff
@@ -108,6 +147,31 @@ public synchronized void setChainHead(StoredBlock newChainHead) {
...
+        }
+
+        logger.trace("[getInMainchain] Found block with hash {} at height {}", bestBlockHash.get(), height);
+        return Optional.of(block);
+    }
+
+    @Override
+    public void setMainChainBlock(int height, Sha256Hash blockHash) {
+        logger.trace("[setMainChainBlock] Set block with hash {} at height {}", blockHash, height);
+        bridgeStorageProvider.setBtcBestBlockHashByHeight(height, blockHash);
```
donequis, line 174: There's a minimum height for btc block that I don't see neither validated nor enforced (I mean, preventing blocks outside that range from being saved here and making sure that starting at that height blocks will be saved)

marcos-iov:
pipeline:run

donequis (approved):

donequis (approved):

josedahlquist (dismissed):

guidohernan93 (approved):

pamgonzalez (dismissed):

marcos-iov:
pipeline:run

marcos-iov:
pipeline:run

josedahlquist (approved):

PR #1435, opened 2021-02-03
Fixing non-synchronized override

nagarev:
## Description

Adding missing **synchronized** modifier to overridden method.

## Motivation and Context

As described in LGTM [documentation](https://lgtm.com/rules/5990064/): "If a synchronized method is overridden in a subclass, the compiler does not require the overriding method to be synchronized. However, if the overriding method is not synchronized, the thread-safety of the subclass may be broken."

## How Has This Been Tested?

Using Unit Tests

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/remasc/RemascTransaction.java
+++ b/rskj-core/src/main/java/co/rsk/remasc/RemascTransaction.java
@@ -73,7 +73,7 @@ public class RemascTransaction extends Transaction {
     }
 
     @Override
-    public RskAddress getSender() {
+    public RskAddress getSender() { // lgtm [java/non-sync-override]
         return REMASC_ADDRESS;
     }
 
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/remasc/RemascTransaction.java
```diff
@@ -73,7 +73,7 @@ public long transactionCost(Constants constants, ActivationConfig.ForBlock activ
     }
 
     @Override
-    public RskAddress getSender() {
+    public synchronized RskAddress getSender() {
```
ajlopezrsk, line 76: Why we need to add this word? It would be better to remove the `synchronized` from parent class `Transaction` using an internal lock for BOTH `Transaction.getSender` methods

ajlopezrsk (changes requested):

rskj-core/src/main/java/co/rsk/remasc/RemascTransaction.java
nagarev: Done!

heitorfm: Just to keep it here, I think we need to talk about if its really necessary to have this synchronization here

nagarev:
Since the methods don't do a whole lot, and that we don't want to degrade the performance unexpectedly by adding the "synchronized word", we can get rid of this alerts at all.

marcciosilva (approved):

Vovchyk (approved):

PR #1442, opened 2021-02-05
Removing unused container warning

nagarev:
## Description

Removing LGTM unused container warning in **OP** class

## Motivation and Context

As described in LGTM [documentation](https://lgtm.com/rules/890083/): "If the contents of a collection or map are never accessed in any way, then it is useless and the code that updates it is effectively dead code. Often, such objects are left over from an incomplete refactoring, or they indicate an underlying logic error."

But in this case, this properties are being serialized on JSON RPC requests... Since this is the case, this warning can be seen as a false positive for us... so we are removing them.

In a future we could refactor the OP class and start serializing through the use of getters instead of fields, but it requires time and testing for it to go well.

## How Has This Been Tested?

Using Unit Tests

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/vm/trace/Op.java
+++ b/rskj-core/src/main/java/org/ethereum/vm/trace/Op.java
@@ -37,8 +37,8 @@ public class Op {
     private long gas;
     private long gasCost;
 
-    private List<String> memory = new ArrayList<>();
-    private List<String> stack = new ArrayList<>();
+    private List<String> memory = new ArrayList<>(); // lgtm [java/unused-container]
+    private List<String> stack = new ArrayList<>(); // lgtm [java/unused-container]
     private Map<String, String> storage = new HashMap<>();
 
     public OpCode getOp() {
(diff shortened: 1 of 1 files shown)


rskj-core/src/main/java/org/ethereum/vm/trace/Op.java
```diff
@@ -37,8 +37,8 @@
     private long gas;
     private long gasCost;
 
-    private List<String> memory = new ArrayList<>();
-    private List<String> stack = new ArrayList<>();
+    private List<String> memory = new ArrayList<>(); // lgtm [java/unused-container]
```
Vovchyk, line 40: why do we need these two fields, if we don't ever read them?

ajlopezrsk: Could you add a comment describing how these variables are used? (JSON serialization)

nagarev: Done!

ajlopezrsk (approved):

PR #1449, opened 2021-02-12
Improve World State Garbage Collector Performance

ajlopez:
## Description
The process of world state garbage collector can be configured to be enable. Usually it is not enabled. This pull request improves the performance of the process.

## Motivation and Context
When the world state garbage collector is working, the state is divided in epochs (in different key-value stores) according to block height of the root state. The current implementation freeze the block processing when the collector eliminates a bucket or epoch. In some cases (in testnet) this time could be 10 o more minutes. This pull request adds the use of a thread to such task, allowing the rest of the node continue the process of blocks.

## How Has This Been Tested?

It was tested by the existing code tests, and also manually using 20_000 blocks long synchronization of testnet, with epoch of 100 blocks (low value, not recommended for production)

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

* **Other information**:
I recommend the test of this new functionality in testnet and mainnet, in long and in normal synchronization, with production configuration parameters.

One possible configuration for testing:

```
blockchain = {
  gc = {
        enabled = true
        epochs = 3
        blocksPerEpoch = 20000
  }
}
```


SergioDemianLerner:
Hi! The design of the garbage collector was made so it didn't require any read lock or write locks even in case of multithreading. I understand that locks were added because in the current implementation the epochs array is rotated, and this interferes with the foreach() iterations over the epochs performed in several parts of the code.
Looking in detail, the only place that would require a write lock is discardOldestEpoch() because of the join operation of database removal and Collections.rotate(). Technically rotate() could be replaced by a rotation operation that preserves semantics quite well by shifting instead of rotating (which duplicates elements while doing so), and making sure that duplicate elements are discarded in each foreach() loop, but I understand that doing so requires knowledge of the thread-safeness of the List<TrieStore>iterator. Another option is to let the epochs vector grow freely and never rotate, and add a head and tail pointer to the list to mark the current epochs being active. Tail pointer incrementation can be made thread-safe using AtomicLong, which is very fast, surely much faster than the locks. 
I don't think the locks will decrease the performance right now, so the implementation seems fine. Although documenting the performance of the read/write operations prior and past this change could be helpful to track potential future performance problems.

PR #1459, opened 2021-02-23
Improve save receipts performance

Vovchyk:
## Description
Store receipts with the same tx hash (but different block hash) separately.

## Motivation and Context

With this change receipts store implementation will be storing receipts with the same tx hash separately (for different blocks). Now current implementation stores them in a list, which is being read and written multiple times in case of many sibling blocks are processed.

## Types of changes

## Checklist:

*fed: save-receipts-perf*


rskj-core/src/test/java/org/ethereum/db/ReceiptStoreImplV2Test.java
```diff
@@ -0,0 +1,339 @@
...
+import org.ethereum.datasource.HashMapDB;
+import org.ethereum.vm.LogInfo;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class ReceiptStoreImplV2Test {
```
ajlopezrsk, line 37: Maybe, we could have only one abstract test class, that invokes a factory method. The method creates a V1 or V2 store, and it is written in two concrete test classes

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -977,14 +978,21 @@ protected Web3InformationRetriever getWeb3InformationRetriever() {
...
+        int receiptsCacheSize = rskSystemProperties.getReceiptsCacheSize();
+        KeyValueDataSource ds = LevelDbDataSource.makeDataSource(Paths.get(rskSystemProperties.databaseDir(), "receipts"));
 
         if (receiptsCacheSize != 0) {
             ds = new DataSourceWithCache(ds, receiptsCacheSize);
         }
 
-        return new ReceiptStoreImpl(ds);
+        //noinspection SwitchStatementWithTooFewBranches
+        switch (rskSystemProperties.receiptsDataFormatVersion()) {
```
ajlopezrsk, line 990: I'm not sure to have this feature as a configuration setting. 

We could use V2 directly in the code. Then, having the fallback behavior, once the new code is distributed, the receipts database will be have the new version, automatically. And no one can set the setting to v1. If it were changed, the database will not work.

On the other hand, the settings is good for testing and comparison using the same code base in the rest of the application

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

rskj-core/src/main/java/org/ethereum/db/ReceiptStoreImplV2.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    private static List<byte[]> parseBlockHashList(byte[] txsBytes) {
+        if (txsBytes == null || txsBytes.length == 0) {
+            return Collections.emptyList();
+        }
+
+        RLPList txHashList = (RLPList) RLP.decode2(txsBytes).get(0);
+        if (txHashList.size() == 0 || txHashList.get(0) instanceof RLPList) {
+            return Collections.emptyList();
```
ajlopezrsk, line 115: I presume than when there is an instance of RLPList, then it is the old format. Maybe we could extract now the block hashes from the old format

rskj-core/src/test/java/org/ethereum/db/ReceiptStoreImplV2Test.java
Vovchyk: Good point. I've made it as a single parameterized test class.

rskj-core/src/main/java/co/rsk/RskContext.java
Vovchyk: removed that feature flag

rskj-core/src/main/java/org/ethereum/db/ReceiptStoreImplV2.java
Vovchyk: yes, correct. Here we don't really need block hashes extracted from older data as they cannot be used as keys for extracting saved receipts with new format.

ajlopezrsk (approved):

rskj-core/src/main/java/co/rsk/cli/tools/ConnectBlocks.java
```diff
@@ -46,9 +47,16 @@ public static void main(String[] args) throws IOException {
 
         String filename = args[0];
 
+        long startTime = System.currentTimeMillis();
+        System.out.println("Started at: " + Instant.ofEpochMilli(startTime).toString());
```
donequis, line 51: Guessing this was for testing locally

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

PR #1460, opened 2021-02-25
Send Transaction and Send Raw Transaction returns an error message when automine is enabled

ajlopez:
## Descriptio
When a transaction is mined immediately (when the automine is enabled, ie using `-Dminer.client.autoMine=true`, if an error occurs (ie out of gas, revert) a message describing the problem is returned

## Motivation and Context
Some node implementation used for testing (like ganache and hardhat) have a node that executes immediately a transaction (like our autominer) but also they return a message in case an error occurs. If the error was raised by a unsatisfied require (in Solidity) with an associated description message, that message is included in the returned JSON object.

## How Has This Been Tested?
It was tested manually, maybe a better code test could be implemented

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

The message added to `eth_sendTransaction` and `eth_sendRawTransaction` was taken from `eth_call` implementation. But it should be changed for compatibility with the ecosystem. Then, the returned message by `eth_call` was changed, too. In this case, the change was only the prefix of the message: the rest of the description is the same as before. But we should check if that change breaks something with our integration tests using tools like Truffle or OpenZeppelin tests

## Checklist:

* **Other information**:

We should research why the previous prefix in `eth_call` returned message was chosen.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
@@ -62,6 +62,7 @@ public class BlockExecutor {
     private final ActivationConfig activationConfig;
 
     private Map<Keccak256, ProgramResult> transactionResults;
+    private boolean registerProgramResults;
 
     public BlockExecutor(
             ActivationConfig activationConfig,
@@ -281,8 +282,6 @@ public class BlockExecutor {
 
         int txindex = 0;
 
-        this.transactionResults = new HashMap<>();
-
         for (Transaction tx : block.getTransactionsList()) {
             logger.trace("apply block: [{}] tx: [{}] ", block.getNumber(), i);
 
@@ -311,7 +310,10 @@ public class BlockExecutor {
             }
 
             executedTransactions.add(tx);
-            this.transactionResults.put(tx.getHash(), txExecutor.getResult());
+
+            if (this.registerProgramResults) {
+                this.transactionResults.put(tx.getHash(), txExecutor.getResult());
+            }
 
             if (vmTrace) {
                 txExecutor.extractTrace(programTraceProcessor);
@@ -411,5 +413,10 @@ public class BlockExecutor {
 
     public ProgramResult getProgramResult(Keccak256 txhash) {
         return this.transactionResults.get(txhash);
+    }
+
+    public void setRegisterProgramResults(boolean value) {
+        this.registerProgramResults = value;
+        this.transactionResults = new HashMap<>();
     }
 }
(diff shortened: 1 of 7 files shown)


ajlopez:
This post could help in the writing of truffle test using truffle-assertion https://kalis.me/assert-reverts-solidity-smart-contract-test-truffle/

(I run truffle test with truffle-assertions and revert message check, and not it works with our node (regtest, automine))

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -306,6 +311,10 @@ private BlockResult executeInternal(
 
             executedTransactions.add(tx);
 
+            if (this.registerProgramResults) {
```
fedejinich, line 314: mmm I think we should stop adding more resposabilities to this executor, we can have a special one for automine and build on top of that

fedejinich (approved):
I don't want to delay this PR, we can improve it later. I'll approve it and raise an issue

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -401,4 +410,13 @@ public static void maintainPrecompiledContractStorageRoots(Repository track, Act
...
         return logBloom.getData();
     }
+
+    public ProgramResult getProgramResult(Keccak256 txhash) {
+        return this.transactionResults.get(txhash);
+    }
+
+    public void setRegisterProgramResults(boolean value) {
+        this.registerProgramResults = value;
+        this.transactionResults = new HashMap<>();
```
Vovchyk, line 420: I'd suggest to initialize `transactionResults` in the line where it's declared and here simply do `clear()`, if needed. `getProgramResult()` will crash if `transactionResults` is not initialized yet.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
ajlopezrsk: Yes, but it would be a greater change. The block executor historically had taken TOO MUCH responsabilities to be refactored, in an improvement task

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
ajlopezrsk: Good catch, done

ajlopezrsk:
*pipeline:run*

alejandrod-iov:
pipeline:run

alejandrod-iov:
pipeline:run

alejandrod-iov:
pipeline:run

ajlopezrsk (approved):

PR #1465, opened 2021-03-08
Returns error when the sent block previously exists in storage.

pamgonzalez:


rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -1,49 +1,6 @@
 package co.rsk.peg;
 
-import static co.rsk.peg.PegTestUtils.createBaseInputScriptThatSpendsFromTheFederation;
```
josedahlquist, line 3: are you sure all these imports are not used?

josedahlquist (approved):
Other than my comment, the change LGTM

PR #1488, opened 2021-03-29
Include activation code for RSKIP 219

pamgonzalez:


rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ConsensusRule.java
```diff
@@ -18,7 +18,8 @@
 
 package org.ethereum.config.blockchain.upgrades;
 
-public enum ConsensusRule {
```
marcos-iov, line 21: Was this change on purpose?

guidohernan93: I think it was an accidental new line, but should not be like that.

pamgonzalez: Thanks guys, I will resolve it

pamgonzalez:
Use PR: https://github.com/rsksmart/rskj/pull/1494

pamgonzalez:
Use PR: https://github.com/rsksmart/rskj/pull/1494

PR #1483, opened 2021-03-22
Return Transaction Nonce as Quantity Value in JSON RPC methods

ajlopez:
## Description

The nonce of a transaction now it is returned WITHOUT leading zeroes

## Motivation and Context

The main motivation is to fix issues with ecosystem tools, like `geth`. In previous changes, other JSON RPC values were adapted to be returned WITHOU leading zeroes, but transaction nonce was not fixed.

## How Has This Been Tested?

With code tests

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Requires Activation Code (Hard Fork)

* **Other information**:

Issue #1484, opened 2021-03-24
RPC endpoint returns "remasc" transaction that makes no sense


samsondav:
Hi, this is Sam from Chainlink. We've been doing some integration work with RSK and run into a bit of trouble. This command:

```curl --data '{"method":"eth_getBlockByNumber","id":"1","params":["0x30fed8", true],"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST https://public-node.rsk.co | jq```

returns a "REMASC" transaction with 0 gas limit, 0 gas used and 0 gas price.

This breaks ethereum compatibility because geth tools such as the client assume that the gas used is greater than 0.

It might make sense for the RPC client to not return this transaction since it does not mesh well with existing etheruem tooling and is useless from a client's perspective.

fedejinich:
mmm I wont say that makes no sense, that transaction is part of the consensus and it might be relevant for the rsk ecosystem. Maybe we can find a middle path were both responses coexists (with gas used > 0 and without it)

PR #1490, opened 2021-03-29
Skip remasc transaction in BlockResultDTO according to configuration

ajlopez:
## Description

Trying to solve #1483 

Use  `-Drpc.skipRemasc=true`

## Motivation and Context

To be done

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
```diff
@@ -131,10 +132,22 @@ public static BlockResultDTO fromBlock(Block b, boolean fullTx, BlockStore block
         List<Transaction> blockTransactions = b.getTransactionsList();
         if (fullTx) {
             for (int i = 0; i < blockTransactions.size(); i++) {
+                Transaction tx = blockTransactions.get(i);
+
+                if (skipRemasc && tx.isRemascTransaction(i, blockTransactions.size())) {
```
fedejinich, line 137: mmm we don't want to skip remasc, I'd rather be more generic, we want to skip tx with gas used = 0. I know remasc is the only one that has gas used = 0, but this should be refactored. You can filter before going into the `for` instead of mantaining duplicated logic. And even if we just want to skip remasc it should be done outside `for` cycle

fedejinich: maybe
```java
List<Transaction> blockTransactions = withGasPriceGreaterThanZero(b.getTransactionsList(), flag or skipRemasc)
```

fedejinich (changes requested):

fedejinich:
isn't fixing [this one](https://github.com/rsksmart/rskj/issues/1484)?

fedejinich:
isn't fixing [this one](https://github.com/rsksmart/rskj/issues/1484)?

PR #1498, opened 2021-04-09
Fast bridge burning utxos fix

marcos-iov:
When a fast bridge Federation UTXO is spent, it probably generates a change UTXO, this change UTXO is then registered via registerBtcTransaction, but the logic in this method detects the transaction as an attempted peg-in, which will be rejected because the FB Federation is a multisig. This will burn the funds because the FB Federation won't register those funds.

Fix:
Extract the standard redeem script from the input using the existing bitcoinj logic, and compare it against the active federations redeem script. If it matches, then don't process the tx as a peg-in

*fed:fb-burning-utxos-fix*

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
@@ -194,7 +194,12 @@ public class BridgeUtils {
             }
         }
 
-        Wallet federationsWallet = BridgeUtils.getFederationsNoSpendWallet(btcContext, activeFederations, false, null);
+        Wallet federationsWallet = BridgeUtils.getFederationsNoSpendWallet(
+            btcContext,
+            activeFederations,
+            false,
+            null
+        );
         Coin valueSentToMe = tx.getValueSentToMe(federationsWallet);
         Coin minimumPegInTxValue = activations.isActive(ConsensusRule.RSKIP219) ?
             bridgeConstants.getMinimumPeginTxValueInSatoshis() :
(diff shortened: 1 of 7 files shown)


josedahlquist (approved):

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -324,8 +325,81 @@ public void testIsPegInTxForTwoFederations() {
     }
 
     @Test
-    public void testIsPegInTx_changeUtxoFromFastBridgeFederation() {
+    public void testIsPegInTx_hasChangeUtxoFromFastBridgeFederation() {
```
pamgonzalez, line 328: Could you include the expected result in the tittle?

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -324,8 +325,81 @@ public void testIsPegInTxForTwoFederations() {
...
+        // Create a tx from the fast bridge fed to the active fed
+        BtcTransaction tx = new BtcTransaction(networkParameters);
+        tx.addOutput(Coin.COIN, activeFederation.getAddress());
+        tx.addInput(Sha256Hash.ZERO_HASH, 0, fastBridgeRedeemScript);
+
+        Assert.assertFalse(BridgeUtils.isPegInTx(tx, activeFederation, btcContext, bridgeConstants));
+    }
+
+    @Test
+    public void testIsPegInTx_hasChangeUtxoFromFastBridgeErpFederation() {
```
pamgonzalez, line 345: Could you include the expected result in the tittle?

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -324,8 +325,81 @@ public void testIsPegInTxForTwoFederations() {
...
+        // Create a tx from the fast bridge erp fed to the active fed
+        BtcTransaction tx = new BtcTransaction(networkParameters);
+        tx.addOutput(Coin.COIN, activeFederation.getAddress());
+        tx.addInput(Sha256Hash.ZERO_HASH, 0, fastBridgeErpRedeemScript);
+
+        Assert.assertFalse(BridgeUtils.isPegInTx(tx, activeFederation, btcContext, bridgeConstants));
+    }
+
+    @Test
+    public void testIsPegInTx_hasChangeUtxoFromErpFederation() {
```
pamgonzalez, line 376: Could you include the expected result in the tittle?

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -189,6 +169,17 @@ public static boolean isPegInTx(BtcTransaction tx, List<Federation> activeFedera
             if (retiredFederationP2SHScript != null && scriptCorrectlySpendsTx(tx, index, retiredFederationP2SHScript)) {
                 return false;
             }
+
+            RedeemScriptParser redeemScriptParser = RedeemScriptParserFactory.get(tx.getInput(index).getScriptSig().getChunks());
```
donequis, line 173: I don't immediatelly see exploit here, but I see issues.

On one side, the `RedeemScriptParserFactory.get` could match an ERP style formed by non real erp members. Meaning that someone can create an ERP style redeem script with valid format, valid federators, but wrong ERP memebers, and here it would match. This can be possible be used to create a fork between updated and non updated nodes pre activation.

I would add some activation guard here and think carefully if we want the redeemscriptparserfactory to know the erp federation and the timelock parameter

marcos-iov:
pipeline:run

donequis (approved):

donequis (changes requested):
Don't know how to cancel approbal

donequis (changes requested):
When checking if the transaction is a pegout or a migration transaction we should check the same we do on pegin for fast bridge transactions.

donequis:
I mean, `isMigrationTx` and `isPegOutTx` should also match fast bridge transactions in a nicer way

pamgonzalez (dismissed):

donequis (approved):

josedahlquist:
pipeline:run

marcos-iov:
pipeline:run

josedahlquist (approved):

PR #1502, opened 2021-04-15
Get rid of stripJar task

Vovchyk:
## Description
This change removes `stripJar` gradle task. Also it includes refactoring of dependency references.

Files with the names like `LICENSE/LICENSE.*`, `NOTICE/NOTICE.*` in the `META-INF/` folder are renamed to include a lib’s name as prefix just to prevent their clashing

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/build.gradle
+++ b/rskj-core/build.gradle
@@ -1,7 +1,3 @@
-import java.util.jar.JarEntry
-import java.util.jar.JarFile
-import java.util.jar.JarOutputStream
-
 plugins {
     id 'application'
 }
@@ -137,9 +133,6 @@ tasks.withType(Jar) {
     reproducibleFileOrder = true
     dirMode = 0775
     fileMode = 0664
-    doLast {
-        stripJar(outputs.files.singleFile)
-    }
 }
 
 task sourcesJar(type: Jar, dependsOn: classes) {
@@ -215,34 +208,12 @@ artifacts {
     archives distTar
 }
 
-def gitCurrentBranch() {
+static def gitCurrentBranch() {
     def process = "git rev-parse --abbrev-ref HEAD".execute()
     return process.text.trim()
 }
 
-def gitCommitHash() {
+static def gitCommitHash() {
     def process = "git rev-parse --short HEAD".execute()
     return process.text.trim()
-}
-
-def gitDateHash() {
-    def process = "git show -s --format=%ct ${gitCommitHash()}".execute()
-    return process.text.toLong()
-}
-
-def stripJar(File originalFile) {
-    def constantTimeForZipEntries = new GregorianCalendar(1980, Calendar.FEBRUARY, 1, 0, 0, 0).timeInMillis
-    def reproducibleFile = File.createTempFile("temp", ".tmp")
-    reproducibleFile.withOutputStream { fout ->
-        def out = new JarOutputStream(fout)
-        def jf = new JarFile(originalFile)
-        jf.entries().unique {it.name}.sort {it.name}.each {
-            def copy = new JarEntry(it.name)
-            copy.time = constantTimeForZipEntries
-            out.putNextEntry(copy)
-            out << jf.getInputStream(it)
-        }
-        out.finish()
-    }
-    reproducibleFile.renameTo originalFile
 }
--- a/rskj-core/build.gradle
+++ b/rskj-core/build.gradle
@@ -38,11 +38,12 @@ repositories {
 }
 
 sourceCompatibility = 1.8
+targetCompatibility = 1.8
 
 mainClassName = 'co.rsk.Start'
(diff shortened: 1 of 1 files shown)


rskj-core/build.gradle
```diff
@@ -189,18 +243,31 @@ task generatePom(dependsOn: jar) {
...
-    classifier = 'all'
+
+    exclude 'META-INF/*.SF', 'META-INF/*.DSA', 'META-INF/*.RSA'
+
+    exclude 'META-INF/NOTICE'
+    exclude 'META-INF/NOTICE.md'
+    exclude 'META-INF/NOTICE.txt'
+    exclude 'META-INF/LICENSE'
+    exclude 'META-INF/LICENSE.md'
+    exclude 'META-INF/LICENSE.txt'
```
lucasvuotto, line 260: We need legal counsul before getting rid of it. Also I'd add a comment stating that we're excluding those files because it generates weird differences in the reproducible build, thing that bogs me a lot. Do we have any clue why this is happening? I haven't found anything yet, sadly.

rskj-core/build.gradle
```diff
@@ -189,18 +243,31 @@ task generatePom(dependsOn: jar) {
...
+
+    exclude 'module-info.class'
```
lucasvuotto, line 262: Are we removing this one for the same reasons as the previous chunk?

rskj-core/build.gradle
```diff
@@ -137,18 +194,15 @@ tasks.withType(Jar) {
     reproducibleFileOrder = true
     dirMode = 0775
     fileMode = 0664
```
lucasvuotto, line 196: Can you move this to the `tasks.withType(AbstractArchiveTask)` block above in line 51, instead of setting it just for `Jar`? In that case, we can get rid of this `tasks.withType(Jar)` block altogether, as `Jar` inherits from `AbstractArchiveTask`. Also, I did a build without this set, and only 2 files those modes while every other file and directory has group write permission disabled, so I feel more inclined to set

```
dirMode = 0755
fileMode = 0644
```

instead.

lucasvuotto (commented):
Overall, I'm fine with the changes. Big thanks for getting rid of the mixture of inline version numbers and variables! :hugs:

Maybe a nitpick, but using the `.set()` method for the deprecated variables feels off compared to the rest of the file. IIRC, `archiveClassifier = 'sources'` instead of `archiveClassifier.ste('sources')` should work too, and would prefer it like that.

rskj-core/build.gradle
Vovchyk: >using the .set() method for the deprecated variables feels off compared to the rest of the file. IIRC, archiveClassifier = 'sources' instead of archiveClassifier.ste('sources') should work too, and would prefer it like that.

Yeah, I'd also prefer this, but I did it due to the annoying message of my IDE - `Cannot assign 'String' to 'Property<String>' `

>We need legal counsul before getting rid of it. 

Agree, created this as a draft just to open the discussion

>Also I'd add a comment stating that we're excluding those files because it generates weird differences in the reproducible build, thing that bogs me a lot. Do we have any clue why this is happening?

The thing is that multiple third-party jars are being merged into the final uber/fat jar and some file names of them may clash. 

Before this change it was done the following way: first, gradle `fatJar` task puts all files from external dependencies into the `*-all.jar` file including duplicates of `.class`, `LICENSE`/`LICENSE.*`, `NOTICE`/`NOTICE.*` etc. files. After that, `stripJar` task removes duplicate entries from the jar keeping only first file with unique name and re-sets timestamps (which is useless in fact, cause `preserveFileTimestamps` flag does its job beforehand). As duplicates are removed, as of now, not all `LICENSE` and `NOTICE` files are being included into the final jar, only one instance.

My change removes the `stripJar` part and sets `EXCLUDE` duplicate strategy for the gradle task via this `duplicatesStrategy = DuplicatesStrategy.EXCLUDE` (`INCLUDE` is by default), which simply removes duplicate entries on its own. Also the change excludes `LICENSE`/`LICENSE.*`, `NOTICE`/`NOTICE.*` files, which is a question to discuss.

Regarding `module-info.class`, there're also a few entries of it from others deps. As of now, only one instance of them goes into the final jar. This file is related to the new java module system. As we provide an app, not a lib, I believe we could also get rid of it - OFC needs to be properly tested.

rskj-core/build.gradle
Vovchyk: Mentioned about this file in the comment above

rskj-core/build.gradle
Vovchyk: done

nagarev (dismissed):
LGTM

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

ajlopezrsk (approved):

PR #1504, opened 2021-04-22
Add new check for peg-out minimum considering fees 

josedahlquist:
A peg-out transaction has a minimum established in the Bridge constants.
This value doesn't take into account the changing nature of the fees.
Ths commit adds a validation to calculate the approximated cost of a peg-out with the current fees per KB of the network.
The cost is calculated considering a transaction with two inputs and two outputs, which is the most common situation ATM.

A peg-out transaction should also generate value for the user, this is, if a peg-out tx costs 400k satoshis and the user is attempting to peg-out 450k, the remainder (50k) won't be of much use for the user. In order to avoid this situation the Bridge now includes a percentage above the fee cost that the peg-out should include, if this is amount is not attained, the peg-out is rejected and funds returned.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
+++ b/rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
@@ -76,6 +76,8 @@ public class BridgeConstants {
 
     protected String oldFederationAddress;
 
+    protected int percentageAboveFeeForPegouts;
+
     public NetworkParameters getBtcParams() {
         return NetworkParameters.fromID(btcParamsString);
     }
@@ -157,4 +159,8 @@ public class BridgeConstants {
     public long getMinSecondsBetweenCallsToReceiveHeader() { return minSecondsBetweenCallsReceiveHeader; }
 
     public int getMaxDepthBlockchainAccepted() { return maxDepthBlockchainAccepted; }
+
+    public int getPercentageAboveFeeForPegouts() {
+        return percentageAboveFeeForPegouts;
+    }
 }
--- a/rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
+++ b/rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
@@ -76,7 +76,7 @@ public class BridgeConstants {
 
     protected String oldFederationAddress;
 
-    protected int percentageAboveFeeForPegouts;
+    protected int minimumPegoutValuePercentageToReceiveAfterFee;
 
     public NetworkParameters getBtcParams() {
         return NetworkParameters.fromID(btcParamsString);
@@ -160,7 +160,7 @@ public class BridgeConstants {
 
     public int getMaxDepthBlockchainAccepted() { return maxDepthBlockchainAccepted; }
 
-    public int getPercentageAboveFeeForPegouts() {
-        return percentageAboveFeeForPegouts;
+    public int getMinimumPegoutValuePercentageToReceiveAfterFee() {
+        return minimumPegoutValuePercentageToReceiveAfterFee;
     }
 }
--- a/rskj-core/src/main/java/co/rsk/config/BridgeMainNetConstants.java
+++ b/rskj-core/src/main/java/co/rsk/config/BridgeMainNetConstants.java
@@ -144,6 +144,8 @@ public class BridgeMainNetConstants extends BridgeConstants {
 
         minSecondsBetweenCallsReceiveHeader = 300;  // 5 minutes in Seconds
         maxDepthBlockchainAccepted = 25;
+
+        percentageAboveFeeForPegouts = 20;
     }
 
     public static BridgeMainNetConstants getInstance() {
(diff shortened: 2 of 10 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -820,8 +821,26 @@ private void emitRejectEvent(Coin value, String senderAddressStr, RejectedPegout
...
     private boolean requestRelease(Address destinationAddress, Coin value, Transaction rskTx) throws IOException {
         if (activations.isActive(RSKIP219)) {
+            int pegoutSize = getRegularPegoutTxSize(getActiveFederation());
+            Coin feePerKB = getFeePerKb();
+            // The pegout transaction has a cost related to its size and the current feePerKB
+            // The actual cost cannot be asserted exactly so the calculation is approximated
+            // On top of this, the remainder after the fee should be enough for the user to be able to operate
+            // For this, the calculation includes an additional percentage to assert for this
+            Coin requireFundsForFee = feePerKB
+                .divide(1000) // Get the s/b
```
donequis, line 831: You will ussually prefer to multiply and then divide

donequis: 1024 is a bit less confusing here, but it's just a design decision

rskj-core/src/main/java/co/rsk/config/BridgeMainNetConstants.java
```diff
@@ -144,6 +144,8 @@
 
         minSecondsBetweenCallsReceiveHeader = 300;  // 5 minutes in Seconds
         maxDepthBlockchainAccepted = 25;
+
+        percentageAboveFeeForPegouts = 20;
```
donequis, line 148: What's the meaning of this value? Sounds like if you need to pay 100 in fee you need to pegout at least 120, while the reasonable for me is that if you need to pay 100 in fees, the minimum pegout should be about 500, so the user doesn't lose that much

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
josedahlquist: The 1000 comes from the interpretation of feePerKB bitcoinj-thin has. See here: https://github.com/rsksmart/bitcoinj-thin/blob/master/src/main/java/co/rsk/bitcoinj/wallet/Wallet.java#L728
I wanted to keep the same unit 😬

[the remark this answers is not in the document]
josedahlquist: in bitcoinj-thin is 1000 😅  https://github.com/rsksmart/bitcoinj-thin/blob/master/src/main/java/co/rsk/bitcoinj/wallet/Wallet.java#L728

rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
```diff
@@ -76,6 +76,8 @@
 
     protected String oldFederationAddress;
 
+    protected int percentageAboveFeeForPegouts;
```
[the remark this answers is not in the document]
josedahlquist: liked the second one 👍

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -501,4 +502,40 @@ public static int extractAddressVersionFromBytes(byte[] addressBytes) throws Bri
         System.arraycopy(addressBytes, 1, hashBytes, 0, 20);
         return hashBytes;
     }
+
+    public static int getRegularPegoutTxSize(@Nonnull Federation federation) {
+        final int INPUT_MULTIPLIER = 2;
+        final int SIGNATURE_MULTIPLIER = 71;
+        final int OUTPUT_MULTIPLIER = 2;
+        final int OUTPUT_SIZE = 25;
```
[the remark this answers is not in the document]
josedahlquist: sure, will do!

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -820,8 +821,26 @@ private void emitRejectEvent(Coin value, String senderAddressStr, RejectedPegout
...
+                .multiply(pegoutSize); // times the size in bytes
+            requireFundsForFee = requireFundsForFee
+                .add(requireFundsForFee
+                    .times(bridgeConstants.getPercentageAboveFeeForPegouts())
+                    .divide(100)
+                ); // add the gap
+
+            // The pegout value should be greater or equals than the max of these two values
+            Coin minValue = Coin.valueOf(Math.max(bridgeConstants.getMinimumPegoutTxValueInSatoshis().value, requireFundsForFee.value));
```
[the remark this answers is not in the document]
josedahlquist: good catch, will add it

rskj-core/src/main/java/co/rsk/config/BridgeMainNetConstants.java
josedahlquist: The value will be updated accordingly before code freeze, I just defined a "mockup" value. I agree with you that 20% sounds to low

marcos-iov (dismissed):

pamgonzalez (dismissed):

donequis (dismissed):

pamgonzalez (approved):

marcos-iov (approved):

Issue #1387, opened 2020-12-07
JSON/RPC endpoint returns HTTP status 500 on error (should be 200)


saurik:
When JSON/RPC endpoints return errors, the resulting body JSON document contains the error message and the HTTP status is supposed to be 200. While I personally can see an argument for why JSON/RPC "should have" returned an HTTP error in this case, it actually doesn't make much sense in other contexts and no matter what 500 would be the wrong code to return.

```
# curl -w '%{http_code}\n' -o/dev/null -s -H 'content-type: application/json' --data '{"id":"","jsonrpc":"2.0","method":"eth_chainId","params":[2]}' https://public-node.testnet.rsk.co/
500
```
```
# curl -L --post301 -w '%{http_code}\n' -o/dev/null -s -H 'content-type: application/json' --data '{"id":"","jsonrpc":"2.0","method":"eth_chainId","params":[2]}' https://ethereumclassic.network/
200
```

patogallaiovlabs:
I'll take a look... make sense what you say, the node behaves like v1.0 spec, for v2.0 all the codes shouldn't be necessary. We can follow sthg like this: https://github.com/hyperledger/besu/issues/549

bguiz:
can we get this addressed in the next release pls?

PR #1522, opened 2021-05-05
Add custom HttpStatusCodeProvider to adjust to JSON-RPC 2.0 conventions

marcciosilva:
## Description
Since the jsonrpc4j keeps on returning internal error HTTP status codes upon JSON-RPC errors (it shouldn't, since the transport layer protocol should be agnostic to JSON-RPC), we add custom logic to handle them.

Inspired on [this](https://github.com/hyperledger/besu/issues/549), any valid JSON-RPC request should generate a JSON-RPC response, the HTTP status code should be 200. No matter if the response is a success or error response.
Calling a method that hasn't been enabled or a method that doesn't exist are valid JSON-RPC requests, therefore the expected response is a JSON-RPC error (and status code = 200)
Any invalid JSON-RPC request (malformed JSON, etc.) should generate a 400 status code.

## Motivation and Context
Issue #1387 

## How Has This Been Tested?
Unit testing

## Types of changes

## Checklist:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/rpc/netty/Web3HttpStatusCodeProvider.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/netty/Web3HttpStatusCodeProvider.java
@@ -27,7 +27,7 @@ public class Web3HttpStatusCodeProvider implements HttpStatusCodeProvider {
                 HttpResponseStatus.OK.code();
     }
 
-    private boolean isBadRequestResultCode(int resultCode) {
+    private static boolean isBadRequestResultCode(int resultCode) {
         return BAD_REQUEST_JSON_ERRORS.contains(resultCode);
     }
 
--- a/rskj-core/src/main/java/co/rsk/rpc/netty/Web3HttpStatusCodeProvider.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/netty/Web3HttpStatusCodeProvider.java
@@ -1,0 +1,48 @@
+package co.rsk.rpc.netty;
+
+import com.googlecode.jsonrpc4j.HttpStatusCodeProvider;
+import io.netty.handler.codec.http.HttpResponseStatus;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+import static com.googlecode.jsonrpc4j.ErrorResolver.JsonError;
+import static com.googlecode.jsonrpc4j.ErrorResolver.JsonError.*;
+
+public class Web3HttpStatusCodeProvider implements HttpStatusCodeProvider {
+
+    private static final List<JsonError> BAD_REQUEST_JSON_ERRORS = Arrays.asList(
+            METHOD_PARAMS_INVALID, INVALID_REQUEST, PARSE_ERROR
+    );
+
+    /**
+     * Only invalid JSON-RPC requests (e.g. a malformed JSON) qualify for a Bad Request HTTP status.
+     * In any other case, the HTTP protocol remains entirely independent from JSON-RPC (since its 2.0 version).
+     * Reference: https://www.jsonrpc.org/
+     */
+    @Override
+    public int getHttpStatusCode(int resultCode) {
+        return isBadRequestResultCode(resultCode) ?
+                HttpResponseStatus.BAD_REQUEST.code() :
+                HttpResponseStatus.OK.code();
+    }
+
+    private boolean isBadRequestResultCode(int resultCode) {
+        Optional<JsonError> optionalJsonError = BAD_REQUEST_JSON_ERRORS
+                .stream()
+                .filter(jsonError -> jsonError.code == resultCode)
+                .findFirst();
+        return optionalJsonError.isPresent();
+    }
+
+    /**
+     * The default implementation wrongly assumes that an HTTP status code is mapped to a single JSON RPC error - this
+     * is not always the case. Either way, this is not used in the current project.
+     */
+    @Override
+    public Integer getJsonRpcCode(int httpStatusCode) {
+        throw new UnsupportedOperationException("Cannot possibly determine a single JSON RPC from an HTTP status code");
+    }
+
+}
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/co/rsk/rpc/netty/Web3HttpStatusCodeProvider.java
```diff
@@ -0,0 +1,48 @@
...
+import io.netty.handler.codec.http.HttpResponseStatus;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+import static com.googlecode.jsonrpc4j.ErrorResolver.JsonError;
+import static com.googlecode.jsonrpc4j.ErrorResolver.JsonError.*;
+
+public class Web3HttpStatusCodeProvider implements HttpStatusCodeProvider {
```
marcciosilva, line 13: Based on [this class](https://github.com/briandilley/jsonrpc4j/blob/master/src/main/java/com/googlecode/jsonrpc4j/DefaultHttpStatusCodeProvider.java) (which we used prior to this change).
FYI: version 1.6 of jsonrpc4j doesn't adjust error handling to the JSON-RPC guidelines, that's why I decided not to update it this PR.

rskj-core/src/main/java/co/rsk/rpc/netty/Web3HttpStatusCodeProvider.java
```diff
@@ -0,0 +1,48 @@
...
+        return optionalJsonError.isPresent();
+    }
+
+    /**
+     * The default implementation wrongly assumes that an HTTP status code is mapped to a single JSON RPC error - this
+     * is not always the case. Either way, this is not used in the current project.
+     */
+    @Override
+    public Integer getJsonRpcCode(int httpStatusCode) {
+        throw new UnsupportedOperationException("Cannot possibly determine a single JSON RPC from an HTTP status code");
```
marcciosilva, line 45: jsonrpc4j maps each supported HTTP status code to a single JSON-RPC code - this is not entirely correct in my opinion, since a single HTTP code could map to N JSON-RPC codes (this even happens in jsonrpc4j).
Since we don't use this method in rskj, I decided to leave it like this.

marcciosilva:
pipeline: run

fedejinich:
pipeline:run

rskj-core/src/main/java/co/rsk/rpc/netty/Web3HttpStatusCodeProvider.java
ajlopezrsk: Take into account that a JSON RPC request could be an array of valid JSON RPC requests. Is this case covered in the process of responses/responses code?

rskj-core/src/main/java/co/rsk/rpc/netty/Web3HttpStatusCodeProvider.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    @Override
+    public int getHttpStatusCode(int resultCode) {
+        return isBadRequestResultCode(resultCode) ?
+                HttpResponseStatus.BAD_REQUEST.code() :
+                HttpResponseStatus.OK.code();
+    }
+
+    private boolean isBadRequestResultCode(int resultCode) {
+        Optional<JsonError> optionalJsonError = BAD_REQUEST_JSON_ERRORS
```
ajlopezrsk, line 32: Why not a simple contains? Or a map of JSON ERROR to boolean (true == it is a bad request result code)?

rskj-core/src/main/java/co/rsk/rpc/netty/Web3HttpStatusCodeProvider.java
marcciosilva: Since I just swapped the `DefaultHttpStatusCodeProvider` (from jsonrpc4j library) for a custom implementation that adheres to the same signatures (but mapping HTTP codes differently), the actual processing of requests is not involved (this is only used in the response handling part of the pipeline).

rskj-core/src/main/java/co/rsk/rpc/netty/Web3HttpStatusCodeProvider.java
marcciosilva: I didn't use a `contains` because it's a list of `JsonError` and what I receive is an integer (somehow I have to reach the `code` member variable).
Let me know what you think about the change I just uploaded (instead of using a JsonError list y directly use a collection of the codes, then I just use a `contains`).

nagarev:
pipeline:run

rskj-core/src/main/java/co/rsk/rpc/netty/Web3ResultHttpResponseHandler.java
```diff
@@ -17,23 +17,30 @@
...
 import io.netty.handler.codec.http.HttpVersion;
+
 import static io.netty.handler.codec.http.HttpHeaders.Names.CONTENT_TYPE;
 import static io.netty.handler.codec.http.HttpHeaders.Values.APPLICATION_JSON;
 
 public class Web3ResultHttpResponseHandler extends SimpleChannelInboundHandler<Web3Result> {
 
+    private HttpStatusCodeProvider httpStatusCodeProvider;
+
+    public Web3ResultHttpResponseHandler(HttpStatusCodeProvider httpStatusCodeProvider) {
```
Vovchyk, line 35: I guess using this 3rd-party interface is perfectly fine here, but as an option it could also be `IntFunction<HttpResponseStatus> errorCodeToHttpStatusMapper`

Vovchyk (dismissed):
lgtm

rskj-core/src/main/java/co/rsk/rpc/netty/Web3ResultHttpResponseHandler.java
marcciosilva: I think it's an interesting suggestion, but considering that we already depend on many classes of com.googlecode.jsonrpc4 for JSON-RPC handling, I think it'd be overkill (besides, I guess rskj shouldn't have the responsibility of defining stuff related to this). Is it ok if I leave it like this?

Vovchyk (approved):

fedejinich (dismissed):
lgtm! 
It'd be nice to cover it also in the integration tests

marcciosilva:
pipeline: run

Vovchyk:
pipeline: run

marcciosilva:
pipeline: run

marcciosilva:
pipeline:run

marcciosilva:
pipeline:run

nagarev:
pipeline:run

PR #1523, opened 2021-05-06
Memory management improvements in import sync

Vovchyk:
## Description
This change reduces amount of RAM required for the `--import sync` feature.

The idea is to release memory that's not needed anymore during the importing process as soon as possible. The idea also is to use `LinkedList`'s instead of `ArrayList`'s during the process to free already processed items (cause some lists may hold millions of entries).

## Motivation and Context
Right now, while starting the node with the `--import` flag, it crashes with OoM error. Java option `-Xmx8g` can be used to mitigate this issue a bit (this simply increases available ram for a java process to 8GBs).

With the change in this PR, min required RAM can be decreased to 4GBs, so the node can import data with the help of the `-Xmx4g` flag.

## How Has This Been Tested?
By running the node via the following commands:
- `java -Xmx4g -jar ./rskj-core/build/libs/rskj-core-3.0.0-SNAPSHOT-all.jar --testnet --import`
- `java -Xmx4g -jar ./rskj-core/build/libs/rskj-core-3.0.0-SNAPSHOT-all.jar --main --import`

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/db/importer/BootstrapImporter.java
+++ b/rskj-core/src/main/java/co/rsk/db/importer/BootstrapImporter.java
@@ -33,9 +33,9 @@ import org.ethereum.util.RLPElement;
 import org.ethereum.util.RLPList;
 
 import java.math.BigInteger;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.LinkedList;
 import java.util.Objects;
+import java.util.Queue;
 
 public class BootstrapImporter {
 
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/trie/Trie.java
```diff
@@ -485,7 +486,7 @@ public Trie delete(String key) {
             internalToMessage();
         }
 
-        return cloneArray(encoded);
+        return encoded;
```
ajlopezrsk, line 489: Is this change really needed for the purpose of this pull request? Yes, cloneArray is a bit of "defensive programming" but really aisle the external provided value from any modification.

Vovchyk: yes, it creates a lot of arrays' copies during manipulations with a trie. Without this change the import process fails with OoM (with 4gb of ram).

Also we don't defend underlying data / array in other classes, e.g. `Keccak256` and `ByteArrayWrapper`. At the end, this is a tradeoff.

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

rskj-core/src/main/java/co/rsk/db/importer/BootstrapImporter.java
```diff
@@ -102,10 +81,58 @@ private static void insertBlocks(BlockStore blockStore,
...
+
+            destinationTrieStore.save(trie);
+        }
+    }
+
+    private static Queue<RLPElement> decodeQueue(byte[] data) {
+        RLPList rlpList = RLP.decodeList(data);
+        int size = rlpList.size();
+
+        LinkedList<RLPElement> result = new LinkedList<>();
```
nagarev, line 131: Can we make this variable of type `Queue<RLPElement>`? Maybe I'm missing something

Vovchyk: Not sure I got the point here. `Queue` is an interface and `LinkedList` is one of its implementations.

nagarev (approved):
LGTM

rskj-core/src/main/java/co/rsk/db/importer/BootstrapImporter.java
Vovchyk: ahh.. I see)

Vovchyk: Done

nagarev (approved):
LGTM

rskj-core/src/main/java/co/rsk/db/importer/BootstrapImporter.java
```diff
@@ -102,10 +81,58 @@ private static void insertBlocks(BlockStore blockStore,
...
+        prepareNodeValues(hashMapDB, RLP.decodeList(Objects.requireNonNull(nodeListQueue.poll()).getRLPData()));
+    }
+
+    private static void prepareNodeData(Queue<byte[]> rlpDataQueue, HashMapDB hashMapDB, RLPList nodesData) {
+        int size = nodesData.size();
+        for (int k = 0; k < size; k++) {
+            RLPElement element = nodesData.get(k);
+            byte[] rlpData = Objects.requireNonNull(element.getRLPData());
+
+            hashMapDB.put(Keccak256Helper.keccak256(rlpData), rlpData);
```
joaquinlpereyra, line 104: Previously, the `fakeDB` had as keys the `.getHash()` of the trie, which passes through quite a bit of serialization and changes, instead of directly hashing the `rlpData`. Are they equivalent somehow? I may be missing something but I think this could mess up other methods such as `TrieStoreImpl.save()`, as `.getHash()` of the fakeStore is used there...

TL;DR: The fake store and the _actual_ store use different keys, which may end up impacting the actual store as the actual store relies on the fake store `getHash` method.

PR #1536, opened 2021-05-17
Moved Methods From BridgeSupport To FederationSupport

kelvinator07:


rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestPowerMock.java
```diff
@@ -1446,6 +1447,7 @@ public void registerBtcTransactionMigrationTx() throws BlockStoreException, Addr
     }
 
     @Test
+    @Ignore
```
pamgonzalez, line 1450: Do we have a story to modify this test case that are ignored now? 
Are you included new test cases for the function that you are moved? 
Remember not use Power mock for that. Thank you

kelvinator07: No Story, But I'm Working On The New Test Cases
Thanks Pam

Issue #1531, opened 2021-05-12
JSON RPC 500 internal error data null


pmprete:
RSK does not return the reason of the error when it's an invalid field and the returned error to the client does not explain anything besides 500 internal error.

I was lucky to have access to the RSK node where i can see the following message

`--
2021-05-12-14:43:23.268 ERROR [web3]  JsonRPC error when for method public default java.lang.String co.rsk.rpc.Web3EthModule.eth_call(org.ethereum.rpc
.Web3$CallArguments,java.lang.String) with arguments [{"value":"0x1c6bf526340000","form":"0x27598400A96D4EE85f86b0931e49cBc02adD6dF0","data":"0xaae40a
2a0000000000000000000000008963f840e6b110c9821b42222990f15e1d6488dc0000000000000000000000004664d4cbd5104a0e974354724cbc8e0d9bd1aca3","to":"0xc19f0882bf
318c9f8767c7d520018888e878417b"}, "latest"]
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "form" (class org.ethereum.rpc.Web3$CallArguments), not marked as
 ignorable (7 known properties: "gasPrice", "value", "gas", "from", "to", "nonce", "data"])
 at [Source: N/A; line: -1, column: -1] (through reference chain: org.ethereum.rpc.Web3$CallArguments["form"])
at com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException.from(UnrecognizedPropertyException.java:62)
at com.fasterxml.jackson.databind.DeserializationContext.handleUnknownProperty(DeserializationContext.java:834)
at com.fasterxml.jackson.databind.deser.std.StdDeserializer.handleUnknownProperty(StdDeserializer.java:1093)
--`

We should transmit the exception in the JSON RPC response

> Unrecognized field "form" (class org.ethereum.rpc.Web3$CallArguments), not marked as
 ignorable (7 known properties: "gasPrice", "value", "gas", "from", "to", "nonce", "data"])
 
 So the user can know the error

PR #1544, opened 2021-05-25
Adding error handler for invalid fields

nagarev:
When doing a JSON RPC request, if an invalid field is sent, then only an internal error is returned.

## Description

When a field is invalid a description of the problem is returned Instead of returning and Internal Error without description, e.g., when a "form" field is sent instead of "from", the following message is returned:

> Unrecognized field "form" (8 known properties: ["gasPrice", "value", "gas", "from", "to", "chainId", "nonce", "data"])

## Motivation and Context

Closes https://github.com/rsksmart/rskj/issues/1531

## How Has This Been Tested?

By using Unit Tests and Integration Tests

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
@@ -1,5 +1,6 @@
 package org.ethereum.rpc.exception;
 
+import co.rsk.jsonrpc.JsonRpcError;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.exc.InvalidFormatException;
 import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
@@ -9,6 +10,7 @@ import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Method;
 import java.util.Arrays;
+import java.util.Iterator;
 import java.util.List;
 
 /**
@@ -24,22 +26,40 @@ public class RskErrorResolver implements ErrorResolver {
         if(t instanceof  RskJsonRpcRequestException) {
             error =  new JsonError(((RskJsonRpcRequestException) t).getCode(), t.getMessage(), null);
         } else if (t instanceof InvalidFormatException) {
-            error = new JsonError(-32603, "Internal server error, probably due to invalid parameter type", null);
+            error = new JsonError(JsonRpcError.INTERNAL_ERROR, "Internal server error, probably due to invalid parameter type", null);
         } else if (t instanceof UnrecognizedPropertyException) {
-            String message = getExceptionMessage((UnrecognizedPropertyException) t);
-            logger.error("JsonRPC error {} for method {} with arguments {}", message, method, arguments);
-            error = new JsonError(-32603, message, null);
+            error = new JsonError(
+                    JsonRpcError.INVALID_PARAMS,
+                    getExceptionMessage((UnrecognizedPropertyException) t),
+                    null);
         } else {
             logger.error("JsonRPC error when for method {} with arguments {}", method, arguments, t);
-            error = new JsonError(-32603, "Internal server error", null);
+            error = new JsonError(JsonRpcError.INTERNAL_ERROR, "Internal server error", null);
         }
         return error;
     }
 
     private String getExceptionMessage(UnrecognizedPropertyException ex) {
-        if (ex.getMessage() == null) return "Invalid parameters";
-        return Arrays.stream(ex.getMessage().split("\\n at "))
-                .findFirst()
-                .orElse("Invalid parameters");
+        if (ex.getPropertyName() == null || ex.getKnownPropertyIds() == null) return "Invalid parameters";
+
+        StringBuilder stringBuilder = new StringBuilder("Unrecognized field \"");
+        stringBuilder.append(ex.getPropertyName());
+        stringBuilder.append("\" (");
+        stringBuilder.append(ex.getKnownPropertyIds().size());
+        stringBuilder.append(" known properties: [");
+
+        Iterator<Object> iterator = ex.getKnownPropertyIds().iterator();
+        while (iterator.hasNext()) {
+            stringBuilder.append("\"");
+            stringBuilder.append(iterator.next());
+            stringBuilder.append("\"");
+            if (iterator.hasNext()) {
+                stringBuilder.append(", ");
+            }
+        }
+
+        stringBuilder.append("])");
+
+        return stringBuilder.toString();
     }
 }
--- a/rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
@@ -2,11 +2,13 @@ package org.ethereum.rpc.exception;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.exc.InvalidFormatException;
+import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
 import com.googlecode.jsonrpc4j.ErrorResolver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.List;
 
 /**
@@ -23,10 +25,21 @@ public class RskErrorResolver implements ErrorResolver {
             error =  new JsonError(((RskJsonRpcRequestException) t).getCode(), t.getMessage(), null);
         } else if (t instanceof InvalidFormatException) {
             error = new JsonError(-32603, "Internal server error, probably due to invalid parameter type", null);
+        } else if (t instanceof UnrecognizedPropertyException) {
+            String message = getExceptionMessage((UnrecognizedPropertyException) t);
+            logger.error("JsonRPC error {} for method {} with arguments {}", message, method, arguments);
+            error = new JsonError(-32603, message, null);
         } else {
             logger.error("JsonRPC error when for method {} with arguments {}", method, arguments, t);
             error = new JsonError(-32603, "Internal server error", null);
         }
         return error;
+    }
+
+    private String getExceptionMessage(UnrecognizedPropertyException ex) {
+        if (ex.getMessage() == null) return "Invalid parameters";
+        return Arrays.stream(ex.getMessage().split("\\n at "))
+                .findFirst()
+                .orElse("Invalid parameters");
     }
 }
--- a/rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
@@ -9,7 +9,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Method;
-import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 
--- a/rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
@@ -38,7 +38,7 @@ public class RskErrorResolver implements ErrorResolver {
         return error;
     }
 
-    private String getExceptionMessage(UnrecognizedPropertyException ex) {
+    private static String getExceptionMessage(UnrecognizedPropertyException ex) {
         if (ex.getPropertyName() == null || ex.getKnownPropertyIds() == null) return "Invalid parameters";
 
         StringBuilder stringBuilder = new StringBuilder("Unrecognized field \"");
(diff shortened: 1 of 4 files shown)


nagarev:
pipeline:run

pmprete (approved):
Good Work! 💯

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
```diff
@@ -23,10 +25,21 @@ public JsonError resolveError(Throwable t, Method method, List<JsonNode> argumen
             error =  new JsonError(((RskJsonRpcRequestException) t).getCode(), t.getMessage(), null);
         } else if (t instanceof InvalidFormatException) {
             error = new JsonError(-32603, "Internal server error, probably due to invalid parameter type", null);
+        } else if (t instanceof UnrecognizedPropertyException) {
+            String message = getExceptionMessage((UnrecognizedPropertyException) t);
+            logger.error("JsonRPC error {} for method {} with arguments {}", message, method, arguments);
+            error = new JsonError(-32603, message, null);
```
marcciosilva, line 31: Maybe we could use a constant for this value? Since it's used several times in this file (and others)

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
```diff
@@ -23,10 +25,21 @@ public JsonError resolveError(Throwable t, Method method, List<JsonNode> argumen
...
         } else {
             logger.error("JsonRPC error when for method {} with arguments {}", method, arguments, t);
             error = new JsonError(-32603, "Internal server error", null);
         }
         return error;
     }
+
+    private String getExceptionMessage(UnrecognizedPropertyException ex) {
```
marcciosilva, line 39: Maybe we can just use Throwable (which already provides the `getMessage` method) and avoid the cast?

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
```diff
@@ -23,10 +25,21 @@ public JsonError resolveError(Throwable t, Method method, List<JsonNode> argumen
...
+        if (ex.getMessage() == null) return "Invalid parameters";
```
marcciosilva, line 40: Maybe we can use a constant for "Invalid parameters", since it's used more than once (and it's being created twice for every call to this method)

rskj-core/src/test/java/org/ethereum/rpc/exception/RskErrorResolverTest.java
```diff
@@ -0,0 +1,128 @@
...
+    private RskErrorResolver rskErrorResolver;
+
+    @Before
+    public void setup() {
+        rskErrorResolver = new RskErrorResolver();
+    }
+
+    @Test
+    public void test_resolveError_givenRskJsonRpcRequestException_returnsJsonErrorAsExpected() throws NoSuchMethodException {
+        // Given
```
marcciosilva, line 33: Love the Given / When / Then comments ❤️

rskj-core/src/test/java/org/ethereum/rpc/exception/RskErrorResolverTest.java
```diff
@@ -0,0 +1,128 @@
...
+        JsonError result = rskErrorResolver.resolveError(exception, methodMock, jsonNodeListMock);
+
+        // Then
+        Assert.assertNotNull(result);
+        Assert.assertEquals(-32603, result.code);
+        Assert.assertEquals("Internal server error", result.message);
+        Assert.assertNull(result.data);
+    }
+
+    public void mockMethod() { }
```
marcciosilva, line 126: Maybe I'd make this private (unless it's necessary because of reflection magics)

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
nagarev: I like it! 🙌🏼

rskj-core/src/test/java/org/ethereum/rpc/exception/RskErrorResolverTest.java
nagarev: In this case it is necessary so it is easier to use, will try a couple of other options and let you know.

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
```diff
@@ -23,10 +25,21 @@ public JsonError resolveError(Throwable t, Method method, List<JsonNode> argumen
...
+        return Arrays.stream(ex.getMessage().split("\\n at "))
```
Vovchyk, line 41: as an alternative we may use `getPropertyName()` of `UnrecognizedPropertyException` to compose the message (instead of parsing)

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
```diff
@@ -23,10 +25,21 @@ public JsonError resolveError(Throwable t, Method method, List<JsonNode> argumen
             error =  new JsonError(((RskJsonRpcRequestException) t).getCode(), t.getMessage(), null);
         } else if (t instanceof InvalidFormatException) {
             error = new JsonError(-32603, "Internal server error, probably due to invalid parameter type", null);
+        } else if (t instanceof UnrecognizedPropertyException) {
+            String message = getExceptionMessage((UnrecognizedPropertyException) t);
+            logger.error("JsonRPC error {} for method {} with arguments {}", message, method, arguments);
```
Vovchyk, line 30: do we need logging an error here?

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
nagarev: Yes, I don't really like the exception message parsing, the problem is that the idea is to show a nice descriptive message (for example, including the allowed fields). I will see what can be done regarding that and update it accordingly.

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
nagarev: Will remove it.

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
nagarev: Done!

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
nagarev: In the end I've opted for doing this cast since I need the exception to extract some data.

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
nagarev: this is not needed anymore.

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
nagarev: This idea of @Vovchyk was a nice one, so I've updated the message, now it is built using more the exception information: not only the property name but also the "known fields".

So, instead of showing this message:

`Unrecognized field "form" (class org.ethereum.rpc.Web3$CallArguments), not marked as ignorable (8 known properties: "gasPrice", "value", "gas", "from", "to", "chainId", "nonce", "data"])`

We show this:

`Unrecognized field "form" (8 known properties: ["gasPrice", "value", "gas", "from", "to", "chainId", "nonce", "data"])`

It is a subtle difference I think, but we hide internal details about the implementation and also fix an issue of the exception where the opening bracket ("[") was missing from the message.

Also, **some important thing**: now we are returning `-32602` (Invalid Parameters) instead of `-32603` (Internal Error).

@pmprete how do you see it know? Do you like the new message? And how about changing the status code? Thanks in advance!

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
nagarev: Done!

nagarev:
pipeline:run

rskj-core/src/test/java/org/ethereum/rpc/exception/RskErrorResolverTest.java
nagarev: 🙌🏼

rskj-core/src/test/java/org/ethereum/rpc/exception/RskErrorResolverTest.java
```diff
@@ -80,17 +81,65 @@ public void test_resolveError_givenUnrecognizedPropertyException_nullExceptionMe
 
         // Then
         Assert.assertNotNull(result);
-        Assert.assertEquals(-32603, result.code);
+        Assert.assertEquals(-32602, result.code);
```
marcciosilva, line 84: Maybe use the constant here? Or is it hardcoded so that if the constant changes we have to update the test?

marcciosilva: This comment applies for all of the usages below

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
```diff
@@ -24,22 +25,40 @@ public JsonError resolveError(Throwable t, Method method, List<JsonNode> argumen
         if(t instanceof  RskJsonRpcRequestException) {
             error =  new JsonError(((RskJsonRpcRequestException) t).getCode(), t.getMessage(), null);
         } else if (t instanceof InvalidFormatException) {
-            error = new JsonError(-32603, "Internal server error, probably due to invalid parameter type", null);
+            error = new JsonError(JsonRpcError.INTERNAL_ERROR, "Internal server error, probably due to invalid parameter type", null);
```
marcciosilva, line 28: Maybe use `JsonRpcInternalError` directly?

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
```diff
@@ -24,22 +25,40 @@ public JsonError resolveError(Throwable t, Method method, List<JsonNode> argumen
...
         } else if (t instanceof UnrecognizedPropertyException) {
-            String message = getExceptionMessage((UnrecognizedPropertyException) t);
-            logger.error("JsonRPC error {} for method {} with arguments {}", message, method, arguments);
-            error = new JsonError(-32603, message, null);
+            error = new JsonError(
```
marcciosilva, line 30: Maybe create a `JsonRpcInvalidParamsError` to match `JsonRpcInternalError`? Maybe it's pointless

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
```diff
@@ -22,11 +25,40 @@ public JsonError resolveError(Throwable t, Method method, List<JsonNode> argumen
...
+    private String getExceptionMessage(UnrecognizedPropertyException ex) {
+        if (ex.getPropertyName() == null || ex.getKnownPropertyIds() == null) return "Invalid parameters";
+
+        StringBuilder stringBuilder = new StringBuilder("Unrecognized field \"");
+        stringBuilder.append(ex.getPropertyName());
+        stringBuilder.append("\" (");
+        stringBuilder.append(ex.getKnownPropertyIds().size());
+        stringBuilder.append(" known properties: [");
+
+        Iterator<Object> iterator = ex.getKnownPropertyIds().iterator();
```
marcciosilva, line 50: Maybe something like this to avoid iterating with the iterator?
It's just a suggestion though, perhaps this is harder to read

```java
ex.getKnownPropertyIds().stream()
    .map(s -> String.format("\"%s\"", s))
    .joining(", ")
```

marcciosilva: Or maybe the stringBuilder is more performant (most likely)

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
```diff
@@ -22,11 +25,40 @@ public JsonError resolveError(Throwable t, Method method, List<JsonNode> argumen
...
+                    null);
         } else {
             logger.error("JsonRPC error when for method {} with arguments {}", method, arguments, t);
-            error = new JsonError(-32603, "Internal server error", null);
+            error = new JsonError(JsonRpcError.INTERNAL_ERROR, "Internal server error", null);
         }
         return error;
     }
+
+    private String getExceptionMessage(UnrecognizedPropertyException ex) {
```
Vovchyk, line 41: `static` ?

Vovchyk (approved):
LGTM 👍

rskj-core/src/main/java/org/ethereum/rpc/exception/RskErrorResolver.java
nagarev: Will do it!

nagarev: Done!

Vovchyk (approved):

PR #1559, opened 2021-06-14
Implemented Class To Handle xpub To Compressed Format

kelvinator07:


rskj-core/src/main/java/co/rsk/PublicKeyUtils.java
```diff
@@ -0,0 +1,40 @@
...
+ * This class contains methods for converting an Extended Public Key To The Compressed Format
+ *
+ * @author Kelvin Isievwore
+ */
+public class PublicKeyUtils {
+
+    public static String convertXpubToCompressedPublicKey(String xpub) {
+
+        // Decode from Base58Check
+        byte[] decodedXpub = decodeFromBase58Check(xpub);
```
marcos-iov, line 18:
```suggestion
        byte[] decodedXpub = Base58.decodeChecked(xpub);
```
We could simply call Base58 class here instead of creating a new function

rskj-core/src/main/java/co/rsk/PublicKeyUtils.java
```diff
@@ -0,0 +1,40 @@
...
+
+        // Grab last 33 bytes
+        byte[] last33Bytes = Arrays.copyOfRange(decodedXpub, decodedXpub.length - 33, decodedXpub.length);
+
+        return toHexString(last33Bytes);
```
marcos-iov, line 23:
```suggestion
        return Hex.toHexString(last33Bytes);
```
Same here

rskj-core/src/main/java/co/rsk/PublicKeyUtils.java
```diff
@@ -0,0 +1,40 @@
...
+    private static byte[] decodeFromBase58Check(String input) {
+        return Base58.decodeChecked(input);
+    }
+
+    private static String toHexString(byte[] input) {
+        return Hex.toHexString(input);
+    }
+
+    public static void main(String[] args) {
+        String xPub = "xpub661MyMwAqRbcG8Zah6TcX3QpP5yJApaXcyLK8CJcZkuYjczivsHxVL5qm9cw8BYLYehgFeddK5WrxhntpcvqJKTVg96dUVL9P7hZ7Kcvqvd";
```
marcos-iov, line 35:
```suggestion
        String xPub = "";
```
Maybe it's best if we leave this field blank for users to complete

rskj-core/src/main/java/co/rsk/PublicKeyUtils.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private static byte[] decodeFromBase58Check(String input) {
+        return Base58.decodeChecked(input);
+    }
+
+    private static String toHexString(byte[] input) {
+        return Hex.toHexString(input);
+    }
+
+    public static void main(String[] args) {
```
marcos-iov, line 34: I'd move the main function to the top of the file, at least for me it makes it easier to use

Issue #1549, opened 2021-06-03
Semantic confusion in Trie.GetNode() could lead to blockchain fork


SergioDemianLerner:
Inspired in https://www.paradigm.xyz/2021/05/booby-trapping-the-ethereum-blockchain/, I found a bug (and potential vulnerability) this method in NodeReference.java:

```
    public Optional<Trie> getNode() {
        if (lazyNode != null) {
            return Optional.of(lazyNode);
        }

        if (lazyHash == null) {
            return Optional.empty();
        }

        Optional<Trie> node = store.retrieve(lazyHash.getBytes());

        lazyNode = node.orElse(null);
        return node;
    }
```
    
When the node cannot be retrieved from storage, it returns an optional empty trie.
However, the inexistence of a node in the database is NOT a a normal situation, but a fatal exception that should shutdown the node as soon as possible. Anything you do in this situation can be used to fork the network.
Returning an optional trie is NOT ok, because many parts of the code assume that if GetNode().isPresent() == false, then this is because the node has no children, NOT because the database is broken.
For example  
```
private Trie put(TrieKeySlice key, byte[] value, boolean isRecursiveDelete) {
....
        Optional<Trie> leftOpt = trie.left.getNode();
        Optional<Trie> rightOpt = trie.right.getNode();
        if (leftOpt.isPresent() && rightOpt.isPresent()) {
            return trie;
        }
        
```
As you see, in Trie.put() getNode().isPresent() is used to test if the node has a child.

The semantic problem is that isPresent() is confused with isPresentInTheDatabase(), but it really means doesTheChildNodeExistsInTheTrie() ?

GetNode should be modified to something like this:

```
 public Optional<Trie> getNode() {
        if (lazyNode != null) {
            return Optional.of(lazyNode);
        }

        if (lazyHash == null) {
            return Optional.empty();
        }

        Optional<Trie> node = store.retrieve(lazyHash.getBytes());

        //// --->> begin new
        if (!node.isPresent()) {
          throw new RuntimeException("DATABASE BROKEN do not continue");
        }
        /// <<<- end new
        lazyNode = node.orElse(null);
        return node;
    }
```

While database corruption is not something normal, there are several protocols that, if implemented incorrectly, could lead to missing key/values in the database. For example, the garbage collector, or future light client sync or fast sync protocols.

joaquinlpereyra:
[...]

PR #1567, opened 2021-06-24
Fixing semantic confusion

nagarev:
Now we throw and exception when a node can not be retrieved from database.

## Description

Fixing the bug reported by @SergioDemianLerner  on issue [#1549](https://github.com/rsksmart/rskj/issues/1549).

## Motivation and Context

Extracted from the issue "When the node cannot be retrieved from storage, it returns an optional empty trie.
However, the inexistence of a node in the database is NOT a a normal situation, but a fatal exception that should shutdown the node as soon as possible. Anything you do in this situation can be used to fork the network."

## How Has This Been Tested?

Unit Tests

## Types of changes

## Checklist:

* **Other information**:

fixes https://github.com/rsksmart/rskj/issues/1549

### DIFF
--- a/rskj-core/src/main/java/co/rsk/trie/NodeReference.java
+++ b/rskj-core/src/main/java/co/rsk/trie/NodeReference.java
@@ -19,22 +19,29 @@ package co.rsk.trie;
 
 import co.rsk.core.types.ints.Uint8;
 import co.rsk.crypto.Keccak256;
+import co.rsk.util.NodeStopper;
 import org.ethereum.crypto.Keccak256Helper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.nio.ByteBuffer;
+import java.util.Objects;
 import java.util.Optional;
 
 public class NodeReference {
 
+    private static final Logger logger = LoggerFactory.getLogger(NodeReference.class);
     private static final NodeReference EMPTY = new NodeReference(null, null, null);
 
     private final TrieStore store;
+    private final NodeStopper nodeStopper;
 
     private Trie lazyNode;
     private Keccak256 lazyHash;
 
-    public NodeReference(TrieStore store, @Nullable Trie node, @Nullable Keccak256 hash) {
+    public NodeReference(TrieStore store, @Nullable Trie node, @Nullable Keccak256 hash, @Nonnull NodeStopper nodeStopper) {
         this.store = store;
         if (node != null && node.isEmptyTrie()) {
             this.lazyNode = null;
@@ -43,6 +50,12 @@ public class NodeReference {
             this.lazyNode = node;
             this.lazyHash = hash;
         }
+
+        this.nodeStopper = Objects.requireNonNull(nodeStopper);
+    }
+
+    public NodeReference(TrieStore store, @Nullable Trie node, @Nullable Keccak256 hash) {
+        this(store, node, hash, exitStatus -> System.exit(1));
     }
 
     public boolean isEmpty() {
@@ -52,6 +65,7 @@ public class NodeReference {
     /**
      * The node or empty if this is an empty reference.
      * If the node is not present but its hash is known, it will be retrieved from the store.
+     * If the node could not be retrieved from the store, the Node is stopped using System.exit(1)
      */
     public Optional<Trie> getNode() {
         if (lazyNode != null) {
@@ -64,7 +78,15 @@ public class NodeReference {
 
         Optional<Trie> node = store.retrieve(lazyHash.getBytes());
 
-        lazyNode = node.orElse(null);
+        // Broken database, can't continue
+        if (!node.isPresent()) {
+            logger.error("Broken database, execution can't continue");
+            nodeStopper.stop(1);
+            return Optional.empty();
+        }
+
+        lazyNode = node.get();
+
         return node;
     }
 
--- a/rskj-core/src/test/java/co/rsk/mine/MinerUtilsTest.java
+++ b/rskj-core/src/test/java/co/rsk/mine/MinerUtilsTest.java
@@ -28,15 +28,27 @@ import org.ethereum.core.Repository;
 import org.ethereum.core.Transaction;
 import org.ethereum.util.ByteUtil;
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mockito;
 
 import java.math.BigInteger;
 import java.util.*;
 
+import static org.junit.Assert.fail;
+import static org.mockito.Mockito.verify;
+
 public class MinerUtilsTest {
 
+    private static final Coin ONE_COIN = Coin.valueOf(1L);
+
     private final TestSystemProperties config = new TestSystemProperties();
+    private MinerUtils minerUtils;
+
+    @Before
+    public void setup() {
+        minerUtils = new MinerUtils();
+    }
 
     @Test
     public void getAllTransactionsTest() {
@@ -69,7 +81,7 @@ public class MinerUtilsTest {
 
         Mockito.when(transactionPool.getPendingTransactions()).thenReturn(txs);
 
-        List<Transaction> res = new MinerUtils().getAllTransactions(transactionPool);
+        List<Transaction> res = minerUtils.getAllTransactions(transactionPool);
 
         Assert.assertEquals(2, res.size());
     }
@@ -83,9 +95,8 @@ public class MinerUtilsTest {
         Map<RskAddress, BigInteger> accountNounces = new HashMap();
         Repository repository = Mockito.mock(Repository.class);
         Mockito.when(repository.getNonce(tx.getSender())).thenReturn(BigInteger.valueOf(0));
-        Coin minGasPrice = Coin.valueOf(1L);
 
-        List<Transaction> res = new MinerUtils().filterTransactions(new LinkedList<>(), txs, accountNounces, repository, minGasPrice);
+        List<Transaction> res = minerUtils.filterTransactions(new LinkedList<>(), txs, accountNounces, repository, ONE_COIN);
         Assert.assertEquals(1, res.size());
     }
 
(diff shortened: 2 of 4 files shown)


rskj-core/src/main/java/co/rsk/mine/MinerUtils.java
```diff
@@ -202,6 +203,9 @@
                 accountNonces.put(txSender, txNonce);
 
                 logger.debug("Accepted tx={} sender: {} value: {} nonce: {}", hash, txSender, txValue, txNonce);
+            } catch (NodeReferenceException nre) {
```
Vovchyk, line 206: Cannot spot this.. What method in this try-catch block can throw `NodeReferenceException`?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/mine/MinerUtils.java
Vovchyk: see it now) probably `originalRepo.getNonce`

marcciosilva, line 206: we mostly use `e` or `ex` for naming exceptions - maybe I'd mantain that convention

rskj-core/src/main/java/co/rsk/trie/NodeReferenceException.java
```diff
@@ -0,0 +1,9 @@
+package co.rsk.trie;
+
+public class NodeReferenceException extends RuntimeException {
+
+    public NodeReferenceException(String s) {
```
marcciosilva, line 5: I'd rename `s` to `message`

rskj-core/src/test/java/co/rsk/mine/MinerUtilsTest.java
```diff
@@ -288,4 +293,30 @@ public void getAllTransactionsCheckOrderTest() {
         Assert.assertEquals(res.get(7).getGasPrice(), Coin.valueOf(100));
         Assert.assertEquals(res.get(8).getGasPrice(), Coin.valueOf(1));
     }
+
+    @Test
+    public void filterTransactions_getNonceThrowsNodeReferenceException_throwsRuntimeException() {
+        // Given
+        RskAddress senderAddressMock = new RskAddress(new byte[20]);
+        Transaction txMock = Tx.create(config, 0, 50000, 5, 0, 0, 0);
```
marcciosilva, line 301: maybe we could extract this to a method? like `getTestTransaction` or `getMockTransaction` since I've seen it used one other time in the same test class (and since we don't care about the values, we could just call it without parameters).
I'd also consider naming the parameters we pass (value, gasLimit, nonce, etc.) but maybe it's not necessary

rskj-core/src/test/java/co/rsk/mine/MinerUtilsTest.java
```diff
@@ -288,4 +293,30 @@ public void getAllTransactionsCheckOrderTest() {
...
+        Mockito.when(txMock.getSender()).thenReturn(senderAddressMock);
+
+        List<Transaction> txList = new LinkedList<>();
+        txList.add(txMock);
+
+        Repository repositoryMock = Mockito.mock(Repository.class);
+        Mockito.doThrow(new NodeReferenceException("")).when(repositoryMock).getNonce(senderAddressMock);
```
marcciosilva, line 308: maybe I'd just use `NodeReferenceException.class` as a parameter for `doThrow` instead of adding an empty message

rskj-core/src/test/java/co/rsk/mine/MinerUtilsTest.java
```diff
@@ -288,4 +293,30 @@ public void getAllTransactionsCheckOrderTest() {
...
+
+        // Then
+        try {
+            new MinerUtils().filterTransactions(new LinkedList<>(), txList, new HashMap<>(), repositoryMock, Coin.valueOf(1L));
```
marcciosilva, line 312: I see the `new MinerUtils()` everywhere on this class - maybe we could make it a class variable and initialize it on a `@Before` method?

marcciosilva, line 312: do the list and map have to be modified? if not, we could use the emptyList() and emptyMap() from the Collections module

marcciosilva, line 312: maybe I'd move `Coin.valueOf(1L);` to a class constant, since I've seen it used multiple times in this class
(just a suggestion, maybe it doesn't add value)

rskj-core/src/test/java/co/rsk/mine/MinerUtilsTest.java
```diff
@@ -288,4 +293,30 @@ public void getAllTransactionsCheckOrderTest() {
...
+            fail("RuntimeException should be thrown!");
+        } catch (RuntimeException ex) {
```
marcciosilva, line 314: maybe inside the catch we could get its cause and assert it's a `NodeReferenceException`?
just to make sure that the cause is what we want (and if someone modifies something in the future, causing another `RuntimeException`, this test will fail)

rskj-core/src/main/java/co/rsk/trie/NodeReference.java
```diff
@@ -51,6 +51,7 @@ public boolean isEmpty() {
     /**
      * The node or empty if this is an empty reference.
      * If the node is not present but its hash is known, it will be retrieved from the store.
+     * If the node could not be retrieved from the store, a NodeReferenceException is thrown.
      */
     public Optional<Trie> getNode() {
```
marcciosilva, line 56: is this only used from `MinerUtils`? just to understand if we have to modify the `catch` clauses on any other class apart from that

nagarev: This method is used in more than one place, but the exception _per se_ is only being captured in MinerUtils and given that `NodeReferenceException` extends from `RuntimeException` there shouldn't be any problem in the rest of them.

rskj-core/src/test/java/co/rsk/mine/MinerUtilsTest.java
nagarev: At this level MinerUtils throws a RuntimeException, the one that throws NodeReferenceException is NodeReference class. Maybe we can modify MinerUtils and instead of throwing another exception, just let it bubble up. I think It could be better.

rskj-core/src/main/java/co/rsk/mine/MinerUtils.java
nagarev: Yep, that's the one. It is very entangled and difficult to see what's happening, but it was the only way I saw to stop the node back at the `Start.java` class.

rskj-core/src/main/java/co/rsk/trie/NodeReferenceException.java
```diff
@@ -0,0 +1,9 @@
+package co.rsk.trie;
```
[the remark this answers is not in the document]
nagarev: Done!

rskj-core/src/test/java/co/rsk/mine/MinerUtilsTest.java
marcciosilva: I agree! because in the end, our custom exception already is a `RuntimeException`

rskj-core/src/main/java/co/rsk/trie/NodeReference.java
marcciosilva: awesome!

donequis:
This doesn't seem to solve the issue. If an exception like this occurs during the normal EVM execution, wouldn't this just revert the transaction? Catching it in `VM.steps()` for example.

rskj-core/src/main/java/co/rsk/trie/NodeReference.java
nagarev: Strategy was updated. Dismissed.

rskj-core/src/main/java/co/rsk/trie/NodeReference.java
```diff
@@ -27,11 +28,11 @@
 
 public class NodeReference {
 
-
     private final TrieStore store;
 
     private Trie lazyNode;
     private Keccak256 lazyHash;
+    private NodeUtils nodeUtils;
```
Vovchyk, line 35: can we make this `final`?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/trie/NodeReference.java
nagarev: Done!

Vovchyk (approved):

rskj-core/src/main/java/co/rsk/trie/NodeReference.java
```diff
@@ -63,7 +73,14 @@ public boolean isEmpty() {
 
         Optional<Trie> node = store.retrieve(lazyHash.getBytes());
 
-        lazyNode = node.orElse(null);
+        // Broken database, can't continue
+        if (!node.isPresent()) {
+            nodeStopper.stop(1);
+            return Optional.empty();
```
joaquinlpereyra, line 79: A really clear and to the point log on _why_ your node is stopping should be present, because if you are here your node won't start again (or rather, will exist always at the same point). Can be very confusing for an unsuspecting user.

nagarev: Done!

rskj-core/src/test/java/co/rsk/mine/MinerUtilsTest.java
nagarev: At the end this changed a lot. Now we don't need this part.

rskj-core/src/main/java/co/rsk/mine/MinerUtils.java
nagarev: Outdated

nagarev:
The message will be discussed in a near feature, for now, we'll leave it as it is.

nagarev:
pipeline:run

nagarev:
pipeline:run

Vovchyk (approved):

PR #1568, opened 2021-06-28
Pegout transactions improvement

marcos-iov:
Include an output with OP_RETURN code and value zero to pegout transactions, in order to be easily identified by the Bridge. This output will have '52534b4f' (hexadecimal representation of 'RSKO') as its only content, for the purpose of distinguishing this output from other possible outputs with OP_RETURN op code.

*fed:pegout-improvement*


donequis:
If a pegout txs is sent just before the RSKIP activation and confirmed later, the change will be rejected and lost.

Issue #1550, opened 2021-06-03
Suboptimal code in deletion of Trie key/values


SergioDemianLerner:
The following method adds/removes data from the Trie:

 ```
private Trie put(TrieKeySlice key, byte[] value, boolean isRecursiveDelete) {
        .....
        
        // only coalesce if node has only one child and no value
        if (trie.valueLength.compareTo(Uint24.ZERO) > 0) {
            return trie;
        }

        Optional<Trie> leftOpt = trie.left.getNode();
        Optional<Trie> rightOpt = trie.right.getNode();
        if (leftOpt.isPresent() && rightOpt.isPresent()) {
            return trie;
        }

        if (!leftOpt.isPresent() && !rightOpt.isPresent()) {
            return trie;
        }
        Trie child;
        byte childImplicitByte;
        if (leftOpt.isPresent()) {
            child = leftOpt.get();
            childImplicitByte = (byte) 0;
        } else { // has right node
            child = rightOpt.get();
            childImplicitByte = (byte) 1;
        }
      ...
```
If the trie needs to coalesce nodes for a delete operation, then both left and right children are loaded from DB by GetNode(). This is not necessary: it's only necessary to know if they exists or not. 
The simplest case to trigger unnecessary node loads is trying to remove a non-existing element (key not found).

The fixed code should be something like this:
 ```
private Trie put(TrieKeySlice key, byte[] value, boolean isRecursiveDelete) {
     
        ...... 
        
        // only coalesce if node has only one child and no value
        if (trie.valueLength.compareTo(Uint24.ZERO) > 0) {
            return trie;
        }
        if (trie.left.isEmpty()==trie.rightisEmpty()) {
            return trie;
        }
        
       Trie child;
       byte childImplicitByte;
        if (!trie.left.isEmpty()) {
            child = trie.left.getNode().get();
            childImplicitByte = (byte) 0;
        } else { // has right node
            child = trie.right.getNode().get();
            childImplicitByte = (byte) 1;
        }
        ...
```

joaquinlpereyra:
I have created a small, quick benchmark to test whether the proposed change actually improves deletion speed. 

```
    @Test
    public void deleteManyValues() {
        Trie trie1 = new Trie();

        Integer amount = 1_000_000;

        for (int k = 0; k < amount; k++) {
            trie1 = trie1.put("key" + k, ("value" + k).getBytes());
        }

        long start = System.nanoTime();
        for (int k = 1; k < amount; k += 2) {
[...]

PR #1575, opened 2021-07-01
Fix sub-optimal retrieval

nagarev:
## Types of changes

## Checklist:

* **Other information**:

closes https://github.com/rsksmart/rskj/issues/1550

### DIFF
--- a/rskj-core/src/main/java/co/rsk/trie/Trie.java
+++ b/rskj-core/src/main/java/co/rsk/trie/Trie.java
@@ -775,24 +775,24 @@ public class Trie {
             return trie;
         }
 
-        Optional<Trie> leftOpt = trie.left.getNode();
-        Optional<Trie> rightOpt = trie.right.getNode();
-        if (leftOpt.isPresent() && rightOpt.isPresent()) {
-            return trie;
-        }
-
-        if (!leftOpt.isPresent() && !rightOpt.isPresent()) {
+        // both left and right exist (or not) at the same time
+        if (trie.left.isEmpty() == trie.right.isEmpty()) {
             return trie;
         }
 
         Trie child;
         byte childImplicitByte;
-        if (leftOpt.isPresent()) {
-            child = leftOpt.get();
+        if (!trie.left.isEmpty()) {
+            child = trie.left.getNode().orElse(null);
             childImplicitByte = (byte) 0;
         } else { // has right node
-            child = rightOpt.get();
+            child = trie.right.getNode().orElse(null);
             childImplicitByte = (byte) 1;
+        }
+
+        // could not retrieve from database
+        if (child == null) {
+            return trie;
         }
 
         TrieKeySlice newSharedPath = trie.sharedPath.rebuildSharedPath(childImplicitByte, child.sharedPath);
--- a/rskj-core/src/main/java/co/rsk/trie/Trie.java
+++ b/rskj-core/src/main/java/co/rsk/trie/Trie.java
@@ -26,11 +26,15 @@ import co.rsk.crypto.Keccak256;
 import co.rsk.metrics.profilers.Metric;
 import co.rsk.metrics.profilers.Profiler;
 import co.rsk.metrics.profilers.ProfilerFactory;
+import co.rsk.util.NodeStopper;
 import org.ethereum.crypto.Keccak256Helper;
 import org.ethereum.db.ByteArrayWrapper;
 import org.ethereum.util.ByteUtil;
 import org.ethereum.util.RLP;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/trie/Trie.java
```diff
@@ -766,23 +766,18 @@ private Trie put(TrieKeySlice key, byte[] value, boolean isRecursiveDelete) {
...
+        // both left and right exist (or not) at the same time
+        if (trie.left.isEmpty() == trie.right.isEmpty()) {
             return trie;
         }
 
         Trie child;
         byte childImplicitByte;
-        if (leftOpt.isPresent()) {
-            child = leftOpt.get();
+        if (!trie.left.isEmpty() && trie.left.getNode().isPresent()) {
```
joaquinlpereyra, line 776: My `null` guard comment was regarding `trie.left`, not `trie.left.getNode()`, which will never return `null` as it always returns an `Optional`. This guard calls `getNode()` which is quite expensive unnecessarily (as it has to perform `getHash()`)

This should be taken with a grain of salt though, even though it is possible that `trie.left` is null (nothing in the code prevents it as far as I see), the old code actually does assume that it is not null, maybe it is best to keep it that way.

So:

* To avoid risks of changing behavior I'd keep `trie.left`, `trie.right` without null checks, as the previous code assumed they were always something and this is most likely what's happening if there's no reason to believe otherwise
* The added guard adds more computation and also very subtly modifies behavior.

nagarev: Previously the code was doing the checkup about the node being not null here:

```java
Optional<Trie> leftOpt = trie.left.getNode();
Optional<Trie> rightOpt = trie.right.getNode();
if (leftOpt.isPresent() && rightOpt.isPresent()) { // <- Checking Here
    return trie;
}
```
so, in the following lines, when we called 

```java
rightOpt.get()
```

We were already covered.

But in the new code we only check that left and right are not empty without checking the contained node _per se_:

```java
if (trie.left.isEmpty() == trie.right.isEmpty()) {
```

So when calling

```java
.getNode().get()
```

we are not sure that the contents are not null.

So I think that:
- one of the points "To avoid risks of changing behavior I'd keep trie.left, trie.right without null checks, as the previous code assumed they were always something..." it's not completely right (because the old code was performing checks on the node being not null).
- we have to decide where and how we want to avoid extra computation. In this case the computation would be made for the cases when we know for sure that one of the left or right references are not empty.

What do you think?

nagarev: And yes `trie.left` and `trie.right` were not being checked for null right before as you pointed.

nagarev: With the checks removed, now if the `Node` element is not present, a `java.util.NoSuchElementException` is thrown by the `get()` method. So I've removed the checks for clarity, and we can start improving the solution from [this new commit](https://github.com/rsksmart/rskj/pull/1575/commits/2580b15ff41217a6abbc50d3bc643f043daae7e3) and on.

joaquinlpereyra: Leaving for further discussion: the underlying issue here is that `isEmpty` and `getNode` are not correlated to what one would expect: an `!trie.isEmpty` **does not** assure that `trie.getNode().isPresent()`, at least according to the code.

IMHO this scenario should not be considered an `option`: looking up a node in the db which does not exist should be considered an exception, no correct alternative exists to continue processing the information.

rskj-core/src/main/java/co/rsk/trie/Trie.java
```diff
@@ -766,26 +766,26 @@ private Trie put(TrieKeySlice key, byte[] value, boolean isRecursiveDelete) {
...
+            child = trie.left.getNode().orElse(null);
             childImplicitByte = (byte) 0;
         } else { // has right node
-            child = rightOpt.get();
+            child = trie.right.getNode().orElse(null);
             childImplicitByte = (byte) 1;
         }
 
+        // could not retrieve from database
+        if (child == null) {
```
Vovchyk, line 785: why did you add this condition? From my understanding, this should never happen.

nagarev: Yep, we think the same, but since this "child" is the result of evaluating an `Optional` field, we have to take that in account. Also note that the code behavior before this change was exactly the same, so this is also for the sake of maintaining it the same.

nagarev:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk (approved):

rskj-core/src/main/java/co/rsk/trie/Trie.java
fedejinich: mmm isn't this an unnecesary check? if you do the `Trie.getNode()` you can get the broken db error

```java
public Optional<Trie> getNode() {
        if (lazyNode != null) {
            return Optional.of(lazyNode);
        }

        if (lazyHash == null) {
            return Optional.empty();
        }

        Optional<Trie> node = store.retrieve(lazyHash.getBytes());

        // Broken database, can't continue
        if (!node.isPresent()) {
            logger.error("Broken database, execution can't continue");
            nodeStopper.stop(1);
            return Optional.empty();
        }

        lazyNode = node.get();

        return node;
    }
```

fedejinich (commented):
apart form the comment I've made, I think the overall it's ok

PR #1576, opened 2021-07-01
Skip REMASC Transaction (Configurable)

fedejinich:
## Description

Continuing this [previous PR](https://github.com/rsksmart/rskj/pull/1490) from @ajlopez, adds support to skip REMASC transaction from blocks on JSON RPC requests. This feature it's disabled by default

Use `-Drpc.skipRemasc=true`

## Motivation and Context

There are some Ethereum's ecosystem tools that relies on transactions with gasUsed > 0 (since there are no transactions with gasUsed > 0 on eth). Checkout this example https://github.com/rsksmart/rskj/issues/1484

In RSK every block has a tranaction, which is part of the consensus, named REMASC. This special case differs from other ones beacause it has `gasUsed == 0`.

In order to provide a better UX and compatibility we provide a mechanism to filter this Transaction from the returned block/s on JSON RPC requests (IE `eth_getBlockByNumber` or `eth_getBlockByHash`).

DISCLAIMER: even though we provide this mechanisim the user must take care and notice that there is a hidden REMASC transaction contained in the block.

## How Has This Been Tested?

Unit tests coming form the angel's PR

## Types of changes

## Checklist:

**Other information**:
* https://developers.rsk.co/rsk/architecture/mining/remasc/

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
@@ -35,6 +35,8 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import static org.ethereum.crypto.HashUtil.EMPTY_TRIE_HASH;
+
 
 public class BlockResultDTO {
     private final String number; // QUANTITY - the block number. null when its pending block.
@@ -143,13 +145,20 @@ public class BlockResultDTO {
             uncles.add(header.getHash().toJsonString());
         }
 
+        // Useful for geth integration
+        byte[] transactionsRoot = skipRemasc &&
+                b.getTransactionsList().size() == 1 &&
+                b.getTransactionsList().get(0).isRemascTransaction(0,1) ?
+                    EMPTY_TRIE_HASH :
+                    b.getTxTrieRoot();
+
         return new BlockResultDTO(
                 isPending ? null : b.getNumber(),
                 isPending ? null : b.getHash(),
                 b.getParentHash(),
                 b.getUnclesHash(),
                 isPending ? null : b.getLogBloom(),
-                b.getTxTrieRoot(),
+                transactionsRoot,
                 b.getStateRoot(),
                 b.getReceiptsRoot(),
                 isPending ? null : b.getCoinbase(),
--- a/rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
@@ -31,6 +31,7 @@ import org.ethereum.rpc.TypeConverter;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
@@ -133,7 +134,7 @@ public class BlockResultDTO {
         // For full tx will present as TransactionResultDTO otherwise just as transaction hash
         List<Object> transactions = IntStream.range(0, blockTransactions.size())
                 .mapToObj(txIndex -> toTransactionResult(txIndex, b, fullTx, skipRemasc))
-                .filter(o -> o != null)
+                .filter(Objects::nonNull)
                 .collect(Collectors.toList());
 
         List<String> uncles = new ArrayList<>();
--- a/rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
@@ -22,7 +22,6 @@ import co.rsk.core.BlockDifficulty;
 import co.rsk.core.Coin;
 import co.rsk.core.RskAddress;
 import co.rsk.crypto.Keccak256;
-import co.rsk.remasc.RemascTransaction;
 import org.ethereum.core.Block;
 import org.ethereum.core.BlockHeader;
 import org.ethereum.core.Transaction;
@@ -32,6 +31,8 @@ import org.ethereum.rpc.TypeConverter;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 
 public class BlockResultDTO {
@@ -128,29 +129,12 @@ public class BlockResultDTO {
 
         Coin mgp = b.getMinimumGasPrice();
 
-        List<Object> transactions = new ArrayList<>();
         List<Transaction> blockTransactions = b.getTransactionsList();
-        if (fullTx) {
-            for (int i = 0; i < blockTransactions.size(); i++) {
-                Transaction tx = blockTransactions.get(i);
-
-                if (skipRemasc && tx.isRemascTransaction(i, blockTransactions.size())) {
-                    continue;
-                }
-
-                transactions.add(new TransactionResultDTO(b, i, blockTransactions.get(i)));
-            }
-        } else {
-            for (int i = 0; i < blockTransactions.size(); i++) {
-                Transaction tx = blockTransactions.get(i);
-
-                if (skipRemasc && tx.isRemascTransaction(i, blockTransactions.size())) {
-                    continue;
-                }
-
-                transactions.add(tx.getHash().toJsonString());
-            }
-        }
+        // For full tx will present as TransactionResultDTO otherwise just as transaction hash
+        List<Object> transactions = IntStream.range(0, blockTransactions.size())
+                .mapToObj(txIndex -> toTransactionResult(txIndex, b, fullTx, skipRemasc))
+                .filter(o -> o != null)
+                .collect(Collectors.toList());
 
         List<String> uncles = new ArrayList<>();
 
@@ -185,6 +169,20 @@ public class BlockResultDTO {
                 b.getHashForMergedMining(),
                 b.getFeesPaidToMiner()
         );
+    }
+
+    private static Object toTransactionResult(int transactionIndex, Block block, boolean fullTx, boolean skipRemasc) {
+        Transaction transaction = block.getTransactionsList().get(transactionIndex);
+
+        if(skipRemasc && transaction.isRemascTransaction(transactionIndex, block.getTransactionsList().size())) {
+            return null;
+        }
+        
+        if(fullTx) {
+            return new TransactionResultDTO(block, transactionIndex, transaction);
+        }
+
+        return transaction.getHash().toJsonString();
     }
 
     public String getNumber() {
--- a/rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
@@ -65,7 +65,7 @@ public class BlockResultDTO {
     private final String paidFees;
     private final String cumulativeDifficulty;
 
-    public BlockResultDTO(
+    private BlockResultDTO(
             Long number,
             Keccak256 hash,
             Keccak256 parentHash,
--- a/rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
@@ -22,6 +22,7 @@ import co.rsk.core.BlockDifficulty;
 import co.rsk.core.Coin;
 import co.rsk.core.RskAddress;
 import co.rsk.crypto.Keccak256;
+import co.rsk.remasc.RemascTransaction;
 import org.ethereum.core.Block;
 import org.ethereum.core.BlockHeader;
 import org.ethereum.core.Transaction;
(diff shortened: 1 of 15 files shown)


fedejinich:
pipeline:run

rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
```diff
@@ -127,17 +129,12 @@ public static BlockResultDTO fromBlock(Block b, boolean fullTx, BlockStore block
...
-            }
-        } else {
-            for (Transaction tx : blockTransactions) {
-                transactions.add(tx.getHash().toJsonString());
-            }
-        }
+        // For full tx will present as TransactionResultDTO otherwise just as transaction hash
+        List<Object> transactions = IntStream.range(0, blockTransactions.size())
+                .mapToObj(txIndex -> toTransactionResult(txIndex, b, fullTx, skipRemasc))
+                .filter(o -> o != null)
```
marcciosilva, line 136: I'd use `.filter(Objects::nonNull)` (I´ve seen it around the node and think it´s more legible=

rskj-core/src/test/java/org/ethereum/rpc/dto/BlockResultDTOTest.java
```diff
@@ -0,0 +1,101 @@
...
+        Blockchain blockChain = objects.getBlockchain();
+
+        BlockBuilder builder = new BlockBuilder(null, null, null).parent(blockChain.getBestBlock());
+        List<Transaction> transactions = new ArrayList<>();
+        RemascTransaction remascTransaction = spy(new RemascTransaction(1));
+        when(remascTransaction.isRemascTransaction(1, 2)).thenReturn(true);
+        Transaction transaction = spy(new RemascTransaction(1));
+        when(transaction.isRemascTransaction(0, 2)).thenReturn(false);
+        transactions.add(transaction);
+        transactions.add(remascTransaction);
```
marcciosilva, line 63: maybe we could extract this code to a function like the following (to make it more readable, like a comment but in code):
```suggestion
        buildBlockWithRemascAndNormalTxs(builder);
```

rskj-core/src/test/java/org/ethereum/rpc/dto/BlockResultDTOTest.java
```diff
@@ -0,0 +1,101 @@
...
+        Assert.assertNotNull(blockResultDTO);
+        Assert.assertEquals(2, blockResultDTO.getTransactions().size());
+    }
+
+    @Test
+    public void getBlockResultDTOWithoutRemascAndTransactionHashes() {
+        BlockResultDTO blockResultDTO = BlockResultDTO.fromBlock(block, false, blockStore, true);
+
+        Assert.assertNotNull(blockResultDTO);
+        Assert.assertEquals(1, blockResultDTO.getTransactions().size());
```
marcciosilva, line 83: if it's not too much work, I'd probably make the transactions constants in the test class, and here I'd assert that we're effectively obtaining the normal one (we only know we're obtaining one - the logic could be wrong and we wouldn't notice)
(the logic is correct, I'm just saying it might be safer this way)

rskj-core/src/test/java/org/ethereum/rpc/dto/BlockResultDTOTest.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Test
+    public void getBlockResultDTOWithRemascAndFullTransactions() {
```
marcciosilva, line 87: this test is doing the same stuff that `getBlockResultDTOWithRemascAndTransactionHashes`, right?
maybe I'd assert that we're getting the full txs we're expecting (and that we're just getting hashes in the other test)

rskj-core/src/test/java/org/ethereum/rpc/dto/BlockResultDTOTest.java
```diff
@@ -0,0 +1,101 @@
...
+        BlockResultDTO blockResultDTO = BlockResultDTO.fromBlock(block, true, blockStore, false);
+
+        Assert.assertNotNull(blockResultDTO);
+        Assert.assertEquals(2, blockResultDTO.getTransactions().size());
+    }
+
+    @Test
+    public void getBlockResultDTOWithoutRemascAndFullTransactions() {
```
marcciosilva, line 95: same thing here - I'd make sure we're actually getting the full txs, because this ends up being the same test as `getBlockResultDTOWithoutRemascAndTransactionHashes`

rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
fedejinich: ohhh nice, i was trying to find that one!

fernandorubal:
Hello! Here Fernando from Qubistry. We did some tests against the Ethereum Go client and it failed when validating the Tx Trie Hash on blocks where Remasc is the only Tx and is hidden from the RPC response. We solved it by returning an empty Tx hash if there's just 1 Tx in the block (the remasc tx) and skipRemasc is enabled, in BlockResultDTO.java.

Instead of 
`b.getTxTrieRoot`

We used
`b.getTransactionsList().size() == 1 && skipRemasc ? EMPTY_TRIE_HASH : b.getTxTrieRoot()`

fedejinich:
> Hello! Here Fernando from Qubistry. We did some tests against the Ethereum Go client and it failed when validating the Tx Trie Hash on blocks where Remasc is the only Tx and is hidden from the RPC response. We solved it by returning an empty Tx hash if there's just 1 Tx in the block (the remasc tx) and skipRemasc is enabled, in BlockResultDTO.java.
> 
> Instead of
> `b.getTxTrieRoot()`
> 
> We used
> `b.getTransactionsList().size() == 1 && skipRemasc ? EMPTY_TRIE_HASH : b.getTxTrieRoot()`

what kind of validations are you trying to do? can you give more context? anything to reproduce?

fernandorubal:
> what kind of validations are you trying to do? can you give more context? anything to reproduce?

We're running some integrations test that uses the Ethereum Go client as library and it failed when, after sending a tx and subscribing to new heads to wait for confirmations, tried to validate the Tx Root Hash of the received new heads:

https://github.com/ethereum/go-ethereum/blob/master/ethclient/ethclient.go

on line 130:

```
if head.TxHash != types.EmptyRootHash && len(body.Transactions) == 0 {
    return nil, fmt.Errorf("server returned empty transaction list but block header indicates transactions")
}
```
When skipRemasc is enabled on the rskj node, the Geth client receives a non-empty tx root hash but no transactions are apparently included (because remasc is hidden).

So what we did is to return an empty tx root hash in those cases and it worked without any other issues

fedejinich:
This is an experimental feature, which just skips the REMASC tx only for the transactions list, not from the blockchain state, adding that could bring potential issues.

The behavior that you're mentioning it's the expected one, the REMASC transaction is part of the RSK consensus rule, you should receive a non empty transaction root hash. Unfortunately this will conflict with this geth tool you're mentioning :(

rskj-core/src/main/java/org/ethereum/rpc/dto/BlockResultDTO.java
fedejinich: https://github.com/rsksmart/rskj/pull/1576/commits/ae3ff0624728affb4ea5983b6f8a358b98dc66aa

rskj-core/src/test/java/org/ethereum/rpc/dto/BlockResultDTOTest.java
fedejinich: https://github.com/rsksmart/rskj/pull/1576/commits/ae3ff0624728affb4ea5983b6f8a358b98dc66aa

rskj-core/src/test/java/org/ethereum/rpc/dto/BlockResultDTOTest.java
fedejinich: https://github.com/rsksmart/rskj/pull/1576/commits/ae3ff0624728affb4ea5983b6f8a358b98dc66aa

rskj-core/src/test/java/org/ethereum/rpc/dto/BlockResultDTOTest.java
fedejinich: https://github.com/rsksmart/rskj/pull/1576/commits/ae3ff0624728affb4ea5983b6f8a358b98dc66aa

rskj-core/src/test/java/org/ethereum/rpc/dto/BlockResultDTOTest.java
fedejinich: https://github.com/rsksmart/rskj/pull/1576/commits/ae3ff0624728affb4ea5983b6f8a358b98dc66aa

fedejinich:
pipeline:run

fernandorubal:
> This is an experimental feature, which just skips the REMASC tx only for the transactions list, not from the blockchain state, adding that could bring potential issues.
> 
> The behavior that you're mentioning it's the expected one, the REMASC transaction is part of the RSK consensus rule, you should receive a non empty transaction root hash. Unfortunately this will conflict with this geth tool you're mentioning :(
> 
> Take account that even if you are skiping that transaction manually (by adding the `empty_root_hash`), you'll be testing against a different blockchain state.
> 
> You can check [this link](https://developers.rsk.co/rsk/architecture/mining/remasc/) to read more about this special transaction.

Thank you, we know about the remasc tx, and the potential risks.

We are working on native Chainlink integration for a while now, that originated this PR in the first place. We started testing this skipRemasc option with Chainlink to see if it was enough to achieve a successful integration, that is where we encountered the follow up error of the tx hash validation. Sadly the changes from this PR won't serve the original purpose because the 0 gasPrice error would be replaced with the tx trie validation error. This is an integration blocker, I would say.

The tx hash change proposed by us was needed for it to work without errors, and didn't provoke any further errors down the line. Our tests even worked on our Testnet node flawlessly, using Chainlink nodes connected natively to it. We know that this is an experimental feature, and that changing the block tx hash isn't a pretty solution, but it won't work otherwise and this PR wouldn't be enough to make native ethereum tools using geth work as initially intended. As long as the potential risks are well noted and taking into consideration that the skipRemasc option will be used only on those special cases, I don't think it should be an issue to include that change.

I think another completely different solution would be to format the remasc tx to comply with a standard Eth tx (valid r, s, v and gasPrice fields, among others) instead of hiding it, but I don't think that's even possible.

fedejinich:
@fernandorubal can you test this new changes?

fedejinich:
pipeline:run

fedejinich:
pipeline:run

fedejinich:
pipeline:run

fernandorubal:
> @fernandorubal can you test this new changes?

I tested it just now, it worked flawlessly! Thank you!!

fedejinich:
> > @fernandorubal can you test this new changes?
> 
> I tested it just now, it worked flawlessly! Thank you!!

nice, thanks fernando!!

fedejinich:
pipeline:run

rskj-core/src/main/java/org/ethereum/rpc/dto/TransactionResultDTO.java
```diff
@@ -79,4 +77,117 @@ public TransactionResultDTO(Block b, Integer index, Transaction tx) {
...
+        this.r = r;
+    }
+
+    public String getS() {
+        return s;
+    }
+
+    public void setS(String s) {
+        this.s = s;
+    }
```
nagarev, line 191: It seems that these "setter methods" are not being used anywhere, I would remove them since it helps with understanding how we use this DTO in the rest of the code. What do you think?

nagarev: Also "getter methods" for BlockNumber, Gas, and GasPrice seems unused, I don't know what to think about this... maybe we have to refactor this and delete these fields?

rskj-core/src/main/resources/expected.conf
```diff
@@ -235,6 +235,8 @@ rpc = {
             }
         }
     ]
+
```
nagarev, line 238: would remove this space here to preserve the style.

rskj-core/src/main/resources/reference.conf
```diff
@@ -354,6 +354,8 @@ rpc {
              enabled: "true"
          }
     ]
+
```
nagarev, line 357: Also here I would remove this space here to preserve the style.

Vovchyk (dismissed):

rskj-core/src/main/java/org/ethereum/rpc/dto/TransactionResultDTO.java
fedejinich: removed the setters, the getters are necessary for testing purposes. Also since it's a dto, it should only be used as dumb presentational object, any other use would break that pattern.
all the involved fields are necessary to match the api spec, there's no need to remove any field

rskj-core/src/main/resources/expected.conf
fedejinich: done

rskj-core/src/main/resources/reference.conf
fedejinich: done

rskj-core/src/main/java/org/ethereum/rpc/dto/TransactionResultDTO.java
fedejinich: and I'm not completely sure but maybe jackson it's using the getters also

horca17:
next steps? thanks

fedejinich:
pipeline:run

fedejinich:
pipeline:run

fedejinich:
pipeline:run

fedejinich:
just rebased

aeidelman (approved):

Issue #1584, opened 2021-07-07
Reuse calculated block hash


Vovchyk:
Right now, when `Block.getHash()` is called, internally it calls `BlockHeader.getHash()`, which in its turn does this:
```java
    public Keccak256 getHash() {
        return new Keccak256(HashUtil.keccak256(getEncoded()));
    }
```
... where `HashUtil.keccak256` is pretty heavy operation. It looks like that during block execution `BlockHeader.getHash()` is being triggered multiple times, which obviously has impact on time that's needed to execute a block.

Suggestion: cache a block hash in a block header object.

PR #1585, opened 2021-07-07
Reused calculated block hash

Vovchyk:
## Description

This change allows to cache a block hash after first call to `BlockHeader.getHash()` and reuse cached value in next calls.

## Motivation and Context

See https://github.com/rsksmart/rskj/issues/1584

## How Has This Been Tested?

unit tests / running node in local env.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
+++ b/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
@@ -115,7 +115,7 @@ public class BlockHeader {
     private volatile boolean sealed;
 
     /* Holds calculated block hash */
-    private volatile Keccak256 hash;
+    private Keccak256 hash;
 
     /* Indicates if the block was mined according to RSKIP-92 rules */
     private final boolean useRskip92Encoding;
--- a/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
+++ b/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
@@ -50,12 +50,12 @@ public class BlockHeader {
     private static final int UMM_LEAVES_LENGTH = 20;
 
     /* The SHA3 256-bit hash of the parent block, in its entirety */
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -293,12 +302,17 @@ public void setLogsBloom(byte[] logsBloom) {
...
             throw new SealedBlockHeaderException("trying to alter logs bloom");
         }
+        this.hash = null;
 
         this.logsBloom = logsBloom;
     }
 
     public Keccak256 getHash() {
-        return new Keccak256(HashUtil.keccak256(getEncoded()));
+        if (this.hash == null) {
```
donequis, line 311: Can I recommend doing it on sealed blocks only?

joaquinlpereyra:
This should work, but potentially opens up the possibility of absolutely destroying a block's hash. I just tried deleting a `this.hash = null` and all tests run correctly. When you delete _all_ of them some tests fail but at first glance have no relationship with the change. 

It'd be interesting to see a test that modifies the block somehow and then checks that the hash has indeed changed.

Vovchyk:
> This should work, but potentially opens up the possibility of absolutely destroying a block's hash.

Yeah, we should be careful here. Btw, similar technique is being used in the `Transaction` class.

> It'd be interesting to see a test that modifies the block somehow and then checks that the hash has indeed changed.

Good point. Added a test for this.

joaquinlpereyra (approved):
🟢

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

heitorfm (approved):

Vovchyk:
Resolves #1585

PR #1588, opened 2021-07-13
Remove the leading zero from the gasLimit field of Block Header notifications

fernandorubal:
## Description

This PR removes the leading zero from the gasLimit field value of the Block Header subscriptions notifications

## Motivation and Context

This PR aims to improve compatibility with Ethereum's native tools like the official Go client library, and solve an issue where subscribing to new heads leads to a json.UnmarshalTypeError "hex number with leading zero digits" on the client when trying to read the gasLimit field from a Block Header notification.

## How Has This Been Tested?

A new test was added in EthSubscriptionNotificationTest.java called gasLimitWithoutLeadingZeros that uses the same logic as the basicRequest test, but checks against a different gasLimit value that returns a trailing zero if not converted using toQuantityJsonHex. Setting a gasLimit of `1000000` would result in `0x0f4240` using toJsonHex.

## Types of changes

## Checklist:

* **Other information**:


fedejinich:
Also I'll add that this change corresponds on retrieving formatted QUANTITY data instead of UNFORMATTED data, according to the json rpc spec https://eth.wiki/json-rpc/API

fedejinich:
nice one!

fedejinich:
also needs tests

Vovchyk:
pipeline: run

fernandorubal:
> Please can you add some tests

I modified the existing EthSubscriptionNotificationTest to work with a gasLimit of 1000000 that errors in case toQuantityJsonHex is not used in gasLimit field.
2f95d68

fedejinich:
Sorry but i didn't get how this affects the new format. Also I'd expect to have tests for retrieving a block header notification without a leading zeros

rskj-core/src/test/java/co/rsk/blockchain/utils/BlockGenerator.java
```diff
@@ -297,8 +301,16 @@ public Block createChildBlock(Block parent, List<Transaction> txs, List<BlockHea
     }
 
     public Block createBlock(int number, int ntxs) {
-        Bloom logBloom = new Bloom();
-        Block parent = getGenesisBlock();
+        return createBlock(number, ntxs, -1);
+    }
+
+    public Block createBlock(int number, int ntxs, long gasLimit){
```
nagarev, line 307: Are we going to allow this new `createBlock` method to be called outside of `BlockGenerator`? If that's the case, I would add another separate test for this instead of updating the old one. If not, then I would change its visibility to `private`, and would also add another test, so we're sure that we are testing the `gasLimit` being `-1` and `!-1`.

rskj-core/src/test/java/co/rsk/blockchain/utils/BlockGenerator.java
```diff
@@ -297,8 +301,16 @@ public Block createChildBlock(Block parent, List<Transaction> txs, List<BlockHea
...
+        Block parent;
+        if (gasLimit == -1){
+            parent = getGenesisBlock();
+        } else {
+            parent = getGenesisBlock(gasLimit);
+        }
```
marcciosilva, line 313: I'd maybe change this section to this, so there's no confusion in the `-1` usage convention (and would only involve changing `1000000` to `1000000L` in the test class you modified):
```suggestion
        return createBlock(number, ntxs, null);
    }

    public Block createBlock(int number, int ntxs, Long gasLimit){
        Block parent = gasLimit == null ? getGenesisBlock() : getGenesisBlock(gasLimit);
```

fernandorubal: a much better and cleaner solution, thank you!

rskj-core/src/test/java/co/rsk/blockchain/utils/BlockGenerator.java
fernandorubal: 36363e3

fernandorubal:
> Sorry but i didn't get how this affects the new format. Also I'd expect to have tests for retrieving a block header notification without a leading zeros

Yes, I think I explained myself too poorly.

The purpose of the original test change was to use a different gasLimit, that returns a trailing zero if not converted to Quantity.

Setting a gasLimit of `1000000` would result in `0x0f4240` if using toJsonHex. The original test gasLimit of `3141592` returns `0x2fefd8` either way, because the first byte is 2f instead of 0f. I hope is clearer now that the original test is restored and a second test, specific to this issue, was added.


36363e3

rskj-core/src/test/java/co/rsk/blockchain/utils/BlockGenerator.java
fernandorubal: https://github.com/rsksmart/rskj/pull/1588/commits/64c10df1c22f076a5cba5e7b6290aebcd3d74c2a

nagarev:
pipeline:run

horca17:
Hello, any news? Thanks

fedejinich:
it's under QA @horca17

nagarev:
Hi @fernandorubal! QA is ready, so you can rebase your branch and we can proceed to merge this! thanks!

fernandorubal:
> Hi @fernandorubal! QA is ready, so you can rebase your branch and we can proceed to merge this! thanks!

Hi, thank you! Rebase done

nagarev:
It seems that your fork `master` branch is not synced with rskj `master` branch. Can you try to sync it first, and then do a `rebase` of your forked `master` branch into your `fernandorubal:gaslimitleadingzero` branch? Then you must push the changes again and it should be ready. Thanks!

fernandorubal:
> It seems that your fork `master` branch is not synced with rskj `master` branch. Can you try to sync it first, and then do a `rebase` of your forked `master` branch into your `fernandorubal:gaslimitleadingzero` branch? Then you must push the changes again and it should be ready. Thanks!

You're right. Since the previous rebase I did, a new commit was added. I did a second rebase, it should be fine now.

nagarev:
Excellent!!! Thanks a lot!

nagarev:
pipeline:run

nagarev (approved):
LGTM

PR #1595, opened 2021-08-02
Added persistence of states cache

Vovchyk:
## Description
This change allows to save and then load a snapshot of states / unitrie cache, so that this speeds up initial loading of unitrie entries / nodes from a disk.

Note: the feature is OFF by default. To enable it just run a node with the `--persist-states-cache-snapshot` flag.

Note2: this is an experimental feature and is not covered with unit tests yet.

## Motivation and Context
Loading tries from db into cache/memory (in a recursive way) when the node starts is a time consuming operation which has impact on performance of block processing (until the cache in memory is populated). The intent of this change is to reduce the time needed to populate this states cache in memory.

## How Has This Been Tested?

By running a local node

## Types of changes

## Checklist:

* **Other information**:


Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
```diff
@@ -247,11 +260,126 @@ public void close() {
...
+                } else if (valueSize == 0) {
+                    value = new byte[0];
+                }
+
+                cache.put(ByteUtil.wrap(key), value);
+            }
+        } catch (IOException e) {
+            cache.clear();
+            boolean delResult = cacheSnapshotFile.delete();
+            logger.error("Cannot read from cache snapshot file. File deletion result: '{}'", delResult, e);
```
SergioDemianLerner, line 322: should it be deleted on error? Maybe it's better to rename it. For example, add the suffix ".err" to its name. Then it's possible to send the file to more advanced users for debugging, or debug locally by renaming the file back.

SergioDemianLerner (approved):

rskj-core/src/main/java/co/rsk/core/bc/BlockChainFlusher.java
```diff
@@ -85,6 +85,15 @@ private void closeAll() {
         logger.trace("closing blockStore.");
         blockStore.close();
         logger.trace("blockStore closed.");
+
+        logger.trace("closing trieStore.");
```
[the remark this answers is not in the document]
Vovchyk: yep, updated the trace msgs

rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
```diff
@@ -247,11 +259,26 @@ public void close() {
...
             uncommittedCache.clear();
             committedCache.clear();
         }
         finally {
             this.lock.writeLock().unlock();
         }
     }
+
+    @Nonnull
+    private static Map<ByteArrayWrapper, byte[]> makeCommittedCache(int cacheSize,
```
nagarev, line 274: I think that maybe this method can be implemented as "initSnapshotHandler" or something like that, since its goal is to load the cache into the handler. I would leave the new constructor as this:

```java
public DataSourceWithCache(@Nonnull KeyValueDataSource base, int cacheSize,
                               @Nullable CacheSnapshotHandler cacheSnapshotHandler) {
    this.cacheSize = cacheSize;
    this.base = Objects.requireNonNull(base);
    this.uncommittedCache = new LinkedHashMap<>(cacheSize / 8, (float)0.75, false); 
    this.committedCache = Collections.synchronizedMap(new MaxSizeHashMap<>(cacheSize, true));
    this.cacheSnapshotHandler = cacheSnapshotHandler;
    initSnapshotHandler();
}
```

And then the other method would be as this:

```java
private void initSnapshotHandler() {
    if (this.cacheSnapshotHandler != null) {
            this.cacheSnapshotHandler.load(this.committedCache);
    }
}
```
Also we can get rid of this last method and apply that logic in the constructor. I think is a little bit more "direct" or easy to understand at plain sight.

Vovchyk: well, this method doesn't really initialise an instance of `SnapshotHandler`, but I'd say rather initialises `committedCache` map. In fact, `cacheSnapshotHandler` value is `nullable` (it is `null` for a few other data sources, like receipts and it is `null` when the feature is OFF (by default)) and class itself is stateless, so I guess it's not needed to be initialised.

I see `makeCommittedCache()` method as a factory that can create an instance of `Map<ByteArrayWrapper, byte[]>` and populate it, if needed.

jjtechuy: pipeline: run

jjtechuy:
pipeline: run

jjtechuy:
pipeline: run

rskj-core/src/main/java/org/ethereum/util/MapSnapshot.java
```diff
@@ -0,0 +1,120 @@
...
+
+    }
+
+    public static class In extends MapSnapshot<DataInputStream> {
+
+        protected In(@Nonnull InputStream input) {
+            super(new DataInputStream(input));
+        }
+
+        public void read(@Nonnull Map<ByteArrayWrapper, byte[]> map) throws IOException {
```
nagarev, line 59: I think we can add some javadoc explaining the order in what the values are read and, for example, the constraints or possible values (keySize should be >= 1, valueSize should be >= -1, etc). I think it could help to understand the code better besides it is really tidy and easy to read, but is not that clear what's happening without having to read the whole function beforehand.

rskj-core/src/main/java/org/ethereum/util/MapSnapshot.java
```diff
@@ -0,0 +1,120 @@
...
+            Objects.requireNonNull(map);
+
+            int count = stream.readInt();
```
nagarev, line 62: I would rename it to "entryCount", only a suggestion btw.

rskj-core/src/main/java/org/ethereum/util/MapSnapshot.java
```diff
@@ -0,0 +1,120 @@
...
+                } else if (valueSize == 0) {
+                    value = new byte[0];
+                }
+
+                map.put(ByteUtil.wrap(key), value);
+            }
+        }
+    }
+
+    public static class Out extends MapSnapshot<DataOutputStream> {
```
nagarev, line 91: I think that adding the "Out" class first could help to get rid of a javadoc (previous comment). But Maybe we also could add some general javadoc to the `MapSnapshot` class, describing how the IN and OUT are related, etc.

rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
nagarev: Yeah, that name `initSnapshotHandler` maybe is not the best to describe what's happening. It is more like a "configuration" or something like that. I mean: the `committedCache` will end up being initialized as `new MaxSizeHashMap<>(cacheSize, true);` no matter what, but the difference between the `cacheSnapshotHandler` being null or not only makes that object to load the already initialized `committedCache`. I think that the "flow" on the constructor would be a little bit clearer if we "configure" the `cacheSnapshotHandler` after all objects are assigned and so, that was the way I've seen it.

rskj-core/src/main/java/org/ethereum/datasource/CacheSnapshotHandler.java
```diff
@@ -0,0 +1,123 @@
...
+            if (!errFile.exists()) {
+                if (!cacheSnapshotFile.renameTo(errFile)) {
+                    logger.error("Cannot rename cache snapshot file", e);
+                }
+            }
+            logger.error("Cannot read from cache snapshot file", e);
+        }
+    }
+
+    public void save(@Nonnull Map<ByteArrayWrapper, byte[]> cache) {
```
nagarev, line 82: I think we can explain here in a javadoc what's happening with the file (why and when is it being renamed and deleted, only a brief explanation).

nagarev:
LGTM, let's discuss the comments and then I will approve, a couple of tests to review yet.

rskj-core/src/main/java/org/ethereum/datasource/CacheSnapshotHandler.java
nagarev: AMAZING!

Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

nagarev (approved):
LGTM

joaquinlpereyra:
👋🏿 Hello. Nice work. 

Wouldn't it be nice to have some kind of checksum for the cache? Either because of a problem with the writing itself or another process messing with the file. Should be easy to do as you always read / write the entirety of it. I tried messing with the values and it reads them no problems, which would lead to a node having values in the storage which shouldn't be there and the user would have to delete the cache manually to make it right.

Vovchyk:
>Wouldn't it be nice to have some kind of checksum for the cache? Either because of a problem with the writing itself or another process messing with the file. Should be easy to do as you always read / write the entirety of it. I tried messing with the values and it reads them no problems, which would lead to a node having values in the storage which shouldn't be there and the user would have to delete the cache manually to make it right.

@joaquinlpereyra, added checksum verification

PR #1603, opened 2021-08-10
AutoPrune message queue

fedejinich:
Currently the only way to remove things from the (NodeMessageHandler) message queue it's by processing them.

This PR introduces the auto prune feature:
- Removes all the messages (from the message queue) after not beeing processed by a certain period.
- Disabled by default

Still need to discuss the auto prune time period


fedejinich:
this was an inspirational PR :D

PR #1614, opened 2021-09-16
Removed Jcenter() from gradle repositories

rmoreliovlabs:
## Description
`Jcenter()` repository was removed from gradle dependencies

## Motivation and Context
On February 3 2021, JFrog announced that they will be shutting down Bintray and JCenter. 

Based on the current timeline, builds that use JCenter will be able to resolve dependencies until February 1, 2022 without changes. After that date, there are no guarantees that you will be able to build your software if you continue to use JCenter.

Some details are here, here and here.

The place where it’s been in use in RskJ

UPDATE 4/27/2021: We listened to the community and will keep JCenter as a read-only repository indefinitely. Our customers and the community can continue to rely on JCenter as a reliable mirror for Java packages.

it looks like that JCenter is gonna provide read-only access to already published dependencies even after Feb 1, 2022, but I believe that it’d be better to migrate away from it in any way due to potential degradation of quality of their service.

## How Has This Been Tested?
Just ran it locally by building the project using `gradlew build` and then running the .jar `/rskj-core/build/libs/rskj-core-3.1.0-SNAPSHOT-all.jar co.rsk.Start`

## Types of changes

## Checklist:

- [ x] My code follows the code style of this project.
- [ x] My change requires a change to the documentation.

* **Other information**:

### DIFF
--- a/rskj-core/build.gradle
+++ b/rskj-core/build.gradle
@@ -31,7 +31,6 @@ tasks.withType(Javadoc) {
 
 repositories {
     mavenCentral()
-    jcenter()
     maven {
         url "https://deps.rsklabs.io"
     }


Vovchyk:
@rmoreliovlabs this can be closed, can't it?

PR #1616, opened 2021-09-21
Updated gradle version to 7.2

rmoreliovlabs:
Updated to Gradle 7 version in order to take advantage of newer tools.

## Description

* Updated gradle wrapper.
* Updated Removed `Jcenter()`.
* Update maven plugin to `maven-publish` instead.
* Added `subprojects` block since `pom {}` is only allowed there.

## Motivation and Context

Upgrade Gradle version.

## How Has This Been Tested?

* I ran ` ./gradlew clean assemble --refresh-dependencies` 
* Finally I ran `gradle build`

## Types of changes

## Checklist:

* **Other information**:


rskj-core/build.gradle
```diff
@@ -204,12 +203,13 @@ task javadocJar(type: Jar) {
 }
 
 jar {
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
```
Vovchyk, line 206: why do we need this?

rmoreliovlabs: Because build fails saying it finds duplicated `build-info.properties` and it doesn't know what to do with it when building the jar. I think this happens due to the version upgrade

Vovchyk: ok. I see. The thing is that we do need some duplicates, for instance, license files. See [here](https://github.com/rsksmart/rskj/blob/master/rskj-core/build.gradle#L291). One thing you can do is as it's done [here](https://github.com/rsksmart/rskj/blob/master/rskj-core/build.gradle#L251).

rskj-core/build.gradle
```diff
@@ -59,72 +58,72 @@ tasks.withType(AbstractArchiveTask) {
...
     libs = [
-            nettyLib: "io.netty:netty-codec-http:${libVersions.nettyVer}",
-            spongyCastleLib: "com.madgag.spongycastle:core:${libVersions.spongyCastleVer}",
-            bouncyCastleLib: "org.bouncycastle:bclcrypto-jdk15on:${libVersions.bouncyCastleVer}",
-            leveldbLib: "org.iq80.leveldb:leveldb:${libVersions.leveldbVer}",
+            nettyLib               : "io.netty:netty-codec-http:${libVersions.nettyVer}",
+            spongyCastleLib        : "com.madgag.spongycastle:core:${libVersions.spongyCastleVer}",
+            bouncyCastleLib        : "org.bouncycastle:bclcrypto-jdk15on:${libVersions.bouncyCastleVer}",
+            leveldbLib             : "org.iq80.leveldb:leveldb:${libVersions.leveldbVer}",
             fusesourceLeveldbJniLib: "org.fusesource.leveldbjni:leveldbjni:${libVersions.fusesourceLeveldbJniVer}",
```
Vovchyk, line 85: could you check if there's a need of `org.iq80.leveldb:leveldb` and `org.fusesource.leveldbjni:leveldbjni` libs in the project? It seems that they are not in use right now.

rmoreliovlabs: Here for `iq80`: https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/org/ethereum/datasource/HashMapDB.java#L23 

Here for `fusesource`: https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/org/ethereum/datasource/LevelDbDataSource.java#L41

Vovchyk: do you think it could be that those two classes are also part of `leveldbjni-all-1.18.3.jar` ?

rmoreliovlabs: It worked removing the these dependencies, already pushed.

Vovchyk (approved):

Vovchyk:
pipeline:run

PR #1619, opened 2021-09-29
Fixed ext_dumpState (it didn't work in Mainnet) / added partial export

SergioDemianLerner:
The function ext_dumpState() didn't work in Mainnet for several reasons:
1) It threw an exception when getCode() returns null
2) It takes too long to finish (more than 30 minutes)
3) It exhaust all available memory

This PR fixes all 3.

## Description

It fixes 1 by returning "" (empty string) when the code is missing.
It fixes 2 and 3 by allowing partial requests.

Three (3) arguments are added:
* String account,
* boolean exportStorageKeys
* boolean exportCode

The argument account, if non-empty, generates a dump limited to a single account. If empty, all accounts are processed.
The argument exportStorageKeys enables or disables the dump of contract storage key/values.
The argument exportCode enables or disables the dump of the contract code.

Also a new field "codeHash" was added to enable first requesting all codeHashes, and then selectively requesting code for those accounts whose code matches a specific hash.

## Motivation and Context
It fixes the method ext_dumpState() which didn't work for Mainnet (and probably for testnet also)

## How Has This Been Tested?

A unit test was added that tests partial dumps.

To iterate over the Trie without getting into the storage cells, a new preorder iterator was added to the Trie class.
Note that it would be recommended that in another PR all iterators are moved away from the Trie class.

## Types of changes

It introduces a breaking change because more arguments were added to dumpState() however since currently this method doesn't work clearly nobody (except me) is using this method, so it is highly improbable that anything breaks with my change.

## Checklist:

- [x ] Tests for the changes have been added (for bug fixes / features)

* **Other information**:

In the future we can support debug_accountRange, which is the way geth enables this functionality.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/NetworkStateExporter.java
+++ b/rskj-core/src/main/java/co/rsk/core/NetworkStateExporter.java
@@ -57,6 +57,10 @@ public class NetworkStateExporter {
     }
 
     public boolean exportStatus(String outputFile) {
+        return exportStatus(outputFile, "",true,true);
+    }
+
+    public boolean exportStatus(String outputFile,String accountKey, boolean exportStorageKeys,boolean exportCode) {
         RepositorySnapshot frozenRepository = repositoryLocator.snapshotAt(blockchain.getBestBlock().getHeader());
 
         File dumpFile = new File(outputFile);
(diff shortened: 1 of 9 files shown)


rskj-core/src/main/java/co/rsk/rpc/Web3RskImpl.java
```diff
@@ -102,10 +102,21 @@ public Web3RskImpl(
...
 
-    public void ext_dumpState() {
+    public void ext_dumpState(String account,boolean exportStorageKeys,boolean exportCode) {
         Block bestBlcock = blockStore.getBestBlock();
-        logger.info("Dumping state for block hash {}, block number {}", bestBlcock.getHash(), bestBlcock.getNumber());
-        networkStateExporter.exportStatus(System.getProperty("user.dir") + "/" + "rskdump.json");
+        logger.info("Dumping state for block hash {}, block number {}, account {}",
+                bestBlcock.getHash(), bestBlcock.getNumber(),account);
+        String name = "rskdump";
+        if (account.length()!=0) {
```
fedejinich, line 110: here you're trying to validate the input, right? I think it can be improved by initializing a  `new RskAddress()` (it has already implemented the validation)

rskj-core/src/main/java/co/rsk/core/NetworkStateExporter.java
```diff
@@ -57,52 +56,90 @@ public NetworkStateExporter(RepositoryLocator repositoryLocator, Blockchain bloc
     }
 
     public boolean exportStatus(String outputFile) {
+        return exportStatus(outputFile, "",true,true);
+    }
+
+    public void exportAllAccounts(ObjectNode mainNode,RepositorySnapshot frozenRepository,boolean exportStorageKeys,boolean exportCode) {
+        for (RskAddress addr : frozenRepository.getAccountsKeys()) {
+            if (!addr.equals(RskAddress.nullAddress())) {
```
fedejinich, line 64: this is duplicated logic, I think you can just define something like this 
```java
private void exportAccount(ObjectNode mainNode, RepositorySnapshot frozenRepository, boolean exportStorageKeys, boolean exportCode, RskAddress addr) {
        if (!addr.equals(RskAddress.nullAddress())) {
            mainNode.set(addr.toString(), createAccountNode(mainNode, addr, frozenRepository, exportStorageKeys, exportCode));
        }
    }
```
and then do the `for` iteration in the `exportStatus`, ending up with something like this:
```java
public boolean exportStatus(String outputFile,String accountKey, boolean exportStorageKeys,boolean exportCode) {
        ...
        try(FileWriter fw = new FileWriter(dumpFile.getAbsoluteFile()); BufferedWriter bw = new BufferedWriter(fw)) {
            ...
            if (accountKey.length() == 0) { // exports all the accounts
                for (RskAddress addr : frozenRepository.getAccountsKeys()) {
                    exportAccount(mainNode, frozenRepository, exportStorageKeys, exportCode, addr);
                }
            } else { // exports just the desired account
                RskAddress addr = new RskAddress(accountKey);
                exportAccount(mainNode, frozenRepository, exportStorageKeys, exportCode, addr);
            }
            ...
            return true;
        } catch (Exception e) {
            ...
            return false;
        }
    }
```

fedejinich (commented):
The overall looks good to me, it fixes the mentioned issues and adds more relevant features for this method.
I agree that even tough it's a breaking change it shouldn't bring any problem on updating it.
Said that I've suggested some minor changes.
Nice catch 👍

fedejinich:
pipeline:run

rskj-core/src/main/java/co/rsk/trie/Trie.java
```diff
@@ -1212,17 +1220,23 @@ public IterationElement next() {
...
-                visiting.push(new IterationElement(rightNodeKey, rightNode));
-            }
-            Trie leftNode = node.retrieveNode((byte) 0x00);
-            if (leftNode != null) {
-                TrieKeySlice leftNodeKey = nodeKey.rebuildSharedPath((byte) 0x00, leftNode.getSharedPath());
-                visiting.push(new IterationElement(leftNodeKey, leftNode));
+
+            int nodeKeyLength = nodeKey.length();
+            // If we're stoping at accounts, do not add children
+            if ((!stopAtAccountDepth) || (nodeKeyLength < (1 + TrieKeyMapper.SECURE_KEY_SIZE + RskAddress.LENGTH_IN_BYTES) * Byte.SIZE))
```
fedejinich, line 1226: you can add some comments to prevent using the iterator in a wrong way

fedejinich:
pipeline:run

rskj-core/src/main/java/co/rsk/core/NetworkStateExporter.java
```diff
@@ -57,52 +56,90 @@ public NetworkStateExporter(RepositoryLocator repositoryLocator, Blockchain bloc
     }
 
     public boolean exportStatus(String outputFile) {
```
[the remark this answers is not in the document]
SergioDemianLerner: I think it is useful as a way to establish which are the default parameters for a full dump.

fedejinich (dismissed):

rskj-core/src/main/java/co/rsk/core/NetworkStateExporter.java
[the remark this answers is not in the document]
SergioDemianLerner: I removed the test case, left the method with default args as documentation.

rskj-core/src/main/java/co/rsk/core/NetworkStateExporter.java
SergioDemianLerner: done!

rskj-core/src/main/java/co/rsk/trie/Trie.java
SergioDemianLerner: done!

SergioDemianLerner: Done!

fedejinich (dismissed):
lgtm! (I've added sonarcloud missing curly braces)

Vovchyk (dismissed):

rmoreliovlabs (approved):

rskj-core/src/main/java/co/rsk/rpc/Web3RskImpl.java
```diff
@@ -102,10 +104,30 @@ public Web3RskImpl(
...
+            RskAddress addrToCheck;
+            try {
+                addrToCheck = new RskAddress(account);
+
+                // Now make sure we don't enable path-traversal vulnerabilities in case
+                // the address format is ever changed.
+                if (account.indexOf(File.separatorChar)>=0) {
```
VaWheel, line 119: `account` is not a part of the `filename` (only its hex representation) so check is probably redundant

VaWheel, line 119: is this check needed since hex representation of address `addrToCheck.toHexString()` is used for filename?

SergioDemianLerner: As the comment explain, It's a second layer of defense in case in the future the RSK address format allows addresses containing '/' or '\'.

VaWheel: In case new format contains a separator char and `addrToCheck.toHexString()` is still used for filename it's ok not to check the separator (as it would be encoded as `2f` or `5c`). In case other encoding of address will be used for filename than yes it may require a check (but this also may become a problem as legitimate addresses containing a separator wouldn't pass the check).

SergioDemianLerner: You may be totally right, but since the code is used only for testing, I think our time discussing this issue is more valuable than the cost of editing the line of code, and the benefit it provides.
So, I'm happy with either removing or leaving the check.

Vovchyk:
pipeline:run

SergioDemianLerner:
rebase looks good.

Vovchyk (approved):

PR #1620, opened 2021-10-01
Added bootstrap node into cli tools

Vovchyk:
## Description

This adds bootstrap node launcher to the list of cli tools

## Motivation and Context

This allows to run an RSK node in the special "bootstrap" mode, which means that the only peer discovery protocol is active, all other capabilities, like syncing, blocks / txs propagation etc., are disabled. This should offload the bootstrap node from other responsibilities and allow it to focus only on helping other peers to boot.

The bootstrap node can be started, for example, via the following command:
```java
java -cp ./rskj-core-3.1.0-SNAPSHOT-all.jar co.rsk.cli.tools.StartBootstrap --testnet
```

## Types of changes

## Checklist:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/Start.java
+++ b/rskj-core/src/main/java/co/rsk/Start.java
@@ -33,11 +33,17 @@ public class Start {
     public static void main(String[] args) {
         setUpThread(Thread.currentThread());
 
+        RskContext ctx = null;
         try {
-            RskContext ctx = new RskContext(args);
+            ctx = new RskContext(args);
             runNode(Runtime.getRuntime(), new PreflightChecksUtils(ctx), ctx);
         } catch (Exception e) {
             logger.error("The RSK node main thread failed, closing program", e);
+
+            if (ctx != null) {
+                ctx.close();
+            }
+
             System.exit(1);
         }
     }
--- a/rskj-core/src/main/java/co/rsk/cli/tools/StartBootstrap.java
+++ b/rskj-core/src/main/java/co/rsk/cli/tools/StartBootstrap.java
@@ -58,20 +58,21 @@ public class StartBootstrap {
                                  @Nonnull PreflightChecksUtils preflightChecks,
                                  @Nonnull Runtime runtime,
                                  @Nonnull NodeStopper nodeStopper) {
-        NodeRunner runner = ctx.getNodeRunner();
         try {
             // make preflight checks
             preflightChecks.runChecks();
 
-            // start node runner
-            runner.run();
-
             // subscribe to shutdown hook
-            runtime.addShutdownHook(new Thread(runner::stop, "stopper"));
+            runtime.addShutdownHook(new Thread(ctx::close, "stopper"));
+
+            // start node runner
+            NodeRunner runner = ctx.getNodeRunner();
+            runner.run();
         } catch (Exception e) {
             logger.error("Main thread of RSK bootstrap node crashed", e);
 
-            runner.stop();
+            ctx.close();
+
             nodeStopper.stop(1);
         }
     }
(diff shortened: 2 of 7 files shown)


Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

rskj-core/src/main/java/co/rsk/Start.java
```diff
@@ -33,11 +33,17 @@
...
+            ctx = new RskContext(args);
             runNode(Runtime.getRuntime(), new PreflightChecksUtils(ctx), ctx);
         } catch (Exception e) {
             logger.error("The RSK node main thread failed, closing program", e);
+
+            if (ctx != null) {
+                ctx.close();
+            }
+
             System.exit(1);
```
nagarev, line 47: Maybe we can put all of this inside a finally block?

nagarev:
Loved it! LGTM

nagarev (dismissed):
LGTM

rskj-core/src/main/java/co/rsk/Start.java
Vovchyk: we should close it only if an exception has been thrown. If no exception, we the context should stay non-closed until a process finishes.

joaquinlpereyra: There's a runtime hook in `runNode` making sure the context is closed on normal shutdown. Actually, I think in case of most exceptions, it will be called twice, but should be ok because it is idempotent.

rskj-core/src/main/java/co/rsk/cli/tools/StartBootstrap.java
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Entry point of RSK bootstrap node.
+ *
+ * The bootstrap node starts one service which only participates in the peer discovery protocol. All other capabilities
+ * (which are usually part of a full node, e.g. blocks propagation, syncing, JSON-RPC interface etc.) are disabled.
+ *
+ * Note: this is an experimental tool
+ */
+public class StartBootstrap {
```
joaquinlpereyra, line 42: I would probably start up the `JSON-RPC` service here too, just because I think it is really helpful for debugging and to get information on the nodes. What do you think?

joaquinlpereyra (changes requested):

rskj-core/src/main/java/co/rsk/cli/tools/StartBootstrap.java
Vovchyk: > There's a runtime hook in runNode making sure the context is closed on normal shutdown. Actually, I think in case of most exceptions, it will be called twice, but should be ok because it is idempotent.

`preflightChecks.runChecks()` may also crash, which is triggered just before registering a hook. You're right, `close()` can be called multiple times without side effects.

> I would probably start up the JSON-RPC service here too, just because I think it is really helpful for debugging and to get information on the nodes. What do you think?

The ones that provide useful information (e.g. `net_peerCount` and `net_peerList`) rely on `ChannelManager`, which under the hood returns "active" peers (the ones that serve other nodes, e.g. send blocks or txs). The idea here is to avoid usage of such services. I guess, if we really need such debugging capabilities, we may be needing separate methods, which would simply return neighbours of a node (those that are connected via the UDP PING-PONG roundtrip).

joaquinlpereyra: TIL. I had assumed `peer{Count,List}` returned the peers, active or not. Disregard my comment then 👍🏿

heitorfm (approved):

PR #1631, opened 2021-10-20
Added RSKIP 271 Activation Code

kelvinator07:


rskj-core/src/main/resources/reference.conf
```diff
@@ -56,6 +56,7 @@ blockchain = {
             rskip218 = iris300
             rskip219 = iris300
             rskip220 = iris300
+            rskip271 = iris300
```
josedahlquist, line 59: You have to create a new network upgrade (for the moment call it `400`) and make sure to set its activation height as `-1` in all networks

PR #1632, opened 2021-10-20
Remove unused TriePutResult class

fedejinich:
This class (created in 2017) is not used by anyone. It should be deleted


fedejinich:
pipeline:run

PR #1633, opened 2021-10-20
Added new BridgeConstant to define the amount of blocks between peg-outs

kelvinator07:


rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
```diff
@@ -163,4 +165,12 @@ public String getOldFederationAddress() {
     public int getMinimumPegoutValuePercentageToReceiveAfterFee() {
         return minimumPegoutValuePercentageToReceiveAfterFee;
     }
+
+    public int getNumberOfBlocksBetweenPegouts() {
+        return numberOfBlocksBetweenPegouts;
+    }
+
+    public void setNumberOfBlocksBetweenPegouts(int numberOfBlocksBetweenPegouts) {
```
marcos-iov, line 173: No need for a setter, this are constant values

rskj-core/src/main/java/co/rsk/config/BridgeDevNetConstants.java
```diff
@@ -157,5 +157,7 @@ public BridgeDevNetConstants(List<BtcECKey> federationPublicKeys) {
         maxDepthBlockchainAccepted = 25;
 
         minimumPegoutValuePercentageToReceiveAfterFee = 20;
+
+        numberOfBlocksBetweenPegouts = 18; // 3 hours of BTC blocks (considering 1 block every 10 minutes)
```
marcos-iov, line 161: It's RSK blocks, so 1 block every 30 seconds

rskj-core/src/main/java/co/rsk/config/BridgeRegTestConstants.java
```diff
@@ -159,6 +159,8 @@ public BridgeRegTestConstants(List<BtcECKey> federationPublicKeys) {
         oldFederationAddress = "2N7ZgQyhFKm17RbaLqygYbS7KLrQfapyZzu";
 
         minimumPegoutValuePercentageToReceiveAfterFee = 20;
+
+        numberOfBlocksBetweenPegouts = 18; // 3 hours of BTC blocks (considering 1 block every 10 minutes)
```
marcos-iov, line 163: In the case of regtest maybe a small value would be a good idea, so we can quickly test pegouts. I'd say around 50 blocks

marcos-iov: In regtest the mining is really fast, we can also disable automatic mining and advance the network at demand. This is useful in integration tests for example

PR #1654, opened 2021-11-11
Changed difficulty min value for testnet

rmoreliovlabs:
Difficulty drop in Testnet should drop to a higher difficulty (current one is too low)

## Description

 Testnet it's processing a lot a blocks when the following condition happens:

1. Miner with a lot of hardware resources gets in.
2. Block Validation Difficulty Increases.
3. Miner with a lot of hardware leaves the net.
4. Difficulty drops really fast at a point that Miner with low resources can process blocks every second

## Motivation and Context
This change will ensure Testnet consensus to happen so fast

## How Has This Been Tested?
Added 2 tests for it

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/config/Constants.java
+++ b/rskj-core/src/main/java/org/ethereum/config/Constants.java
@@ -50,10 +50,34 @@ public class Constants {
     private final boolean seedCowAccounts;
     private final int durationLimit;
     private final BlockDifficulty minimumDifficulty;
+    private final BlockDifficulty minimumDifficultyForRskip290;
     private final BlockDifficulty fallbackMiningDifficulty;
     private final BigInteger difficultyBoundDivisor;
     private final int newBlockMaxSecondsInTheFuture;
     public final BridgeConstants bridgeConstants;
+    private final ActivationConfig activationConfig;
+
+    public Constants(
+            byte chainId,
+            boolean seedCowAccounts,
+            int durationLimit,
+            BlockDifficulty minimumDifficulty,
+            BlockDifficulty fallbackMiningDifficulty,
+            BigInteger difficultyBoundDivisor,
+            int newBlockMaxSecondsInTheFuture,
+            BridgeConstants bridgeConstants,
+            ActivationConfig activationConfig) {
+        this.chainId = chainId;
+        this.seedCowAccounts = seedCowAccounts;
+        this.durationLimit = durationLimit;
+        this.minimumDifficulty = minimumDifficulty;
+        this.fallbackMiningDifficulty = fallbackMiningDifficulty;
+        this.difficultyBoundDivisor = difficultyBoundDivisor;
+        this.newBlockMaxSecondsInTheFuture = newBlockMaxSecondsInTheFuture;
+        this.bridgeConstants = bridgeConstants;
+        this.activationConfig = activationConfig;
+        this.minimumDifficultyForRskip290 = new BlockDifficulty(new BigInteger("550000000"));
+    }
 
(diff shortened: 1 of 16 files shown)


rskj-core/src/main/java/org/ethereum/config/Constants.java
```diff
@@ -50,10 +50,12 @@
     private final boolean seedCowAccounts;
     private final int durationLimit;
     private final BlockDifficulty minimumDifficulty;
+    private final BlockDifficulty minimumDifficultyForRskip290 = new BlockDifficulty(new BigInteger("550000000"));
```
joaquinlpereyra, line 53: This works, but it deviates from the way the other variables are set. I would consider setting it in the `testnet` constructor.

rmoreliovlabs: Done. Updated.

rskj-core/src/main/java/org/ethereum/config/Constants.java
```diff
@@ -84,7 +109,13 @@ public int getDurationLimit() {
     }
 
     public BlockDifficulty getMinimumDifficulty() {
```
Vovchyk, line 111: this method is not in use anymore, can we delete it?

rskj-core/src/main/resources/config/main.conf
```diff
@@ -9,6 +9,7 @@ blockchain.config {
         twoToThree = 2018000,
         papyrus200 = 2392700
         iris300 = 3614800
+        hop400 = 0
```
Vovchyk, line 12: until we have a final activation number, we can put `-1` here. Otherwise, it'd be activated starting from block `0`, which is a consensus break

rskj-core/src/main/resources/config/testnet.conf
```diff
@@ -8,7 +8,8 @@ blockchain.config {
         wasabi100 = 0,
         twoToThree = 504000,
         papyrus200 = 863000,
-        iris300 = 2060500
+        iris300 = 2060500,
+        hop400 = 0
```
Vovchyk, line 12: `-1` ?

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/config/Constants.java
Vovchyk: I might be wrong, but it seems to me that @joaquinlpereyra meant [this](https://github.com/rsksmart/rskj/pull/1654/files#diff-f405fc1a0176d94e7c3c2572e30caaadbd8b571f2cac6892358bb0948de6c96aR252) place

rskj-core/src/main/java/org/ethereum/config/Constants.java
```diff
@@ -210,11 +242,16 @@ public static Constants devnetWithFederation(List<BtcECKey> federationPublicKeys
...
                 BigInteger.valueOf(50),
                 540,
-                new BridgeDevNetConstants(federationPublicKeys)
+                new BridgeDevNetConstants(federationPublicKeys),
+                new BlockDifficulty(new BigInteger("550000000"))
         );
     }
 
     public static Constants testnet() {
+        return testnet(null);
```
joaquinlpereyra, line 251: This constructor is used only in tests and looks dangerous to use outside of them (will always return previous difficulty, even when activation has passed). Maybe delete it?

rmoreliovlabs:
pipeline:run

Vovchyk (approved):

PR #1658, opened 2021-11-15
Storage Rent: Node tracking

fedejinich:
This PR adds a mechanism to track all the unitrie nodes involved in each transaction execution.

To do so, I've introduced a new `RentManager` which is initialized in each new transaction. The tracking mechanism it's done by calling the `RentManager` with the desired key after each `put(key)` or `find(key)` operation. 
This is not the most performant approach, it duplicates the search up to the desired key. Ideally, it should track each node without the need to invoke the `findNodes(key)` method.

DIAGRAM WILL BE INCLUDED


smishraIOV:
Looks good! Easy to follow. I added some comments (for own understanding and discussion). Made a couple of minor changes to some test assertions. Pushed to new branch ([in personal repo](https://github.com/smishraIOV/rskj/tree/sr-rentManager-shree)).

fedejinich:
pipeline:run

fedejinich:
note: code smells are related to some non-relevant parts of rskj (will be tackled before merging it). So, there's no need to worry about it

Issue #1458, opened 2021-02-23
Inconsistent number conversion processing transactions


ajlopez:
They are some JSON RPC methods that process transactions with data provided as JSON, like `eth_sendTransaction`, `eth_call`, `eth_estimateGas` (the method `eth_sendRawTransaction` receives the encoded transaction in an hexadecimal string)

Let's take one of this arguments: `gas`

It comes as a string, like `"1000000"`, `"0xa000"`, but also `"a000"` is accepted in SOME methods (not all, others reject this value without `0x` prefix). This is the inconsistency.

In the method `EthModule.call` 

```java
public String call(Web3.CallArguments args, String bnOrId)
```

an internal  method is used: `callConstant`:

```java
    private ProgramResult callConstant(Web3.CallArguments args, Block executionBlock) {
        CallArgumentsToByteArray hexArgs = new CallArgumentsToByteArray(args);
        return reversibleTransactionExecutor.executeTransaction(
                executionBlock,
                executionBlock.getCoinbase(),
                hexArgs.getGasPrice(),
                hexArgs.getGasLimit(),
                hexArgs.getToAddress(),
                hexArgs.getValue(),
                hexArgs.getData(),
                hexArgs.getFromAddress()
        );
    }
```

The `args.gas` field is string. The `hexArgs.getGasLimit()` code is:

```java
    public byte[] getGasLimit() {
        // maxGasLimit value is 100000000000000
        String maxGasLimit = "0x5AF3107A4000";
        byte[] gasLimit = stringHexToByteArray(maxGasLimit);
        if (args.gas != null && args.gas.length() != 0) {
            gasLimit = stringHexToByteArray(args.gas);
        }

        return gasLimit;
    }
```

But NOTABLY, the `stringHexToByteArray` method converts the numeric string (ie `"1000"`) AS HEXADECIMAL:

```java
    public static byte[] stringHexToByteArray(String x) {
        String result = x;
        if (x.startsWith("0x")) {
            result = x.substring(2);
        }
        if (result.length() % 2 != 0) {
            result = "0" + result;
        }
        return Hex.decode(result);
    }
```

So, the result is a byte array with `{ 0x10, 0x00 }`

This contrasts with the send transaction behavior. `eth_sendTransaction` receives the same `args` but it processes them differently. Internally, it executes `EthModuleTransactionBase.sendTransaction`, that use `TypeConverter.stringNumberAsBigInt(args.gas)` to convert the gas string.

This is the code of the conversion method:

[...]

PR #1659, opened 2021-11-17
JSON RPC methods converting args more consistently

rmoreliovlabs:
There are some JSON RPC methods that process transactions with data provided as JSON, like eth_sendTransaction, eth_call, eth_estimateGas (the method eth_sendRawTransaction receives the encoded transaction in an hexadecimal string)

## Description
They are some JSON RPC methods that process transactions with data provided as JSON, like `eth_sendTransaction`, `eth_call`, `eth_estimateGas` (the method `eth_sendRawTransaction` receives the encoded transaction in an hexadecimal string)

Let's take one of this arguments: `gas`

It comes as a string, like `"1000000"`, `"0xa000"`, but also `"a000"` is accepted in SOME methods (not all, others reject this value without `0x` prefix). This is the inconsistency.

In the method `EthModule.call` 

```java
public String call(Web3.CallArguments args, String bnOrId)
```

an internal  method is used: `callConstant`:

```java
    private ProgramResult callConstant(Web3.CallArguments args, Block executionBlock) {
        CallArgumentsToByteArray hexArgs = new CallArgumentsToByteArray(args);
        return reversibleTransactionExecutor.executeTransaction(
                executionBlock,
                executionBlock.getCoinbase(),
                hexArgs.getGasPrice(),
                hexArgs.getGasLimit(),
                hexArgs.getToAddress(),
                hexArgs.getValue(),
                hexArgs.getData(),
                hexArgs.getFromAddress()
        );
    }
```

The `args.gas` field is string. The `hexArgs.getGasLimit()` code is:

```java
    public byte[] getGasLimit() {
        // maxGasLimit value is 100000000000000
        String maxGasLimit = "0x5AF3107A4000";
        byte[] gasLimit = stringHexToByteArray(maxGasLimit);
        if (args.gas != null && args.gas.length() != 0) {
            gasLimit = stringHexToByteArray(args.gas);
        }

        return gasLimit;
    }
```

But NOTABLY, the `stringHexToByteArray` method converts the numeric string (ie `"1000"`) AS HEXADECIMAL:

```java
    public static byte[] stringHexToByteArray(String x) {
        String result = x;
        if (x.startsWith("0x")) {
            result = x.substring(2);
        }
        if (result.length() % 2 != 0) {
            result = "0" + result;
        }
        return Hex.decode(result);
    }
```

So, the result is a byte array with `{ 0x10, 0x00 }`

This contrasts with the send transaction behavior. `eth_sendTransaction` receives the same `args` but it processes them differently. Internally, it executes `EthModuleTransactionBase.sendTransaction`, that use `TypeConverter.stringNumberAsBigInt(args.gas)` to convert the gas string.

This is the code of the conversion method:

```java
    public static BigInteger stringNumberAsBigInt(String input) {
        if (input.startsWith("0x")) {
            return TypeConverter.stringHexToBigInteger(input);
        } else {
            return TypeConverter.stringDecimalToBigInteger(input);
        }
    }
```

IT TAKES INTO ACCOUNT THE PREFIX 0x, so, `"1000"` will be parse AS DECIMAL, not hexadecimal.

This inconsistency does not affect the consensus, and apparently only affects non-critical methods (query methods). 

TO CHECK: other fields conversion, like `args.gasPrice` 

_Extracted from: https://github.com/rsksmart/rskj/issues/1458_

## Motivation and Context
It will add more consistency at the moment of parsing args from our RPC methods

## How Has This Been Tested?
Yes added some tests and also tested on postman

## Types of changes

## Checklist:

* **Other information**:


rmoreliovlabs:
pipeline:run

Vovchyk (dismissed):

rskj-core/src/main/java/org/ethereum/rpc/TypeConverter.java
```diff
@@ -67,6 +68,31 @@ private static BigInteger stringDecimalToBigInteger(String input) {
         return Hex.decode(result);
     }
 
+    public static byte[] strHexOrStrNumberToByteArray(String strHexOrStrNumber) {
```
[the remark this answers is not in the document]
rmoreliovlabs: Comment addressed

rskj-core/src/main/java/org/ethereum/rpc/TypeConverter.java
```diff
@@ -82,11 +108,11 @@ public static String toJsonHex(byte[] x) {
     }
 
     public static String toJsonHex(Coin x) {
-        return x != null ? x.asBigInteger().toString() : "" ;
+        return x != null ? x.asBigInteger().toString() : "";
```
joaquinlpereyra, line 111: I believe this should `0x0` in case `x == null`.

rskj-core/src/main/java/org/ethereum/rpc/TypeConverter.java
```diff
@@ -82,11 +108,11 @@ public static String toJsonHex(byte[] x) {
...
     }
 
     public static String toJsonHex(String x) {
-        return "0x"+x;
+        return "0x" + x;
```
joaquinlpereyra, line 115: Maybe add a sanity check here that `x` is actually a number in hex? `Character.digit` could be useful. Or convert string to big integer and just call `toQuantityJsonHex(BigInteger n)`.

rskj-core/src/main/java/org/ethereum/rpc/TypeConverter.java
VaWheel: should we also prepend "0x" as BigInteger.toString() returns hex without "0x" ?

rskj-core/src/main/java/org/ethereum/rpc/TypeConverter.java
rmoreliovlabs: Updated

rskj-core/src/main/java/org/ethereum/rpc/TypeConverter.java
rmoreliovlabs: Updated

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

rmoreliovlabs:
pipeline:run

Vovchyk (approved):

PR #1685, opened 2022-01-05
Created checktimeLockExpiration method in BridgeSupport

kelvinator07:


rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -133,6 +136,9 @@
     private FastBridgeFederationInformation fastBridgeFederationInformationsToSave = null;
     private long receiveHeadersLastTimestamp = 0;
 
+    private Long nextUtxoExpirationCheckpointHeight;
```
josedahlquist, line 139: I suggest using long instead of Long as we should never receive a null value here.

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -884,6 +890,53 @@ public void saveReceiveHeadersLastTimestamp() {
...
+
+        if (nextUtxoExpirationCheckpointHeight == null) {
+            nextUtxoExpirationCheckpointHeight = safeGetFromRepository(NEXT_UTXO_EXPIRATION_CHECKPOINT_HEIGHT, BridgeSerializationUtils::deserializeOptionalLong).orElse(0L);
+        }
+
+        return Optional.of(nextUtxoExpirationCheckpointHeight);
+    }
+
+    public void setNextUtxoExpirationCheckpointHeight(Long nextUtxoExpirationCheckpointHeight) {
+        if (activations.isActive(RSKIP264) && nextUtxoExpirationCheckpointHeight != null) {
```
josedahlquist, line 906: I suggest changing the type to long to discard possible null values, and instead check that the value is always positive.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -2499,6 +2500,17 @@ public long getActiveFederationCreationBlockHeight() {
         return activeFederationCreationBlockHeightOpt.orElse(0L);
     }
 
+    public void checktimeLockExpiration() {
```
josedahlquist, line 2503: this should probably be protected instead

josedahlquist (approved):
Please consider changing the datatypes of the storage to long. Not blocker though.
Good work!

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
kelvinator07: Using long Will Require Me To Initialise It To Zero
Then Check If Its > 0 Before Setting Or Saving To Storage.. Is That Okay?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
kelvinator07: Done

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
marcos-iov: It will initialize to `0` by default, then just check value > 0 in the setter. Makes it simpler

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -2499,6 +2500,17 @@ public long getActiveFederationCreationBlockHeight() {
         return activeFederationCreationBlockHeightOpt.orElse(0L);
     }
 
+    protected void checktimeLockExpiration() {
+        if (activations.isActive(RSKIP264)) {
+            long height = provider.getNextUtxoExpirationCheckpointHeight().orElse(0L);
```
marcos-iov, line 2505:
```suggestion
            long nextUtxoExpirationCheckpointHeight = provider.getNextUtxoExpirationCheckpointHeight().orElse(0L);
```

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -2499,6 +2500,17 @@ public long getActiveFederationCreationBlockHeight() {
...
+            long timestamp = provider.getNextUtxoExpirationCheckpointTimestamp().orElse(0L);
```
marcos-iov, line 2506:
```suggestion
            long nextUtxoExpirationCheckpointTimestamp = provider.getNextUtxoExpirationCheckpointTimestamp().orElse(0L);
```
Feel free to use other names, but let's try to use something meaningful

kelvinator07: Resolved

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
kelvinator07: Resolved

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
kelvinator07: Resolved

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
kelvinator07: Resolved

marcos-iov:
This needs rebase now

PR #1687, opened 2022-01-05
Sender cache on tx validation

nagarev:
Adding cache for block transaction validation

## Description

While some time ago a caché for tx.getSender was introduced, it’s not being used in block txs validation

## Motivation and Context

A cache here would improve node performance on transaction validations

## How Has This Been Tested?

- Unit Tests
- Performance Tests

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/test/java/co/rsk/core/bc/BlockValidatorBuilder.java
+++ b/rskj-core/src/test/java/co/rsk/core/bc/BlockValidatorBuilder.java
@@ -26,6 +26,7 @@ import co.rsk.db.StateRootsStoreImpl;
 import co.rsk.trie.TrieStore;
 import co.rsk.util.TimeProvider;
 import co.rsk.validators.*;
+import org.ethereum.core.SignatureCache;
 import org.ethereum.datasource.HashMapDB;
 import org.ethereum.db.BlockStore;
 import org.mockito.Mockito;
@@ -76,7 +77,7 @@ public class BlockValidatorBuilder {
         this.blockTxsValidationRule = new BlockTxsValidationRule(new RepositoryLocator(
                 trieStore,
                 new StateRootHandler(config.getActivationConfig(), new StateRootsStoreImpl(new HashMapDB()))
-        ));
+        ), null); // TODO -> Set a useful mock here
         return this;
     }
 
--- a/rskj-core/src/main/java/co/rsk/RskContext.java
+++ b/rskj-core/src/main/java/co/rsk/RskContext.java
@@ -1055,7 +1055,7 @@ public class RskContext implements NodeContext, NodeBootstrapper {
             Constants commonConstants = getRskSystemProperties().getNetworkConstants();
             blockParentDependantValidationRule = new BlockParentCompositeRule(
                     new BlockTxsFieldsValidationRule(),
-                    new BlockTxsValidationRule(getRepositoryLocator()),
+                    new BlockTxsValidationRule(getRepositoryLocator(), getBlockTxSignatureCache()),
                     new PrevMinGasPriceRule(),
                     new BlockParentNumberRule(),
                     new BlockDifficultyRule(getDifficultyCalculator()),
--- a/rskj-core/src/main/java/co/rsk/validators/BlockTxsValidationRule.java
+++ b/rskj-core/src/main/java/co/rsk/validators/BlockTxsValidationRule.java
@@ -23,6 +23,7 @@ import co.rsk.db.RepositoryLocator;
 import co.rsk.db.RepositorySnapshot;
 import co.rsk.panic.PanicProcessor;
 import org.ethereum.core.Block;
+import org.ethereum.core.SignatureCache;
 import org.ethereum.core.Transaction;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -47,9 +48,11 @@ public class BlockTxsValidationRule implements BlockParentDependantValidationRul
     private static final PanicProcessor panicProcessor = new PanicProcessor();
 
     private final RepositoryLocator repositoryLocator;
+    private final SignatureCache signatureCache;
 
-    public BlockTxsValidationRule(RepositoryLocator repositoryLocator) {
+    public BlockTxsValidationRule(RepositoryLocator repositoryLocator, SignatureCache signatureCache) {
         this.repositoryLocator = repositoryLocator;
+        this.signatureCache = signatureCache;
     }
 
     @Override
@@ -77,7 +80,7 @@ public class BlockTxsValidationRule implements BlockParentDependantValidationRul
                 return false;
             }
 
-            RskAddress sender = tx.getSender();
+            RskAddress sender = signatureCache != null ? tx.getSender(signatureCache) : tx.getSender();
             BigInteger expectedNonce = curNonce.get(sender);
             if (expectedNonce == null) {
                 expectedNonce = parentRepo.getNonce(sender);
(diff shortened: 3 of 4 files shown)


nagarev:
pipeline:run

nagarev:
pipeline:run

rskj-core/src/main/java/org/ethereum/core/BlockTxSignatureCache.java
```diff
@@ -21,53 +21,51 @@
...
 
     public BlockTxSignatureCache(SignatureCache internalCache) {
+        super(MAX_CACHE_SIZE, false);
         this.internalCache = internalCache;
-        addressesCache = new MaxSizeHashMap<>(MAX_CACHE_SIZE,false);
     }
 
     @Override
-    public RskAddress getSender(Transaction transaction) {
+    public synchronized RskAddress getSender(Transaction transaction) {
```
illuque-iov, line 37: If I'm getting it correctly, this synchronization is not required. After removing the manual initialization of `addressesCache`, both it and `internalCache` are initialised with `Collections.synchronizedMap` and you are not iterating over a "collection-converted" version of the map.

Am I missing a scenario when this is required maybe?

Vovchyk: I guess `Collections.synchronizedMap()` would only make those two collections synchronised, meaning their operations, like `put()` or `remove()`, would be synchronised and thread-safe, but I believe the idea here is to make the entire `getSender` method synchronised, so no race conditions will occur in there

illuque-iov: Yes, what you mention about the collections is correct.

It is the race condition what I am not able to spot, or at least a problematic one, but most probably I'm failing to find it :)

The only thing I can imagine is the time between `addressesCache.get` and `addressesCache.put`, where two simultaneous threads could consider the address as new and both would compute the sender. Not sure what is better in this case, this extra computation or the synchronization penalty, but I guess this was measured or considered.

nagarev: Yup, that was the idea. Maybe we can get rid of this modifier on `getSender` for `ReceivedTxSignatureCache` since there are no "cache write" inside it so a race condition would not occur, but in the rest of the cases (`storeSender` method on `ReceivedTxSignatureCache` and `storeSender` and `getSender` methods on `BlockTxSignatureCache`) should be protected against race conditions, since both methods do write to the cache. Maybe the worst case scenario is not that bad, but better to avoid it.

nagarev:
pipeline:run

rskj-core/src/main/java/org/ethereum/core/BlockTxSignatureCache.java
Vovchyk: those collections are instances the `MaxSizeHashMap` class, which has a side affect in the `put` method - it removes older items if the cache is full.

For instance, [this](https://github.com/rsksmart/rskj/blob/IRIS-3.1.0/rskj-core/src/main/java/org/ethereum/core/BlockTxSignatureCache.java#L50-L54) sort of code would be problematic in multi-threaded environments, as `addressesCache.put()` may delete an item in one thread, so `internalCache.getSender(transaction)` would return `null` in some other one, which obviously breaks the logic.

Also by having `SignatureCache` class synchronised at its own level, I think we would not need to make those two collections synchronised.

In general, I personally prefer to have such kind of "safeguards" in place, especially when behaviour methods or its dependencies may change later.

Vovchyk: I'd also rewrite [this](https://github.com/rsksmart/rskj/blob/IRIS-3.1.0/rskj-core/src/main/java/org/ethereum/core/BlockTxSignatureCache.java#L50-L54) kind of blocks - we could only call `internalCache.getSender()` and check its result for `null` ( without calling `internalCache.containsTx()` )

illuque-iov: I see your point in the [previous message](https://github.com/rsksmart/rskj/pull/1687#discussion_r783722101) @Vovchyk 👍

nagarev: I've made an additional refactor regarding [this](https://github.com/rsksmart/rskj/pull/1687#discussion_r783728513).

rskj-core/src/main/java/org/ethereum/core/SignatureCache.java
```diff
@@ -20,27 +20,35 @@
...
+import co.rsk.util.MaxSizeHashMap;
 
+import java.util.Collections;
 import java.util.Map;
 
 public abstract class SignatureCache {
-    protected Map<Transaction, RskAddress> addressesCache;
 
-    public abstract RskAddress getSender(Transaction transaction);
+    protected Map<Keccak256, RskAddress> addressesCache;
```
Vovchyk, line 32: as an option, `addressesCache` could be marked as `final` - to emphasise that its initialisation should be done in the ctor

rskj-core/src/main/java/org/ethereum/core/SignatureCache.java
```diff
@@ -20,27 +20,35 @@
...
 
-    public boolean containsTx(Transaction transaction) {
-        return addressesCache.containsKey(transaction);
+    public SignatureCache(int maxCacheSize, boolean accessOrder) {
+        addressesCache = Collections.synchronizedMap(
```
Vovchyk, line 35: just thinking that we may not need this collection to be synchronized anymore?

rskj-core/src/test/java/co/rsk/core/bc/BlockValidatorBuilder.java
```diff
@@ -76,7 +77,7 @@ public BlockValidatorBuilder addBlockTxsValidationRule(TrieStore trieStore) {
         this.blockTxsValidationRule = new BlockTxsValidationRule(new RepositoryLocator(
                 trieStore,
                 new StateRootHandler(config.getActivationConfig(), new StateRootsStoreImpl(new HashMapDB()))
-        ));
+        ), null); // TODO -> Set a useful mock here
```
Vovchyk, line 80: good point )

Vovchyk (approved):

Vovchyk (approved):

rskj-core/src/main/java/org/ethereum/core/SignatureCache.java
nagarev: If we're. enforcing the sync via the method modifiers, we can get rid of this I think, yes.

illuque-iov (dismissed):

rskj-core/src/test/java/co/rsk/core/bc/BlockValidatorBuilder.java
nagarev: We should! 😂 Done!

rskj-core/src/main/java/org/ethereum/core/SignatureCache.java
nagarev: Updated!

rskj-core/src/main/java/org/ethereum/core/SignatureCache.java
nagarev: Done!

nagarev:
pipeline:run

nagarev:
pipeline:run

illuque-iov (approved):

nagarev:
pipeline:run

nagarev:
pipeline:run

Vovchyk (approved):

Issue #1582, opened 2021-07-06
RemascTransaction is not always considered as RemascTransaction


fedejinich:
There are some cases when we want to check if a transaction is a REMASC transaction, to do so we use `isRemascTransaction()` method from Transaction class.

```java
enum MessageType {
...
BODY_RESPONSE_MESSAGE(15) {
        @Override
        public Message createMessage(BlockFactory blockFactory, RLPList list) {
        ...
        List<Transaction> transactions = new ArrayList<>();
        for (int k = 0; k < rlpTransactions.size(); k++) {
            byte[] txdata = rlpTransactions.get(k).getRLPData();
            Transaction tx = new ImmutableTransaction(txdata);

            if (tx.isRemascTransaction(k, rlpTransactions.size())) {
                tx = new RemascTransaction(txdata);
            }

            transactions.add(tx);
        }
        ...
 },
...
}
```
But there are also some other cases when we check this just by using an `instanceof RemascTransaction`

```java
class MinnerUtil {
...
public List<org.ethereum.core.Transaction> filterTransactions(List<Transaction> txsToRemove, List<Transaction> txs, Map<RskAddress, BigInteger> accountNonces, RepositorySnapshot originalRepo, Coin minGasPrice) {
       ...
        if (!(tx instanceof RemascTransaction) && tx.getGasPrice().compareTo(minGasPrice) < 0) {
            logger.warn("Rejected tx={} because of low gas account {}, removing tx from pending state.", hash, txSender);

            txsToRemove.add(tx);
            continue;
        }
        ...
    }
...
}
```
Nowadays if a `RemascTransaction` is created with the `blockNumber` contrstructor, then `isRemascTransaction()` will always return false. This particular case it's caused because the data provided to this transaction it's setted by using and "unsafe" constructor which relies on `BytesUtils.cloneBytes`, preventing nulls for the data field and producing unexpected behavior.

```java
class Transaction {
    protected Transaction(byte[] nonce, Coin gasPriceRaw, byte[] gasLimit, RskAddress receiveAddress, Coin valueRaw, byte[] data,
                byte chainId, final boolean localCall) {
        this.nonce = ByteUtil.cloneBytes(nonce);
        this.gasPrice = gasPriceRaw;
        this.gasLimit = ByteUtil.cloneBytes(gasLimit);
        this.receiveAddress = receiveAddress;
        this.value = valueRaw;
        this.data = ByteUtil.cloneBytes(data);
        this.chainId = chainId;
[...]

PR #1688, opened 2022-01-06
Documented RemascTransaction check usage across the platform

rmoreliovlabs:
Documented RemascTransaction check usage across the platform

## Description
Basically we are documenting the usage of:

```
obj instanceof RemascTransaction
```
and 
```
transaction.isRemascTransaction(x, y)
```

Since they are used on multiple scenarios but with the goal, we need to make sure things are clear.

## Motivation and Context
This issue https://github.com/rsksmart/rskj/issues/1582, wants to refactor the check to determine whether a transaction is remasc or not but the refactor is not worth since it's being used inconsistently, but not in a bad way, in multiple places.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/remasc/RemascTransaction.java
+++ b/rskj-core/src/main/java/co/rsk/remasc/RemascTransaction.java
@@ -28,7 +28,19 @@ import org.ethereum.vm.PrecompiledContracts;
 import static org.ethereum.rpc.TypeConverter.toUnformattedJsonHex;
 
 /**
+ * <p>
  * Tx that invokes Remasc's processMinersFees method.
+ * </p>
+ * <p>
+ * If you would like an easier to know if a transaction is a Remasc transaction rather than using
+ * #{@link Transaction#isRemascTransaction(int, int)}, you could use:
+ * <blockquote>obj instanceOf RemascTransaction</blockquote>
+ * </p>
+ * <p>
+ * <b>NOTE:</b> Consider that this only will work if you create an instance of this class, otherwise using
+ * #{@link Transaction#isRemascTransaction(int, int)} is more safe.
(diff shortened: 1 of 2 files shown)


Vovchyk (dismissed):

rskj-core/src/main/java/co/rsk/remasc/RemascTransaction.java
```diff
@@ -28,7 +28,19 @@
 import static org.ethereum.rpc.TypeConverter.toUnformattedJsonHex;
 
 /**
+ * <p>
  * Tx that invokes Remasc's processMinersFees method.
+ * </p>
+ * <p>
+ * If you would like an easier to know if a transaction is a Remasc transaction rather than using
```
joaquinlpereyra, line 35: It is not clear to me when should I use one or the other 😕

joaquinlpereyra: No concerns. I just think this could be explained in the docs:

- Use `Transaction.isRemascTransaction` if you have not yet parsed the transaction and it is still in binary format. 
- Use `instance of` to know if a parsed transaction is a remasc transaction.

Note that `instance of` may return true even though `isRemascTransaction` is false... this is the most confusing case and the one I think could be clearer.

rmoreliovlabs: Thanks for the explanation, updated the PR, please let me know if that works.

rmoreliovlabs:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk (approved):

rmoreliovlabs:
pipeline:run

PR #1689, opened 2022-01-10
Allow max 1 connection per host on PeerExplorer keeping last one

illuque-iov:
## Description

Added a configuration flag in order to allow or reject multiple connections of the same host (IP+port) to PeerExplorer (and, as a consequence, to DistanceTable).

When multiple Pings from the same host with different _nodeId_ are received, we have two scenarios depending on the mentioned flag value:
1. If set to _true_: the current behaviour remains, therefore multiple connections are allowed for same host (as many as different _nodeIds_ are received
2. If set to _false_: just the last connection remains as active, the other ones are disconnected / removed

## Motivation and Context

Related issue https://rsklabs.atlassian.net/browse/CORE-648

The issue is based on the fact that attacker can establish many connections using single IP:PORT and that _sendPing_ sends only one PING request to a single _IP:PORT_ at a time. Therefore, attacker can fill _distanceTable_ using different _NodeId_ and maintain connections with only one _PONG_ at a time.

## How Has This Been Tested?

Tested both with unit tests and locally with a RSKj node connected to _regtest_: I've run a Python script that sends 500 pings changing the _NodeID_. With the flag set to _true_ multiple connections are accepted; with it set to _false_, just the last one remains. I've added a second device to the test to check the behaviour with a different host and its connection was added successfully, again keeping just last one if multiple were sent.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
+++ b/rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
@@ -23,8 +23,8 @@ import co.rsk.net.discovery.message.*;
 import co.rsk.net.discovery.table.NodeDistanceTable;
 import co.rsk.net.discovery.table.OperationResult;
 import co.rsk.net.discovery.table.PeerDiscoveryRequestBuilder;
-import co.rsk.util.ExecState;
 import co.rsk.scoring.PeerScoringManager;
+import co.rsk.util.ExecState;
 import co.rsk.util.IpUtils;
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.lang3.StringUtils;
@@ -60,6 +60,10 @@ public class PeerExplorer {
     private final Map<String, PeerDiscoveryRequest> pendingFindNodeRequests = new ConcurrentHashMap<>();
 
     private final Map<NodeID, Node> establishedConnections = new ConcurrentHashMap<>();
+
+    private final Map<String, NodeID> knownHosts;
+    private final boolean allowMultipleConnectionsPerHost;
+
     private final Integer networkId;
 
     private final ECKey key;
@@ -85,7 +89,7 @@ public class PeerExplorer {
     public PeerExplorer(List<String> initialBootNodes,
                         Node localNode, NodeDistanceTable distanceTable, ECKey key,
                         long reqTimeOut, long updatePeriod, long cleanPeriod, Integer networkId,
-                        PeerScoringManager peerScoringManager) {
+                        PeerScoringManager peerScoringManager, boolean allowMultipleConnectionsPerHost) {
         this.localNode = localNode;
         this.key = key;
         this.distanceTable = distanceTable;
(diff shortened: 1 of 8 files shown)


Vovchyk:
pipeline:run

illuque-iov:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

rskj-core/src/main/resources/config/regtest.conf
```diff
@@ -24,12 +24,14 @@ peer {
         # the peer window will show
         # only what retrieved by active
         # peer [true/false]
-        enabled = false
+        enabled = true
```
Vovchyk, line 27: is it by accident? we don't need this, do we?

rskj-core/src/main/resources/config/regtest.conf
```diff
@@ -24,12 +24,14 @@ peer {
...
 
         # List of the peers to start
         # the search of the online peers
         # values: [ip:port]
         ip.list = [ ]
+
+        allowMultipleConnectionsPerHost = true
```
Vovchyk, line 34: `regtest.conf` is only one network config among others, like `mainnet`, `testnet` etc. (and I believe PeerExplorer is disabled in `regtest`). Check out `reference.conf` file, which is a base config for all networks.

Vovchyk (changes requested):

rskj-core/src/main/resources/config/regtest.conf
illuque-iov: Yes, I forgot to remove it. Fixed, thx!

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
```diff
@@ -379,24 +386,46 @@ private void removeConnections(List<PeerDiscoveryRequest> expiredRequests) {
...
+        } else {
+            this.challengeManager.startChallenge(result.getAffectedEntry().getNode(), senderNode, this);
+        }
+    }
+
+    private void disconnectPeerIfDuplicatedByNodeId(Node senderNode, String ip, int port) {
+        NodeID oldNodeIdForHost = this.knownHosts.get(senderNode.getAddressAsString());
+        boolean existsWithDifferentId = oldNodeIdForHost != null && !oldNodeIdForHost.equals(senderNode.getId());
+        if (existsWithDifferentId) {
+            logger.debug("Disconnecting peer with old id: ip[{}] port[{}] id [{}]", ip, port, oldNodeIdForHost);
```
Vovchyk, line 426: just thinking.. probably we may log a `warning` or `info` here (instead of `debug`)?

Vovchyk (approved):

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
illuque-iov: Good point, I think it would be a weird behavior so I will make it a `warn`.

Vovchyk (dismissed):

illuque-iov:
pipeline:run

illuque-iov:
pipeline:run

illuque-iov:
pipeline:run

illuque-iov:
pipeline:run

illuque-iov:
pipeline:run

Vovchyk (approved):

Issue #1524, opened 2021-05-06
Add a command-line switch similar to -D, but parsed in Main args[]


SergioDemianLerner:
The -D switch is parsed by the Java virtual machine but its function in rskj has nothing to do with the VM.

I suggest that a new switch, such as "-x", is added as a standard argument allowing the same functionality that performed by -D. After some time, the -D switch can be deprecated in favor of "-x".

PR #1694, opened 2022-01-15
Added -X as an alternative way to read args in java command line

rmoreliovlabs:
## Description
In order to improve developer experience, we have added a flag `-X` that will be used to read arguments in the command line. An alternative to `-D`, parsed by our application.

## Motivation and Context
This issue was the motiviation https://github.com/rsksmart/rskj/issues/1524.

We needed a simple way to access the same arguments natively. It's not necessary to break backwards compatibility, there can be accessed by two ways.

For example: -Xdababase.dir=xxxx

The problem in letting the VM handle these arguments is that it is error prone. IntelliJ has a different field to specify VM arguments.

## How Has This Been Tested?
* Running this from Intellij and adding `-Xdatabase.dir=/home/rsk/data` to `CLI arguments` under `Edit Configurations`.
* Should work also usin OS CLI by running: `java -cp rskj-core-<VERSION>.jar co.rsk.Start -Xdatabase.dir=xxxx`

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/cli/CliArgs.java
```diff
@@ -92,6 +100,9 @@ public Parser(Class<O> optionsClass, Class<F> flagsClass) {
                                 throw new IllegalArgumentException("You must provide a flag name, e.g. --quiet");
                             }
                             flags.add(getFlagByName(args[i].substring(2, args[i].length())));
+                        } else if (args[i].charAt(1) == 'X' || args[i].charAt(1) == 'x') {
+                            String arg = args[i].replace("-X", "").replace("-x", "");
+                            paramValueMap.putAll(parseArgToMap(arg));
```
illuque-iov, line 105: Minor things, but `args[i].charAt(1)`, used three times, could be a variable like `secondChar`.
And you could do the compare like this: `secondChar.equalsIgnoreCase("X")`

In addition, maybe it would be a good idea to extract some constants for things like `X` and even `-`: they are used several times as simple chars.

rskj-core/src/main/java/co/rsk/cli/CliArgs.java
```diff
@@ -138,5 +149,46 @@ private O getOptionByName(String optionName) {
...
+
+            String param = paramValue[0];
+            String value = paramValue[1];
+            String[] paramParts = param.split("\\.");
+
+            Map<String, Object> paramValueMap = new HashMap<>();
+
+            for (int i = 0; i < paramParts.length; i++) {
+                if (i == 0) {
+                    paramValueMap.put(paramParts[i], value);
```
illuque-iov, line 181: I think it is worth it to move this case outside the loop and start it in `i=1`

rskj-core/src/main/java/co/rsk/cli/CliArgs.java
rmoreliovlabs: I don't think it would be efficient to use `equalsIgnoreCase` since I want to know if the chars `-` and `X` are in the these exact positions. There could be more X in other positions, so I think it's better to follow the pattern from previous checks in this case, code would be more consistent, I think I can do something to refactor a bit more by putting these chars into a variable and normalize the char to avoid replacing to times below within `parseArgToMap`

rskj-core/src/main/java/co/rsk/cli/CliArgs.java
```diff
@@ -138,5 +150,43 @@ private O getOptionByName(String optionName) {
...
+         *     {
+         *         "database": {
+         *             "dir": "/home/rsk/core"
+         *         }
+         *     }
+         * </blockquote>
+         * @param arg to parse
+         * @return a string arg parsed to an equivalent map having the same structure system properties would have.
+         */
+        private Map<String, Object> parseArgToMap(String arg) {
```
Vovchyk, line 167: wondering if we really need to parse it on our own?.. Check [this](https://github.com/rsksmart/rskj/blob/IRIS-3.1.0/rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java#L41) out - seems the `typesafe` lib has its own capability to parse path-based properties

rmoreliovlabs: Done

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
```diff
@@ -119,6 +119,13 @@ private Config getConfigFromCliArgs() {
             config = entry.getKey().withConfig(config, entry.getValue());
         }
 
+        if (!cliArgs.getParamValueMap().isEmpty()) {
```
Vovchyk, line 122: can you also cover this part with unit tests?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/config/ConfigLoader.java
rmoreliovlabs: Done

Vovchyk (approved):

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk (approved):

rmoreliovlabs:
pipeline:run

rmoreliovlabs:
pipeline:run

rmoreliovlabs:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

PR #1700, opened 2022-01-24
Updated getStateForDebugging bridge method to also return the next pegout creation block number

kelvinator07:


marcos-iov (commented):
I think we could also include `queuedPegoutsCount`? What do you think @josedahlquist ?

marcos-iov (approved):

josedahlquist (approved):

PR #1723, opened 2022-02-25
Added transaction execution list edges to block header

javicid:
Adds new header field to enable the parallel execution of transactions.


julianlen (approved):
LGTM!

PR #1724, opened 2022-02-25
Get gasPrice from DB on node start

illuque-iov:
Get gasPrice value from DB (`bestBlock.getMinimumGasPrice()`) when node starts and no block received neither windows filled.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/listener/GasPriceTracker.java
+++ b/rskj-core/src/main/java/org/ethereum/listener/GasPriceTracker.java
@@ -22,6 +22,7 @@ package org.ethereum.listener;
 import co.rsk.core.Coin;
 import co.rsk.remasc.RemascTransaction;
 import org.ethereum.core.Block;
+import org.ethereum.core.Blockchain;
 import org.ethereum.core.Transaction;
 import org.ethereum.core.TransactionReceipt;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/org/ethereum/listener/GasPriceTracker.java
```diff
@@ -90,7 +97,9 @@ private void onTransaction(Transaction tx) {
 
     public synchronized Coin getGasPrice() {
         if (window[0] == null) { // not filled yet
-            return defaultPrice;
+            return Optional.ofNullable(blockchain.getBestBlock())
```
Vovchyk, line 100: yeah, this is one way to do it (simplest one). But we could also load more blocks, e.g. in the ctor. (parent of the best one, parent of parent and so on..) from the db until the `window` array is filled up (the algorithm uses this array of 512 items as a main source of historical data [here](https://github.com/rsksmart/rskj/blob/91e5f60de29160ca8b96542bd0e59b0a57c31f13/rskj-core/src/main/java/org/ethereum/listener/GasPriceTracker.java#L96-L98)). And [this](https://github.com/rsksmart/rskj/blob/91e5f60de29160ca8b96542bd0e59b0a57c31f13/rskj-core/src/main/java/org/ethereum/listener/GasPriceTracker.java#L65-L76) is a place where that array is being populated.

illuque-iov: Addressed 👍

Vovchyk (dismissed):

illuque-iov:
pipeline:run

illuque-iov:
pipeline:run

Issue #1715, opened 2022-02-16
TypeConverter, HexUtil are confusing


fedejinich:
## TypeConverter

It's deprecated and it's doing mostly the same as HexUtils. I think it should be removed it makes unnecessary noise.

## HexUtil

- What's the purpose of it? It was created after TypeConverter and it's doing practically the same thing.
- `HexUtil.leftPad()` it's a byte array manipulation, it's not related to Hex types, I think it should be moved to `ByteUtil`
- `HexUtils.toQuantityJsonHex(byte[] x)` it has no use apart from tests. What's the purpose?

I think there is a lot of space to improve this class, it needs a deeper analysis.

## My thoughts 

I think these classes are very useful but they have a lot of noise, it would be nice to make a good code cleanup.

heitorfm:
Yes, HexUtils was the first phase of the refactoring. For better naming, placing and organization.
In the next moment we will remove the TypeConverter completely.

heitorfm:
done

PR #1727, opened 2022-03-08
refactor: replacing TypeConverter with HexUtils

heitorfm:
replacing TypeConverter with HexUtils

## Description
This is the replacing of TypeConverter
Also touchs on the issue https://github.com/rsksmart/rskj/issues/1715

## Motivation and Context
better placing and naming

## How Has This Been Tested?
unit tests

## Types of changes


rskj-core/src/test/java/co/rsk/core/RskAddressTest.java
```diff
@@ -91,7 +92,7 @@ public void invalidNullAddressBytes() {
         new RskAddress((byte[]) null);
     }
 
-    @Test(expected = NullPointerException.class)
+    @Test(expected = RskJsonRpcRequestException.class)
```
Vovchyk, line 95: `RskAddress` is a core class that's being used across the system, not only in the `RPC` interface (eg. in the config and db classes). Not sure `RskJsonRpcRequestException` (which is part of the `rpc` package) is a good candidate to be thrown by the `RskAddress`'s ctor.

Another thing is that `RskAddress` is one of the most used classes in the system.. we should be very careful while changing one exception that's being thrown onto another, as this may lead to breaking compatibility in the `json-rpc` interface or an intentional hf.

Vovchyk (changes requested):

rskj-core/src/test/java/co/rsk/core/RskAddressTest.java
heitorfm: Nice catch ! Gona review it, thanks.

heitorfm:
pipeline:run

heitorfm:
pipeline:run

Vovchyk (approved):

PR #1732, opened 2022-03-11
Trie class refactor for clarity

SergioDemianLerner:
In preparation for future performance improvements, many inner classes that bloated the Trie class have been moved to individual source files. 

## Description
This is a refactoring. No new functionality was added.

## Motivation and Context

The Trie class became too unnecessarily large. By splitting this file into separate classes, some non-risky changes can be made into the separate files without the danger of changing the trie class.

## How Has This Been Tested?
Same tests than before. No new test is necessary.

## Checklist:

- [x] My code follows the code style of this project. (Almost no code changes apart from movements have been made)


Vovchyk:
pipeline:run

rskj-core/src/main/java/co/rsk/trie/SharedPathSerializer.java
```diff
@@ -0,0 +1,153 @@
...
+
+        int lshared = getPathBitsLength(message);
+
+        int lencoded = PathEncoder.calculateEncodedLength(lshared);
+        byte[] encodedKey = new byte[lencoded];
+        message.get(encodedKey);
+        return TrieKeySlice.fromEncoded(encodedKey, 0, lshared, lencoded);
+    }
+
+    private static VarInt readVarInt(ByteBuffer message) {
```
joaquinlpereyra, line 123: Duplicated from Trie? Maybe better to extract into a `VarInt` static class or similar, in name of the refactor for clarity.

rskj-core/src/main/java/co/rsk/trie/IterationElement.java
```diff
@@ -0,0 +1,46 @@
...
+        public Trie getNode() {
+            return node;
+        }
+
+        public final TrieKeySlice getNodeKey() {
+            return nodeKey;
+        }
+
+        public String toString() {
+            byte[] encodedFullKey = nodeKey.expand();
```
joaquinlpereyra, line 39: Was the change from `encode` to `expand` intentional?

SergioDemianLerner: Encode meaning depends on the type of encoding you're using. Everything is "encoding". 
In the next PR I plan to replace the TrieKeySlice of another class which uses different encoding, the word "encode" will be even more confusing. The word "expand" means encode it in a way where 1s and 0s are expanded one element in each byte.

rskj-core/src/main/java/co/rsk/trie/SharedPathSerializer.java
SergioDemianLerner: ok. I'll do it

rskj-core/src/main/java/co/rsk/trie/IterationElement.java
SergioDemianLerner: I think the implementation of "expand" is exactly equal to the previous implementation of "encode".  So logs won't change. I can check this.

nagarev: I think the implementation is a little bit different, since the old `encode` method calls to `PathEncoder.encode(...`. Did we check the thing regarding the logs changing yet?

rskj-core/src/main/java/co/rsk/trie/SharedPathSerializer.java
nagarev: Are we going to extract this method or it will be added in a future PR?

nagarev:
LGTM, there's a couple of comments to address and we are good to go. The refactor is really good!

BTW, I would have preferred to focus this PR on moving the classes outside the `Trie.class` only, and then made the rest of the changes in a separate PR, so they're easier to spot and we can focus on them exclusively.

rskj-core/src/main/java/co/rsk/trie/IterationElement.java
Vovchyk: @SergioDemianLerner , any updates on this?

Vovchyk: @SergioDemianLerner , it seems that this is the only pending thing for this PR)

` nodeKey.expand()`'s implementation is following:
```
    // Returns the key slice in expanded form (byte array of ones and zeros)
    public byte[] expand() {
        return Arrays.copyOfRange(expandedKey, offset, limit);
    }
```
and the impl of the `encode()` method:
```
    public byte[] encode() {
        // TODO(mc) avoid copying by passing the indices to PathEncoder.encode
        return PathEncoder.encode(Arrays.copyOfRange(expandedKey, offset, limit));
    }
```

also it looks like that `PathEncoder.encode(bytes) != bytes`, which means that `IterationElement.toString()` would return a bit different strings, and as a result, it'd have an impact on values being logged:
```
public String toString() {
            byte[] encodedFullKey = nodeKey.expand();
```

is this intentional? shouldn't we use an old format that's being returned by `nodeKey.encode()`?

SergioDemianLerner: Now I see what you mean. The change was intentional to let me more easily debug. This is because I'm using different Trie key classes and I want all of them to log in the same simple-to-understand format (not the internal format). But I'm ok to revert it if you are use to reading logs in a specific format. But I guess nobody (except me) cares what was the previous logging format.
So I leave that question here: do you care of the new logging format ?

Vovchyk: yeah, agree. Probably no one else besides you has been using this format.

Vovchyk (approved):

Vovchyk:
pipeline:run

PR #1733, opened 2022-03-15
Added logs to processPegoutsIndividually, processPegoutsInBatch methods

kelvinator07:


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1078,6 +1078,7 @@ private void processPegoutsIndividually(
             .stream()
             .anyMatch(entry -> walletBalance.isGreaterThan(entry.getAmount()) || walletBalance.equals(entry.getAmount()));
         if (!canProcessAtLeastOnePegout) {
+            logger.warn("[processPegoutsIndividually] wallet balance {} cannot process at least one pegout", walletBalance);
```
nathanieliov, line 1081:
```suggestion
            logger.warn("[processPegoutsIndividually] wallet balance {} cannot process even one pegout", walletBalance);
```

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1172,6 +1174,7 @@ private void processPegoutsInBatch(
             if (releaseRequestQueue.getEntries().isEmpty()) {
```
nathanieliov, line 1174:
```suggestion
            if (pegoutEntries.isEmpty()) {
```

Just for consistency. I got a bit confused when see this instead of the variable `pegoutEntries`.

kelvinator07: releaseRequestQueue.getEntries() Is Used Because pegoutEntries Might Have Been Updated On Line 1144

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
kelvinator07: I Think At **Least One** Is Clearly..
For Example, We Have 5 Requests, A Check Is Done To Make Sure It Can Process At Least One Of Them Else It Logs That Warning And Returns. Pending When We Have Enough Funds

nathanieliov (approved):
LGTM

marcos-iov:
I think we could add more logs, the idea is to be able to debug what's happening via logs.

For example, in processPegoutsInBatch
- Going to create a batched pegout transaction for {} requests, total amount {}
- No pegouts requests in queue
- Max size exceeded, going to divide requests list in half
- Update next pegout creation height to block {}

Maybe you can think of some more, keep in mind what would be useful for you if you wanted to know the status of the pegouts by looking at the logs

kelvinator07:
> I think we could add more logs, the idea is to be able to debug what's happening via logs.
> 
> For example, in processPegoutsInBatch
> 
> * Going to create a batched pegout transaction for {} requests, total amount {}
> * No pegouts requests in queue
> * Max size exceeded, going to divide requests list in half
> * Update next pegout creation height to block {}
> 
> Maybe you can think of some more, keep in mind what would be useful for you if you wanted to know the status of the pegouts by looking at the logs

Thanks Marcos. It Has Been Updated With More Useful Logs.

marcos-iov (approved):

josedahlquist (approved):

PR #1734, opened 2022-03-17
Research/header generation from master

julianlen:


rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
```diff
@@ -0,0 +1,249 @@
...
+import java.util.*;
+
+public class ParallelizeTransactionHandler {
+    private final TransactionBucket sequentialBucket;
+    private final HashMap<ByteArrayWrapper, Set<Short>> bucketByWrittenKey;
+    private final HashMap<ByteArrayWrapper, Set<Short>> bucketByReadKey;
+    private final Map<RskAddress, Set<Short>> bucketBySender;
+    private final List<TransactionBucket> parallelBuckets;
+
+    public ParallelizeTransactionHandler(short buckets, long blockGasLimit) {
```
javicid, line 35: I think that it makes more sense to pass a bucket gas limit here

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -513,7 +672,9 @@ private BlockResult executeSequential(
             logger.trace("track commit");
 
             long gasUsed = txExecutor.getGasUsed();
-            totalGasUsed += gasUsed;
+            long totalGasUsedInBucket = parallelizeTransactionHandler.getGasUsedIn(bucketId.get());
+
+            totalGasUsedInBucket += gasUsed;
```
javicid, line 677: doesn't the gas bucket already include the tx gas used?

rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
```diff
@@ -0,0 +1,249 @@
...
+import org.ethereum.db.ByteArrayWrapper;
+import org.ethereum.vm.GasCost;
+
+import java.util.*;
+
+public class ParallelizeTransactionHandler {
+    private final TransactionBucket sequentialBucket;
+    private final HashMap<ByteArrayWrapper, Set<Short>> bucketByWrittenKey;
+    private final HashMap<ByteArrayWrapper, Set<Short>> bucketByReadKey;
+    private final Map<RskAddress, Set<Short>> bucketBySender;
```
javicid, line 32: can one sender be in multiple buckets?

rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
```diff
@@ -0,0 +1,249 @@
...
+import co.rsk.core.RskAddress;
+import org.ethereum.core.Transaction;
+import org.ethereum.db.ByteArrayWrapper;
+import org.ethereum.vm.GasCost;
+
+import java.util.*;
+
+public class ParallelizeTransactionHandler {
+    private final TransactionBucket sequentialBucket;
+    private final HashMap<ByteArrayWrapper, Set<Short>> bucketByWrittenKey;
```
javicid, line 30: can a key be written in multiple buckets?

rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
```diff
@@ -0,0 +1,249 @@
...
+        }
+        this.sequentialBucket = new TransactionBucket(buckets, gasLimit);
+    }
+
+    public boolean sequentialBucketHasGasAvailable(Transaction tx) {
+        long txGasLimit = GasCost.toGas(tx.getGasLimit());
+        return this.sequentialBucket.hasGasAvailable(txGasLimit);
+    }
+
+    public Optional<Short> addTransaction(Transaction tx, Set<ByteArrayWrapper> newReadKeys, Set<ByteArrayWrapper> newWrittenKeys, long gasUsedByTx) {
```
javicid, line 52: should this throw an exception if there is not enough gas in the sequential bucket?

rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
```diff
@@ -0,0 +1,249 @@
...
+        return txs;
+    }
+
+    public short[] getTransactionsPerBucketInOrder() {
+        List<Short> bucketSizes = new ArrayList<>();
+        short bucketEdges = 0;
+
+        for (TransactionBucket bucket: this.parallelBuckets) {
+            if (bucket.getTransactions().isEmpty()) {
+                break;
```
javicid, line 112: shouldn't this be a `continue` instead of a `break`?

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -627,4 +627,8 @@ public byte[] getUmmRoot() {
     }
 
     public short[] getTxExecutionListsEdges() { return this.txExecutionListsEdges; }
+
+    public void setTxExecutionListsEdges(short[] bucketOrder) {
```
javicid, line 631: maybe rename param to `edges` for consistency

javicid (commented):
Great job! Just some suggestions

rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
julianlen: Two different transactions that write/read the same key can be in both Sequential and Parallel buckets.

rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
julianlen: Two different transactions that write/read the same key can be in both Sequential and Parallel buckets.

rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
julianlen: Indeed, `break` was because the selection of the next available bucket was given the next available instead of the one with less gas.

rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
julianlen: Line 49 is only a query for the available gas. Then, when the handler tries to add a transaction into the sequential bucket, if the sequential has no gas, it returns an `Optional.empty()`.

rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
julianlen: Nope, you were right

rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
julianlen: Yep, I changed it!

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
julianlen: Indeed! Good catch

PR #1748, opened 2022-04-04
Add tests rejecting funds for retiring federation in registerFastBridgeBtcTransaction

nathanieliov:


rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -646,6 +646,139 @@ private BigInteger sendFundsToAnyAddress(
...
+
+        Address retiringFederationAddress = PegTestUtils.getFastBridgeAddressFromRedeemScript(
+            bridgeConstants,
+            retiringFederation.getRedeemScript(),
+            Sha256Hash.wrap(fastBridgeDerivationHash.getBytes())
+        );
+
+        BtcTransaction btcTx = btcTransactionProvider.provide(bridgeConstants, activeFederationAddress, retiringFederationAddress);
+        btcTx.addInput(
+            Sha256Hash.wrap(fastBridgeDerivationHash.getBytes()),
```
josedahlquist, line 725: just a comment, the use of the fastBridgeDerivationHash here may lead to a mistake... It is not related and won't happen in reality. I would use a zero hash instead.
```suggestion
            Sha256Hash.ZERO_HASH,
```

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -1280,6 +1413,99 @@ public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_after_RS
         Assert.assertEquals(co.rsk.core.Coin.fromBitcoin(valueToSend).asBigInteger(), result);
     }
 
+    @Test
+    public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_surpasses_locking_cap_after_RSKIP293() throws BlockStoreException, BridgeIllegalArgumentException, IOException {
```
josedahlquist, line 1417: We should also assert the pegout is created.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -646,6 +646,139 @@ private BigInteger sendFundsToAnyAddress(
...
+            null,
+            null,
+            lbcAddress.getBytes(),
+            null,
+            null,
+            null,
+            null
+        );
+
+        return bridgeSupport.registerFastBridgeBtcTransaction(
```
josedahlquist, line 769: You could assert the bridge has a pegout created after executing this method.

nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
nathanieliov: Done

josedahlquist (approved):

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -646,6 +649,168 @@ private BigInteger sendFundsToAnyAddress(
...
+        when(provider.getOldFederation()).thenReturn(retiringFederation);
+        ReleaseTransactionSet releaseTransactionSet = new ReleaseTransactionSet(new HashSet<>());
+        when(provider.getReleaseTransactionSet()).thenReturn(releaseTransactionSet);
+
+        Address userRefundBtcAddress = PegTestUtils.createRandomBtcAddress(bridgeConstants.getBtcParams());
+        Address lpBtcAddress = PegTestUtils.createRandomBtcAddress(bridgeConstants.getBtcParams());
+
+        Repository repository = spy(createRepository());
+        // Set current bridge balance in order to create the condition to surpass the locking cap when needed
+        doReturn(co.rsk.core.Coin.valueOf(1)).when(repository).getBalance(any());
```
marcos-iov, line 680: If we are doing this then it doesn't really make much sense to receive the locking cap value as a parameter, right?

I feel it would much clearer if we set a given locking cap value (as it is now) and then simply try lo peg-in a value above it

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -646,6 +649,168 @@ private BigInteger sendFundsToAnyAddress(
...
+            .withProvider(provider)
+            .withBridgeConstants(bridgeConstants)
+            .withBtcLockSenderProvider(btcLockSenderProvider)
+            .withActivations(activations)
+            .withBtcBlockStoreFactory(mockFactory)
+            .withExecutionBlock(executionBlock)
+            .withRepository(repository)
+            .build();
+
+        bridgeSupport = spy(bridgeSupport);
```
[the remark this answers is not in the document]
nathanieliov: I took that from legacy code, and didn't realize it. I'm changing it.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -646,6 +649,168 @@ private BigInteger sendFundsToAnyAddress(
...
+        BridgeStorageProvider provider = mock(BridgeStorageProvider.class);
+        when(provider.getNewFederation()).thenReturn(activeFederation);
+        when(provider.getOldFederation()).thenReturn(retiringFederation);
+        ReleaseTransactionSet releaseTransactionSet = new ReleaseTransactionSet(new HashSet<>());
+        when(provider.getReleaseTransactionSet()).thenReturn(releaseTransactionSet);
+
+        Address userRefundBtcAddress = PegTestUtils.createRandomBtcAddress(bridgeConstants.getBtcParams());
+        Address lpBtcAddress = PegTestUtils.createRandomBtcAddress(bridgeConstants.getBtcParams());
+
+        Repository repository = spy(createRepository());
```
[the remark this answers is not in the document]
nathanieliov: I just realize this is not necessary.  I can test the surpass locking cap scenario using a repository with a balance of 0. This really doesn't affect how this scenario is being tested now.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -646,6 +649,168 @@ private BigInteger sendFundsToAnyAddress(
...
+                Arrays.asList(activeFederationAddress, retiringFederationAddress));
+            Coin amountToRefund = BridgeUtils.getAmountSentToAddresses(activations,
+                bridgeConstants.getBtcParams(),
+                btcContext,
+                releaseTx,
+                Arrays.asList(
+                    result.longValue() == FastBridgeTxResponseCodes.REFUNDED_LP_ERROR.value()? lpBtcAddress: userRefundBtcAddress
+                ));
+
+            Assert.assertTrue("Pegout value should be bigger than zero and smaller than pegin value", amountToRefund.isGreaterThan(Coin.ZERO) && amountToRefund.isLessThan(amountSent));
```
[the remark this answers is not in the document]
nathanieliov: I discussed it with Jose. He told me that "calculate the fee is a complex task and no necessary for this case", and suggest me to simple do the above evaluation. So, I see from your comment that you disagree.

Currently, I honestly don't know how to calculate fees. Do you mind to tell me how can I do it?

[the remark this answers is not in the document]
marcos-iov: I agree that calculating the exact fee is quite complicated. The fees depend on the size of the transaction, and that depends on the number of inputs, outputs, signatures, etc.

What I suggest is to use an estimated fee, something like `peg-in value * 0,9 <= amountToRefund < peg-in value`. That is something we do in integration tests for example. It's far from perfect but at least I think it's an approximation better than checking between 0 and peg-in value.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -1280,6 +1445,99 @@ public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_after_RS
         Assert.assertEquals(co.rsk.core.Coin.fromBitcoin(valueToSend).asBigInteger(), result);
     }
 
+    @Test
+    public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_surpasses_locking_cap_after_RSKIP293() throws BlockStoreException, BridgeIllegalArgumentException, IOException {
```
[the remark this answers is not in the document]
nathanieliov: You mean tests pre RSKIP 293 that assert funds sent to a retiring federation were not refunded?  I'll take a look I think that test may be missing. But regarding funds sent to an active federation, those are refunded properly since before RSKIP293.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -1280,6 +1445,99 @@ public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_after_RS
...
+                BtcTransaction tx = new BtcTransaction(bridgeConstants.getBtcParams());
+                tx.addOutput(valueToSend, retiringFederationAddress);
+                return tx;
+            },
+            false,
+            bridgeConstantsRegtest.getMaxRbtc()
+        );
+        Assert.assertEquals(FastBridgeTxResponseCodes.REFUNDED_USER_ERROR.value(), result.longValue());
+
+        // Refund should be transfer to the liquidity provider
```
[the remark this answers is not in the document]
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
nathanieliov: I agree. I got bias from the legacy code and didn't notice it can be done in that way. I already changed it. Take a look and let me know what you think.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -669,15 +670,19 @@ private BigInteger sendFundsSurpassesLockingCapToAnyAddress(
...
         Address userRefundBtcAddress = PegTestUtils.createRandomBtcAddress(bridgeConstants.getBtcParams());
         Address lpBtcAddress = PegTestUtils.createRandomBtcAddress(bridgeConstants.getBtcParams());
 
-        Repository repository = spy(createRepository());
-        // Set current bridge balance in order to create the condition to surpass the locking cap when needed
-        doReturn(co.rsk.core.Coin.valueOf(1)).when(repository).getBalance(any());
+        Repository repository = createRepository();
+        // For simplicity of this test, the half of the max rbtc value is set as the current balance for the repository
+        // This means an amount sent over the half of the max rbtc value will surpass the locking cap which is set equal to that value
+        repository.addBalance(PrecompiledContracts.BRIDGE_ADDR, co.rsk.core.Coin.fromBitcoin(bridgeConstantsRegtest.getMaxRbtc().div(2)));
```
marcos-iov, line 683: To be honest, I find this solution a bit confusing. The idea is to test that the locking cap is exceeded but you are setting the locking cap as the max value. So it's actually like there's no locking cap.

I think the best solution is to receive the desired locking cap value as a parameter here, so each test can determine how much funds are included in the transaction and what the locking cap should be. That way you can determine on each test if the locking cap is exceeded or not.

We can discuss it further if you like

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -1446,61 +1451,59 @@ public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_after_RS
     }
 
     @Test
-    public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_surpasses_locking_cap_after_RSKIP293() throws BlockStoreException, BridgeIllegalArgumentException, IOException {
-        // Refund should be transfer to the user
+    public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_surpasses_locking_cap_should_refund_user_before_RSKIP293() throws BlockStoreException, BridgeIllegalArgumentException, IOException {
```
marcos-iov, line 1453: It's actually not refunding, right?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -1446,61 +1451,59 @@ public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_after_RS
...
-                tx.addOutput(valueToSend, retiringFederationAddress);
+                tx.addOutput(bridgeConstantsRegtest.getMaxRbtc(), retiringFederationAddress);
                 return tx;
             },
             false,
-            bridgeConstantsRegtest.getMaxRbtc()
+            false
         );
-        Assert.assertEquals(FastBridgeTxResponseCodes.REFUNDED_USER_ERROR.value(), result.longValue());
+        Assert.assertEquals(FastBridgeTxResponseCodes.UNPROCESSABLE_TX_VALUE_ZERO_ERROR.value(), result.longValue());
```
marcos-iov, line 1463: You get zero value because there are no funds sent to the active federation. It's never reaching the check for locking cap that it's what we actually want to test, isn't it?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -1280,6 +1449,95 @@ public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_after_RS
...
+                return tx;
+            },
+            false,
+            true
+        );
+        Assert.assertEquals(FastBridgeTxResponseCodes.REFUNDED_USER_ERROR.value(), result.longValue());
+    }
+
+    @Test
+    public void registerFastBridgeBtcTransaction_sum_funds_sent_to_active_and_retiring_fed_surpasses_locking_cap_before_RSKIP293() throws BlockStoreException, BridgeIllegalArgumentException, IOException {
```
marcos-iov, line 1481:
```suggestion
    public void registerFastBridgeBtcTransaction_sum_funds_sent_to_active_and_retiring_fed_surpasses_locking_cap_before_RSKIP293_no_refund() throws BlockStoreException, BridgeIllegalArgumentException, IOException {
```

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -1280,6 +1449,95 @@ public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_after_RS
         Assert.assertEquals(co.rsk.core.Coin.fromBitcoin(valueToSend).asBigInteger(), result);
     }
 
+    @Test
+    public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_surpasses_locking_cap_should_refund_user_before_RSKIP293() throws BlockStoreException, BridgeIllegalArgumentException, IOException {
```
marcos-iov, line 1453: I think we are only missing one test pre RSKIP that sends funds to both active and retiring federation, locking cap is exceeded, and only funds sent to the active federation are refunded.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -646,6 +649,172 @@ private BigInteger sendFundsToAnyAddress(
...
+
+        BtcBlockStoreWithCache btcBlockStore = mock(BtcBlockStoreWithCache.class);
+        BtcBlockStoreWithCache.Factory mockFactory = mock(BtcBlockStoreWithCache.Factory.class);
+        when(mockFactory.newInstance(repository, bridgeConstants, provider, activations)).thenReturn(btcBlockStore);
+        Block executionBlock = Mockito.mock(Block.class);
+        when(executionBlock.getNumber()).thenReturn(10L);
+
+        BtcLockSender btcLockSender = mock(BtcLockSender.class);
+        BtcLockSenderProvider btcLockSenderProvider = mock(BtcLockSenderProvider.class);
+        when(btcLockSenderProvider.tryGetBtcLockSender(any())).thenReturn(Optional.of(btcLockSender));
```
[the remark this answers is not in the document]
nathanieliov: No, it isn't needed. Already removed.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -801,11 +802,15 @@ private BigInteger sendFundsSurpassesLockingCapToAnyAddress(
...
                 btcContext,
                 releaseTx,
-                Arrays.asList(
-                    result.longValue() == FastBridgeTxResponseCodes.REFUNDED_LP_ERROR.value()? lpBtcAddress: userRefundBtcAddress
+                Collections.singletonList(
+                    result.longValue() == FastBridgeTxResponseCodes.REFUNDED_LP_ERROR.value() ? lpBtcAddress : userRefundBtcAddress
                 ));
 
-            Assert.assertTrue("Pegout value should be bigger than zero and smaller than pegin value", amountToRefund.isGreaterThan(Coin.ZERO) && amountToRefund.isLessThan(amountSent));
+            // For simplicity of this test an estimated fee of 1% of the amount is being used
```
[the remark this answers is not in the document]
nathanieliov: The comment is wrong.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
nathanieliov: Resolved

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -646,6 +649,172 @@ private BigInteger sendFundsToAnyAddress(
...
+                Collections.singletonList(
+                    result.longValue() == FastBridgeTxResponseCodes.REFUNDED_LP_ERROR.value() ? lpBtcAddress : userRefundBtcAddress
+                ));
+
+            // For simplicity of this test we are using as estimated fee of 10% of the amount sent in order to check
+            // that the amount to refund is at least above the amount sent minus the estimated fee
+            Coin estimatedFee = amountSent.divide(10);
+            Coin estimatedAmountToRefund = amountSent.minus(estimatedFee);
+
+            Assert.assertTrue("Pegout value should be bigger than the estimated fee(" + estimatedFee + ") and " +
```
marcos-iov, line 809:
```suggestion
            Assert.assertTrue("Pegout value should be bigger than the estimated amount to refund(" + estimatedAmountToRefund + ") and " +
```

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -1280,6 +1449,118 @@ public void registerFastBridgeBtcTransaction_funds_sent_to_retiring_fed_after_RS
...
+                tx.addOutput(Coin.FIFTY_COINS, retiringFederationAddress);
+                return tx;
+            },
+            true
+        );
+        Assert.assertEquals(co.rsk.core.Coin.fromBitcoin(Coin.FIFTY_COINS.div(2)).asBigInteger(), result);
+    }
+
+    @Test
+    public void registerFastBridgeBtcTransaction_funds_sent_to_active_and_retiring_fed_surpasses_locking_cap_before_RSKIP293_only_funds_sent_to_active_fed_should_be_refunded_to_LP() throws BlockStoreException, BridgeIllegalArgumentException, IOException {
```
marcos-iov, line 1469: I don't think this is working as expected.

The method `sendFundsSurpassesLockingCapToAnyAddress` checks that the refunded value is between amountSent and amountSent - fees, right? And amountSent is calculated considering the value sent to both active and retiring federations https://github.com/rsksmart/rskj/blob/reject-funds-for-retiring-fed/rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java#L791

But we expect to receive a refund only for the amount sent to the active federation, right?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
```diff
@@ -646,6 +649,172 @@ private BigInteger sendFundsToAnyAddress(
...
+        BtcBlockStoreWithCache btcBlockStore = mock(BtcBlockStoreWithCache.class);
+        BtcBlockStoreWithCache.Factory mockFactory = mock(BtcBlockStoreWithCache.Factory.class);
+        when(mockFactory.newInstance(repository, bridgeConstants, provider, activations)).thenReturn(btcBlockStore);
+        Block executionBlock = Mockito.mock(Block.class);
+        when(executionBlock.getNumber()).thenReturn(10L);
+
+        BridgeSupport bridgeSupport = bridgeSupportBuilder
+            .withProvider(provider)
+            .withBridgeConstants(bridgeConstants)
+            .withBtcLockSenderProvider(mock(BtcLockSenderProvider.class))
```
marcos-iov, line 695: Just a minor comment, BridgeSupportBuilder uses a mock by default no need to set it here

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
nathanieliov: Good catch.

This has room for improvement, but even so it is working as expected. When the RSKIP293 isn't active the method `getAmountSentToAddresses` only look for funds sent to one address, the first address in the list. This is the legacy behaviour. So, this is comparing only with the funds sent to the activeFedAddress instead of both fedAddresses.

However, as you I believe this is confusing too. What do you think if I change `getAmountSentToAddresses` method to throw an exception when RSKIP293 isn't active and is being called with a list of two addresses?

image

nathanieliov: Anyway, I'm also going to add a condition in `sendFundsSurpassesLockingCapToAnyAddress ` method to only pass the active federation when RSKIP293 isn't active. This will make it clear that this is the only address being used to get the amount sent.

marcos-iov: I like the idea of changing `sendFundsSurpassesLockingCapToAnyAddress` to get the amount sent only to the active federation if RSKIP 293 is not active. 

I wouldn't throw an exception in `getAmountSentToAddresses` when RSKIP 293 is not active, that was not the behavior pre hop and could cause the network to fork.

Thanks!

nathanieliov: Perfect. Thanks to you.

Changes pushed.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
nathanieliov: Done.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportFlyoverTest.java
nathanieliov: Done.

marcos-iov (approved):

PR #1750, opened 2022-04-11
Fixed issue with InetAddressBlock#contains when provided bits are 32

illuque-iov:
Fix issue on address comparison for 32 bits accuracy. It was causing a bug when (un)banning addresses with subnet /32.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/scoring/InetAddressBlock.java
```diff
@@ -26,7 +26,10 @@ public class InetAddressBlock {
     public InetAddressBlock(InetAddress address, int bits) {
         this.description = address.getHostAddress() + "/" + bits;
         this.bytes = address.getAddress();
-        this.nbytes = this.bytes.length - (bits + 7) / 8;
+
+        int tmpNBytes = this.bytes.length - (bits + 7) / 8;
```
Vovchyk, line 30: looks like it won't work, for instance, with `bits=31` for some address

rskj-core/src/main/java/co/rsk/scoring/InetAddressBlock.java
```diff
@@ -11,30 +11,39 @@
  * <p>
  * Created by ajlopez on 11/07/2017.
  */
+// TODO when InetAddressBlock==subnet gets clarified, update this code-base accordingly, probably some fields and methods will no longer be needed and other will need updates (equals, hashcode...)
 public class InetAddressBlock {
```
Vovchyk, line 15: do you think we could create another class for cidr blocks?.. Just not to mix stuff in one single class. And mark this one as deprecated with a `todo` to refactor / substitute it with the new one in future?

illuque-iov: Actually, I was in doubts between the two approaches.

Finally I decided to keep it this way as it meant less changes (an easier to review PR) and enabling this functionality in the future for the channel initialisation was as simple as renaming the new method to `contains` and removing the old one.

But I agree we are mixing stuff here and can be confusing if we do not address the TODO in the short term, so I will better split it in two as suggested 👍

illuque-iov: Done 👍

rskj-core/src/main/java/co/rsk/scoring/InetAddressCidrBlock.java
```diff
@@ -0,0 +1,118 @@
...
+    private final int mask;
+    private final int maskBytesCount;
+
+    /**
+     * Creates an InetAddressBlock given an address and the number of bits to ignore
+     *
+     * @param address the address
+     * @param bits    the numbers of bits to ignore
+     */
+    public InetAddressCidrBlock(InetAddress address, int bits) {
```
Vovchyk, line 47: this ctor is public one.. can we validate `bits` argument here? There's already some validation in `co.rsk.scoring.InetAddressUtils#parse`, but an instance can be created directly by this ctor.

rskj-core/src/test/java/co/rsk/scoring/InetAddressCidrBlockTest.java
```diff
@@ -0,0 +1,225 @@
...
+        Assert.assertNotEquals(block1, block3);
+        Assert.assertNotEquals(block1, block5);
+
+        Assert.assertNotEquals(null, block1);
+
+        Assert.assertEquals(block1.hashCode(), block4.hashCode());
+    }
+
+    @Test
+    public void subnetWithCidr8IPV4() throws UnknownHostException {
```
Vovchyk, line 63: could you also add a test for cidr `/0` ?

Vovchyk (changes requested):

rskj-core/src/test/java/co/rsk/scoring/InetAddressCidrBlockTest.java
illuque-iov: Missed that one, thanks. Fixed.

rskj-core/src/main/java/co/rsk/scoring/InetAddressCidrBlock.java
illuque-iov: Good point! I added the suggested validation and also made it package-protected as it was only used from `InetAddressUtils#parse` and `InetAddressCidrBlockTest` tests

Vovchyk (approved):

PR #1755, opened 2022-04-13
Healthcheck endpoint

nagarev:
## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/Start.java
+++ b/rskj-core/src/main/java/co/rsk/Start.java
@@ -17,6 +17,7 @@
  */
 package co.rsk;
 
+import co.rsk.rpc.netty.HttpServer;
 import co.rsk.util.PreflightChecksUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -56,6 +57,10 @@ public class Start {
         runtime.addShutdownHook(new Thread(ctx::close, "stopper"));
 
         runner.run();
+
+        // TODO, WIP -> Add to context and remove from here (For testing purposes only)
+        HttpServer httpServer = new HttpServer();
+        httpServer.start();
     }
 
     static void setUpThread(@Nonnull Thread thread) {
--- a/rskj-core/src/main/java/co/rsk/rpc/netty/HttpServer.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/netty/HttpServer.java
@@ -1,0 +1,41 @@
+package co.rsk.rpc.netty;
+
+import io.netty.bootstrap.ServerBootstrap;
+import io.netty.channel.ChannelFuture;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.channel.socket.nio.NioServerSocketChannel;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class HttpServer {
+
+    private static final Logger logger = LoggerFactory.getLogger(HttpServer.class);
+
+    public void start() {
+
+        EventLoopGroup bossGroup = new NioEventLoopGroup();
+        EventLoopGroup workerGroup = new NioEventLoopGroup();
+
+        try {
+
+            ServerBootstrap serverBootstrap = new ServerBootstrap();
+            serverBootstrap.group(bossGroup, workerGroup)
+                    .channel(NioServerSocketChannel.class)
+                    .childHandler(new HttpServerInitializer());
+
+            logger.info("HTTP Server Ready");
+
+            ChannelFuture channelFuture = serverBootstrap.bind(8080).sync();
+            channelFuture.channel().closeFuture().sync();
+
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        } finally {
+            bossGroup.shutdownGracefully();
+            workerGroup.shutdownGracefully();
+        }
+
+    }
+
+}
--- a/rskj-core/src/main/java/co/rsk/rpc/netty/HttpServerHandler.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/netty/HttpServerHandler.java
@@ -1,0 +1,57 @@
+package co.rsk.rpc.netty;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.SimpleChannelInboundHandler;
+import io.netty.handler.codec.http.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+
+import static io.netty.handler.codec.http.HttpHeaders.Names.CONTENT_TYPE;
+import static io.netty.handler.codec.http.HttpHeaders.Names.CONTENT_LENGTH;
+
+public class HttpServerHandler extends SimpleChannelInboundHandler<HttpObject> {
+
+    private static final Logger logger = LoggerFactory.getLogger(HttpServerHandler.class);
+
+    @Override
+    protected void channelRead0(ChannelHandlerContext ctx, HttpObject msg) throws Exception {
+
+        if (msg instanceof HttpRequest) {
+            logger.info("Client address: {}", ctx.channel().remoteAddress());
+
+            HttpRequest httpRequest = (HttpRequest) msg;
+
+            URI uri = new URI(httpRequest.getUri());
+
+            if ("/favicon.ico".equals(uri.getPath())) {
+                return;
+            }
+
+            ByteBuf content = Unpooled.copiedBuffer("Test", StandardCharsets.UTF_8);
+
+            if ("/ping".equals(uri.getPath())) {
+                content = Unpooled.copiedBuffer("pong", StandardCharsets.UTF_8);
+                logger.info("Ping Received");
+            }
+
+            DefaultFullHttpResponse response = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK, content);
+
+            response.headers().set(CONTENT_TYPE, "text/plain;charset=UTF-8");
+            response.headers().set(CONTENT_LENGTH, content.readableBytes());
+
+            ctx.writeAndFlush(response);
+        }
+
+    }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+        ctx.close();
+    }
+
+}
--- a/rskj-core/src/main/java/co/rsk/rpc/netty/HttpServerInitializer.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/netty/HttpServerInitializer.java
@@ -1,0 +1,17 @@
+package co.rsk.rpc.netty;
+
+import io.netty.channel.ChannelInitializer;
+import io.netty.channel.ChannelPipeline;
+import io.netty.channel.socket.SocketChannel;
+import io.netty.handler.codec.http.HttpServerCodec;
+
+public class HttpServerInitializer extends ChannelInitializer<SocketChannel> {
+
+    @Override
+    protected void initChannel(SocketChannel ch) {
+        ChannelPipeline pipeline = ch.pipeline();
+        pipeline.addLast("httpServerCodec", new HttpServerCodec());
+        pipeline.addLast("httpServerHandler", new HttpServerHandler());
+    }
+
+}


rskj-core/src/main/java/co/rsk/rpc/netty/http/HttpServer.java
```diff
@@ -0,0 +1,75 @@
...
+
+public class HttpServer implements InternalService {
+    private static final Logger logger = LoggerFactory.getLogger(HttpServer.class);
+
+    private int inetPort;
+    private ModuleConfigDTO moduleConfigDTO;
+
+    private HttpServer() { }
+
+    public HttpServer(int inetPort, ModuleConfigDTO moduleConfigDTO) {
```
illuque-iov, line 38: hey @nagarev, probably I’m missing something here, but QQ anyway…

why are we adding a new `HttpServer` and all related classes instead of creating a `co.rsk.rpc.modules.debug.HealthModule` or even just adding a new endpoint to `co.rsk.rpc.modules.debug.DebugModule`? what is the intention or the benefit?

rskj-core/src/main/java/co/rsk/rpc/netty/http/modules/HealthCheckModule.java
```diff
@@ -0,0 +1,35 @@
...
+import co.rsk.rpc.netty.http.HttpUtils;
+import io.netty.handler.codec.http.DefaultFullHttpResponse;
+import io.netty.handler.codec.http.HttpMethod;
+
+public class HealthCheckModule implements HttpModule {
+
+    @Override
+    public DefaultFullHttpResponse processRequest(String uri, HttpMethod method) {
+        if ("/health-check/ping".equals(uri) && method.equals(HttpMethod.GET)) {
+            return HttpUtils.createResponse("pong");
```
illuque-iov, line 29: Maybe this PR is just the first step of a health-check but... shouldn't we be checking things like enabled modules, transactions and blocks being recorded correctly, etc?

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
```diff
@@ -138,6 +138,18 @@ public boolean isMinerServerFixedClock() {
         return configFromFiles.getBoolean("miner.server.isFixedClock");
     }
 
+    public boolean isHttpServerEnabled() {
+        return configFromFiles.getBoolean("http.server.enabled");
```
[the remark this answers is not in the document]
nagarev: Yup, at that moment this was only a draft PR.

rskj-core/src/main/java/co/rsk/rpc/netty/http/HttpServer.java
nagarev: This is a functional requirement.

rskj-core/src/test/java/co/rsk/rpc/netty/rest/RestServerDispatcherTest.java
```diff
@@ -0,0 +1,91 @@
...
+        doReturn(null).when(healthCheckModuleMock).processRequest(uri, method);
+
+//        Field healthCheckModuleField = httpServerDispatcher.getClass()
+//                .getDeclaredField("healthCheckModule");
+//        healthCheckModuleField.setAccessible(true);
+//        Field modifiers = Field.class.getDeclaredField("modifiers");
+//        modifiers.setAccessible(true);
+//        modifiers.setInt(healthCheckModuleField,
+//                healthCheckModuleField.getModifiers() & ~Modifier.FINAL);
+//        healthCheckModuleField.set(httpServerDispatcher, healthCheckModuleField);
```
illuque-iov, line 82: You left this commented out code for any reason? If so please mention that

illuque-iov: My bad, just remembered you mentioned you will be working on tests after opening the PR.

rskj-core/src/main/java/co/rsk/rpc/netty/rest/RestServer.java
```diff
@@ -0,0 +1,82 @@
...
+
+            ServerBootstrap serverBootstrap = new ServerBootstrap();
+            serverBootstrap.group(bossGroup, workerGroup)
+                    .channel(NioServerSocketChannel.class)
+                    .childHandler(new RestServerInitializer(restModuleConfigDTO));
+
+            logger.info("REST Server Ready");
+
+            ChannelFuture channelFuture = serverBootstrap.bind(inetHost, inetPort).sync();
+            channelFuture.channel().closeFuture().sync();
```
illuque-iov, line 64: Maybe logging that server is down after this.

rskj-core/src/main/java/co/rsk/rpc/netty/rest/RestServerHandler.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    protected void channelRead0(ChannelHandlerContext ctx, HttpObject msg) throws Exception {
+
+        if (msg instanceof HttpRequest) {
+            logger.info("Client address: {}", ctx.channel().remoteAddress());
+
+            HttpRequest httpRequest = (HttpRequest) msg;
+
+            DefaultFullHttpResponse response = dispatcher.dispatch(httpRequest);
```
illuque-iov, line 44: Now that I think of it... what do you think about replacing `RestServerDispatcher` class by a `RestModules` enum? Since `RestModule` is an interface we could do something like this:

```
HttpRequest httpRequest = (HttpRequest) msg;
String uri = new URI(httpRequest.getUri()).getPath();

RestModules restModule = RestModules.lookup(uri, restModuleConfigDTO);

DefaultFullHttpResponse response;
if (restModule == null) { // uri not found or module disabled
    response = 404;
} else {
    response = restModule.processRequest(uri, httpRequest.getMethod());
}

if (response == null) {
    logger.info("Omitted response.");
    return;
}

ctx.writeAndFlush(response);
```

where `RestModules` is

```
public enum RestModules {

    HEALTHCHECK(new HealthCheckModule(), "/health-check", RestModuleConfigDTO::isHealthCheckModuleEnabled);

    private final RestModule restModule;

    private final String uri;

    private final Function<RestModuleConfigDTO, Boolean> isActive;

    RestModules(RestModule restModule, String uri, Function<RestModuleConfigDTO, Boolean> isActive) {
        this.restModule = restModule;
        this.uri = uri;
        this.isActive = isActive;
    }

    public static RestModules lookup(String moduleUri, RestModuleConfigDTO restModuleConfig) {
        for (RestModules module : RestModules.values()) {
            if (moduleUri.startsWith(module.uri) && module.isActive.apply(restModuleConfig)) {
                return module;
            }
        }
        return null;
    }

    public DefaultFullHttpResponse processRequest(String uri, HttpMethod method) {
        return restModule.processRequest(uri, method);
    }
}
```

WDYT?

rskj-core/src/test/java/co/rsk/rpc/netty/rest/RestServerDispatcherTest.java
nagarev: This is a WIP test, I'm finishing those right now.

rskj-core/src/main/java/co/rsk/rpc/netty/rest/RestServerHandler.java
nagarev: Did a couple of changes to use some of the suggestions, I think it ended up being nice.

rskj-core/src/main/java/co/rsk/rpc/netty/rest/RestServer.java
nagarev: Done!

rskj-core/src/main/java/co/rsk/rpc/netty/rest/RestUtils.java
```diff
@@ -0,0 +1,69 @@
...
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.HttpVersion;
+
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
+import static io.netty.handler.codec.http.HttpHeaders.Names.CONTENT_LENGTH;
+import static io.netty.handler.codec.http.HttpHeaders.Names.CONTENT_TYPE;
+
+public class RestUtils {
```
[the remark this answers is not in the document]
nagarev: Since I've already created an abstract class for modules to inherit (due to the dispatcher refactoring) I will use it to include these methods there as you suggested (I personally think that the inheritance way is the most direct way of doing it).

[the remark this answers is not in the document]
nagarev: Was thinking on adding it to the recently created `RestModule` abstract class, but it seems to me that those utility methods are really short and basic, and at the same time they're not directly related to the modules _per se_, but to the `REST` server in general (they're needed to create responses and keep everything as standard as possible).

So in this case I personally prefer to keep it as it is (testing it is really simple). But for sure I prefer any of the other strategies when working with bigger or not so specific classes.

WDYT?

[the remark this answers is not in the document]
illuque-iov: Fair enough, I have a preference for the other approach but as you mentioned, these methods are small. If this class starts growing, maybe we can change the strategy.

rskj-core/src/main/java/co/rsk/rpc/netty/rest/RestServerHandler.java
illuque-iov: Well, TBH I don't see the need of having both `RestModuleLoader` and `RestServerDispatcher`. I think they share the same responsibility and are small enough to have just one class. Having both makes the code more complex to follow IMO. 

What benefit you see on having both?

I think module lookup and activation check could be moved to `RestModuleLoader` on a method like `public static RestModule getRestModule(String uri)`. Then, in `co.rsk.rpc.netty.rest.RestServerHandler#channelRead0`, you would simply:
- call `restModule.processRequest` if module found and active
- create the NOT FOUND response if module not found or inactive

WDYT?

nagarev:
pipeline:run

illuque-iov (dismissed):

nagarev:
pipeline:run

nagarev:
pipeline:run

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -1024,6 +1027,10 @@ public synchronized List<InternalService> buildInternalServices() {
             internalServices.add(getPeerScoringReporterService());
         }
 
+        if (getRskSystemProperties().isRestServerEnabled()) {
```
asoto-iov, line 1030: About this logic, the RestServer it is not a "Service" itself. The thing providing the service or the feature is in this case the Health check module. With this logic we could have active a rest server without any real feature using it. 

I would update how this service is handled make the health check module the the one to start the rest server if needed.

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
```diff
@@ -159,6 +161,29 @@ public double minerMinFeesNotifyInDollars() {
         return getDouble("miner.minFeesNotifyInDollars", 0);
     }
 
+    public boolean isRestServerEnabled() {
```
asoto-iov, line 164: Here we have 5 methods to return different properties related with the rest server and the HealthCheck module, maybe we can use one single objet go handle the properties.

rskj-core/src/main/java/co/rsk/rest/RestServer.java
```diff
@@ -0,0 +1,84 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(RestServer.class);
+
+    private final EventLoopGroup bossGroup;
+    private final EventLoopGroup workerGroup;
+
+    private final InetAddress inetHost;
+    private final int inetPort;
+    private final RestModuleConfigDTO restModuleConfigDTO;
+
+    public RestServer(InetAddress inetHost, int inetPort, RestModuleConfigDTO restModuleConfigDTO) {
```
asoto-iov, line 42: As mentioned before the config could be in a single class.

rskj-core/src/main/java/co/rsk/rest/RestServerDispatcher.java
```diff
@@ -0,0 +1,75 @@
...
+    public RestServerDispatcher(List<RestModule> moduleList) {
+        Objects.requireNonNull(moduleList, "Module List can not be null");
+        this.moduleList = Collections.unmodifiableList(moduleList);
+    }
+
+    public DefaultFullHttpResponse dispatch(HttpRequest request) throws URISyntaxException {
+
+        String uri = new URI(request.getUri()).getPath();
+
+        RestModule restModule = moduleList.stream()
```
asoto-iov, line 49: On thing, we are only filtering by path but not by method (GET,PUT...) and this is done internally in the module. If we decide to return 404 if the method doesn't match then we can filter directly here. and avoid any other logic.

rskj-core/src/main/java/co/rsk/rest/RestServerHandler.java
```diff
@@ -0,0 +1,64 @@
...
+    protected void channelRead0(ChannelHandlerContext ctx, HttpObject msg) throws Exception {
+
+        if (msg instanceof HttpRequest) {
+            logger.info("Client address: {}", ctx.channel().remoteAddress());
+
+            HttpRequest httpRequest = (HttpRequest) msg;
+
+            DefaultFullHttpResponse response = dispatcher.dispatch(httpRequest);
+
+            if (response == null) {
```
asoto-iov, line 48: I think it could be better to return a default response in any case.

rskj-core/src/main/java/co/rsk/rest/RestUtils.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
+import static io.netty.handler.codec.http.HttpHeaders.Names.CONTENT_LENGTH;
+import static io.netty.handler.codec.http.HttpHeaders.Names.CONTENT_TYPE;
+
+public class RestUtils {
+
+    private static final String DEFAULT_CONTENT_TYPE = "text/plain;charset=UTF-8";
```
asoto-iov, line 34: is there a reason to use text/plain instead of JSON? Wouldn't be better?

rskj-core/src/main/java/co/rsk/rest/modules/HealthCheckModule.java
```diff
@@ -0,0 +1,39 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package co.rsk.rest.modules;
+
+import co.rsk.rest.RestUtils;
+import io.netty.handler.codec.http.DefaultFullHttpResponse;
+import io.netty.handler.codec.http.HttpMethod;
+
+public class HealthCheckModule extends RestModule {
```
asoto-iov, line 24: I would separate the  Module logic from the Rest implementation, so if for any reason we move from rest to json-rpc  (for example) we only would have to change one part and not he hole module implementation.

rskj-core/src/main/java/co/rsk/rest/modules/HealthCheckModule.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public HealthCheckModule(String uri, boolean active) {
+        super(uri, active);
+    }
+
+    @Override
+    public DefaultFullHttpResponse processRequest(String uri, HttpMethod method) {
+        if ("/health-check/ping".equals(uri) && method.equals(HttpMethod.GET)) {
```
asoto-iov, line 32: If the method doesn't match the path, shouldn't we return NotFound or Method Not Allowed response? 
Also the path verification is being done twice.

rskj-core/src/main/java/co/rsk/rest/modules/RestModuleLoader.java
```diff
@@ -0,0 +1,44 @@
...
+public class RestModuleLoader {
+
+    private final List<RestModule> restModuleList;
+
+    public RestModuleLoader(RestModuleConfigDTO restModuleConfigDTO) {
+        Objects.requireNonNull(restModuleConfigDTO, "REST Module Config can not be null");
+
+        restModuleList = new ArrayList<>();
+
+        restModuleList.add(new HealthCheckModule("/health-check",
```
asoto-iov, line 36: Here would be a nice place to do the filtering by enabled/disabled module. If it is not active then we wouldn't need it into the list. what do you think?

rskj-core/src/main/resources/config/testnet.conf
```diff
@@ -51,6 +51,19 @@ peer {
     networkId = 8100
 }
 
+rest {
```
asoto-iov, line 54: Instead of rest I would use a generic name. Here you are providing a service or services in the future and the REST server is the channel or the way to provide then and not the other way. 
I was thinking in something like: 
`management-service {
    modules {
        health-check {
            enabled = false
        }
    }
    rest-config {
        bind_address = localhost
        port = 4446        
    }
}`

So the config from the modules and the rest server would be separated.

rskj-core/src/main/java/co/rsk/rest/RestServerInitializer.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public RestServerInitializer(RestModuleConfigDTO restModuleConfigDTO) {
+        this.restModuleConfigDTO = restModuleConfigDTO;
+    }
+
+    @Override
+    protected void initChannel(SocketChannel ch) {
+        ChannelPipeline pipeline = ch.pipeline();
+        pipeline.addLast("httpServerCodec", new HttpServerCodec());
+        pipeline.addLast("httpServerHandler", new RestServerHandler(restModuleConfigDTO));
```
asoto-iov, line 38: If the logic is increased maybe it would have sense to have an error handler.

rskj-core/src/main/java/co/rsk/rest/modules/RestModule.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package co.rsk.rest.modules;
+
+import io.netty.handler.codec.http.DefaultFullHttpResponse;
+import io.netty.handler.codec.http.HttpMethod;
+
+public abstract class RestModule {
```
asoto-iov, line 23: Here we could add an api rest method list to be sure that all the modules are defining them, and a maybe generic method to validate the uri and the rest method, as it would be the same for the modules.

illuque-iov:
I generally agree with @asoto-iov's comments.

My general feeling, matching somehow with my previously provided feedback, is that this PR may work but it's too basic or too MVP maybe for a production code. Applying the suggestions would improve this and it shouldn't take too.

Having said that it's up to you to do another iteration over the code or not.

illuque-iov (commented):
I generally agree with @asoto-iov's comments.

My general feeling after Angel's comments and revisiting the PR after my last approval is still that this PR, while working, it's too basic or too MVP maybe for a production code., I think applying the suggestions would improve this and it shouldn't take too long.

WDYT?

rskj-core/src/main/java/co/rsk/rest/RestServerDispatcher.java
```diff
@@ -0,0 +1,75 @@
...
+                .filter(module -> uri.startsWith(module.getUri())).findFirst().orElse(null);
+
+        if (restModule == null) {
+            logger.info("Handler Not Found.");
+            return RestUtils.createResponse(NOT_FOUND, HttpResponseStatus.NOT_FOUND);
+        }
+
+        if (restModule.isActive()) {
```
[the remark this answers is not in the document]
nagarev: Modules are already being loaded at constructor level, and yes, this list is expected to grow in the future, module status (active/inactive) is used by Dispatcher to decide what to do.

rskj-core/src/main/java/co/rsk/rest/RestServerDispatcher.java
nagarev: The path being checked here is only the start bit of it, since that part defines what's the module that has to take care of the incoming request. Method and full path are taken care by the module itself.

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
nagarev: I've chosen to follow the way it's been done before, we can create an object too, let's discuss that.

rskj-core/src/main/java/co/rsk/RskContext.java
nagarev: The idea behind this `REST` server is for it to be a standalone server, for now, the only module will be `HealthCheck` one, but in the future, if we have to add new a new `REST` module, we'll simply have to add a new module and that's it.

Modules can be configured independently, things like status (active/inactive) and or different things can be tweaked via configuration file.

Modules don't have the responsibility to start the `Rest` server, but the other way around.

What we can do, thinking about responsibilities, is to add an "auto shutdown feature" if no modules are active (why we would have a running and listening server is all modules are inactive?).

rskj-core/src/main/java/co/rsk/rest/RestServer.java
nagarev: this will be discussed as previously mentioned.

rskj-core/src/main/java/co/rsk/rest/RestServerHandler.java
nagarev: This is an omitted response when, for some reason, a `null` response is returned by the `RestServerDispatcher`, and so, response is omitted. It works as a defense mechanism when something unexpected happens inside previously mentioned component.

rskj-core/src/main/java/co/rsk/rest/RestServerInitializer.java
nagarev: Yes, we can add it whenever it became necessary, the idea of having this logic here is precisely that: adding some code clarity and separation of concerns. For now, I believe that the different components are taking care of different kind of errors in a place that "makes sense".

rskj-core/src/main/java/co/rsk/rest/RestUtils.java
nagarev: We're not always returning json data on all responses, so the more general `text/plain;charset=UTF-8` is being used instead of `application/json` one.

rskj-core/src/main/java/co/rsk/rest/modules/HealthCheckModule.java
nagarev: We already have other types of servers, and if we decide to move some modules between them, we would have to refactor plenty of things, since implementations between servers differs a lot.

As a side note, the whole idea of having a `REST` server is to have some clear separation between different modules or things you can do with the node. In this case, `Health check` feature is going to be accessed only via its `REST API`.

rskj-core/src/main/java/co/rsk/rest/modules/HealthCheckModule.java
nagarev: If the module returns `null`, the `Dispatcher` is in charge of handle it, so we don't have to add that logic in every module.

Will update the string comparing part so it only checks for the  important part of the URI instead of all of it.

rskj-core/src/main/java/co/rsk/rest/modules/RestModule.java
nagarev: The logic of `URI` comparison and request dispatching is done at `Dispatcher` level, by evaluating the base URI and module status (active/inactive).

rskj-core/src/main/java/co/rsk/rest/modules/RestModuleLoader.java
nagarev: In current implementation `Dispatcher` is in charge of checking module status (active/inactive). In the other hand, module quantity will be always a small, so I think we can have the complete list here, and check status when needed.

If, in the future, we observe some overhead that can be avoided, we can add some logic at `Dispatcher` level so it knows when to ignore some module beforehand.

rskj-core/src/main/resources/config/testnet.conf
nagarev: We decided to go with the `REST` name to group all modules that will be accessed via their `REST API`, so they're, in fact, as their name implies, `REST` modules. Configuration reflects that fact.

nagarev:
pipeline:run

rskj-core/src/main/resources/config/main.conf
```diff
@@ -55,6 +55,19 @@ peer {
     networkId = 775
 }
 
+rest {
```
Vovchyk, line 58: do we need these network specific configs? it seems to me that they are duplicates on the one defined in reference.conf

Vovchyk (changes requested):

PR #1758, opened 2022-04-18
Affecting consensus changes on RSK in the following bridge methods: registerFastBridgeTransaction(flyover pegin) and registerBtcTransaction(regular pegin). Also, adding a new bridge method: getActivePowpegRedeemScript. 

nathanieliov:
Changes in registerFastBridgeTransaction method(flyover pegin):

- Take into account the retiring federation(not just the active federation as the current behavior). Now funds sent to the retiring federation will be considered.
- Add validation that checks the minimum of the total amount and per UTXOs.
- Add support for segwit outputs. Currently, parsing transactions with segwit outputs is resulting in a parsing error.

Changes in registerBtcTransaction method(regular pegin):

- Add validation that checks the minimum per UTXOs instead of the total amount.

Add new bridge method: getActivePowpegRedeemScript.
- Basically, this method allows users to get the active powpeg redeem script instead of having to create it by themselves.


nathanieliov:
pipeline:run

nathanieliov:
pipeline:run

nathanieliov:
rsk:HOP-4.0.0-rc
fed:HOP-4.0.0-rc
ci:hop-rc

nathanieliov:
pipeline:run

rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
```diff
@@ -450,6 +450,17 @@ public enum BridgeMethods {
             activations -> activations.isActive(RSKIP134),
             fixedPermission(true)
     ),
+    GET_ACTIVE_POWPEG_REDEEM_SCRIPT(
+            CallTransaction.Function.fromSignature(
+                    "getActivePowpegRedeemScript",
+                    new String[]{},
+                    new String[]{"bytes"}
+            ),
+            fixedCost(3_000L), // TODO: DEFINE THE CORRECT FIXED COST FOR THS METHOD
```
donequis, line 459: Please, fix this before merge

rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
```diff
@@ -450,6 +450,17 @@ public enum BridgeMethods {
...
+            (BridgeMethodExecutorTyped) Bridge::getActivePowpegRedeemScript,
+            activations -> activations.isActive(RSKIP293),
+            fixedPermission(false)
```
donequis, line 462: Why isn't this local only? AFAIK is for external use, so it could be limited to that.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -2475,6 +2469,17 @@ public Coin getLockingCap() {
...
 
+    /**
+     * Returns the redeemScript of the current active federation
+     *
+     * @return Returns the redeemScript of the current active federation
+     */
+    public Optional<Script> getActivePowpegRedeemScript() {
+        return activations.isActive(RSKIP293) ?
+            Optional.of(getActiveFederation().getRedeemScript()) :
+            Optional.empty();
```
donequis, line 2480: I suggest throwing error or at least logging an error on this kind of unrecoverable situations

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -138,6 +158,201 @@ public static Wallet getFederationsSpendWallet(
...
+    }
+
+    /**
+     * @param context
+     * @param addresses
+     * @return a simple wallet instance with the give list of address added as watched addresses
+     */
+    private static WatchedBtcWallet createWatchedBtcWalletFromAddresses(Context context, List<Address> addresses) {
+        WatchedBtcWallet wallet = new WatchedBtcWallet(context);
+        long now = Utils.currentTimeMillis() / 1000L;
```
donequis, line 290: Use block.timestamp or a constant if needed. The library doesn't seem to use the value except for logging a warning, but it feels cleaner to avoid System.currentTime

rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
nathanieliov: This will be completed in another task/PR. We already added the performance tests to get the data to define the cost(https://github.com/rsksmart/rskj/pull/1759).

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
nathanieliov: Do you mean situations where the RSKIP is not active yet?

We are already logging some steps of the call of the method.

image

What else do you think would be worthy to log?

rskj-core/src/main/java/co/rsk/peg/BridgeMethods.java
nathanieliov: This is doing a simple operation that poses no risk of misuse of the node resources. Thus, we understand this is not needed.

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
nathanieliov: If this is only being used for logging a warning, I don't see any issue with continue using it for getting the current milliseconds. In fact, the library does the same internally.
 
 
image

josedahlquist (approved):

PR #1760, opened 2022-04-20
Handler's been refactorized

julianlen:
We made the next change in the PrallelilizeTransactionHandle class:
- bucketBySender: instead of storing the list of buckets where a tx is located given the sender if the sender is in the sequential, no changes are made. If the sender was already stored in a parallel bucket and the Handler is adding another tx with the same sender to the sequential, the sequential number is settled given the sender.
- getBucketCandidates():  Changed the algorithm to a simpler one that returns the bucketId:
  -  If the tx sender is in the sequential, goes to the sequential.
  -  If there is a collision with more than one key placed in two different buckets, goes to the sequential. Otherwise, if it is the same bucket, goes to that bucket. If there is no collision, it goes to the bucket with gas the less.

- We added a  Double Dispatch, then deleted in the [PR](https://github.com/rsksmart/rskj/pull/1764)

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
@@ -22,6 +22,7 @@ import co.rsk.core.Coin;
 import co.rsk.core.RskAddress;
 import co.rsk.core.TransactionExecutorFactory;
 import co.rsk.core.TransactionListExecutor;
+import co.rsk.core.bc.ParallelizeTransactionHandler.TransactionBucket;
 import co.rsk.crypto.Keccak256;
 import co.rsk.db.RepositoryLocator;
 import co.rsk.metrics.profilers.Metric;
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
```diff
@@ -128,107 +110,118 @@ public short[] getTransactionsPerBucketInOrder() {
...
 
 
-    private Set<Short> getBucketCandidates(Transaction tx, Set<ByteArrayWrapper> newReadKeys, Set<ByteArrayWrapper> newWrittenKeys) {
-        Set<Short> bucketCandidates = new HashSet<>();
+    private TransactionBucket getBucketCandidates(Transaction tx, Set<ByteArrayWrapper> newReadKeys, Set<ByteArrayWrapper> newWrittenKeys) {
+        Optional<TransactionBucket> bucketCandidate = getBucketBySender(tx);
 
-        getBucketBySender(tx).ifPresent(bucketCandidates::addAll);
+        if (bucketCandidate.isPresent() && bucketCandidate.get().isSequential()) {
+            return getSequentialBucket();
```
javicid, line 158:
```suggestion
            return bucketCandidate;
```

rskj-core/src/main/java/co/rsk/core/bc/ParallelizeTransactionHandler.java
```diff
@@ -128,107 +110,118 @@ public short[] getTransactionsPerBucketInOrder() {
...
 
         // read - written
         for (ByteArrayWrapper newReadKey : newReadKeys) {
             if (bucketByWrittenKey.containsKey(newReadKey)) {
-                bucketCandidates.add(bucketByWrittenKey.get(newReadKey));
+                TransactionBucket bucket = bucketByWrittenKey.get(newReadKey);
+
+                if (bucketCandidate.isPresent() && !bucketCandidate.get().equals(bucket)) {
+                   return getSequentialBucket();
+                } else {
```
javicid, line 168: I think this else block can be removed or replaced by `else if (!bucketCandidate.isPresent())`

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -718,8 +702,7 @@ private BlockResult executeSequential(
 
             TransactionReceipt receipt = new TransactionReceipt();
             receipt.setGasUsed(gasUsed);
-            receipt.setCumulativeGas(totalGasUsedInBucket);
-
+            bucketGasAccumulated.ifPresent(receipt::setCumulativeGas);
```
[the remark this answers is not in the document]
julianlen: Well, actually, this's weird. Because it's always present, otherwise [this](https://github.com/rsksmart/rskj/blob/de130e96e0d59e8cd1e1665d894a902ecf7b5899/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java#L666) line is either returning or continuing the loop.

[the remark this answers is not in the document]
javicid: What if `acceptInvalidTransactions` is true?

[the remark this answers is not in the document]
julianlen: I got it, you're right!

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -698,16 +692,7 @@ private BlockResult executeSequential(
...
-
-            try {
-                totalGasUsedInBucket = parallelizeTransactionHandler.getGasUsedIn(bucketId.get());
-            } catch (RuntimeException e) {
-                logger.warn("block: [{}] execution was interrupted", block.getNumber());
-                logger.trace("", e);
-                profiler.stop(metric);
-                return BlockResult.INTERRUPTED_EXECUTION_BLOCK_RESULT;
-            }
+            totalGasUsed = parallelizeTransactionHandler.getGasUsedInSequential();
```
javicid, line 695: Rename to `gasUsedInSequential` and add a new variable `gasUsedInBlock += gasUsed`

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -760,7 +751,7 @@ private BlockResult executeSequential(
                 executedTransactions,
                 receipts,
                 bucketOrder,
-                totalGasUsed,
+                totalGasUsed, // totalBlock = parallel1 + parallel2 + sequential?
```
javicid, line 754: Replace this with new variable `gasUsedInBlock`

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
julianlen: Remember that a transaction fits the block if there is enough gas in the sequential bucket, the totalGasUsed variable is checked on the init() method

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
julianlen: hmm but this variable is checked in the `init()` method to know if a transaction fits in the block.

PR #1780, opened 2022-05-26
Research/parallel-tx/Block validation

julianlen:


rskj-core/src/main/java/co/rsk/core/bc/ReadWrittenKeysTracker.java
```diff
@@ -19,20 +19,27 @@
...
 import java.util.Set;
 
 //TODO(JULI):
 // * Next step should be to check whether a key is written in the cache but also deleted in the same transaction. This key shouldn't be considered as a written key.
 
 public class ReadWrittenKeysTracker implements IReadWrittenKeysTracker {
+    private final Map<ByteArrayWrapper, Long> threadByReadKey;
+    private final Map<ByteArrayWrapper, Long> threadByWrittenKey;
     private Set<ByteArrayWrapper> temporalReadKeys;
     private Set<ByteArrayWrapper> temporalWrittenKeys;
```
javicid, line 34: You can use the keyset of each map as the set of read/written keys. There is no need to keep separate Sets.

PR #1786, opened 2022-05-31
Added parallel transaction execution

javicid:
Adds a function to execute transactions in parallel based on the edges defined in the block header.


julianlen (approved):
LGTM!

PR #1794, opened 2022-06-03
Research/main parallel tx refactor validblock

ilanolkies:
Refactored tests and a little bit of the implementation. Also corrected one test that hit two cases at the same time, now hits one of the two.

Things we need to further look at:
- Why would we check if the block has `edges`. If it is validating edges it means it is already marked as post-RSKIP144, so we should asume edges exist
- Discussion already opened, why could a block have less than max edges?

### DIFF
--- a/rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionListsEdgesRule.java
+++ b/rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionListsEdgesRule.java
@@ -19,28 +19,30 @@ public class ValidTxExecutionListsEdgesRule implements BlockValidationRule {
     @Override
     public boolean isValid(Block block) {
         short[] edges = block.getHeader().getTxExecutionListsEdges();
-        int nTxs = block.getTransactionsList().size();
 
-        if (edges == null) {
+        if (edges == null || edges.length == 0) {
             return true;
         }
+
         if (edges.length > Constants.getMaxTransactionExecutionThreads()) {
             logger.warn("Invalid block: number of execution lists edges is greater than number of execution threads ({} vs {})",
                         edges.length, Constants.getMaxTransactionExecutionThreads());
             return false;
         }
-        short prev = 0;
 
+        short prev = 0;
         for (short edge : edges) {
             if (edge <= prev) {
                 logger.warn("Invalid block: execution lists edges are not in ascending order");
                 return false;
             }
-            if (edge > nTxs) {
-                logger.warn("Invalid block: execution list edge is out of bounds: {}", edge);
-                return false;
-            }
             prev = edge;
+        }
+
+        int txListSize = block.getTransactionsList().size();
+        if (edges[edges.length-1] > txListSize) {
+            logger.warn("Invalid block: execution list edge is out of bounds");
+            return false;
         }
         return true;
     }
--- a/rskj-core/src/test/java/co/rsk/validators/ValidTxExecutionListsEdgesTest.java
+++ b/rskj-core/src/test/java/co/rsk/validators/ValidTxExecutionListsEdgesTest.java
@@ -78,7 +78,7 @@ public class ValidTxExecutionListsEdgesTest {
     }
 
     @Test
-    public void blockWithEmptyListDefined() {
+    public void blockWithRepeatedEdge() {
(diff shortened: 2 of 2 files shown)


rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionListsEdgesRule.java
```diff
@@ -19,29 +19,31 @@ public class ValidTxExecutionListsEdgesRule implements BlockValidationRule {
     @Override
     public boolean isValid(Block block) {
         short[] edges = block.getHeader().getTxExecutionListsEdges();
-        int nTxs = block.getTransactionsList().size();
 
-        if (edges == null) {
+        if (edges == null || edges.length == 0) {
```
javicid, line 23: Why remove the second check?

rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionListsEdgesRule.java
```diff
@@ -19,29 +19,31 @@ public class ValidTxExecutionListsEdgesRule implements BlockValidationRule {
...
-                return false;
-            }
             prev = edge;
         }
+
+        int txListSize = block.getTransactionsList().size();
+        if (edges[edges.length-1] > txListSize) {
+            logger.warn("Invalid block: execution list edge is out of bounds");
+            return false;
+        }
```
javicid, line 46: Nice!

javicid, line 46: Nice!

rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionListsEdgesRule.java
```diff
@@ -19,29 +19,31 @@ public class ValidTxExecutionListsEdgesRule implements BlockValidationRule {
...
             if (edge <= prev) {
                 logger.warn("Invalid block: execution lists edges are not in ascending order");
                 return false;
             }
-            if (edge > nTxs) {
-                logger.warn("Invalid block: execution list edge is out of bounds: {}", edge);
-                return false;
-            }
             prev = edge;
         }
```
javicid, line 40: A cleaner version of this loop:

```
for (int i = 0; i < edges.length - 1; i++) {
        if (edges[i] > edges[i + 1]) {
            return false;
        }
    }
```

javicid, line 40: A cleaner version of this loop:

```
for (int i = 0; i < edges.length - 1; i++) {
        if (edges[i] > edges[i + 1]) {
            return false;
        }
    }
```

rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionListsEdgesRule.java
ilanolkies: 😄

rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionListsEdgesRule.java
ilanolkies: What did I remove? 🤔 If length is zero why would we iterate and all that.

`Constants.getMaxTransactionExecutionThreads()` must be >1

rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionListsEdgesRule.java
```diff
@@ -19,29 +19,34 @@ public class ValidTxExecutionListsEdgesRule implements BlockValidationRule {
...
                 return false;
             }
-            if (edge > nTxs) {
-                logger.warn("Invalid block: execution list edge is out of bounds: {}", edge);
-                return false;
-            }
-            prev = edge;
         }
+
+        if (edges[edges.length-1] > block.getTransactionsList().size()) {
```
ilanolkies, line 45:
```suggestion
        if (edges[edges.length-1] > block.getTransactionsList().size() - 1) {
```

This would keep the REMASC tx, that is the last tx, outside of the parallel buckets

PR #1797, opened 2022-06-06
Improve logs for flyover pegin rejection and funds migration

nathanieliov:
- Improve logs of flyover pegin rejection
- Rearrange some conditions and logs on `processFundsMigration` method to improve the trace of logs.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -919,21 +919,22 @@ public class BridgeSupport {
         List<UTXO> availableUTXOs = getRetiringFederationBtcUTXOs();
         Federation activeFederation = getActiveFederation();
 
-        logger.trace("[processFundsMigration] Going to process migration...");
+        if (federationIsInMigrationAge(activeFederation)) {
+            long migrationAge = rskExecutionBlock.getNumber() - activeFederation.getCreationBlockNumber();
+            logger.trace("[processFundsMigration] Active federation (age={}) is in migration age.", migrationAge);
+            if (hasMinimumFundsToMigrate(retiringFederationWallet)){
+                logger.info(
+                    "[processFundsMigration] Retiring federation has funds to migrate: {}.",
+                    retiringFederationWallet.getBalance().toFriendlyString()
+                );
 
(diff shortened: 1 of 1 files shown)


rmoreliovlabs (approved):

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -919,21 +919,22 @@ private void processFundsMigration(Transaction rskTx) throws IOException {
...
-        if (federationIsInMigrationAge(activeFederation) && hasMinimumFundsToMigrate(retiringFederationWallet)) {
-            logger.info(
-                "[processFundsMigration] Active federation (age={}) is in migration age and retiring federation has funds to migrate: {}.",
-                rskExecutionBlock.getNumber() - activeFederation.getCreationBlockNumber(),
-                retiringFederationWallet.getBalance().toFriendlyString()
-            );
+        if (federationIsInMigrationAge(activeFederation)) {
+            long migrationAge = rskExecutionBlock.getNumber() - activeFederation.getCreationBlockNumber();
+            logger.trace("[processFundsMigration] Active federation (age={}) is in migration age.", migrationAge);
+            if (hasMinimumFundsToMigrate(retiringFederationWallet)){
```
rmoreliovlabs, line 925: nit:
Shouldn't we log if when the condition is false? maybe saying
```
[processFundsMigration] Retiring federation doesn't have funds to migrate: {}.
```

rmoreliovlabs (approved):

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
nathanieliov: The thing is that it most likely will flood the logs with no useful information. Why? because the method `updateCollections` is called many times, which then calls `processFundsMigration` each time, and as the migration of funds is only done when the above conditions are met, which is something that only happens during a migration, it will be logging `Retiring federation doesn't have funds to migrate: {}.` almost all the time.

rmoreliovlabs: Ok, sounds reasonable to me.

Thanks for explaning.

josedahlquist (approved):

PR #1799, opened 2022-06-06
 Research/parallel-tx/splitting executions

julianlen:


rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -260,13 +260,26 @@ private boolean validateLogsBloom(BlockHeader header, BlockResult result) {
         return Arrays.equals(calculateLogsBloom(result.getTransactionReceipts()), header.getLogsBloom());
     }
 
+    @VisibleForTesting
+    public BlockResult executeForMining(Block block, BlockHeader parent, boolean discardInvalidTxs) {
```
javicid, line 264: This method is unnecessary as it is only used in tests. Can be replaced by a call to the other executeForMining.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -260,13 +260,26 @@ private boolean validateLogsBloom(BlockHeader header, BlockResult result) {
...
+
     @VisibleForTesting
     public BlockResult execute(Block block, BlockHeader parent, boolean discardInvalidTxs) {
         return execute(block, parent, discardInvalidTxs, false);
     }
 
     public BlockResult execute(Block block, BlockHeader parent, boolean discardInvalidTxs, boolean ignoreReadyToExecute) {
-        return executeInternal(null, 0, block, parent, discardInvalidTxs, ignoreReadyToExecute);
+        return executeParallelInternal(null, 0, block, parent, discardInvalidTxs, ignoreReadyToExecute);
     }
```
javicid, line 283: Unless I missed something, these two methods can be removed and replaced for calls to `executeParallelInternal`, which can be public and renamed to simply `execute` because it doesn't always perform the execution in parallel.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -585,14 +597,11 @@ private void addFeesToRemasc(LongAccumulator remascFees, Repository track) {
         }
     }
 
-    private BlockResult executeSequential(
-            @Nullable ProgramTraceProcessor programTraceProcessor,
-            int vmTraceOptions,
+    private BlockResult executeForMiningInternal(
```
javicid, line 600: Could also be named to something like `executeForMiningAfterRSKIP144`

javicid (changes requested):

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
julianlen: Indeed, it's weird why it is like this in master `@VisibleForTest` and Public together make no sense

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
julianlen: Right!

javicid (approved):

PR #1806, opened 2022-06-13
Account tx rate limit algorithm iter 2

illuque-iov:
Improvements:
- Improved rejected tx logging
- Take into account market status in order to activate `gasPrice` factor: blocks should have at least a 90% completion for market to be considered active
- Allow propagation of the first transaction of an account regardless consumed gas, otherwise its propagation could take very long, or not occur at all, depending on some conditions: if the tx consumes more than the initial granted Virtual Gas, it will be rejected, therefore not being propagated. It would only be propagated if it is re-tried again when enough was was accumulated. But the nodes receiving it after this broadcast would behave in the same way the previous node initially did, causing a delay or a break in the broadcasting. This problem affects only to first transaction, as for the next ones the account will either exist (with some accumulated gas) or be created (with max quota) on the map for all nodes (that received the tx).

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/net/handler/quota/TxQuota.java
+++ b/rskj-core/src/main/java/co/rsk/net/handler/quota/TxQuota.java
@@ -67,15 +67,20 @@ public class TxQuota {
      *
      * @param virtualGasToConsume Gas to be consumed
      * @param tx                  Tx being executed, used for logging purposes only
+     * @param blockNumber         Block number for logging purposes
      * @return True if there was enough accumulated gas
      */
-    public synchronized boolean acceptVirtualGasConsumption(double virtualGasToConsume, Transaction tx) {
+    public synchronized boolean acceptVirtualGasConsumption(double virtualGasToConsume, Transaction tx, long blockNumber) {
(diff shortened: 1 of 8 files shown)


Vovchyk (approved):

tiassumpcao (commented):
@illuque-iov isFirstTxSender makes sense to allow propagation to start without bumps. added to quota so next time that sender makes a TX it will either:
(a) have greater nonce and more accumulated gas next time (considering sufficient timestamp deltas)
(b) if sender account is abusive and attempts multiple consecutive TXs in a short period of time, gets limited by availableVirtualGas

quota updates and refresh seems ok, just wanted to double check the following is the expected for quota creation rule:

max quota => source account & not new
min quota => new source account 
min quota => any dest account

illuque-iov:
> 

Hi @tiassumpcao 

Your assumptions about quota creation are correct.

Regarding `isFirstTxSender`, the idea behind that is to prevent the slow propagation (or no propagation at all) of the very first transaction form an account if it exceeds the min virtual gas, as explained in the PR description and code comments.

PR #1811, opened 2022-06-16
Research/parallel-tx/optimize writes

ilanolkies:
> This cases should not be considered as writing a key, since they doesn't affect the output of the parallel execution:
> - Adding 0 balance
> - Write a key with the value that was already set

Ref: https://github.com/rsksmart/RSKIPs/pull/296/files#diff-886654e1645d9fd29f3e21aa0b8e9b71fb109544883d9adb4187853983657ddbR76-R78

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
+++ b/rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
@@ -391,6 +391,7 @@ public class MutableRepository implements Repository {
     }
 
     private void internalPut(byte[] key, byte[] value) {
+        if (Arrays.equals(value, mutableTrie.get(key))) return; // writes the same value
         tracker.addNewWrittenKey(new ByteArrayWrapper(key));
         mutableTrie.put(key, value);
     }
--- a/rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
+++ b/rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
@@ -299,6 +299,10 @@ public class MutableRepository implements Repository {
     public synchronized Coin addBalance(RskAddress addr, Coin value) {
         AccountState account = getAccountStateOrCreateNew(addr);
 
+        if (value.equals(new Coin(BigInteger.ZERO))) {
+            return account.getBalance();
+        }
+
         Coin result = account.addToBalance(value);
         updateAccountState(addr, account);
 
--- a/rskj-core/src/test/java/co/rsk/db/RepositoryTrackingTest.java
+++ b/rskj-core/src/test/java/co/rsk/db/RepositoryTrackingTest.java
@@ -6,8 +6,10 @@ import co.rsk.core.bc.ReadWrittenKeysTracker;
 import co.rsk.trie.Trie;
 import co.rsk.trie.TrieStore;
 import co.rsk.trie.TrieStoreImpl;
+import org.bouncycastle.util.encoders.Hex;
 import org.ethereum.datasource.HashMapDB;
 import org.ethereum.db.MutableRepository;
+import org.ethereum.vm.DataWord;
 
 import static org.junit.Assert.assertEquals;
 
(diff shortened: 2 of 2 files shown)


rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
```diff
@@ -299,6 +299,10 @@ public synchronized Coin getBalance(RskAddress addr) {
     public synchronized Coin addBalance(RskAddress addr, Coin value) {
         AccountState account = getAccountStateOrCreateNew(addr);
 
+        if (value.equals(new Coin(BigInteger.ZERO))) {
+            return account.getBalance();
+        }
+
         Coin result = account.addToBalance(value);
         updateAccountState(addr, account);
```
ilanolkies, line 307: Another possibility is:

```suggestion
        Coin result = account.addToBalance(value);

        if (!value.equals(new Coin(BigInteger.ZERO))) updateAccountState(addr, account);
```

rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
```diff
@@ -299,6 +299,10 @@ public synchronized Coin getBalance(RskAddress addr) {
     public synchronized Coin addBalance(RskAddress addr, Coin value) {
         AccountState account = getAccountStateOrCreateNew(addr);
 
+        if (value.equals(new Coin(BigInteger.ZERO))) {
+            return account.getBalance();
```
julianlen, line 303: What does it happen if you send 0 balance to a non-existent account? Should it be created?

julianlen: Another question: you add 0 balance to an account and then it updates its state. When it updates its state it invokes `internalPut` with the new AccountState encoded. Yet, the new AccountState should be equal to the old one. With the change in the `internalPut` method, it shouldn't be added, should it?

rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
```diff
@@ -299,6 +299,10 @@ public synchronized Coin getBalance(RskAddress addr) {
     public synchronized Coin addBalance(RskAddress addr, Coin value) {
         AccountState account = getAccountStateOrCreateNew(addr);
 
+        if (value.equals(new Coin(BigInteger.ZERO))) {
```
javicid, line 302: Here you can use Coin.ZERO

javicid (approved):

rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
julianlen: Well we can avoid this line and put all of the responsibility to the `internalPut`, if addBalance(0) doesn't change anything, the internalPut method won't change anything

ilanolkies: > What does it happen if you send 0 balance to a non-existent account? Should it be created?

> I had the same question. I guess it doesn't matter that it is not created?

Now the `addBalance` method is not changed

SergioDemianLerner: In Ethereum, currently when you add 0 to an account that has not been created, it is not created.
AFAIK, in RSK, it is.

rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
```diff
@@ -387,6 +387,7 @@ private byte[] getAccountData(RskAddress addr) {
     }
 
     private void internalPut(byte[] key, byte[] value) {
+        if (Arrays.equals(value, mutableTrie.get(key))) return; // writes the same value
```
SergioDemianLerner, line 390: Be careful with this.

If the value of thread 1 writes value X to storage key A, and thread 2 writes value Y to storage key A, then it doesn't matter what the previous value of the cell was !
There two thread cannot be executed in parallel!
 Because if Y is different from X, the end result will depend on the order.
 What we check is program INTENTIONs not what is changed or what is not changed. This is different from adding 0 to a balance, which always preserves the balance, no matter the previous value.

ilanolkies: Uh I get it! Very bad impl here

So there is no way to detect if the intention is "writing the same value it was already there"?

PR #1822, opened 2022-06-30
Research/parallel-tx/Fix isRemascTransaction validation and pass correct index

ilanolkies:
## Description

- `txindex` started in 1. So `isLastTx` was false
- `data` was empty but not null in `SnapshotManagerTest` (idk why) so `checkRemascTxZeroValues` was also false

## Motivation and Context

The remasc transaction was added to a parallel thread instead of the sequential one because it was wrongly identified as non-remasc tx

## How Has This Been Tested?

All tests pass now 😄 (see #1794)

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
@@ -641,6 +641,7 @@ public class BlockExecutor {
                     return getBlockResultAndLogExecutionInterrupted(block, metric, tx);
                 }
                 loggingDiscardedBlock(block, tx);
+                txindex++;
                 continue;
             }
 
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
@@ -608,7 +608,7 @@ public class BlockExecutor {
 
             TransactionExecutor txExecutor = transactionExecutorFactory.newInstance(
                     tx,
-                    txindex++,
+                    txindex,
                     block.getCoinbase(),
                     track,
                     block,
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -629,7 +629,7 @@ private BlockResult executeForMiningAfterRSKIP144(
             }
 
             Optional<Long> sublistGasAccumulated;
-            if (tx.isRemascTransaction(txindex, transactionsList.size())) {
+            if (tx.isRemascTransaction(txindex-1, transactionsList.size())) {
```
julianlen, line 632: Instead of substracting to txIndex, txIndex should be incremented at the end of the method.

julianlen (commented):
We should add a new test for the new check to the `isRemascTx`. In addition, here is a test for the sublists:

```
@Test
    public void remascShouldGoAlwaysToSequentialSublist() {
        if (!activeRskip144) {
            return;
        }
        Block parent = blockchain.getBestBlock();
        Block block = getBlockWithNIndependentTransactions(0, BigInteger.valueOf(21000), true);
        BlockResult blockResult = executor.executeAndFill(block, parent.getHeader());
        short[] shorts = {};
        Assert.assertArrayEquals(shorts, blockResult.getTxEdges());
        Assert.assertEquals(1, blockResult.getExecutedTransactions().size());
    }
```

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -670,6 +670,7 @@ private BlockResult executeForMiningAfterRSKIP144(
             loggingExecuteTxAndReceipt(block, i, tx);
 
             i++;
+            txindex++;
```
javicid, line 673: txindex must be incremented also if we discard the tx (right before the `continue` statement at line 628)

javicid (commented):
I think that creating a REMASC tx with `new RemascTransaction(blockNumber)` also returns a transaction that is not REMASC when checked. We could take the opportunity to fix that as well if it is not too much work.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -625,6 +625,7 @@ private BlockResult executeForMiningAfterRSKIP144(
                 }
 
                 loggingDiscardedBlock(block, tx);
+                txindex++;
```
javicid, line 628: shouldn't we increment txindex also at line 644?

javicid (changes requested):

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
ilanolkies: Ohh true!

PR #1828, opened 2022-07-11
Fix wrong Sync state detection

illuque-iov:
## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/net/BlockSyncService.java
+++ b/rskj-core/src/main/java/co/rsk/net/BlockSyncService.java
@@ -75,7 +75,7 @@ public class BlockSyncService {
 
     /**
      * Does initial preprocessing of the {@code block}.
-     * 
+     *
      * @return block and its ancestors (if any), which are not connected yet. Returns an empty list,
      * if the block is too advanced, already connected or some of its ancestors are still being awaited from a network.
      */
@@ -120,7 +120,7 @@ public class BlockSyncService {
 
     public BlockProcessResult processBlock(@Nonnull Block block, Peer sender, boolean ignoreMissingHashes) {
         final Instant start = Instant.now();
-        
+
         // Validate block header first to see if its PoW is valid at all
         if (!isBlockHeaderValid(block)) {
             logger.warn("Invalid block with number {} {} from {} ", block.getNumber(), block.getHash(), sender);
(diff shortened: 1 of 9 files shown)


rskj-core/src/main/java/co/rsk/net/BlockSyncService.java
```diff
@@ -150,7 +150,23 @@ private void tryReleaseStore(long bestBlockNumber) {
         }
     }
 
+    public boolean isReadyToAcceptBlocks() {
+        if (blockchain.getSize() == 1) { // empty blockchain
+            return false;
```
illuque-iov, line 155: I'm not sure about this check. Intuitively, I would say that an empty blockchain should be ready to accept blocks, but I've seen dedicated tests suggesting the contrary:
- `co.rsk.net.NodeBlockProcessorTest#noSyncingWithEmptyBlockchain`
- `co.rsk.net.AsyncNodeBlockProcessorTest#noSyncingWithEmptyBlockchain`

joaquinlpereyra: I have looked into this a bit. What I can say is that:
1. `isReadyToAcceptBlocks` refers specifically to a single type of message, `NewBlockHashMessage`, and probably the method should be renamed to reflect that.
2. Is possible, although surprising, that the intention behind the tests was to prevent blocks from receiving new block hashes they did not requests while they where not synced, similar to what is being checked now by checking the current known block number with zero. 

(2) should be confirmed. The code is a bit hard to follow. What I recommend is setting up a node and checking the logs. Writing down in detail the sync process wouldn't hurt either.

rskj-core/src/main/java/co/rsk/net/BlockSyncService.java
```diff
@@ -150,7 +150,23 @@ private void tryReleaseStore(long bestBlockNumber) {
...
+        }
+
+        if (getLastKnownBlockNumber() == 0) { // recently restarted, no info about lastKnownBlock yet
```
joaquinlpereyra, line 158: This should return false here? We don't want new block hashes when we recently started, we are busy syncing.

rskj-core/src/test/java/co/rsk/core/SnapshotManagerTest.java
```diff
@@ -133,7 +140,6 @@ public void revertToNonExistentSnapshot() {
     @Test
     public void revertToSnapshot() {
         addBlocks(10);
-
```
joaquinlpereyra, line 136: More tests besides cosmetic changes would be nice :) also, documenting why some tests are there (specially the surprising ones found on this PR) would be great.

illuque-iov: Cosmetic changes were due to the IDE auto-formatting, the only change was on line 379: `Assert.assertFalse(processor.isReadyToAcceptBlocks());`

This PR is preliminar to showcase the spotted scenarios, not production ready, so changes are expected and also more tests.

rskj-core/src/main/java/co/rsk/net/BlockSyncService.java
illuque-iov: Thanks Joaco, will take a look.

PR #1844, opened 2022-08-02
Added more services to get public IP from

illuque-iov:
Also added a dummy fallback IP in case any of the services works.

IMO using a dummy IP should be the last option since:
- we don't know if anyone is using a modified node of RSK where the IP is being used (unlikely but possible)
- we could start using the IP within the codebase without realising it could be fake (unlikely too I guess, given nodes could be in AWS, etc.)

That's why I thought adding more services was a good idea. Those services are suggested [here](https://www.baeldung.com/java-get-ip-address) and on a quick check they seem legit (used [this service](https://www.emailveritas.com/url-checker) to check it)

Let me know what do you think.

## Types of changes

## Checklist:

* **Other information**:


illuque-iov:
Closing as we want to remove external service dependency and we will follow a different approach

Issue #1849, opened 2022-08-03
The node has no RPC command to flush all state in memory into the databases in disk


SergioDemianLerner:
It is very important after a long synchronization and also for realistic performance comparisons to make sure that all blockchain state has been flushed to disk. 
I encounter this need when testing FlatDB performance. FlatDB uses memory caches that are periodically flushed to disk.

I also implemented this RPC call. I will submit a PR soon.

PR #1855, opened 2022-08-05
Enable config/command-line switch replacing embedded JAR resources with external files

SergioDemianLerner:
We provide a method to force rskj to load internal resource files from external files. This is controlled by configuration or command-line switches. 

## Description

We add a new section in the configuration file like this:

```
path-remaps {
 genesis =  /tmp/genesis/
}
```
This sections can specify a folder (without filename) or a filename with path, depending on the key.

The supported path remappings are:

* `build-info` (specifies a path to a folder, the file name is always "build-info.properties")
* `remasc` (the remasc configuration file)
* `rskbitcoincheckpoints` (the path to the folder containing a "*.checkpoints" file)
* `genesis` (the path to the folder containing a genesis file. The file name is taken from the config key "genesis")

If a key is not given, no remapping is performed.

## Motivation and Context

The benchmarking tool requires using different genesis files. To replace them, it's necessary to put the new genesis file somewhere in the class path and use a number of obscure methods to force Java to load it out of the JAR file. Many times it has failed. When it fails, it's impossible to diagnose the cause. 

The channelling of all resource loads through a single method  `loadConfigurationFile()` has other benefits:
* It enables much better debugging, as we can later introduce a logger element that outputs which files have been loaded and from where. For example, the new method allowed me to detect that `remasc` file is loaded each time when a block is processed (inefficient) and the `rskbitcoincheckpoints` file is not loaded on startup, but only when the certain bridge methods are called. The existence of both files should be checked at startup for sanity.
* It enables to mock resource loading and and test if the correct files are loaded.

## How Has This Been Tested?

Tested was limited to:
* passing the `-Xpath-remaps.genesis=/tmp/genesis` command line argument and checking that it works as expected.
* Using a configuration file having a `path-remaps` section.

## Types of changes

## Checklist:

There is no documentation for none of the configuration options in the code, so no documentation was added to the code.

I will contact the developer's site maintainer to update the page https://developers.rsk.co/rsk/node/configure/reference/

**Other information**:
I suggest that after or concurrent with this merge, some tests are added to verify the correct placement of resource files.


SergioDemianLerner:
This PR solves issue https://github.com/rsksmart/rskj/issues/1849

illuque-iov:
pipeline:run

illuque-iov:
pipeline:run

illuque-iov:
pipeline:run

illuque-iov:
pipeline:run

illuque-iov:
pipeline:run

illuque-iov:
pipeline:run

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -1386,7 +1389,9 @@ protected StateRootsStore buildStateRootsStore() {
     protected synchronized RskSystemProperties buildRskSystemProperties() {
         checkIfNotClosed();
 
-        return new RskSystemProperties(new ConfigLoader(cliArgs));
+        RskSystemProperties props = new RskSystemProperties(new ConfigLoader(cliArgs));
```
rmoreliovlabs, line 1392: We should use `getRskSystemProperties()` instead?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -148,6 +149,8 @@ public class BridgeSupport {
 
     private static final String INVALID_ADDRESS_FORMAT_MESSAGE = "invalid address format";
 
+    private static final ConfigFileLoader.ResourceLoader RESOURCE_LOADER = BridgeSupport.class::getResourceAsStream;
```
rmoreliovlabs, line 152: Maybe rename this to `BRIDGE_RESOURCE_LOADER`?

rmoreliovlabs (changes requested):
Can we add more tests around new code?

PR #1861, opened 2022-08-09
Store a new pegout entry when a pegout is created

nathanieliov:


rskj-core/src/test/java/co/rsk/peg/PegoutCreationEntryTest.java
```diff
@@ -0,0 +1,65 @@
+package co.rsk.peg;
+
+import co.rsk.bitcoinj.core.Sha256Hash;
+import co.rsk.crypto.Keccak256;
+import junit.framework.TestCase;
+
+public class PegoutCreationEntryTest extends TestCase {
```
marcos-iov, line 7: Why does it extend from TestCase?

rskj-core/src/test/java/co/rsk/peg/PegoutCreationEntryTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public void testGetPegoutCreationEntryBtcTxHash() {
+        Sha256Hash sha256Hash = Sha256Hash.ZERO_HASH;
```
marcos-iov, line 10: Let's use different values for btc hash and rsk hash. Just in case

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
```diff
@@ -776,15 +776,38 @@ public static byte[] serializeCoinbaseInformation(CoinbaseInformation coinbaseIn
...
+        if (data == null) {
+            return Optional.empty();
+        }
+
+        RLPList rlpList = (RLPList) RLP.decode2(data).get(0);
+        if (rlpList.size() != 1) {
+            throw new RuntimeException(String.format("Invalid serialized keccak256. Expected 1 element, but got %d", rlpList.size()));
+        }
+
+        return Optional.of(new Keccak256(rlpList.get(0).getRLPRawData()));
```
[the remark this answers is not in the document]
nathanieliov: Just to get an empty array when array has length equal to 0, instead of getting null. 

Implementation:
image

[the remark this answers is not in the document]
marcos-iov: I see, thanks!

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
```diff
@@ -776,15 +776,38 @@ public static byte[] serializeCoinbaseInformation(CoinbaseInformation coinbaseIn
...
+    }
+
+    public static byte[] serializeKeccak256(Keccak256 data) {
+        if (data == null){
+            return new byte[]{};
         }
-        return Sha256Hash.wrap(element.getRLPData());
+        return RLP.encodeList(RLP.encodeElement(data.getBytes()));
```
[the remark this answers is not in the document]
nathanieliov: I thought it was the correct way to do it. I left a question in the channel regarding this and the answer was to take a look at one of the existing methods which were using that approach to serialize/deserialize the value.

However, I did some tests changing it to use only the element encoding and it worked. So, I refactored the code and now it's being treated as a single element, as it really is, instead of as a list.

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
```diff
@@ -1019,6 +1020,39 @@ public void deserializeInteger() {
         Assert.assertEquals(1200, BridgeSerializationUtils.deserializeInteger(RLP.encodeBigInteger(BigInteger.valueOf(1200))).intValue());
     }
 
+    @Test
+    public void deserializeKeccak256() {
+        Keccak256 originalHash = PegTestUtils.createHash3(2);
+        byte[] encodedHash = RLP.encodeList(RLP.encodeElement(originalHash.getBytes()));
+
+        Optional<Keccak256> result = BridgeSerializationUtils.deserializeKeccak256(encodedHash);
```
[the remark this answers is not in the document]
nathanieliov: I agree this is not the best way to test it, and having the expected result hardcoded may be a better option, but as you mentioned, we need to discuss the source where we are going to get that expected result.

I found this tool that we may use: https://github.com/ethereum/pyrlp

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -3308,6 +3308,200 @@ public void saveNextPegoutHeight_after_RSKIP271() {
...
+
+        Sha256Hash hash = Sha256Hash.ZERO_HASH;
+
+        DataWord storageKeyForPegoutCreationIndex = DataWord.fromLongString("pegoutCreationIndex-" + hash);
+        when(
+            repository.getStorageBytes(
+                PrecompiledContracts.BRIDGE_ADDR,
+                storageKeyForPegoutCreationIndex
+            )
+        ).thenReturn(BridgeSerializationUtils.serializeKeccak256(Keccak256.ZERO_HASH));
```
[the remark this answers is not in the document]
nathanieliov: Why not? Shouldn't both pre and post-fork tests?

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -3308,6 +3308,200 @@ public void saveNextPegoutHeight_after_RSKIP271() {
...
+
+        verify(repository, never()).getStorageBytes(PrecompiledContracts.BRIDGE_ADDR, storageKeyForPegoutCreationIndex);
+    }
+
+    @Test
+    public void getPegoutCreationEntry_after_RSKIP298() {
+        Repository repository = mock(Repository.class);
+        BridgeStorageProvider provider = new BridgeStorageProvider(
+            repository, PrecompiledContracts.BRIDGE_ADDR,
+            config.getNetworkConstants().getBridgeConstants(), activationsAllForks
```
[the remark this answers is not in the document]
nathanieliov: Well, good point. Assuming a given behavior should stay in a certain way forever even after a release or a new consensus rule override it or discard it, we need to maintain those tests that guarantee that behavior as expected at that point of time to not get the node stuck when trying to sync.

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -3308,6 +3308,200 @@ public void saveNextPegoutHeight_after_RSKIP271() {
...
+
+        DataWord storageKeyForPegoutCreationIndex = DataWord.fromLongString("pegoutCreationIndex-" + hash);
+
+        provider.setPegoutCreationEntry(new PegoutCreationEntry(hash, keccak256));
+        provider.savePegoutCreationEntry();
+
+        verify(repository, never()).addStorageBytes(
+            PrecompiledContracts.BRIDGE_ADDR,
+            storageKeyForPegoutCreationIndex,
+            BridgeSerializationUtils.serializeKeccak256(keccak256)
```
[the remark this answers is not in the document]
nathanieliov: I don't like this change. Remember, what happened with the security issue on flyover pegin, the test was exactly doing this. This may hide a wrong behavior as happened on flyover.

rskj-core/src/test/java/co/rsk/peg/PegoutCreationEntryTest.java
nathanieliov: For no reason in particular. I just generated the test class using the IDE helper generator. I can modify the test class to be like the existing ones.

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -3308,6 +3308,200 @@ public void saveNextPegoutHeight_after_RSKIP271() {
...
+                PrecompiledContracts.BRIDGE_ADDR,
+                storageKeyForPegoutCreationIndex
+            )
+        ).thenReturn(BridgeSerializationUtils.serializeKeccak256(Keccak256.ZERO_HASH));
+
+        assertEquals(Optional.empty(), provider.getPegoutCreationEntry(
+            hash
+        ));
+
+        verify(repository, never()).getStorageBytes(PrecompiledContracts.BRIDGE_ADDR, storageKeyForPegoutCreationIndex);
```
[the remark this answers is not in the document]
nathanieliov: I prefer to keep with the correct value. Putting into practice the recommendations from the coininspect team

rskj-core/src/main/java/co/rsk/peg/PegoutCreationEntry.java
```diff
@@ -0,0 +1,52 @@
+package co.rsk.peg;
+
+import co.rsk.bitcoinj.core.Sha256Hash;
+import co.rsk.crypto.Keccak256;
+
+public class PegoutCreationEntry {
+    private Sha256Hash pegoutCreationEntryBtcTxHash;
+    private Keccak256 pegoutCreationEntryRskTxHash;
```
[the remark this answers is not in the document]
nathanieliov: Agree 100%

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -943,6 +940,44 @@ protected int getReleaseRequestQueueSize() throws IOException {
...
 
+    public void setPegoutCreationEntry(PegoutCreationEntry pegoutCreationEntry){
+        if (!activations.isActive(RSKIP298)){
+            return;
+        }
+
+        this.pegoutCreationEntry = pegoutCreationEntry;
+    }
+
+    public Optional<Keccak256> getPegoutCreationEntry(Sha256Hash btcTxHash){
```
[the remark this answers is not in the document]
nathanieliov: Well, by what the name of the method implies(`getPegoutCreationRskTxHashByBtcTxHash`) it should return only`rskTxHash`. So, I prefer to keep it as it is now.

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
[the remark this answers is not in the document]
marcos-iov: Sorry, I didn't quite select the right lines. The test is needed yes, I meant that what's not needed is the whole repository setup since the method will return empty immediately.

But, seeing the assertion you make below then it does make sense :)

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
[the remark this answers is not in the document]
marcos-iov: Cool

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
```diff
@@ -1019,38 +1021,89 @@ public void deserializeInteger() {
...
+        Keccak256 originalHash = PegTestUtils.createHash3(2);
+        byte[] encodedHash = RLP.encodeElement(originalHash.getBytes());
+
+        Optional<Keccak256> result = BridgeSerializationUtils.deserializeKeccak256(encodedHash);
+        Assert.assertTrue(result.isPresent());
+        Assert.assertEquals(originalHash, result.get());
+    }
+
+    @Test
+    public void deserializeKeccak256_nullValue() {
```
marcos-iov, line 1035: What do you think of a new test `deserializeKeccak256_zeroHash`?

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
```diff
@@ -1019,38 +1021,89 @@ public void deserializeInteger() {
...
 
-        Sha256Hash result = BridgeSerializationUtils.deserializeSha256Hash(encodedHash);
-        Assert.assertEquals(originalHash, result);
+        Optional<Sha256Hash> result = BridgeSerializationUtils.deserializeSha256Hash(encodedHash);
+        Assert.assertTrue(result.isPresent());
+        Assert.assertEquals(originalHash, result.get());
     }
 
     @Test
     public void deserializeSha256Hash_nullValue() {
```
marcos-iov, line 1094: deserializeSha256Hash_zeroHash test?

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -3308,6 +3308,209 @@ public void saveNextPegoutHeight_after_RSKIP271() {
         );
     }
 
+    @Test
+    public void getPegoutCreationEntry_before_RSKIP298() {
```
marcos-iov, line 3312:
```suggestion
    public void getPegoutCreationRskTxHashByBtcTxHash_before_RSKIP298() {
```

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -3308,6 +3308,209 @@ public void saveNextPegoutHeight_after_RSKIP271() {
...
+
+        assertEquals(Optional.empty(), provider.getPegoutCreationRskTxHashByBtcTxHash(
+            hash
+        ));
+
+        verify(repository, never()).getStorageBytes(PrecompiledContracts.BRIDGE_ADDR, storageKeyForPegoutCreationIndex);
+    }
+
+    @Test
+    public void getPegoutCreationEntry_after_RSKIP298() {
```
marcos-iov, line 3337:
```suggestion
    public void getPegoutCreationRskTxHashByBtcTxHash_after_RSKIP298() {
```

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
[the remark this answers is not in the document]
marcos-iov: Makes sense to me, thanks!

rskj-core/src/test/java/co/rsk/peg/PegoutCreationEntryTest.java
```diff
@@ -0,0 +1,74 @@
+package co.rsk.peg;
+
+import co.rsk.bitcoinj.core.Sha256Hash;
+import co.rsk.crypto.Keccak256;
+import org.apache.commons.lang3.RandomUtils;
```
marcos-iov, line 5: Let's not add any more dependencies to this package. I think there was a plan to remove it. Just used a fixed seed value when calling createHash method

rskj-core/src/test/java/co/rsk/peg/PegoutCreationEntryTest.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Test
+    public void testTestHashCode() {
+        Sha256Hash sha256Hash = PegTestUtils.createHash(4);
+        Keccak256 keccak256 = PegTestUtils.createHash3(4);
+        PegoutCreationEntry pegoutCreationEntry = new PegoutCreationEntry(sha256Hash, keccak256);
+        PegoutCreationEntry pegoutCreationEntryWithSameValues = new PegoutCreationEntry(sha256Hash, keccak256);
+
+        assertEquals(pegoutCreationEntry.hashCode(), pegoutCreationEntryWithSameValues.hashCode());
```
marcos-iov, line 56: I think it would be better to test it against the expected value. This way it's not really doing anything, it's like `assertEquals(a, a)`

rskj-core/src/test/java/co/rsk/peg/PegoutCreationEntryTest.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Test
+    public void testTestToString() {
+        Sha256Hash sha256Hash = Sha256Hash.ZERO_HASH;
+        Keccak256 keccak256 = Keccak256.ZERO_HASH;
+        PegoutCreationEntry pegoutCreationEntry = new PegoutCreationEntry(sha256Hash, keccak256);
+        PegoutCreationEntry pegoutCreationEntryWithSameValues = new PegoutCreationEntry(sha256Hash, keccak256);
+
+        assertEquals(pegoutCreationEntry.toString(), pegoutCreationEntryWithSameValues.toString());
```
marcos-iov, line 66: Same here, better test against the expected value

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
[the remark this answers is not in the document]
marcos-iov: I think that will work, it's an 'official' library so I guess it makes sense to put some trust into it. Thanks!

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
```diff
@@ -1019,38 +1023,122 @@ public void deserializeInteger() {
...
+        Optional<Keccak256> result = BridgeSerializationUtils.deserializeKeccak256(expectedSerializedHashBytes);
+        Assert.assertTrue(result.isPresent());
+
+        Keccak256 expectedHash = new Keccak256(HexUtils.decode(expectedHashHex.getBytes()));
+        Assert.assertEquals(expectedHash, result.get());
+    }
+
+    @Test
+    public void serializeKeccak256() {
+        Keccak256 originalHash = PegTestUtils.createHash3(2);
```
josedahlquist, line 1060: might be worth leaving a comment stating that calling `createHash3(2)` creates a a Keccak256 hash forced to be `02` and 30 zeroes...
At first I didn't realize this was the case and did not understand how this test could pass 😬

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -3308,6 +3308,242 @@ public void saveNextPegoutHeight_after_RSKIP271() {
...
+
+        ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
+        when(activations.isActive(ConsensusRule.RSKIP298)).thenReturn(true);
+        BridgeStorageProvider provider = new BridgeStorageProvider(
+            repository, PrecompiledContracts.BRIDGE_ADDR,
+            config.getNetworkConstants().getBridgeConstants(), activations
+        );
+
+        Sha256Hash hash = Sha256Hash.ZERO_HASH;
+        Keccak256 keccak256 = Keccak256.ZERO_HASH;
```
josedahlquist, line 3348: it might be better to use hashes different than zero to avoid false positives

rskj-core/src/test/java/co/rsk/peg/BridgeSupportReleaseBtcTest.java
```diff
@@ -483,6 +483,85 @@ public void release_verify_fee_above_fee_and_minimum_is_accepted() throws IOExce
         testPegoutMinimumWithFeeVerification(Coin.COIN, Coin.FIFTY_COINS, true);
     }
 
+    @Test
+    public void test_processPegoutsIndividually_before_RSKIP298_activation() throws IOException {
```
josedahlquist, line 487:
```suggestion
    public void test_pegout_creation_before_RSKIP298_activation_does_not_set_index() throws IOException {
```

rskj-core/src/test/java/co/rsk/peg/BridgeSupportReleaseBtcTest.java
```diff
@@ -483,6 +483,85 @@ public void release_verify_fee_above_fee_and_minimum_is_accepted() throws IOExce
...
+            .build();
+
+        Transaction rskTx = buildUpdateTx();
+        bridgeSupport.updateCollections(rskTx);
+
+        verify(provider, never()).setPegoutCreationEntry(any());
+    }
+
+    @Test
+    public void test_processPegoutsInBatch_after_RSKIP298() throws IOException {
```
josedahlquist, line 517:
```suggestion
    public void test_pegout_creation_after_RSKIP298_activation_sets_index() throws IOException {
```

rskj-core/src/test/java/co/rsk/peg/BridgeSupportReleaseBtcTest.java
```diff
@@ -483,6 +483,85 @@ public void release_verify_fee_above_fee_and_minimum_is_accepted() throws IOExce
...
+        when(provider.getReleaseTransactionSet()).thenReturn(new ReleaseTransactionSet(Collections.emptySet()));
+
+        BridgeSupport bridgeSupport = bridgeSupportBuilder
+            .withBridgeConstants(bridgeConstants)
+            .withProvider(provider)
+            .withActivations(activationMock)
+            .build();
+
+        Transaction rskTx = buildUpdateTx();
+        bridgeSupport.updateCollections(rskTx);
```
josedahlquist, line 511: I would be worth asserting the pegouts are created as expected but the index is not set. This test should be equivalent to the other one expect it just does not set the index.

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
```diff
@@ -1019,38 +1023,122 @@ public void deserializeInteger() {
...
+    public void deserializeKeccak256() {
+        String expectedValueHex = "0200000000000000000000000000000000000000000000000000000000000000";
+        String serializedValueHex = "a00200000000000000000000000000000000000000000000000000000000000000";
+        byte[] serializedValueBytes = HexUtils.decode(serializedValueHex.getBytes());
+
+        Optional<Keccak256> result = BridgeSerializationUtils.deserializeKeccak256(serializedValueBytes);
+        Assert.assertTrue(result.isPresent());
+
+        Keccak256 expectedHash = new Keccak256(HexUtils.decode(expectedValueHex.getBytes()));
+        Assert.assertEquals(expectedHash, result.get());
```
marcos-iov, line 1036:
```suggestion
        Keccak256 keccak256Hash = new Keccak256("0200000000000000000000000000000000000000000000000000000000000000");
        byte[] serializedKeccak256Hash = Hex.decode("a00200000000000000000000000000000000000000000000000000000000000000");
        
        Optional<Keccak256> result = BridgeSerializationUtils.deserializeKeccak256(serializedKeccak256Hash);

        Assert.assertTrue(result.isPresent());
        Assert.assertEquals(keccak256Hash, result.get());
```
Let me know what you think, I believe it's simpler and more straight forward

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
```diff
@@ -1019,38 +1023,122 @@ public void deserializeInteger() {
...
+
+    @Test
+    public void serializeKeccak256() {
+        Keccak256 originalHash = PegTestUtils.createHash3(2);
+
+        byte[] result = BridgeSerializationUtils.serializeKeccak256(originalHash);
+
+        String expectedEncodedHashHex = "a00200000000000000000000000000000000000000000000000000000000000000";
+        byte[] encodedExpectedHash = HexUtils.decode(expectedEncodedHashHex.getBytes());
+        Assert.assertArrayEquals(encodedExpectedHash, result);
```
marcos-iov, line 1066:
```suggestion
        Keccak256 keccak256Hash = new Keccak256("0200000000000000000000000000000000000000000000000000000000000000");
        byte[] serializedKeccak256Hash = Hex.decode("a00200000000000000000000000000000000000000000000000000000000000000");

        byte[] result = BridgeSerializationUtils.serializeKeccak256(keccak256Hash);

        Assert.assertArrayEquals(serializedKeccak256Hash, result);
```
What do you think?

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
```diff
@@ -25,20 +25,24 @@
...
 import co.rsk.peg.resources.TestConstants;
 import co.rsk.peg.utils.MerkleTreeUtils;
 import co.rsk.peg.flyover.FlyoverFederationInformation;
 import co.rsk.peg.whitelist.LockWhitelist;
 import co.rsk.peg.whitelist.LockWhitelistEntry;
 import co.rsk.peg.whitelist.OneOffWhiteListEntry;
+import co.rsk.util.HexUtils;
 import com.google.common.collect.Lists;
 import com.google.common.primitives.UnsignedBytes;
+import org.apache.commons.lang3.RandomUtils;
```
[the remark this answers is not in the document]
nathanieliov: No. It should be removed.

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
nathanieliov: A bit simpler. It works.

rskj-core/src/test/java/co/rsk/peg/BridgeSerializationUtilsTest.java
nathanieliov: Make it a bit more clear. I have no prob to implement it that way.

jeremy-then (approved):
LGTM.

marcos-iov (approved):

PR #1863, opened 2022-08-10
Allow the blockchain database to be used in read-only mode (with transient in-memory changes)

SergioDemianLerner:
This PR enables to configure the rskj node to open all blockchain databases in readonly mode (blockchain.readonly = true). The objective is to help debug synchronization multiple times without the need to restore gigabytes of data each time. 
 

## Description

We added a new configuration option: blockchain.readonly. When this is true:

* Al blockchain-related databases are open in read only mode.
* All blockchain underlying DataSourceWithCache are configured to prevent writes to the base data source, which makes them work as non-persisted databases. All modifications are kept in memory using its committedCache map.

Also we discovered a bug in the DataSourceWithCache logic of independent interest, which was fixed. Previously this bug consequence was minor: that certain elements (deletes) were not correctly cached. In readonly mode, the bug was critical.
Databases that are opened in read-only mode cannot be created (if they do not exist, an exception will be thrown).
Writes to read-only databases throw IllegalArgumentException.
 
## Motivation and Context

It's very common for testing and debugging to require re-executing certain actions from the same starting point. Since currently a blockchain consumes almost 70 GB, any test on using a real blockchain requires backing up the state and copying it over and over each time a new test is attempted. This is time consuming, and bad for the SSD.

This simple change allows to run any test multiple times (i.e. ExecuteBlocks) without worrying about inadvertently destroying or changing the state. 

## How Has This Been Tested?

Several unit test cases were added.
Old test cases were improved.
## Types of changes

## Checklist:


rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
```diff
@@ -56,14 +65,21 @@ public DataSourceWithCache(@Nonnull KeyValueDataSource base, int cacheSize) {
...
 
     public DataSourceWithCache(@Nonnull KeyValueDataSource base, int cacheSize,
-                               @Nullable CacheSnapshotHandler cacheSnapshotHandler) {
+                               @Nullable CacheSnapshotHandler cacheSnapshotHandler,
+                               boolean preventWritesToBase) {
+        this.preventWritesToBase = preventWritesToBase;
         this.cacheSize = cacheSize;
         this.base = Objects.requireNonNull(base);
-        this.uncommittedCache = new LinkedHashMap<>(cacheSize / 8, (float) 0.75, false);
+        this.uncommittedCache = new HashMap<>();
```
illuque-iov, line 73: What is the rationale for this change? Is order not relevant?

Asking because there is a test that checks for insertion order and randomly fails after this change: `org.ethereum.datasource.DataSourceWithCacheTest#putTwoKeyValuesWrittenInOrder`

rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
```diff
@@ -33,12 +33,21 @@
...
+ */
 public class DataSourceWithCache implements KeyValueDataSource {
 
     private static final Logger logger = LoggerFactory.getLogger("datasourcewithcache");
 
     private final int cacheSize;
+    private final boolean preventWritesToBase;
     private final KeyValueDataSource base;
+    // When preventWritesToBase=true, uncommittedCache will forever contain
+    // the changes, and committedCache will still work as a cache.
```
illuque-iov, line 50: Wouldn't it be better to not use uncommittedCache at all in readonly mode? At least, we would be consuming less memory.

On a quick look it would just be a matter of updating org.ethereum.datasource.DataSourceWithCache#putKeyValue to update committedCache when readonly.

rskj-core/src/main/java/org/ethereum/datasource/TransientDataSource.java
```diff
@@ -0,0 +1,236 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+// Datasource without cache. Only stored changes and avoid flushing to base.
+public class TransientDataSource implements KeyValueDataSource {
```
illuque-iov, line 35: This class seems to not be used.

rskj-core/src/main/java/org/ethereum/datasource/TransientMap.java
```diff
@@ -0,0 +1,212 @@
...
+            throw new UnsupportedOperationException();
+        }
+
+        public V get(Object key) {
+            if (m==null)
+                return added.get(key);
+
+            if (deleted.contains(key))
+                return null;
+            V ret = added.get(key);
```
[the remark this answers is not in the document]
illuque-iov: Ok, `HTreeMap` does not allow `null`, but we should enforce that in this class as well as, being it for general use, it could cause weird behaviors.

rskj-core/src/main/java/org/ethereum/datasource/TransientMap.java
```diff
@@ -0,0 +1,212 @@
...
+            } else {
+                ret = added.remove(key);
+            }
+            return ret;
+        }
+
+        public void putAll(Map<? extends K, ? extends V> am) {
+
+            if (m==null) {
+                return;
```
illuque-iov, line 114: Why are we exiting here if `m==null` and not doing the same for `put()`?

illuque-iov: And why not calling put() in the loop? Do we want a different behavior here?

rskj-core/src/main/java/org/ethereum/datasource/TransientMap.java
```diff
@@ -0,0 +1,212 @@
...
+            if (this.keySet == null) {
+                //TODO:  this.keySet = Collections.unmodifiableSet(this.m.keySet());
+            }
+
+            return this.keySet;
+        }
+
+        public Set<Entry<K, V>> entrySet() {
+            if (this.entrySet == null) {
+                //TODO: this.entrySet = unmodifiableEntrySet(this.m.entrySet());
```
illuque-iov, line 142: What is the idea for this TODOs?

rskj-core/src/main/java/org/ethereum/datasource/TransientMap.java
```diff
@@ -0,0 +1,212 @@
...
+
+            }
+        }
+
+        public int size() {
+
+            if (m==null) {
+                return added.size();
+            } else {
+                // todo: mix
```
illuque-iov, line 36: What is missing here?

rskj-core/src/main/java/org/ethereum/datasource/TransientMap.java
```diff
@@ -0,0 +1,212 @@
...
+                return added.put(key,value);
+
+            V ret = m.get(key);
+            if (ret!=null) {
+                if (deleted.contains(key)) {
+                    deleted.remove(key);
+                    changed.put(key,value);
+                    return null;
+                }
+                return changed.put(key,value);
```
illuque-iov, line 87: I think we are note returning a correct previous value when the entry is new in `changed`, since put returns `null`.

Shouldn't we be returning the value provided by `m` in that case to comply with Map interface?

rskj-core/src/main/java/org/ethereum/datasource/TransientMap.java
```diff
@@ -0,0 +1,212 @@
...
+                //TODO:  this.values = Collections.unmodifiableCollection(this.m.values());
+            }
+
+            return this.values;
+        }
+
+
+        public V getOrDefault(Object k, V defaultValue) {
+            V ret = get(k);
+            if (k==null)
```
illuque-iov, line 159: Shouldn't it be `ret == null`?

rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
SergioDemianLerner: Order is irrelevant.
See https://iov-labs.slack.com/archives/C747J6HN3/p1660151397008399 for a discussion

rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
SergioDemianLerner: I cannot find this line in the final code.

rskj-core/src/main/java/org/ethereum/datasource/TransientDataSource.java
SergioDemianLerner: You removed it. I must check my branch. I think is was an alternate implementation for a transient data source. Using a single map

rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
illuque-iov: Thanks. Then I will update the test.

rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
illuque-iov: Never mind, I got it wrong back then. We need the `uncommittedCache` as the actual memory store and `committedCache` as well because it can contain data from the snapshot.

rskj-core/src/main/java/org/ethereum/datasource/TransientDataSource.java
illuque-iov: Yeah, removed it because it was not used, do you have future plans for it?

rskj-core/src/main/java/org/ethereum/datasource/TransientMap.java
illuque-iov: I've reused put() in putAll(), please open a new thread if you think we shouldn't @SergioDemianLerner.

rskj-core/src/main/java/org/ethereum/datasource/TransientMap.java
illuque-iov: I've implemented this, please open a new thread if you think we shouldn't @SergioDemianLerner.

PR #1865, opened 2022-08-10
Research/main parallel tx revision

julianlen:


rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -461,26 +450,23 @@ private BlockResult executeParallel(
...
-                    deletedAccounts,
+                    new HashSet<>(),
                     discardInvalidTxs,
                     acceptInvalidTransactions,
-                    receipts,
-                    executedTransactions,
-                    transactionResults,
+                    new HashMap<>(),
+                    new HashMap<>(),
+                    new HashMap<>(),
```
javicid, line 463: I would remove these from the constructor as they break encapsulation and can cause problems in the future. Even if it means that we have to manually collect and merge the structures of the last execution. It's just another loop iteration at the end.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -549,20 +558,65 @@ private BlockResult executeParallel(
...
         profiler.stop(metric);
         logger.trace("End executeParallel.");
         return result;
     }
 
-    private void addFeesToRemasc(LongAccumulator remascFees, Repository track) {
-        long fee = remascFees.get();
-        if (fee > 0) {
-            track.addBalance(PrecompiledContracts.REMASC_ADDR, Coin.valueOf(fee));
+    private boolean detectCollisions(ReadWrittenKeysTracker tracker) {
```
javicid, line 571: Shouldn't this method be inside the tracker?

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -549,20 +558,65 @@ private BlockResult executeParallel(
...
+        threads.addAll(readKeysByThread.keySet());
+        threads.addAll(writtenKeysByThread.keySet());
+
+        for (Long threadId : threads) {
+
+            Set<ByteArrayWrapper> temporalReadKeys = readKeysByThread.getOrDefault(threadId, new HashSet<>());
+            Set<ByteArrayWrapper> temporalWrittenKeys = writtenKeysByThread.getOrDefault(threadId, new HashSet<>());
+
+            if (readKeysSize == 0 && writtenKeysSize == 0) {
+                readKeys = temporalReadKeys;
```
javicid, line 590: Be careful with this as it copies by reference.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -549,20 +558,65 @@ private BlockResult executeParallel(
...
+                readKeysSize = temporalReadKeys.size();
+                writtenKeys = temporalWrittenKeys;
+                writtenKeysSize = temporalWrittenKeys.size();
+                continue;
+            }
+
+            writtenKeys.removeAll(temporalReadKeys);
+            writtenKeys.removeAll(temporalWrittenKeys);
```
javicid, line 598: Use `Collections.disjoint` better and you can use two nested loops to compare all threads 1v1

javicid (changes requested):

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
julianlen: The last execution needs the merged state. Many of the transactions within the sequential sublist depend on the final state of the parallel execution.

PR #1874, opened 2022-09-04
Fix node startup when database dir folder is empty

rmoreliovlabs:
## Description
Fixing exception thrown when running a node with empty database folder. However it should generate the `dbKind.properties` files

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/datasource/KeyValueDataSource.java
```diff
@@ -140,7 +140,7 @@ static void validateDbKind(DbKind currentDbKind, String databaseDir, boolean dat
         File dir = new File(databaseDir);
         boolean databaseDirExists = dir.exists() && dir.isDirectory();
 
-        if (!databaseDirExists) {
+        if (!databaseDirExists || dir.list().length == 0) {
```
illuque-iov, line 143: `dir.list()` may return null if `databaseDir` is not a directory, causing a NPE.

This is not so important, as it is run just on startup and it would mean a misconfiguration, but I think it is worth fixing it now that we are working on it.

rskj-core/src/test/java/org/ethereum/datasource/KeyValueDataSourceTest.java
```diff
@@ -126,6 +132,19 @@ public void updateBatchWithNulls() {
         keyValueDataSource.updateBatch(updatedValues, Collections.emptySet());
     }
 
+    @Test
+    public void testShouldValidateKindWithEmptyDbDirAndResetDbFalseSuccessfully() {
```
illuque-iov, line 136: These two new tests are run 10 times, as this is a parameterised test, without any change on its params. This extra job is not needed, so maybe it's worth moving them to another file.

illuque-iov (changes requested):

rskj-core/src/main/java/org/ethereum/datasource/KeyValueDataSource.java
rmoreliovlabs: `&& dir.isDirectory()` would make this impossible to happen

illuque-iov: What if it is a regular file?
`dir.exists() == true`
`dir.isDirectory() == false`
`databaseDirExists == false` => `!databaseDirExists == true` and `dir.list().length == 0` will be checked

rskj-core/src/test/java/org/ethereum/datasource/KeyValueDataSourceTest.java
rmoreliovlabs: Done

rskj-core/src/main/java/org/ethereum/datasource/KeyValueDataSource.java
rmoreliovlabs: if `!databaseDirExists == true` it ends up there since it is an or conditional

rmoreliovlabs: Added a validation if `database.dir` is not a folder. Suggested by @Vovchyk

illuque-iov: > if `!databaseDirExists == true` it ends up there since it is an`OR` conditional

You're right, I don't know what I understood there... 🤯

Vovchyk, line 143: could you please cover this with unit tests?

Vovchyk (changes requested):

Vovchyk:
pipeline:run

Vovchyk (changes requested):
it looks like there is a bug in the DbKind file generation flow.. seems due to `KeyValueDataSource.validateDbKind()` being called before reset is performed.

I don't see `dbKind.properties` in the db folder with these flags: `-Dkeyvalue.datasource=rocksdb` and `--reset`

rmoreliovlabs:
> it looks like there is a bug in the DbKind file generation flow.. seems due to `KeyValueDataSource.validateDbKind()` being called before reset is performed.
> 
> I don't see `dbKind.properties` in the db folder with these flags: `-Dkeyvalue.datasource=rocksdb` and `--reset`

Addressed

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -313,8 +313,14 @@ public synchronized NodeRunner getNodeRunner() {
             RskSystemProperties rskSystemProperties = getRskSystemProperties();
             if (rskSystemProperties.databaseReset() || rskSystemProperties.importEnabled()) {
                 FileUtil.recursiveDelete(rskSystemProperties.databaseDir());
+
+                File file = new File(rskSystemProperties.databaseDir());
+
+                file.mkdir();
```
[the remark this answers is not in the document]
Vovchyk: but we do create the folder in `KeyValueDataSource.generatedDbKindFile()` if it's missing

besides that it seems that the case of `! (rskSystemProperties.databaseReset() || rskSystemProperties.importEnabled())` is not respected but this

Vovchyk (changes requested):

Vovchyk (approved):

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk (approved):

illuque-iov (approved):

PR #1879, opened 2022-09-13
Update netty

rmoreliovlabs:
## Description
Update netty version

## Motivation and Context
On scope of Java upgrade task, we tried to upgrade Netty library to the stable and recommended 4.1.78.Final

All tests ran fine but after a merge with master, one of the added tests started failing. The cause is that com.squareup.okhttp.ws.WebSocketListener#onOpen is not called in co.rsk.rpc.netty.Web3WebSocketServerTest#smokeTest(byte[], java.lang.String) with the new version but it is on the old one.

Apparently it only happens for tests using the default path /websocket (#smokeTest and #smokeTestWithBigJson) and, when launched in batch, only for the first one using this path fails, the second one seems to be working fine. When launched isolated, both fail. This suggests that we are sharing some state between them.

Therefore the upgrade was reverted. Further investigation is needed to ensure it is only an issue with the test or anything else could be affected.

## Types of changes

## Checklist:

* **Other information**:
PowPeg PR: https://github.com/rsksmart/powpeg-node/pull/107

*fed:update-netty-version*

### DIFF
--- a/rskj-core/src/main/java/co/rsk/rpc/netty/RskWebSocketServerProtocolHandler.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/netty/RskWebSocketServerProtocolHandler.java
@@ -15,7 +15,10 @@ public class RskWebSocketServerProtocolHandler extends WebSocketServerProtocolHa
 
     public RskWebSocketServerProtocolHandler(String websocketPath, int maxFrameSize) {
         // there are no subprotocols nor extensions
-        super(websocketPath, null, false, maxFrameSize, true);
+        // Note: A port:host can only have one webSocketPath, and that's a Netty limitation.
+        // For more information about
+        // Netty limitations: https://stackoverflow.com/questions/8778806/how-to-handle-different-url-websocket-connections-in-netty/71757361#71757361
+        super(websocketPath, null, false, maxFrameSize, false, true);
     }
 
     @Override
--- a/rskj-core/build.gradle
+++ b/rskj-core/build.gradle
@@ -58,7 +58,7 @@ tasks.withType(AbstractArchiveTask) {
 
 ext {
     libVersions = [
-            nettyVer               : '4.0.56.Final',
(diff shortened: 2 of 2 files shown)


rskj-core/src/main/java/co/rsk/rpc/netty/RskWebSocketServerProtocolHandler.java
```diff
@@ -15,7 +15,7 @@ public class RskWebSocketServerProtocolHandler extends WebSocketServerProtocolHa
 
     public RskWebSocketServerProtocolHandler(String websocketPath, int maxFrameSize) {
         // there are no subprotocols nor extensions
-        super(websocketPath, null, false, maxFrameSize, true);
+        super(websocketPath, null, false, maxFrameSize, true, true);
```
illuque-iov, line 18: @rmoreliovlabs one small thing, please add some comment explaining why we set this flag and the Netty limitation (`a port:host can only have one webSocketPath, and that's a Netty limitation.`)

This way the next dev changing this will be aware ;)

Vovchyk (dismissed):

rskj-core/src/main/java/co/rsk/rpc/netty/RskWebSocketServerProtocolHandler.java
rmoreliovlabs: Done

rmoreliovlabs:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk (approved):

Vovchyk:
pipeline:run

rmoreliovlabs:
pipeline:run

rmoreliovlabs:
pipeline:run

rmoreliovlabs:
pipeline:run

rmoreliovlabs:
pipeline:run

Vovchyk (approved):

PR #1880, opened 2022-09-14
Only emitting the add_signature event when the transaction was actual…

jeremy-then:
Only emitting the add_signature event when the transaction was actually signed.

The test are covering 4 scenarios:

1) Pre RSKIP326, the event should be emitted as usual
2) Pre RSKIP326, the event should be emitted even if no signature was added
3) Post RSKIP326, the event will not be emitted if the transaction was not signed for any reason (invalid signature, etc)
3) Post RSKIP326, the event should only emit if the transaction was actually signed


rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -5576,6 +5583,176 @@ public void addSignatureCreateEventLog() throws Exception {
         verify(eventLogger, times(1)).logAddSignature(federatorPubKey, btcTx, rskTxHash.getBytes());
     }
 
+    @Test
+    public void addSignature_invalidSignatureBeforeRSKIP326_eventStillEmitted() throws Exception {
```
marcos-iov, line 5587: I like the idea you proposed about creating an auxiliary test method to avoid repeating the same logic in every test.

It could be something like this 
```
private void test_addSignature_EventEmmited(boolean isRskip326Active, boolean isSignatureValid, boolean federatorAlreadySigned) {
```
That would allow us to test different cases with the same code I think

marcos-iov (commented):
Do you think we could move all the `addSignature` related tests to a new class? We've been trying to reduce the size of BridgeSupportTest class that is already huge.

You could create a new class `BridgeSupportAddSignatureTest`

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -0,0 +1,727 @@
...
+
+        Federation activeFederation = new Federation(
+                FederationTestUtils.getFederationMembersWithBtcKeys(federation1Keys),
+                Instant.ofEpochMilli(1000L),
+                0L,
+                btcRegTestParams
+        );
+
+        BridgeStorageProvider provider = mock(BridgeStorageProvider.class);
+        BridgeSupport bridgeSupport = new BridgeSupport(
```
kelvinator07, line 81: You Should Use The BridgeSupportBuilder Class Whenever You Need An Instance Of The BridgeSupport For Testing

marcos-iov: @kelvinator07 These are pre-existing tests that were moved to this new class. 

I think we should always use BridgeSupportBuilder in new tests, for the existing ones it's a loooot of work to change them all. What I usually do is just change a couple of the existing ones when adding new tests. Gradually we'll get there one day

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -0,0 +1,727 @@
...
+                .withBridgeConstants(bridgeConstantsRegtest)
+                .withProvider(new BridgeStorageProvider(
+                        track,
+                        contractAddress,
+                        bridgeConstantsRegtest,
+                        activationsAfterForks
+                ))
+                .withRepository(track)
+                .withEventLogger(eventLogger)
+                .withExecutionBlock(mock(Block.class))
```
marcos-iov, line 652: BridgeSupportBuilder sets a mock by default for every param, so this isn't really needed

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -0,0 +1,727 @@
...
+        track.commit();
+
+        track = repository.startTracking();
+        ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
+        List<LogInfo> logs = new ArrayList<>();
+        BridgeEventLogger eventLogger = new BrigeEventLoggerLegacyImpl(bridgeConstantsRegtest, activations, logs);
+
+        BridgeSupport bridgeSupport = bridgeSupportBuilder
+                .withBridgeConstants(bridgeConstantsRegtest)
+                .withProvider(new BridgeStorageProvider(
```
marcos-iov, line 644: Shouldn't it use the `provider` object created above? Why create a new one?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -0,0 +1,727 @@
...
+            List derEncodedSigs2 = new ArrayList();
+            for (int i = 0; i < numberOfInputsToSign; i++) {
+                derEncodedSigs2.add(derEncodedSig2);
+            }
+            bridgeSupport.addSignature(findPublicKeySignedBy(federation.getBtcPublicKeys(), privateKeysToSignWith.get(1)), derEncodedSigs2, keccak256.getBytes());
+        }
+        bridgeSupport.save();
+        track.commit();
+
+        provider = new BridgeStorageProvider(repository, PrecompiledContracts.BRIDGE_ADDR, bridgeConstantsRegtest, activationsBeforeForks);
```
marcos-iov, line 695: Any idea why it's creating another provider object here?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -0,0 +1,727 @@
...
+        when(activations.isActive(ConsensusRule.RSKIP326)).thenReturn(rskip326Active);
+
+        BridgeSupport bridgeSupport = bridgeSupportBuilder
+                .withBridgeConstants(bridgeConstantsRegtest)
+                .withProvider(provider)
+                .withRepository(track)
+                .withEventLogger(eventLogger)
+                .withBtcLockSenderProvider(new BtcLockSenderProvider())
+                .withExecutionBlock(mock(Block.class))
+                .withBtcBlockStoreFactory(null)
```
marcos-iov, line 350: These are probably not needed

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -0,0 +1,727 @@
...
+        }
+    }
+
+    private BtcECKey findPublicKeySignedBy(List<BtcECKey> pubs, BtcECKey pk) {
+        for (BtcECKey pub : pubs) {
+            if (Arrays.equals(pk.getPubKey(), pub.getPubKey())) {
+                return pub;
+            }
+        }
+        return pk;
```
marcos-iov, line 724: This is a bit strange. If it can't find the public key that corresponds to the given private key, then it returns the private?

I think it would make more sense to return an Optional, what do you think? This way it may lead to confussion

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -0,0 +1,727 @@
...
+        addSignatureFromValidFederator(keys, 2, true, false, "InvalidParameters");
+    }
+
+    @Test
+    public void addSignatureNonCanonicalSignature() throws Exception {
+        List<BtcECKey> keys = Collections.singletonList(BridgeRegTestConstants.REGTEST_FEDERATION_PRIVATE_KEYS.get(0));
+        addSignatureFromValidFederator(keys, 1, false, false, "InvalidParameters");
+    }
+
+    private void test_addSignature_EventEmitted(boolean rskip146Active, boolean rskip326Active, byte[] signature, int wantedNumberOfInvocations) throws Exception {
```
marcos-iov, line 312: `rskip146Active` is always true when you call it. And I think it makes sense, we just want to test how many times BridgeEventLogger was called not which implementation.

So I would just this parameter and set RSKIP146 to active when creating the activations mock

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -0,0 +1,727 @@
...
+                .withProvider(provider)
+                .withRepository(track)
+                .withEventLogger(eventLogger)
+                .withBtcLockSenderProvider(new BtcLockSenderProvider())
+                .withExecutionBlock(mock(Block.class))
+                .withBtcBlockStoreFactory(null)
+                .withActivations(activations)
+                .build();
+
+        BtcECKey privateKeyToSignWith = BridgeRegTestConstants.REGTEST_FEDERATION_PRIVATE_KEYS.get(0);
```
marcos-iov, line 354:
```suggestion
        int indexOfKeyToSignWith = 0;
        BtcECKey privateKeyToSignWith = BridgeRegTestConstants.REGTEST_FEDERATION_PRIVATE_KEYS.get(indexOfKeyToSignWith);
```
Then we can use this same index value to get the public key

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -0,0 +1,727 @@
...
+            byte[] program = chunks.get(chunks.size() - 1).data;
+            Script redeemScript = new Script(program);
+            Sha256Hash sigHash = btcTx.hashForSignature(0, redeemScript, BtcTransaction.SigHash.ALL, false);
+            BtcECKey.ECDSASignature sig = privateKeyToSignWith.sign(sigHash);
+            derEncodedSigs = Collections.singletonList(sig.encodeToDER());
+        } else {
+            derEncodedSigs = Collections.singletonList(signature);
+        }
+
+        BtcECKey federatorPubKey = findPublicKeySignedBy(federation.getBtcPublicKeys(), privateKeyToSignWith);
```
marcos-iov, line 370:
```suggestion
        BtcECKey federatorPubKey = BridgeRegTestConstants.REGTEST_FEDERATION_PUBLIC_KEYS.get(indexOfKeyToSignWith);
```
What do you think?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -0,0 +1,727 @@
...
+                .withExecutionBlock(mock(Block.class))
+                .withBtcBlockStoreFactory(null)
+                .withActivations(activations)
+                .build();
+
+        BtcECKey privateKeyToSignWith = BridgeRegTestConstants.REGTEST_FEDERATION_PRIVATE_KEYS.get(0);
+
+        List derEncodedSigs;
+
+        if(signature == null) {
```
marcos-iov, line 358: Just an idea. Instead of receiving the signature in this method, you could receive a boolean value `useValidSignature`. 

Then it's this method's responsibility to either create a valid or a malformed signature. The caller only determines what type of value he wants to test.

What do you think?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -0,0 +1,727 @@
...
+        }
+
+        BtcECKey federatorPubKey = findPublicKeySignedBy(federation.getBtcPublicKeys(), privateKeyToSignWith);
+        bridgeSupport.addSignature(federatorPubKey, derEncodedSigs, rskTxHash.getBytes());
+
+        verify(eventLogger, times(wantedNumberOfInvocations)).logAddSignature(federatorPubKey, btcTx, rskTxHash.getBytes());
+    }
+
+    @Test
+    public void addSignature_calledTwiceWithSameFederator_emitEventOnlyOnce() throws Exception {
```
marcos-iov, line 377: Couldn't this case be covered using `test_addSignature_EventEmitted` method? You could simply add a boolean `shouldSignTwice` .

That would also allow us to test this case before and after RSKIP326, this test only covers the after activation

marcos-iov (commented):
Happy to see we got over 500 lines removed from BridgeSupportTest! Huge improvement

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
jeremy-then: I thought of it, just thought that it would have too many parameters.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -7141,127 +6702,6 @@ private void assertLockingCap(
...
-        provider.save();
-        track.commit();
-
-        track = repository.startTracking();
-        ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
-        List<LogInfo> logs = new ArrayList<>();
-        BridgeEventLogger eventLogger = new BrigeEventLoggerLegacyImpl(bridgeConstantsRegtest, activations, logs);
-        BridgeSupport bridgeSupport = getBridgeSupport(
-            bridgeConstantsRegtest,
-            new BridgeStorageProvider(
```
jeremy-then, line 7181: @marcos-iov , the original test was creating a new `BridgeStorageProvider` instance even though it declares one above. I'm not sure why this is this way.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
jeremy-then: The original test was like that. Take a look: https://github.com/rsksmart/rskj/pull/1880/files#diff-bca5614b0a0ce7ba81d79322090b66cee8d9289e3430798d67767568abc8491aL7181

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
marcos-iov: I think we are missing a test pre RSKIP326, right? It should emit the event twice in that case

marcos-iov (approved):

kelvinator07 (approved):

PR #1881, opened 2022-09-14
Using iterator instead of loading all keys to avoid java heap errors

rmoreliovlabs:
## Description
Improving db migration process by using iterator instead of getting all the keys

## Motivation and Context
In large databases it is not a good idea to get all existing keys. Instead we are using an iterator to iterate key by key.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
+++ b/rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
@@ -19,13 +19,10 @@ package co.rsk.cli.tools;
 
 import co.rsk.RskContext;
 import co.rsk.cli.CliToolRskContextAware;
+import org.ethereum.datasource.DataSourceKeyIterator;
 import org.ethereum.datasource.DbKind;
 import org.ethereum.datasource.KeyValueDataSource;
-import org.ethereum.db.ByteArrayWrapper;
-import org.ethereum.util.ByteUtil;
 import org.ethereum.util.FileUtil;
-import org.iq80.leveldb.DBIterator;
-import org.rocksdb.RocksIterator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -187,38 +184,22 @@ public class DbMigrate extends CliToolRskContextAware {
 
         logger.info("Migrating data source: {}", sourceKeyValueDataSource.getName());
 
-        Object iterator = sourceKeyValueDataSource.iterator();
+        DataSourceKeyIterator iterator = sourceKeyValueDataSource.keyIterator();
 
-        if (iterator instanceof RocksIterator) {
-            RocksIterator rocksIterator = (RocksIterator) iterator;
+        while (iterator.hasNext()) {
+            byte[] data = iterator.next().getData();
 
-            for (rocksIterator.seekToFirst(); rocksIterator.isValid(); rocksIterator.next()) {
-                byte[] data = rocksIterator.key();
+            targetKeyValueDataSource.put(
+                    data,
+                    sourceKeyValueDataSource.get(data)
+            );
+        }
 
-                targetKeyValueDataSource.put(
-                        data,
-                        sourceKeyValueDataSource.get(data)
-                );
-            }
-
-            rocksIterator.close();
-        } else {
-            DBIterator dbIterator = (DBIterator) iterator;
-
-            for (dbIterator.seekToFirst(); dbIterator.hasNext(); dbIterator.next()) {
-                byte[] data = dbIterator.peekNext().getKey();
-                targetKeyValueDataSource.put(
-                        data,
-                        sourceKeyValueDataSource.get(data)
-                );
-            }
-
-            try {
-                dbIterator.close();
-            } catch (IOException e) {
-                logger.error("An error happened closing DB Iterator", e);
-                throw new RuntimeException(e);
-            }
+        try {
+            iterator.close();
+        } catch (Exception e) {
+            logger.error("An error happened closing DB Key Iterator", e);
+            throw new RuntimeException(e);
         }
 
         sourceKeyValueDataSource.close();
--- a/rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
+++ b/rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
@@ -184,19 +184,17 @@ public class DbMigrate extends CliToolRskContextAware {
 
         logger.info("Migrating data source: {}", sourceKeyValueDataSource.getName());
 
-        DataSourceKeyIterator iterator = sourceKeyValueDataSource.keyIterator();
+        try (DataSourceKeyIterator iterator = sourceKeyValueDataSource.keyIterator()) {
+            iterator.seekToFirst();
 
-        while (iterator.hasNext()) {
-            byte[] data = iterator.next().getData();
+            while (iterator.hasNext()) {
+                byte[] data = iterator.next().getData();
 
-            targetKeyValueDataSource.put(
-                    data,
-                    sourceKeyValueDataSource.get(data)
-            );
-        }
-
-        try {
-            iterator.close();
+                targetKeyValueDataSource.put(
+                        data,
+                        sourceKeyValueDataSource.get(data)
+                );
+            }
         } catch (Exception e) {
             logger.error("An error happened closing DB Key Iterator", e);
             throw new RuntimeException(e);
--- a/rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
+++ b/rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
@@ -21,7 +21,11 @@ import co.rsk.RskContext;
 import co.rsk.cli.CliToolRskContextAware;
 import org.ethereum.datasource.DbKind;
 import org.ethereum.datasource.KeyValueDataSource;
+import org.ethereum.db.ByteArrayWrapper;
+import org.ethereum.util.ByteUtil;
 import org.ethereum.util.FileUtil;
+import org.iq80.leveldb.DBIterator;
+import org.rocksdb.RocksIterator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -37,10 +41,10 @@ import java.util.stream.Stream;
 /**
  * The entry point for db migration CLI tool
  * This is an experimental/unsupported tool
- *
+ * <p>
  * Required cli args:
  * - args[0] - database target where we are going to insert the information from the current selected database.
- *
+ * <p>
  * We do support the migrations between the following databases:
  * - LevelDb (leveldb as argument)
  * - RocksDb (rocksdb as argument)
@@ -183,12 +187,39 @@ public class DbMigrate extends CliToolRskContextAware {
 
         logger.info("Migrating data source: {}", sourceKeyValueDataSource.getName());
 
-        sourceKeyValueDataSource.keys().forEach(sourceKey ->
+        Object iterator = sourceKeyValueDataSource.iterator();
+
+        if (iterator instanceof RocksIterator) {
+            RocksIterator rocksIterator = (RocksIterator) iterator;
+
+            for (rocksIterator.seekToFirst(); rocksIterator.isValid(); rocksIterator.next()) {
+                byte[] data = rocksIterator.key();
+
                 targetKeyValueDataSource.put(
-                        sourceKey.getData(),
-                        sourceKeyValueDataSource.get(sourceKey.getData())
-                )
-        );
+                        data,
+                        sourceKeyValueDataSource.get(data)
+                );
+            }
+
+            rocksIterator.close();
+        } else {
+            DBIterator dbIterator = (DBIterator) iterator;
+
+            for (dbIterator.seekToFirst(); dbIterator.hasNext(); dbIterator.next()) {
+                byte[] data = dbIterator.peekNext().getKey();
+                targetKeyValueDataSource.put(
+                        data,
+                        sourceKeyValueDataSource.get(data)
+                );
+            }
+
+            try {
+                dbIterator.close();
+            } catch (IOException e) {
+                logger.error("An error happened closing DB Iterator", e);
+                throw new RuntimeException(e);
+            }
+        }
 
         sourceKeyValueDataSource.close();
         targetKeyValueDataSource.close();
--- a/rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
+++ b/rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
@@ -185,8 +185,6 @@ public class DbMigrate extends CliToolRskContextAware {
         logger.info("Migrating data source: {}", sourceKeyValueDataSource.getName());
 
         try (DataSourceKeyIterator iterator = sourceKeyValueDataSource.keyIterator()) {
-            iterator.seekToFirst();
-
             while (iterator.hasNext()) {
                 byte[] data = iterator.next();
 
--- a/rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
+++ b/rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
@@ -188,7 +188,7 @@ public class DbMigrate extends CliToolRskContextAware {
             iterator.seekToFirst();
 
             while (iterator.hasNext()) {
-                byte[] data = iterator.next().getData();
+                byte[] data = iterator.next();
 
                 targetKeyValueDataSource.put(
                         data,
--- a/rskj-core/src/main/java/org/ethereum/datasource/DataSourceKeyIterator.java
+++ b/rskj-core/src/main/java/org/ethereum/datasource/DataSourceKeyIterator.java
@@ -19,10 +19,8 @@
 
 package org.ethereum.datasource;
 
-import org.ethereum.db.ByteArrayWrapper;
-
 import java.util.*;
 
-public interface DataSourceKeyIterator extends Iterator<ByteArrayWrapper>, AutoCloseable {
+public interface DataSourceKeyIterator extends Iterator<byte[]>, AutoCloseable {
     void seekToFirst();
 }
--- a/rskj-core/src/main/java/org/ethereum/datasource/DataSourceKeyIterator.java
+++ b/rskj-core/src/main/java/org/ethereum/datasource/DataSourceKeyIterator.java
@@ -1,0 +1,28 @@
+/*
+ * This file is part of RskJ
+ * Copyright (C) 2017 RSK Labs Ltd.
+ * (derived from ethereumJ library, Copyright (c) 2016 <ether.camp>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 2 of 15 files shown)


rskj-core/src/main/java/org/ethereum/datasource/KeyValueDataSource.java
```diff
@@ -49,6 +50,8 @@ public interface KeyValueDataSource extends DataSource {
 
     Set<ByteArrayWrapper> keys();
 
+    DbIterator iterator();
```
Vovchyk, line 53: how about introducing a new interface like this?

```java
interface DataSourceKeyIterator extends Iterator<byte[]>, AutoCloseable { }
```

and returning it here instead of `DbIterator` ?

`DbIterator` doesn't provide any interface, so when being used you have to cast it to specific implementations

rmoreliovlabs: Hmm, not sure how this would work for levelbd and rocksdb implementation. Their iterator implementation is different.

rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
```diff
@@ -183,12 +187,39 @@ private void migrate(DbMigrationInformation dbMigrationInformation) {
 
         logger.info("Migrating data source: {}", sourceKeyValueDataSource.getName());
 
-        sourceKeyValueDataSource.keys().forEach(sourceKey ->
+        Object iterator = sourceKeyValueDataSource.iterator();
```
asoto-iov, line 190: What do you think about hide the specific Iterator implementations? 
We could have a common interface and specific wrapper for each database kind. 
With that we can avoid the cast of the different types and avoid the logic duplication as the content of both if else is very similar. 
Also the changes into specific iterators won't affect to this class.

rskj-core/src/main/java/org/ethereum/datasource/KeyValueDataSource.java
rmoreliovlabs: Done

rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
rmoreliovlabs: Done

rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
```diff
@@ -183,12 +184,23 @@ private void migrate(DbMigrationInformation dbMigrationInformation) {
...
 
-        sourceKeyValueDataSource.keys().forEach(sourceKey ->
-                targetKeyValueDataSource.put(
-                        sourceKey.getData(),
-                        sourceKeyValueDataSource.get(sourceKey.getData())
-                )
-        );
+        DataSourceKeyIterator iterator = sourceKeyValueDataSource.keyIterator();
+
+        while (iterator.hasNext()) {
```
asoto-iov, line 189: It would be nice to have a test to verify if both source and target datasources have the same content.

rskj-core/src/main/java/org/ethereum/datasource/RocksDbKeyIterator.java
```diff
@@ -0,0 +1,56 @@
...
+import org.rocksdb.RocksDB;
+import org.rocksdb.RocksIterator;
+
+import java.util.NoSuchElementException;
+
+public class RocksDbKeyIterator implements DataSourceKeyIterator {
+    private RocksIterator iterator;
+
+    public RocksDbKeyIterator(RocksDB db) {
+        this.iterator = db.newIterator();
```
asoto-iov, line 32: It would be needed to position the iterator at the beginning before using it calling seekToFirst()

From the api
> public [RocksIterator](https://javadoc.io/static/org.rocksdb/rocksdbjni/6.27.3/org/rocksdb/RocksIterator.html) newIterator()
Return a heap-allocated iterator over the contents of the database. The result of newIterator() is initially invalid (caller must call one of the Seek methods on the iterator before using it).

Caller should close the iterator when it is no longer needed. The returned iterator should be closed before this db is closed.

rskj-core/src/main/java/org/ethereum/datasource/RocksDbKeyIterator.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public void close() throws Exception {
+        this.iterator.close();
+    }
+
+    @Override
+    public boolean hasNext() {
+        return this.iterator.isValid();
```
asoto-iov, line 42: the isValid method checks the status of the current iterator and it does nothing with the next entry.

rskj-core/src/main/java/org/ethereum/datasource/RocksDbKeyIterator.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public ByteArrayWrapper next() throws NoSuchElementException {
```
asoto-iov, line 46: After check the API the behaviour of the next() iterator is not the same as the "standard" java iterators. 
Wit this implementation we can retrieve the data at an invalid position. It is always needed to validate after do the next.

rskj-core/src/main/java/org/ethereum/datasource/DataSourceKeyIterator.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package org.ethereum.datasource;
+
+import org.ethereum.db.ByteArrayWrapper;
+
+import java.util.*;
+
+public interface DataSourceKeyIterator extends Iterator<ByteArrayWrapper>, AutoCloseable {
+    void seekToFirst();
```
asoto-iov, line 27: is this method only needed to initialise the RockDbIterator or is it used for other reasons?

rskj-core/src/main/java/org/ethereum/datasource/RocksDbKeyIterator.java
```diff
@@ -0,0 +1,56 @@
...
+package org.ethereum.datasource;
+
+import org.ethereum.db.ByteArrayWrapper;
+import org.ethereum.util.ByteUtil;
+import org.rocksdb.RocksDB;
+import org.rocksdb.RocksIterator;
+
+import java.util.NoSuchElementException;
+
+public class RocksDbKeyIterator implements DataSourceKeyIterator {
```
asoto-iov, line 28: It is needed to review the logic of this implementation. The RocksIterator behaviour is totally different from the standard Java Iterator. 

The standard Java Iterator return the first element after do the next() but the RocksIterator needs to be initialised before using it, so calling next by the first time will position the iterator in the second element.

rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
rmoreliovlabs: Unit tests checks that

rskj-core/src/main/java/org/ethereum/datasource/RocksDbKeyIterator.java
rmoreliovlabs: Yeah looks like I forgot to put it there. Done

rskj-core/src/main/java/org/ethereum/datasource/DataSourceKeyIterator.java
rmoreliovlabs: LevelDB and RocksDB iterators

rskj-core/src/main/java/org/ethereum/datasource/RocksDbKeyIterator.java
rmoreliovlabs: Sure

rskj-core/src/main/java/org/ethereum/datasource/RocksDbKeyIterator.java
rmoreliovlabs: Yeah looks that is the usage we have in other places.

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
```diff
@@ -200,6 +200,11 @@ public Set<ByteArrayWrapper> keys() {
                 .collect(Collectors.toCollection(HashSet::new));
     }
 
+    @Override
+    public DataSourceKeyIterator keyIterator() {
+        return new DefaultKeyIterator(keys());
```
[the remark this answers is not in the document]
Vovchyk: or we could simply throw RuntimeException("Not implemented yet"), as we don't need this method at the moment

rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
```diff
@@ -183,12 +184,24 @@ private void migrate(DbMigrationInformation dbMigrationInformation) {
...
+            byte[] data = iterator.next().getData();
+
+            targetKeyValueDataSource.put(
+                    data,
+                    sourceKeyValueDataSource.get(data)
+            );
+        }
+
+        try {
+            iterator.close();
```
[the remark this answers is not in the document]
rmoreliovlabs: done

rskj-core/src/main/java/org/ethereum/datasource/DataSourceWithCache.java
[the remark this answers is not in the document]
rmoreliovlabs: Done

rskj-core/src/main/java/org/ethereum/datasource/RocksDbKeyIterator.java
```diff
@@ -0,0 +1,65 @@
...
+        return this.iterator.isValid();
+    }
+
+    @Override
+    public byte[] next() throws NoSuchElementException {
+        if (!this.hasNext()) {
+            throw new NoSuchElementException();
+        }
+
+        this.iterator.next();
```
asoto-iov, line 57: This implementation won't work. Here the first element is always being miss and the last element is never being validated.
An example with the current implementation:

Having an array `[a,b]` during the first iteration when we call  the next() method: 
```
//In the first iteration after call seekToFirst() the iterator is pointing to 'a'
public byte[] next() throws NoSuchElementException {
        if (!this.hasNext()) {  //RocksIterator.isValid() is called returning true
            throw new NoSuchElementException();
        }

        this.iterator.next(); // after this call the iterator is pointing to the next element -> 'b'
        return this.iterator.key(); // 'b' will be returned without being validated. 
    }
```

RocksIterator does not work as the standard Java Iterator. With the java Iterator we call `next()` to get the value of the first element (and move the pointer to the next element) but the next() implementation of the RocksIterator just moves the pointer to the next element and returns void, so we need to get the value before do the next() call. 

The solution would be something like this:
```
public byte[] next() throws NoSuchElementException {
        if (!this.hasNext()) { //validate the current position
            throw new NoSuchElementException(); 
        }
        byte[] currentKey = this.iterator.key();//save the current value 
        this.iterator.next();//advance to the next position
        return currentKey; //return current value
    }
```
In the second call, the RocksIterator is already pointing to the second position which is being validated and returned. In case the array has only one element, after call next() one time the pointer is already pointing to the next element and when we call hasNext() the isValid method will return false, so everything should be ok. 

Sorry for the very long comment.

rmoreliovlabs: Thanks!

Fixed.

rskj-core/src/test/java/org/ethereum/datasource/DataSourceWithCacheTest.java
```diff
@@ -40,6 +40,16 @@ public void getAfterMiss() {
...
         dataSourceWithCache.get(randomKey);
 
+        try (DataSourceKeyIterator iterator = dataSourceWithCache.keyIterator()){
+
+            iterator.seekToFirst();
+
+            assertTrue(iterator.hasNext());
+            assertArrayEquals(iterator.next(), randomKey);
+        } catch (Exception e) {
+            e.printStackTrace();
```
Vovchyk, line 50: I guess we could do `Assert.fail(...)` here in case we get an exception?

rskj-core/src/test/java/org/ethereum/datasource/KeyValueDataSourceTest.java
```diff
@@ -53,14 +53,34 @@ public void setup() {
...
+            while (iterator.hasNext()) {
+                expectedValue = iterator.next();
+                if (ByteUtil.wrap(expectedValue).equals(ByteUtil.wrap(randomKey))) {
+                    break;
+                }
+            }
+
+            assertArrayEquals(expectedValue, randomKey);
+        } catch (Exception e) {
+            e.printStackTrace();
```
Vovchyk, line 82: I wouldn't silently ignore this exception..

rskj-core/src/test/java/org/ethereum/datasource/LevelDbDataSourceTest.java
```diff
@@ -54,6 +48,16 @@ public void testBatchUpdating() throws IOException {
...
         assertEquals(batchSize, dataSource.keys().size());
 
+        try (DataSourceKeyIterator iterator = dataSource.keyIterator()){
+
+            iterator.seekToFirst();
+
+            assertTrue(iterator.hasNext());
+            assertNotNull(iterator.next());
+        } catch (Exception e) {
+            e.printStackTrace();
```
Vovchyk, line 58: same here

rskj-core/src/test/java/org/ethereum/datasource/RocksDbDataSourceTest.java
```diff
@@ -53,7 +47,17 @@ public void testBatchUpdating() throws IOException {
...
-        
+
+        try (DataSourceKeyIterator iterator = dataSource.keyIterator()){
+
+            iterator.seekToFirst();
+
+            assertTrue(iterator.hasNext());
+            assertNotNull(iterator.next());
+        } catch (Exception e) {
+            e.printStackTrace();
```
Vovchyk, line 58: and here

rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
```diff
@@ -183,12 +184,21 @@ private void migrate(DbMigrationInformation dbMigrationInformation) {
 
         logger.info("Migrating data source: {}", sourceKeyValueDataSource.getName());
 
-        sourceKeyValueDataSource.keys().forEach(sourceKey ->
+        try (DataSourceKeyIterator iterator = sourceKeyValueDataSource.keyIterator()) {
+            iterator.seekToFirst();
```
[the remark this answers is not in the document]
asoto-iov: Fo the DBIterator instance I think it is not needed and for the RocksDbIterator I think it could be done internally as it has no any extra utility outside in our use case.

rskj-core/src/main/java/org/ethereum/datasource/LevelDbKeyIterator.java
```diff
@@ -0,0 +1,74 @@
...
+package org.ethereum.datasource;
+
+import org.iq80.leveldb.DB;
+import org.iq80.leveldb.DBIterator;
+
+import java.util.NoSuchElementException;
+
+public class LevelDbKeyIterator implements DataSourceKeyIterator {
+    private final DBIterator iterator;
+    private boolean blockIteration;
```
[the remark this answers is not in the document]
asoto-iov: In this method we could simply call `return this.iterator.next()`

`org.iq80.leveldb.DBIterator` it is already implementing the [java.util.Iterator](https://docs.oracle.com/en/java/javase/11/docs/api/index.html) interface. 

The peekNext is an additional method that has the same behaviour as the RocksIteraotr.next() method but it is not really needed here. Actually if you check the implementation of the DBIterator.next() method we use, it is doing similar to the proposed solution by Vlad.

rskj-core/src/main/java/org/ethereum/datasource/RocksDbKeyIterator.java
```diff
@@ -0,0 +1,74 @@
...
+package org.ethereum.datasource;
+
+import org.rocksdb.RocksDB;
+import org.rocksdb.RocksIterator;
+
+import java.util.NoSuchElementException;
+
+public class RocksDbKeyIterator implements DataSourceKeyIterator {
+    private final RocksIterator iterator;
+    private boolean blockIteration;
```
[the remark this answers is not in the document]
asoto-iov: I think this attribute class it is not really needed. I proposed something similar to what you wrote in the LevelsDB iterator implementation of the next method avoiding to have this extra class attribute.

asoto-iov:
I think we should review the unit tests as they are not covering the functionality. The RocksDB Iterator implementation was not working fine and the test were not failing. The iterator functionality were not being tested.

Also something similar is may be happening with the DB migration tests. If we were missing one element, something should fail, or am I wrong here?

rskj-core/src/main/java/org/ethereum/datasource/DataSourceKeyIterator.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/cli/tools/DbMigrate.java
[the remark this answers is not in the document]
rmoreliovlabs: Addressed

rskj-core/src/main/java/org/ethereum/datasource/LevelDbKeyIterator.java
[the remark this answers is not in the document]
rmoreliovlabs: Done

rskj-core/src/test/java/org/ethereum/datasource/DataSourceWithCacheTest.java
rmoreliovlabs: Addressed

rskj-core/src/test/java/org/ethereum/datasource/LevelDbDataSourceTest.java
rmoreliovlabs: Addressed

Vovchyk (dismissed):

Vovchyk (approved):

asoto-iov (dismissed):

Vovchyk (approved):

Vovchyk:
pipeline:run

rskj-core/src/main/java/org/ethereum/datasource/DefaultKeyIterator.java
```diff
@@ -0,0 +1,50 @@
...
+public class DefaultKeyIterator implements DataSourceKeyIterator {
+    private final Iterator<ByteArrayWrapper> iterator;
+
+    public DefaultKeyIterator(Collection<ByteArrayWrapper> collection) {
+        this.iterator = collection.iterator();
+    }
+
+    @Override
+    public void close() throws Exception {
+    }
```
VaWheel, line 36: why don't we close iterator here?

Vovchyk (approved):

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

PR #1882, opened 2022-09-16
Adds new pegout_confirmed definition

jeremy-then:
Adds new pegout_confirmed definition

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/utils/BrigeEventLoggerLegacyImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/utils/BrigeEventLoggerLegacyImpl.java
@@ -1,9 +1,6 @@
 package co.rsk.peg.utils;
 
-import co.rsk.bitcoinj.core.Address;
-import co.rsk.bitcoinj.core.BtcECKey;
-import co.rsk.bitcoinj.core.BtcTransaction;
-import co.rsk.bitcoinj.core.Coin;
+import co.rsk.bitcoinj.core.*;
 import co.rsk.config.BridgeConstants;
 import co.rsk.core.RskAddress;
 import co.rsk.crypto.Keccak256;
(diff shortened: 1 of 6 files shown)


rskj-core/src/main/java/co/rsk/peg/utils/BrigeEventLoggerLegacyImpl.java
```diff
@@ -150,6 +147,11 @@ public void logBatchPegoutCreated(BtcTransaction btcTx, List<Keccak256> rskTxHas
         throw new DeprecatedMethodCallException("Calling BrigeEventLoggerLegacyImpl.logBatchPegoutCreated method after RSKIP146 activation");
     }
 
+    @Override
+    public void logPegoutConfirmed(Sha256Hash btcTxHash, long pegoutCreationRskBlockNumber) {
```
marcos-iov, line 151: We should add a test for a call to this method pre RSKIP 146

marcos-iov (commented):
Left a minor comment. Solid work, well done!

rskj-core/src/main/java/co/rsk/peg/utils/BrigeEventLoggerLegacyImpl.java
jeremy-then: This pr is only about declaring the new event and adding this simple test. More tests for this event will be added in a different story which is about actually emitting this event.

kelvinator07:
Hi @jeremy-then You Should Also Add a Test To BridgeEventLoggerLegacyImplTest

rskj-core/src/main/java/co/rsk/peg/utils/BrigeEventLoggerLegacyImpl.java
marcos-iov: That's fine by me as long as you add the details to the corresponding story or create a new one, to make sure we don't forget about this.

It's a simple test anyway, just checking that you get an exception when calling this method pre RSKIP146

marcos-iov (approved):

kelvinator07 (approved):

PR #1812, opened 2022-06-16
Update to java 17 mockito inline

illuque-iov:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Requires Activation Code (Hard Fork)

* **Other information**:

*fed:update_to_java_17_mockito_inline*

PR #1883, opened 2022-09-18
Upgrade jackson and jsonrpc libs

rmoreliovlabs:
## Description
Upgrade jackson and jsonrpc libs.

Jackson ObjectMapper is handling null values in this new version and some tests were expecting NullPointerExceptions and JsonMappingExceptions. I just made those methods expecting Jackson Object Mapper exceptions to have a similar behavior to keep it working as previously expected.

## Motivation and Context
We are using an old version of com.fasterxml.jackson.core:jackson-databind, currently set to 2.8.7,  being updated to 2.8.11 on scope of [Java upgrade effort](https://github.com/rsksmart/rskj/pull/1812). There is already a 2.13 version and 2.14 is coming soon. 

The reason for not updating now is that we’ve detected some changes on method outputs. As an example, ObjectMapper.readTree is now returning null in cases when it was throwing an NullPointerException (tested on co.rsk.config.RemascConfigFactoryTest#createRemascConfigInvalidConfig). Even if code compiles and tests pass (after fixing the previously mentioned difference), we could be creating some bugs that would remain unnoticed until runtime, as we do a wide usage of this library.

It looks like the change was introduced on 2.10.0, as everything seemed to be working fine up to 2.9.10. According to Jackson release strategy, minor versions should not introduce any breaking change, so probably this was a bugfix or alike. However, we think this upgrade would need further and deeper review just in case.

Update

After investigating a bit, it looks like we might also have an incompatibility with com.github.briandilley.jsonrpc4j, that depends on Jackson. This library could only be updated to 1.5.3, as with latest 1.6 it was depending on Jackson's 2.10.2 and causing the same issue mentioned above even when keeping Jackson’s declared dependency at original 2.8.7

## Types of changes

## Checklist:

* **Other information**:
PowPeg PR: https://github.com/rsksmart/powpeg-node/pull/112

*fed:upgrade-jackson-and-json-rpc-libs*


illuque-iov (changes requested):
Have you run the tests after the change?

This one, among other, is failing for me: `co.rsk.config.RemascConfigFactoryTest#createRemascConfigInvalidConfig`

It most probably will fail also in CI/CD when the Gradle verification issue is fixed.

rmoreliovlabs:
> Have you run the tests after the change?
> 
> This one, among other, is failing for me: `co.rsk.config.RemascConfigFactoryTest#createRemascConfigInvalidConfig`
> 
> It most probably will fail also in CI/CD when the Gradle verification issue is fixed.

The previous behavior of the library produced a nullpointer exception when you sent null to `public <T> T treeToValue(TreeNode n, Class<T> valueType)` in the `n` argument. New version's implementation checks for null and if `n` is null then returns `null` hence avoid NullPointerException to happen.

I tried to fix this issue by following the same behavior after `treeToValue` is executed.

rmoreliovlabs:
pipeline:run

rskj-core/src/test/java/co/rsk/rpc/netty/Web3WebSocketServerTest.java
```diff
@@ -203,8 +203,13 @@ private byte[] getJsonRpcDummyMessage(String value) {
 
         byte[] request = new byte[0];
         try {
-            request = OBJECT_MAPPER.writeValueAsBytes(OBJECT_MAPPER.treeToValue(
-                    JSON_NODE_FACTORY.objectNode().setAll(jsonRpcRequestProperties), Object.class));
+            Object object = OBJECT_MAPPER.treeToValue(JSON_NODE_FACTORY.objectNode().setAll(jsonRpcRequestProperties), Object.class);
+
+            if (object == null) {
+                throw new NullPointerException();
```
illuque-iov, line 209: I wouldn't throw exceptions in a Test, I would do nothing here actually, its under our control and this should not be reached but for our mistake.

rskj-core/src/test/java/co/rsk/rpc/netty/Web3HttpServerTest.java
```diff
@@ -118,6 +119,10 @@ private void smokeTest(String contentType, String host, InetAddress rpcAddress,
             String responseBody = response.body().string();
             JsonNode jsonRpcResponse = OBJECT_MAPPER.readTree(responseBody);
 
+            if (jsonRpcResponse.isEmpty()) {
+                throw JsonMappingException.from(jsonRpcResponse.traverse(), "Empty response");
```
illuque-iov, line 123: I wouldn't throw exceptions in a Test, I would do nothing here actually, its under our control and this should not be reached but for our mistake.

rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcWeb3ServerHandler.java
```diff
@@ -86,6 +86,12 @@ private ByteBuf buildErrorContent(int errorCode, String errorMessage) throws Jso
         errorProperties.put("code", jsonNodeFactory.numberNode(errorCode));
         errorProperties.put("message", jsonNodeFactory.textNode(errorMessage));
         JsonNode error = jsonNodeFactory.objectNode().set("error", jsonNodeFactory.objectNode().setAll(errorProperties));
-        return Unpooled.wrappedBuffer(mapper.writeValueAsBytes(mapper.treeToValue(error, Object.class)));
+        Object object = mapper.treeToValue(error, Object.class);
+
+        if (object == null) {
+            throw new NullPointerException();
```
illuque-iov, line 92: I know this is unlikely to be reached, but I would rather return sth like the following instead of throwing, otherwise the other party will not receive anything as response from the node (`ctx.fireChannelRead...` will be skipped).

`return Unpooled.wrappedBuffer(mapper.writeValueAsBytes("unknown error"));`

illuque-iov (commented):
As a general note, IMO the wrapper we discussed is not needed, we only call `mapper.treeToValue` from 3 places in the production code, I think we should deal with the new behavior from Jackson.

rskj-core/src/test/java/co/rsk/rpc/netty/Web3WebSocketServerTest.java
rmoreliovlabs: Not necessary anymore with the latest changes

rskj-core/src/test/java/co/rsk/rpc/netty/Web3HttpServerTest.java
rmoreliovlabs: Not necessary anymore with the latest changes

rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcWeb3ServerHandler.java
rmoreliovlabs: Not necessary anymore with the latest changes

illuque-iov:
IMO wrapping just to keep previous library behavior is not ideal, if Jackson changed it is probably for a reason. I would instead let each calling place decide what to do based on business rules. But no big deal, if you prefer this approach I'm ok.

But I still think we could now fix the "problem" with the NPE I've mentioned [here](https://github.com/rsksmart/rskj/pull/1883#discussion_r977929637.

illuque-iov (commented):
IMO wrapping just to keep previous library behavior is not ideal, if Jackson changed it is probably for a reason. I would instead let each calling place decide what to do based on business rules. But no big deal, if you prefer this approach I'm ok.

But I still think we could now fix the "problem" with the NPE I've mentioned [here](https://github.com/rsksmart/rskj/pull/1883#discussion_r977929637).

Vovchyk:
> IMO wrapping just to keep previous library behavior is not ideal, if Jackson changed it is probably for a reason. I would instead let each calling place decide what to do based on business rules. But no big deal, if you prefer this approach I'm ok.
> 
> But I still think we could now fix the "problem" with the NPE I've mentioned [here](https://github.com/rsksmart/rskj/pull/1883#discussion_r977929637).

imho, I'd try to preserve compatibility with the previous versions of the api, unless there's another reason not to do it.

rewriting each place where null's may occur is error prone, so perhaps keeping this check for null in one place is simpler

illuque-iov:
> > IMO wrapping just to keep previous library behavior is not ideal, if Jackson changed it is probably for a reason. I would instead let each calling place decide what to do based on business rules. But no big deal, if you prefer this approach I'm ok.
> > But I still think we could now fix the "problem" with the NPE I've mentioned [here](https://github.com/rsksmart/rskj/pull/1883#discussion_r977929637).
> 
> imho, I'd try to preserve compatibility with the previous versions of the api, unless there's another reason not to do it.
> 
> rewriting each place where null's may occur is error prone, so perhaps keeping this check for null in one place is simpler

Yeah, I generally agree, I explained myself wrong probably. I meant wrapping was not ideal in this scenario, where wrapped method is only being called in 3 places and where the check for null that we are doing is just throwing manually a NPE because the library was doing that before. For example, doing this is preventing us from properly handling the exception in this two places:
1.  [Not releasing content](https://github.com/rsksmart/rskj/blob/4c18c64b83049deee9b534886be5601f46456562/rskj-core/src/main/java/co/rsk/rpc/netty/RskWebSocketJsonRpcHandler.java#L103) as mentioned [here](https://github.com/rsksmart/rskj/pull/1883#discussion_r977929637)
2. [Not returning anything to the caller](https://github.com/rsksmart/rskj/blob/4c18c64b83049deee9b534886be5601f46456562/rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcWeb3ServerHandler.java#L73) as mentioned [here](https://github.com/rsksmart/rskj/pull/1883#discussion_r977957523)

rskj-core/src/main/java/co/rsk/util/JacksonParserUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.TreeNode;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+public final class JacksonParserUtil {
```
Vovchyk, line 12: it looks like that only `mapper.treeToValue()` allows `null`'s (returns null if null is provided). The others ( `readTree()` ones ) do throw an exception if provided argument is `null`.

So may be ( as @illuque-iov mentioned in the other comment ) we don't need this new class and can handle null's on the case-by-case basis ( sees we have 2 or 3 places )

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/util/JacksonParserUtil.java
rmoreliovlabs: I don't think so @Vovchyk, they have this new NullNode and MissingNode clases that inherits from JsonNode for `readTree` so exception are not going to be thrown for null and empty string content.

rskj-core/src/test/java/co/rsk/util/JacksonParserUtilTest.java
```diff
@@ -0,0 +1,65 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.util.Map;
+
+public class JacksonParserUtilTest {
```
Vovchyk, line 30: seems that `null` / `empty` cases are not covered in these tests

Vovchyk (changes requested):

rskj-core/src/test/java/co/rsk/util/JacksonParserUtilTest.java
rmoreliovlabs: done

rskj-core/src/main/java/co/rsk/util/JacksonParserUtil.java
```diff
@@ -0,0 +1,55 @@
...
+    private JacksonParserUtil() {
+    }
+
+    public static <T> T treeToValue(ObjectMapper mapper, TreeNode treeNode, Class<T> aClass) throws IllegalArgumentException, JsonProcessingException {
+        T result = mapper.treeToValue(treeNode, aClass);
+
+        if (result == null) {
```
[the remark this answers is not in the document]
rmoreliovlabs: done

rskj-core/src/main/java/co/rsk/util/JacksonParserUtil.java
```diff
@@ -0,0 +1,55 @@
...
+
+        return result;
+    }
+
+    public static JsonNode readTree(ObjectMapper mapper, String content) throws IOException {
+        if (content == null || content.isEmpty()) {
+            throw JsonMappingException.from(mapper.getDeserializationContext(), "Input is empty");
+        }
+
+        return mapper.readTree(content);
```
VaWheel, line 45: why don't we check if JsonNode is empty here?

rskj-core/src/main/java/co/rsk/util/JacksonParserUtil.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public static JsonNode readTree(ObjectMapper mapper, byte[] content) throws IOException {
+        if (content == null || content.length == 0) {
+            throw JsonMappingException.from(mapper.getDeserializationContext(), "Input is empty");
+        }
+
+        return mapper.readTree(content);
```
VaWheel, line 53: also here, why don't we check if JsonNode is empty?

rskj-core/src/main/java/co/rsk/util/JacksonParserUtil.java
rmoreliovlabs: because if the content is not empty, then the library should return something

rskj-core/src/main/java/co/rsk/util/JacksonParserUtil.java
rmoreliovlabs: same as a above

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

rmoreliovlabs:
pipeline:run

Vovchyk:
pipeline:run

rmoreliovlabs:
pipeline:run

Vovchyk:
pipeline:run

rmoreliovlabs:
pipeline:run

rmoreliovlabs:
pipeline:run

illuque-iov (approved):

PR #1885, opened 2022-09-21
Refactored block template execution

Vovchyk:
## Description

This removes unnecessary save to store results of block template execution

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
@@ -78,20 +78,20 @@ public class BlockExecutor {
      * @param parent       The parent of the block.
      */
     public BlockResult executeAndFill(Block block, BlockHeader parent) {
-        BlockResult result = execute(block, parent, true, false);
+        BlockResult result = execute(block, parent, true, false, false);
         fill(block, result);
         return result;
     }
 
     @VisibleForTesting
     public void executeAndFillAll(Block block, BlockHeader parent) {
-        BlockResult result = execute(block, parent, false, true);
+        BlockResult result = execute(block, parent, false, true, false);
         fill(block, result);
     }
 
     @VisibleForTesting
     public void executeAndFillReal(Block block, BlockHeader parent) {
-        BlockResult result = execute(block, parent, false, false);
+        BlockResult result = execute(block, parent, false, false, false);
         if (result != BlockResult.INTERRUPTED_EXECUTION_BLOCK_RESULT) {
             fill(block, result);
         }
@@ -122,7 +122,7 @@ public class BlockExecutor {
      */
     @VisibleForTesting
     public boolean executeAndValidate(Block block, BlockHeader parent) {
-        BlockResult result = execute(block, parent, false, false);
+        BlockResult result = execute(block, parent, false, false, false);
 
         return this.validate(block, result);
     }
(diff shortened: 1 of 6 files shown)


rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -219,11 +219,11 @@ private boolean validateLogsBloom(BlockHeader header, BlockResult result) {
 
     @VisibleForTesting
     public BlockResult execute(Block block, BlockHeader parent, boolean discardInvalidTxs) {
-        return execute(block, parent, discardInvalidTxs, false);
+        return execute(block, parent, discardInvalidTxs, false, true);
     }
 
-    public BlockResult execute(Block block, BlockHeader parent, boolean discardInvalidTxs, boolean ignoreReadyToExecute) {
-        return executeInternal(null, 0, block, parent, discardInvalidTxs, ignoreReadyToExecute);
+    public BlockResult execute(Block block, BlockHeader parent, boolean discardInvalidTxs, boolean ignoreReadyToExecute, boolean saveState) {
```
illuque-iov, line 225: WDYT about splitting this method into two, one called `executeSavingState` and another one `executeSkippingStateSave` (or sth like that) instead of passing a boolean? I think it is more clear.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -219,11 +219,11 @@ private boolean validateLogsBloom(BlockHeader header, BlockResult result) {
 
     @VisibleForTesting
     public BlockResult execute(Block block, BlockHeader parent, boolean discardInvalidTxs) {
```
illuque-iov, line 221: Minor thing, but this `@VisibleForTesting` could be package-private.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
asoto-iov, line 225: This method is receiving a bunch of input parameters. For the future if this continue growing we could create something like a BlockExecutorParameters and collect them there to avoid the method signature continue growing, and avoiding possible errors in the order of the passed parameters. 

Adding the Builder pattern to the ParametersClass we could have something like:
```
BlockExecutorParameters.builder()
                .discardInvalidTx(true)
                .ignoreReadyToExecute(true)
                .ignoreSaveState(true)
                .build;
```
For me is easier to read when coding. Also we could create specific methods to create specific "configurations" for example a default() or other different ones and use then if the same behaviour is expected when calling executor from different places avoiding duplication.

illuque-iov: I like @asoto-iov approach mentioned [here](https://github.com/rsksmart/rskj/pull/1885#discussion_r976781685), I think it is the best solution.

The reason I suggested this intermediate approach is to avoid a refactor, as it would be changing the same places you have already changed. Up to you.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
Vovchyk: done

Vovchyk:
pipeline:run

illuque-iov (approved):

illuque-iov (approved):

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

PR #1890, opened 2022-09-27
Set methods in BridgeEventLogger as default

jeremy-then:


marcos-iov (dismissed):
Well done!

kelvinator07 (approved):

rmoreliovlabs (changes requested):
Could you write some description and context about what and why we are adding these changes?

marcos-iov (approved):

PR #1894, opened 2022-09-29
Affecting consensus changes on RSK adding the following bridge method: getPegoutCreationRskTxHashByBtcTxHash

nathanieliov:
A short summary of the changes:

- Add methods and logic into BridgeStorageProvider class to manage the new peg-out creation index.
Methods created: getNextPegoutHeight, setNextPegoutHeight, saveNextPegoutHeight

Index structure
```
index key: (DataWord) Combination of the static string pegoutTxIndex- and the given Bitcoin transaction hash (Sha256hash -> 32 bytes). The resulting string should be hashed using Keccak256 to have a valid 32 bytes long DataWord.
index value: RSK transaction hash (Keccak256 -> 32 bytes)
```

- Add new bridge method `getPegoutCreationRskTxHashByBtcTxHash`, to fetch a rskTxHash(the rsk transaction where the peg-out was created)  given a btcTxHash(which belongs to a peg-out).

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
@@ -802,17 +802,17 @@ public class BridgeSerializationUtils {
 
     public static byte[] serializeSha256Hash(Sha256Hash hash) {
         if (hash == null){
-            return null;
+            return new byte[]{};
         }
         return RLP.encodeElement(hash.getBytes());
     }
 
-    public static Sha256Hash deserializeSha256Hash(byte[] data) {
+    public static Optional<Sha256Hash> deserializeSha256Hash(byte[] data) {
         RLPElement element = RLP.decodeFirstElement(data, 0);
         if (element == null) {
-            return null;
+            return Optional.empty();
         }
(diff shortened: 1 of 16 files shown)


nathanieliov:
pipeline:run

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
```diff
@@ -801,15 +801,37 @@ public static byte[] serializeCoinbaseInformation(CoinbaseInformation coinbaseIn
     }
 
     public static byte[] serializeSha256Hash(Sha256Hash hash) {
+        if (hash == null){
```
donequis, line 804: Can't this be wrapped on an alternative method? Checking if this kind of changes affects consensus and there's no need for them

nathanieliov: This is being fixed in this PR: https://github.com/rsksmart/rskj/pull/1928/files#diff-3bd14645f5f2bfbb02415a9c680c5c58e003143f285a357f760071d4ea0cd117

nathanieliov:
pipeline:run

PR #1895, opened 2022-09-29
Research/main parallel tx remasc

julianlen:
As it is now, a remascFees variable accumulates the fees for a block and sends them to REMASC after processing all the transactions, REMASC tx included.

So far, this is not an issue since the payment to miners is performed after some blocks called Maturity. For the Regtest network, the Maturity is 10. Nevertheless, if we consider the case Maturity becomes 0, miners won't perceive the fees paid in the same block; but in the next block. 

On this PR I moved the payment just before the REMASC tx execution. There were some edge cases in which there was no REMASC tx in the block. Hence, no fees were added. So, I added a check where if the Block Executor processes the last transaction, even though it is not the REMASC one, it increments the balance of the REMASC account with the fees paid in the block.

Finally, for every place where REMASC receives the fees, I added a check to see if REMASC is enabled.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/TransactionListExecutor.java
+++ b/rskj-core/src/main/java/co/rsk/core/TransactionListExecutor.java
@@ -3,6 +3,7 @@ package co.rsk.core;
 import co.rsk.crypto.Keccak256;
 import org.ethereum.core.*;
 import org.ethereum.vm.DataWord;
+import org.ethereum.vm.PrecompiledContracts;
 import org.ethereum.vm.program.ProgramResult;
 import org.ethereum.vm.trace.ProgramTraceProcessor;
 import org.slf4j.Logger;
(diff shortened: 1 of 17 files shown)


rskj-core/src/main/java/co/rsk/core/TransactionListExecutor.java
```diff
@@ -65,17 +68,25 @@ public TransactionListExecutor(
...
-        Coin totalPaidFees = Coin.ZERO;
 
         for (Transaction tx : transactions) {
+
+            int numberOfTransactions = block.getTransactionsList().size();
+            boolean isRemascTransaction = tx.isRemascTransaction(this.i, numberOfTransactions);
+
+            if (this.remascEnabled && isRemascTransaction) {
+                addFeesToRemasc();
+            }
```
javicid, line 88: A suggestion for you to consider: what if we iterate over transactions.size - 1 and then do the if REMASC check outside of the loop? This would avoid checking on each tx.

Issue #948, opened 2019-08-07
Implement Gas Exactimation


juli:
RSKj does not implement the binary search solution implemented by [geth](https://github.com/ethereum/go-ethereum/blob/master/accounts/abi/bind/backends/simulated.go#L267) and [parity](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/client/client.rs#L1658) in `eth_estimateGas`, some transactions can fail unexpectedly with OOG.

Truffle Suite recently published a less CPU intensive solution they called [Gas Exactimation](https://github.com/trufflesuite/ganache-cli/releases/tag/v6.4.2)

Latest implementation: https://github.com/trufflesuite/ganache-core/blob/3ec34fe972dc17a5600eb566c8c309cd5ddb38bd/lib/utils/gasEstimation.js

This issue will be more relevant if RSKj implements EIP-150 63/64 rule (#944)

joaquinlpereyra-iov:
The problem with our current implementation is that it does take into account gas-returning OPCODES like `SUICIDE`, returning the _gas used_, not the gas you must _send_. 

So if you use the gas returned by `eth_estimageGas`, you may get OOG.

SergioDemianLerner:
Also does not take into account gas refunds because of SSTORE changes.
And also gas amounts locked by CALL and the 2300 pre-send amount required by CALLs with value.

My suggestion to solve this issue is to code Gas Exactimation [but it must be analyzed for correctness, because it's written mainly for Ethereum and the 63/64 rule]: 

OR, as a simpler alternative that will work 99.99% of the time (I'm trying hard to find examples where this wouldn't work):
- Change TransactionExecutor to return not only the gas consumed but also the gas paid in advance before refunds.
- Instrument the VM to return the highest difference D=(passed gas-returned gas) for all CALLs IN THE FIRST CALL LEVEL (not nested)
For example, 
if CALL passes 50000 gas to child and child consumes 20000, then the D is 50000-20000=30000.
If another CALL passes 6000 gas to child and child consumes 500, then D is still 30000 because 5500 is lower than the previous value D. At the end the value D must be added to the gas consumed to set the transaction gas limit.
- Add 2300 to the transaction limit IF there is any CALL at non-top level that transfers value and passes exactly 2300 gas.

pmprete:
This issue is still relevant, and is causing https://github.com/rsksmart/rskj/issues/1279

PR #1898, opened 2022-10-04
Initial implementation of RSKIP-209

illuque-iov:
## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1532,14 +1533,42 @@ private MessageCall getMessageCall(DataWord gas, DataWord codeAddress, Activatio
...
+        // TODO:I next two lines can be skipped for RSKIP validation. Temp vars so RSKIP pseudocode var names match with the used ones. Later `remainingGas` and `gasCost` should be reused
+        long availableGas = remainingGas;
+        long standardCallCosts = gasCost;
+
         if (computeGas) {
-            gasCost = GasCost.add(gasCost, calleeGas);
-            spendOpCodeGas();
+            if (activations.isActive(RSKIP209)) {
+                program.spendGas(standardCallCosts, op.name());
+                long callGas = Math.min(specifiedGasPlusMin, availableGas); // TO_CHECK_1 @sergio to take into account transfer stipend, we cannot use `top-of-stack item` but `specifiedGasPlusMin`
```
SergioDemianLerner, line 1543: Yes, but I wonder why we were previously using specifiedGasPlusMin and not min(specifiedGas,minimumTransferGas) (instead of adding them).
I really can't understand why it was computed that way.

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1532,14 +1533,42 @@ private MessageCall getMessageCall(DataWord gas, DataWord codeAddress, Activatio
...
+                long lockGas = availableGas / 64;
+                long maxGasToPassCallee = availableGas - lockGas; // TO_CHECK_2 @sergio we are not taking into account transfer stipend here
+                if (callGas > maxGasToPassCallee) {
```
SergioDemianLerner, line 1546: Do you mean the STIPEND_CALL ?

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1532,14 +1533,42 @@ private MessageCall getMessageCall(DataWord gas, DataWord codeAddress, Activatio
...
+                    Map<Integer, Long> lockedGasByDepth = this.program.getLockedGasByDepth();
+                    long prevGasLock = lockedGasByDepth.getOrDefault(program.getCallDeep(), 0L);
+                    long forbiddenGas = callGas - maxGasToPassCallee;
+                    if (forbiddenGas > prevGasLock) { // locked for depth can only increase
+                        long expectedLock = forbiddenGas - prevGasLock;
+                        program.spendGas(expectedLock, "Locked for CALL depth " + program.getCallDeep());
+                        program.futureRefundGas(expectedLock); // TO_CHECK_4 @sergio this is how we were handling future refunds
+                        lockedGasByDepth.put(program.getCallDeep(), forbiddenGas);
```
SergioDemianLerner, line 1554: I think that is the correct way.

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1532,14 +1533,42 @@ private MessageCall getMessageCall(DataWord gas, DataWord codeAddress, Activatio
...
+                    }
+                }
+                // TODO:I next two lines can be skipped for RSKIP validation. Temp vars so RSKIP pseudocode var names match with the used ones. Later `calleeGas` and `availableGas` should be reused
+                calleeGas = maxGasToPassCallee;
+                remainingGas = availableGas;
+            } else {
+                // TO_CHECK_6 @sergio aren't we missing consuming `maxGasToPassCallee` after RSKIP209? following lines show previous behavior:
+                // we were increasing `gasCost` with `maxGasToPassCallee` and consuming it inside `spendOpCodeGas()`, now we are just expending `gasCost`+`expectedLock`
+                gasCost = GasCost.add(gasCost, calleeGas);
+                spendOpCodeGas();
```
SergioDemianLerner, line 1565: This two lines should be executed for both paths. RSKIP209 and not RSKIP209.

Why only if RSKIP209 is not activated?

rskj-core/src/main/java/org/ethereum/vm/VM.java
SergioDemianLerner: callGas seems to hold the same value as calleeGas here. Why recompute ? It confuses the reader thinking both are different.

rskj-core/src/main/java/org/ethereum/vm/VM.java
illuque-iov: Yes, that was my doubt, as back then I was not sure while checking the RSKIP pseudocode.

But now I have more context on how this piece of code works (thanks to other task), so I will do another iteration on this algorithm and then we can take another look.

rskj-core/src/main/java/org/ethereum/vm/VM.java
illuque-iov: > Yes, but I wonder why we were previously using specifiedGasPlusMin and not min(specifiedGas,minimumTransferGas) (instead of adding them).
> I really can't understand why it was computed that way.

I think we were incrementing the `specifiedGas` with the stipend in order to make the estimation bigger when dealing with a call with value, so as to not get OOG for these cases, as the stipend is needed during execution even if later refunded. Maybe [this comment](https://github.com/rsksmart/rskj/issues/948#issuecomment-520960211) provides some light.

---

> callGas seems to hold the same value as calleeGas here. Why recompute ? It confuses the reader thinking both are different.

Agree, left it just to match your pseudocode and ease the review, but there was a `TODO` to remove it

PR #1903, opened 2022-10-11
Improve db tests

illuque-iov:
Adding missing tests to DB layer

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/test/java/org/ethereum/datasource/DataSourceWithCacheTest.java
```diff
@@ -130,6 +157,51 @@ public void putTwoKeyValuesWrittenInOrder() {
...
+
+            AtomicBoolean threadStarted = new AtomicBoolean(false);
+            ExecutorService executor = Executors.newSingleThreadExecutor();
+            Future<?> future = executor.submit(new Thread(() -> {
+                threadStarted.set(true);
+                dataSourceWithCache.put(randomKey1, randomValueAfterLock);
+            }));
+
+            // wait for thread to be started and put a value while thread is locked
+            Awaitility.await().timeout(Duration.ofMillis(100)).pollDelay(Duration.ofMillis(10)).untilAtomic(threadStarted, equalTo(true));
```
Vovchyk, line 184: isn't `100ms`  too short timeout? tests may fail due to system being overloaded, what do you think?

Vovchyk (changes requested):

rskj-core/src/test/java/org/ethereum/datasource/DataSourceWithCacheTest.java
illuque-iov: yeah, you are right, even if it is waiting just for the thread to start, not to finish, it wouldn't harm if we increase that value (being this a test and having such short polls) and this may prevent some edge scenarios where it is not possible create more threads or whatever 👍

Vovchyk (approved):

Vovchyk (dismissed):

Vovchyk (approved):

PR #1900, opened 2022-10-06
Pegout events improvements

marcos-iov:
Pegout events improvements

## Description

This pr includes the improvements for some of the pegout events described in this RSKIP: https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP326.md

1) A new pegout event called `pegout_confirmed ` is emitted when a pegout is moved from waiting for confirmation to waiting for signature structure.
2) The signature of the `release_request_received` event is changed for the `btcDestinationAddress` parameter to be of type string instead of bytes.
3) The event `add_signature` is now only being emitted when the federator actually signed the transaction (before it was being emitted first, before even checking if a signature was added or not).

## Motivation and Context

Check RSKIP326: https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP326.md

## How Has This Been Tested?

We added unit tests.
We added regression and integration tests in the FITs for these events. Check: 

https://github.com/rootstock/federator-integration-tests/pulls?q=is%3Apr+is%3Aclosed

These tests are not running in FIT master yet because it depends on the changes in this pr. They have been tested locally and on the branches remotely.

The following pr will activate the FIT changes related to these changes after this pr is merged to rskj and powpeg-node master branches:

https://github.com/rootstock/federator-integration-tests/pull/134

This is the related powpeg-node pr that must be merged along this rskj pr: https://github.com/rsksmart/powpeg-node/pull/104

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Requires Activation Code (Hard Fork)

* **Other information**:
*fed:pegout-events-impv-integration*
*fit:test-pegout_confirmed-event*

PR #1907, opened 2022-10-14
Pegout events impv integration rebased

jeremy-then:


marcos-iov:
Duplicates https://github.com/rsksmart/rskj/pull/1900

PR #1915, opened 2022-11-04
Fixed issue with eth_call triggered for pending block

Vovchyk:
## Description

This fixes the issue when requesting a "pending" block data via JSON-RPC api (only reproducible when mining is enabled). The issue was introduced in this [PR](https://github.com/rsksmart/rskj/pull/1885).

## Motivation and Context

In #1885 implicit saving of a trie during block template creation was removed. But it turns out that some json-rpc methods were dependant on this, meaning that they expected the sate of a pending block to be sitting in the store.

This pull request adds one more method - `buildPending()` - which is specifically dedicated to the purpose of building a pending block with its final state (a trie) and cache them in memory for reuse if needed.

## Types of changes

## Checklist:

*ci:rskj-420-rc-ver*

### DIFF
--- a/rskj-core/src/main/java/co/rsk/mine/BlockToMineBuilder.java
+++ b/rskj-core/src/main/java/co/rsk/mine/BlockToMineBuilder.java
@@ -41,6 +41,7 @@ import org.ethereum.db.BlockStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nonnull;
 import java.math.BigInteger;
 import java.util.*;
 
@@ -102,6 +103,24 @@ public class BlockToMineBuilder {
     }
 
     /**
+     * Creates a pending block based on the parent block header. Pending block is temporary, not connected to the chain and
+     * includes txs from the mempool.
+     *
+     * @param parentHeader block header a "pending" block is based on.
+     *
+     * @return "pending" block result.
+     * */
+    @Nonnull
+    public BlockResult buildPending(@Nonnull BlockHeader parentHeader) {
+        Coin minimumGasPrice = minimumGasPriceCalculator.calculate(parentHeader.getMinimumGasPrice());
+        List<BlockHeader> uncles = getUnclesHeaders(parentHeader);
+        List<Transaction> txs = getTransactions(new ArrayList<>(), parentHeader, minimumGasPrice);
+        Block newBlock = createBlock(Collections.singletonList(parentHeader), uncles, txs, minimumGasPrice, null);
+
+        return executor.executeAndFill(newBlock, parentHeader);
+    }
+
+    /**
      * Creates a new block to mine based on the previous mainchain blocks.
      *
      * @param mainchainHeaders last best chain blocks where 0 index is the best block and so on.
@@ -109,17 +128,7 @@ public class BlockToMineBuilder {
      */
     public BlockResult build(List<BlockHeader> mainchainHeaders, byte[] extraData) {
         BlockHeader newBlockParentHeader = mainchainHeaders.get(0);
-        List<BlockHeader> uncles = FamilyUtils.getUnclesHeaders(
-                blockStore,
-                newBlockParentHeader.getNumber() + 1,
-                newBlockParentHeader.getHash(),
-                miningConfig.getUncleGenerationLimit()
-        );
-
-
-        if (uncles.size() > miningConfig.getUncleListLimit()) {
-            uncles = uncles.subList(0, miningConfig.getUncleListLimit());
-        }
+        List<BlockHeader> uncles = getUnclesHeaders(newBlockParentHeader);
 
         Coin minimumGasPrice = minimumGasPriceCalculator.calculate(newBlockParentHeader.getMinimumGasPrice());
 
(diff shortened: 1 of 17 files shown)


Vovchyk:
pipeline:run

rskj-core/src/main/java/co/rsk/rpc/ExecutionBlockRetriever.java
```diff
@@ -126,4 +127,30 @@ private static BlockResult newBlockResult(Block block) {
...
+
+    @Override
+    public void stop() {
+        emitter.removeListener(listener);
+    }
+
+    private class CachedResultCleaner extends EthereumListenerAdapter {
+        @Override
+        public void onPendingTransactionsReceived(List<Transaction> transactions) {
+            cleanCachedResult();
```
illuque-iov, line 161: What happens if `transactions` is empty because received transaction was not valid? Should we be clearing the cache in this case?

illuque-iov: Ok, just saw that is not possible, the event is not emitted in that case ([here](https://github.com/rsksmart/rskj/blob/995ac70e109b164b63992dd7d253c970da36fdce/rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java#L189)) 👍

rskj-core/src/main/java/co/rsk/mine/BlockToMineBuilder.java
```diff
@@ -101,6 +102,24 @@ public BlockToMineBuilder(
...
+     *
+     * @return "pending" block result.
+     * */
+    @Nonnull
+    public BlockResult buildPending(@Nonnull BlockHeader parentHeader) {
+        Coin minimumGasPrice = minimumGasPriceCalculator.calculate(parentHeader.getMinimumGasPrice());
+        List<BlockHeader> uncles = getUnclesHeaders(parentHeader);
+        List<Transaction> txs = getTransactions(new ArrayList<>(), parentHeader, minimumGasPrice);
+        Block newBlock = createBlock(Collections.singletonList(parentHeader), uncles, txs, minimumGasPrice, null);
+
```
illuque-iov, line 119: I think we have some duplication between this method and `co.rsk.mine.BlockToMineBuilder#build`. We could address it, WDYT?

rskj-core/src/main/java/co/rsk/rpc/ExecutionBlockRetriever.java
```diff
@@ -69,27 +66,26 @@ public BlockResult retrieveExecutionBlock(String bnOrId) {
         }
 
         if (PENDING_ID.equals(bnOrId)) {
```
illuque-iov, line 70: Minor thing, but I think this logic should be moved to its own method for the sake of readability.

rskj-core/src/main/java/co/rsk/rpc/ExecutionBlockRetriever.java
```diff
@@ -69,27 +66,26 @@ public BlockResult retrieveExecutionBlock(String bnOrId) {
...
-            Block bestBlock = blockchain.getBestBlock();
-            if (cachedBlock == null || !bestBlock.isParentOf(cachedBlock)) {
-
-                // If the miner server is not running there is no one to update the mining mainchain view,
-                // thus breaking eth_call with 'pending' parameter
-                //
-                // This is just a provisional fix not intended to remain in the long run
-                if (!minerServer.isRunning()) {
-                    miningMainchainView.addBest(bestBlock.getHeader());
+            synchronized (pendingBlockLock) {
```
illuque-iov, line 78: I think we also have some duplication here with the lines above comparing BB with cache, we could maybe extract a common method `getValidCachedResult` or similar.

Vovchyk:
pipeline:run

rskj-core/src/main/java/co/rsk/rpc/ExecutionBlockRetriever.java
Vovchyk: agree, done

rskj-core/src/main/java/co/rsk/rpc/ExecutionBlockRetriever.java
Vovchyk: done

rskj-core/src/main/java/co/rsk/mine/BlockToMineBuilder.java
Vovchyk: the thing is that these lines are a bit specific to each method, so I wouldn't extract common stuff into a separate method. Also methods themselves are not big

illuque-iov: If I get it right, the following 3 lines are conceptually performing the same operation: creating a pending block.
```
        Coin minimumGasPrice = minimumGasPriceCalculator.calculate(newBlockParentHeader.getMinimumGasPrice());
        final List<Transaction> txs = getTransactions(txsToRemove, newBlockParentHeader, minimumGasPrice);
        final Block newBlock = createBlock(mainchainHeaders, uncles, txs, minimumGasPrice, extraData);
```
IMO they could be made to a common `createPendingBlock` method without issues, just minor adaptations on variable creation. But not big deal, ofc.

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

rskj-core/src/main/java/co/rsk/mine/BlockToMineBuilder.java
Vovchyk: actually `buildPending()` creates a "pending" one that's used by json-rpc. Another one - `build()` - is used by minerServer for creating block templates. The second one has some side-effects, eg. removing txs from the mempool,  using list of `mainchainHeaders`, setting some extraData etc. That's why I said that they are a bit specific + small portion of semi-duplication (3 lines)

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

illuque-iov (approved):

PR #1919, opened 2022-11-08
New erp differential

joaquinlpereyra:
differential tests for new ERPP2SH. as usual, mostly a baseline from which to build more tests.


joaquinlpereyra:
I'll close this one as it seems my branch was too behind.

PR #1925, opened 2022-11-14
RSKIP 351

ilanolkies:
## Description
Implementation for RSKIP 351. Summary:
- some info of the block header is passed as header extension on body message
- the encoded block header has the hash of the extension in the rlp list on logs bloom element position padded to the same size to keep header-only clients compatible

## Motivation and Context
- new info is added to the block header, this rskip prevents block header message to become larger

## How Has This Been Tested?
Tests added activating/deactivating RSKIP351 or Fingerroot500 associated hard fork

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
+++ b/rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
@@ -1,21 +1,3 @@
-/*
- * This file is part of RskJ
- * Copyright (C) 2017 RSK Labs Ltd.
- * (derived from ethereumJ library, Copyright (c) 2016 <ether.camp>)
- *
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU Lesser General Public License as published by
- * the Free Software Foundation, either version 3 of the License, or
- * (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
- * GNU Lesser General Public License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public License
- * along with this program. If not, see <http://www.gnu.org/licenses/>.
- */
 package org.ethereum.core;
 
 import co.rsk.config.RskMiningConstants;
@@ -43,14 +25,20 @@ import static org.ethereum.util.ByteUtil.toHexStringOrEmpty;
  * Block header is a value object containing
  * the basic information of a block
  */
-public class BlockHeader {
+public abstract class BlockHeader {
+    /* RSKIP 351 */
+    public abstract byte getVersion();
+    public abstract BlockHeaderExtension getExtension();
+    public abstract void setExtension(BlockHeaderExtension extension);
+    // fields from block header extension
+    public abstract byte[] getLogsBloom();
+    public abstract void setLogsBloom(byte[] logsBloom);
+    // encoding to use in logs bloom field on header response message
+    public abstract byte[] getLogsBloomFieldEncoded();
 
     private static final int HASH_FOR_MERGED_MINING_PREFIX_LENGTH = 20;
     private static final int FORK_DETECTION_DATA_LENGTH = 12;
     private static final int UMM_LEAVES_LENGTH = 20;
-
-    /* RSKIP 351 version */
-    private final byte version;
 
     /* The SHA3 256-bit hash of the parent block, in its entirety */
     private final byte[] parentHash;
@@ -72,8 +60,6 @@ public class BlockHeader {
      * list portion, the trie is populate by [key, val] --> [rlp(index), rlp(tx_recipe)]
      * of the block */
     private byte[] receiptTrieRoot;
-    /* The bloom filter for the logs of the block */
-    private byte[] logsBloom;
     /**
      * A scalar value corresponding to the difficulty level of this block.
      * This can be calculated from the previous block’s difficulty level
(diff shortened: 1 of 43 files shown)


rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -318,20 +291,28 @@ public Keccak256 getHash() {
...
+    public byte[] getEncodedForHash() {
+        return this.getEncoded(true, !useRskip92Encoding, true);
     }
 
     @Nullable
     public Coin getMinimumGasPrice() {
         return this.minimumGasPrice;
     }
 
-    public byte[] getEncoded(boolean withMergedMiningFields, boolean withMerkleProofAndCoinbase) {
```
julianlen, line 334: As there are two versions so far, could we use the RSKIPs to differentiate between versions? Also, if we have 5 different versions, there'll be 5 classes

rskj-core/src/main/java/org/ethereum/core/GenesisHeader.java
```diff
@@ -8,7 +8,7 @@
 import org.ethereum.util.ByteUtil;
 import org.ethereum.util.RLP;
 
-public class GenesisHeader extends BlockHeader {
+public class GenesisHeader extends BlockHeaderV0 {
```
julianlen, line 11: why did you use BlockHeaderV0 instead of BlockHeader?

julianlen (commented):
LGTM! Made some comments, tho. It's pending Sergio's answer

rskj-core/src/main/java/org/ethereum/core/GenesisHeader.java
ilanolkies: We can make Genesis behave like a V0 since hard fork activation is >0 in testnet and mainnet

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -1,21 +1,3 @@
-/*
```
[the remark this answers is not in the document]
ilanolkies: Done https://github.com/rsksmart/rskj/pull/1930/commits/2627c74daeea5e0911960417a5fe6b8aca1dfbda

rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
ilanolkies: - As there are two versions so far, could we use the RSKIPs to differentiate between versions?
  I tried to use `if (RSKIP351 is active) newBehavior()` like because of two things: it prevents adding if clause each time the new behavior is needed + net module does not have easy access to activation config. So the result is that block Factory will set the behavior when creating the block instance. It's kind of null object pattern
- Also, if we have 5 different versions, there'll be 5 classes.
  That is the idea. if we have 5 different versions we avoid tons of `if (RSKIP351 is active) newBehavior()`

rskj-core/src/main/java/org/ethereum/core/BlockHeaderV1.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+        this.hash = null;
+
+        this.extension.setLogsBloom(logsBloom);
+    }
+
+    @Override
+    public byte[] getLogsBloomFieldEncoded() {
+        byte[] ecnoded = new byte[Bloom.BLOOM_BYTES];
+        ecnoded[0] = 0x1;
```
SergioDemianLerner, line 52: What is this 0x1 at position 0 ?

SergioDemianLerner: Why is ecnoded of size BLOOM_BYTES (256) when it should be 33 (if you add the 0x1)

SergioDemianLerner: Why is the field called "ecnoded" and not "encoded" ?

ilanolkies: Based on the RSKIP compressed encoding description https://github.com/rsksmart/RSKIPs/blob/miniheader/IPs/RSKIP351.md#compressed-header

> When serializing the block header version 1 to obtain the block hash, the field logsBloom is replaced by an RLP list of
>  - The blockVersion field (integer) (1 byte)
>  - The Keccak hash of the bloom filter data and the txExecutionSublistsEdges as a list of two elements (32 bytes).

Additionally, based on RSKIP 194

> To avoid forcing a costly firmware upgrade for all PowHSMs, the blockVersion is stored in the logsBloom filter, which is unchecked by the hardware.

The “get logs bloom field” will return 256 bytes that go in the rlp encoding of the compressed header. That is why it is created with the logs bloom size. It was easier to modify fixed size bytes than creating a new rlp list that fits exactly in a given size

I underdtand we need to keep the same 256 bytes size for that field. If not, we can just remove the padding and keep it 33 bytes size

(ecnoded is a typo)

SergioDemianLerner: I log bloom field does not need to be 256 bytes in length. In fact, we aim to reduce it because nobody uses it.
Second, I don't think the version should be stored inside the logBloom field: it think it deserves a separate field in the RLP. That gives much more flexibility (i.e. more than one byte for versioning)

SergioDemianLerner: Let me check again the RSKIP

ilanolkies: Ok so required changes are:
- make logs bloom field 32 bytes size
- add new rlp field for version

(it might be easier to implement logs bloom field as 32 bytes for the hash + rest of bytes for version though, but i like more your proposal)

ilanolkies:
Closing in favor of #1930

PR #1927, opened 2022-11-15
Move bridge storage keys to an enum

nathanieliov:


rskj-core/src/main/java/co/rsk/peg/storageprovider/BridgeStorageIndexKeys.java
```diff
@@ -0,0 +1,63 @@
...
+        return keyAsString;
+    }
+
+    public DataWord getKeyDataWord() {
+        return keyDataWord;
+    }
+
+    public DataWord getCompoundKeyDataWord(String delimiter, String key) {
+        return DataWord.fromLongString(keyAsString + delimiter + key);
+    }
```
marcos-iov, line 62:
```suggestion
    private String key;

    BridgeStorageIndexKeys(String key) {
        this.key = key;
    }

    public DataWord getKey() {
        return DataWord.fromString(key);
    }

    public DataWord getCompoundKey(String delimiter, String identifier) {
        return DataWord.fromLongString(key + delimiter + identifier);
    }
```
Many suggestions in one but I think it makes it more transparent. We avoid having the same data repeated in different formats, remove the unused method getKeyAsString, and avoid using the data type as part of the field name. What do you think?

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -1059,6 +1029,10 @@ private Federation deserializeFederationAccordingToVersion(
         return BridgeSerializationUtils.deserializeFederation(data, networkParameters);
     }
 
+    private <T> T safeGetFromRepository(BridgeStorageIndexKeys keyAddress, RepositoryDeserializer<T> deserializer) {
```
marcos-iov, line 1032: Not sure it's worth adding all these new methods. We could keep using the old ones by calling `getKey()` method. At least we avoid having these methods duplicated

marcos-iov (commented):
Great work, I can tell there was a lot of effort put into it

rskj-core/src/main/java/co/rsk/peg/storageprovider/BridgeStorageIndexKeys.java
nathanieliov: Nice. Looks good.

I left the `getKeyAsString` for future uses, but I'm ok with removing it.

Regarding the naming convention, understood. The class looks simpler and cleaner now.

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
nathanieliov: But this is not duplicating these methods, it is overloading them, and in effect also avoiding having to modify all the places I touched to add a `getKey()`. The purpose of doing it in this way is to force the caller to use the enum which makes the code self-explanatory on which key is being used, what I would like will be to keep only the new method added using the enum. I can add this if we agree to have always an enum key for any entry we want to add to the storage. What do you think?

rskj-core/src/main/java/co/rsk/peg/storageprovider/BridgeStorageIndexKeys.java
```diff
@@ -0,0 +1,63 @@
+package co.rsk.peg.storageprovider;
```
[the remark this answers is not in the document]
nathanieliov: Done. Class moved to `.co.rsk.peg` package

rskj-core/src/main/java/co/rsk/peg/storageprovider/BridgeStorageIndexKeys.java
nathanieliov: Changes done.

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
marcos-iov: That makes sense, let's keep only the one that receives the enum then. Thanks

marcos-iov:
Check the code smells reported by Sonar, it's an easy fix

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
nathanieliov:
```suggestion
 private <T> T getFromRepository(BridgeStorageIndexKeys keyAddress, RepositoryDeserializer<T> deserializer) throws IOException {
        return getFromRepository(keyAddress.getKey(), deserializer);
    }

    private <T> T getFromRepository(DataWord keyAddress, RepositoryDeserializer<T> deserializer) throws IOException {
        byte[] data = repository.getStorageBytes(contractAddress, keyAddress);
        return deserializer.deserialize(data);
    }
```

After doing some analysis on how to get rid of the second method, I came to the conclusion that it may not be possible to do it. The second method it's only being used by the compound keys and so far I see no other way to it different on how is being done today.

```suggestion

    private DataWord getStorageKeyForBtcTxHashAlreadyProcessed(Sha256Hash btcTxHash) {
        return BTC_TX_HASH_AP.getCompoundKeyDataWord("-", btcTxHash.toString());
    }

    private DataWord getStorageKeyForCoinbaseInformation(Sha256Hash btcTxHash) {
        return COINBASE_INFORMATION.getCompoundKeyDataWord("-", btcTxHash.toString());
    }

    private DataWord getStorageKeyForBtcBlockIndex(Integer height) {
        return BTC_BLOCK_HEIGHT.getCompoundKeyDataWord("-", height.toString());
    }
```

```
Optional<Long> height = getFromRepository(getStorageKeyForBtcTxHashAlreadyProcessed(btcTxHash), BridgeSerializationUtils::deserializeOptionalLong);
        if (!height.isPresent()) {
            return height;
        }
```

marcos-iov: What do you think of this?
```
public BridgeStorageIndexKeys getCompoundKey(String delimiter, String identifier) {
        return new BridgeStorageIndexKeys(key + delimiter + identifier);
    }
```

Would that make sense? Replacing the existing `getCompoundKeyDataWord` method fro the one above? What do you think?

Also, we might want to consider renaming the class to `BridgeStorageIndexKey` (singular) since it's going to represent a single key. What do you think?

rskj-core/src/main/java/co/rsk/peg/BridgeStorageIndexKey.java
```diff
@@ -0,0 +1,57 @@
...
+
+    BridgeStorageIndexKey(String key) {
+        this.key = key;
+    }
+
+    public DataWord getKey() {
+        return DataWord.fromString(key);
+    }
+
+    public DataWord getCompoundKeyDataWord(String delimiter, String identifier) {
```
kelvinator07, line 54: The **delimiter** Seems To Be a Constant ("-"). Is It Possible That It Changes In The Future? If No, Then We Can Make It a Constant Variable In This Class.

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
nathanieliov: In conclusion, after having a call and discussing different approaches, we decided to leave it as it is now, and leave this possible refactor for later due to the complexity for compound keys.

rskj-core/src/main/java/co/rsk/peg/BridgeStorageIndexKey.java
nathanieliov: I think it's better to have this flexibility and let the caller decide which delimiter to use instead of abstracting him from it.

kelvinator07 (approved):

PR #1938, opened 2022-11-30
RSKIP 351 updates

ilanolkies:
- Make extension data (ex logs bloom field) shortest as possible (32 bytes for version 1)
- Add version to RLP encoding (1 extra byte now). Now decode header will read version from RLP
- Express compressed on encode/decode (see `decodeHeader(RLPList rlpHeader, boolean compressed, boolean sealed)`)
- Rename methods and variables (use _extension data_ and _encoded compressed_ terms)
- Make header v0 read from extensionData (now logs bloom is stored in what is called the _extension data_ that both headers v0 and v1 have)

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
+++ b/rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
@@ -154,13 +154,20 @@ public class BlockFactory {
             ummRoot = rlpHeader.get(r++).getRLPRawData();
         }
 
-        byte version = !activationConfig.isActive(ConsensusRule.RSKIP351, blockNumber) ? 0x0 : (compressed
-                ? RLP.decodeList(logsBloomField).get(0).getRLPData()[0]
-                : rlpHeader.get(r++).getRLPData()[0]);
+        byte version = 0x0;
+
+        if (activationConfig.isActive(ConsensusRule.RSKIP351, blockNumber)) {
+            version = compressed
+                    ? RLP.decodeList(logsBloomField).get(0).getRLPData()[0]
+                    : rlpHeader.get(r++).getRLPData()[0];
+        }
 
         short[] txExecutionSublistsEdges = null;
-        if (rlpHeader.size() > r && activationConfig.isActive(ConsensusRule.RSKIP144, blockNumber)) {
-            txExecutionSublistsEdges = ByteUtil.rlpToShorts(rlpHeader.get(r++).getRLPRawData());
+
+        if (!activationConfig.isActive(ConsensusRule.RSKIP351, blockNumber) || !compressed) {
+            if (rlpHeader.size() > r && activationConfig.isActive(ConsensusRule.RSKIP144, blockNumber)) {
+                txExecutionSublistsEdges = ByteUtil.rlpToShorts(rlpHeader.get(r++).getRLPRawData());
+            }
         }
 
         byte[] bitcoinMergedMiningHeader = null;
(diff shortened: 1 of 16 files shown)


rskj-core/src/main/java/org/ethereum/core/BlockHeader.java
```diff
@@ -366,7 +370,11 @@ public byte[] getEncoded(boolean withMergedMiningFields, boolean withMerkleProof
             fieldToEncodeList.add(RLP.encodeElement(this.ummRoot));
         }
 
-        this.addExtraFieldsToEncoded(useExtensionEncoding, fieldToEncodeList);
+        if (this.getVersion() > 0) {
+            fieldToEncodeList.add(RLP.encodeByte(this.getVersion()));
```
SergioDemianLerner, line 374: Why the version field is encoded always ? It should only be encoded in uncompressed format. In compressed format, the version would be in the logsBloom field. Right ?

ilanolkies: I was standing on that while implementing and I didn’t lime how it looked 🙂 I think this has the same result of what we discussed. Now logs bloom field is 32 bytes long for version > 0  in that case you should read version from RLP element at the end

ilanolkies: Ok so we are going to keep the RLP size of the _compressed encoding V1_ equal to the RLP size of the _encoding V0_ (prior hard fork).

As `txEdges` and `version` are added on V1 and that `logsBloom` field is replaced by `RLP(version, extensionHash)`, I'm adding this test


```java
@Test
    public void compressedEncodingV1HasSameRLPSizeAsFullEncodedV0WithoutEdges () {
        // this test is added to assert that the rlp header size does not change
        // in the hard fork, assuming both RSKIP 351 and RSKIP 144 are activated
        // together
        byte[] logsBloom = new byte[Bloom.BLOOM_BYTES];
        logsBloom[0] = 0x01;
        logsBloom[1] = 0x02;
        logsBloom[2] = 0x03;
        logsBloom[3] = 0x04;

        BlockHeaderV0 headerV0 = (BlockHeaderV0) createBlockHeaderWithVersion((byte) 0x0);
        headerV0.setLogsBloom(logsBloom);
        headerV0.setTxExecutionSublistsEdges(null);

        BlockHeaderV1 headerV1 = (BlockHeaderV1) createBlockHeaderWithVersion((byte) 0x1);
        headerV1.setLogsBloom(logsBloom);
        headerV1.setTxExecutionSublistsEdges(null);

        RLPList compressedEncodingV1 = RLP.decodeList(headerV1.getEncodedCompressed());
        RLPList fullEncodedV0 = RLP.decodeList(headerV0.getFullEncoded());

        Assertions.assertEquals(fullEncodedV0.size(), compressedEncodingV1.size());
    }
```

rskj-core/src/test/java/co/rsk/core/BlockHeaderTest.java
```diff
@@ -492,17 +528,46 @@ public void fullEncodedV1IsTheSameAsEncodedForHeaderMessageButLogsBloomAndEdges
...
+        }
+    }
 
-        Assertions.assertEquals(fullEncoded.size() - 1, encodedForHeaderMessage.size());
+    @Test
+    public void compressedEncodingV1HasSameRLPSizeAsFullEncodedV0WithoutEdges () {
+        // this test is added to assert that the rlp header size does not change
+        // in the hard fork, assuming both RSKIP 351 and RSKIP 144 are activated
+        // together
+        byte[] logsBloom = new byte[Bloom.BLOOM_BYTES];
```
patogallaiovlabs, line 553: vars like this one or `edges` that are repeated initialization, could be declared static with some meaninful name

patogallaiovlabs (commented):
Minor suggestions

rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
```diff
@@ -154,9 +154,20 @@ private BlockHeader decodeHeader(RLPList rlpHeader, boolean sealed) {
             ummRoot = rlpHeader.get(r++).getRLPRawData();
         }
 
+        byte version = 0x0;
+
+        if (activationConfig.isActive(ConsensusRule.RSKIP351, blockNumber)) {
```
SergioDemianLerner, line 159: I think we discussed not using the blocknumber for detecting the version, but using the length of the field itself.

rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
```diff
@@ -117,7 +117,7 @@ private BlockHeader decodeHeader(RLPList rlpHeader, boolean sealed) {
             receiptTrieRoot = EMPTY_TRIE_HASH;
         }
 
-        byte[] logsBloom = rlpHeader.get(6).getRLPData();
+        byte[] logsBloomField = rlpHeader.get(6).getRLPData(); // logs bloom when decoding extended, list(version, hash(extension)) when compressed
```
SergioDemianLerner, line 120: maybe this variable should be called: extendedDataOrLogsBloom

rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
SergioDemianLerner: Anyway, don't think this is a blocker

ilanolkies: The implementation has the property of allowing parsing the blockchain with the length and without the activation. I think this is a more clear hard fork detection anyway

PR #1947, opened 2022-12-12
Not dispose Options instance in RocksDbDataSource until db is closed

casiojapi:
## Description

Not dispose Options instance in RocksDbDataSource until db is closed

## Motivation and Context

As of now, we are disposing and creating a new instance of the Options class every time we init RocksDB. According to the RocksDB’s javadoc we should not do that until db instance is closed.

## How Has This Been Tested?

Haven’t created unit tests for this specific change because I would like the team to review the changes made and see if there’s any other optimal solution. But all the existing unit tests are going through perfectly, and everything appears to be working fine by running the node locally.

## Types of changes

- [x]  Bug fix (non-breaking change which fixes an issue)
- [ ]  New feature (non-breaking change which adds functionality)
- [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x]  My code follows the code style of this project.
- [ ]  My change requires a change to the documentation.
- [ ]  I have updated the documentation accordingly.
- [ ]  Tests for the changes have been added (for bug fixes / features)
- [ ]  Requires Activation Code (Hard Fork)
- **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/datasource/RocksDbDataSource.java
+++ b/rskj-core/src/main/java/org/ethereum/datasource/RocksDbDataSource.java
@@ -75,19 +75,13 @@ public class RocksDbDataSource implements KeyValueDataSource {
         return ds;
     }
 
+
+
     @Override
     public void init() {
         resetDbLock.writeLock().lock();
         Metric metric = profiler.start(Profiler.PROFILING_TYPE.DB_INIT);
-
-        if (options == null) {
-            options = new Options();
-            options.setCreateIfMissing(true);
-            options.setCompressionType(CompressionType.NO_COMPRESSION);
-            options.setArenaBlockSize(GENERAL_SIZE);
-            options.setWriteBufferSize(GENERAL_SIZE);
-            options.setParanoidChecks(true);
-        }
+        instantiateOptionsInternal();
 
         try {
             logger.debug("~> RocksDbDataSource.init(): {}", name);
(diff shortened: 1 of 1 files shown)


casiojapi:
pipeline:run

casiojapi:
pipeline:run

casiojapi:
pipeline:run

casiojapi:
pipeline:run

casiojapi:
pipeline:run

casiojapi:
pipeline:run

Vovchyk:
pipeline:run

rskj-core/src/main/java/org/ethereum/datasource/RocksDbDataSource.java
```diff
@@ -86,13 +91,6 @@ public void init() {
 
             Objects.requireNonNull(name, "no name set to the db");
 
-            options.setCreateIfMissing(true);
-            options.setCompressionType(CompressionType.NO_COMPRESSION);
-            options.setArenaBlockSize(GENERAL_SIZE);
-            options.setWriteBufferSize(GENERAL_SIZE);
-
-            options.setParanoidChecks(true);
-
```
Vovchyk, line 95: probably it's no need to pass `options` into the `openDb()` method - it cap be used as a class field directly, wdyt?

rskj-core/src/main/java/org/ethereum/datasource/RocksDbDataSource.java
```diff
@@ -370,4 +368,14 @@ public void close() {
     public void flush() {
         // All is flushed immediately: there is no uncommittedCache to flush
     }
+
+    private static Options createOptions() {
+        Options tempOptions = new Options();
```
Vovchyk, line 373: just curious.. why did you call this local variable **temp** Options?

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/datasource/RocksDbDataSource.java
casiojapi: tbh I just didn't want to use the same "options" name as in the class field. And in my mind it made sense because that variable would be temporary (being in the local scope of the method) till I return it.

rskj-core/src/main/java/org/ethereum/datasource/RocksDbDataSource.java
casiojapi: yep, you are absolutely right

rskj-core/src/main/java/org/ethereum/datasource/RocksDbDataSource.java
casiojapi: But yes, I'll change it back to "options". I guess it's more readable that way.

casiojapi:
pipeline:run

Vovchyk (approved):

illuque-iov (dismissed):

Vovchyk (approved):

Vovchyk:
pipeline:run

PR #1948, opened 2022-12-13
RSKIP240 - Storage rent

fedejinich:
# RSKIP240 - Storage Rent

This PR implements the Storage Rent [RSKIP240](https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP240.md).

## Introduced changes

- Track trie nodes used within transaction execution.
  -  Tracked at `MutableRepositoryTracked`.
- `Trie` changes: added timestamp field.
- `MutableTrie` operations:  support reading and writing rent timestamp.
- Rent payment: a mechanism to pay rent for tracked trie-nodes.

#### TransactionExecutor

Added a `StorageRentManager` instance to pay for storage rent. 

#### StorageRentManager

This class performs the rent payment. It gets all the tracked trie keys from the `MutableRepository`, then converts those keys to `RentedNode`, calculates cumulative rent, then subtracts the gas consumed by the process (or OOG), and finally invokes the `transactionTrack`  to update rent timestamps.

Note `RentedNodes` rents are fetched from the `blockTrack` and not from the `transactionTrack` because we want the storage rent data before executing the block. Also, rent is updated in the `transactionTrack` because it should be updated if the transaction succeeds.

Note 2: after some iterations, I've noticed that the `RentedNode` class is unnecessary; it adds complexity without adding any relevant value.

#### MutableRepository

Wrapped all the read and write operations into "internal" methods (overridden in `MutableRepositoryTracked`). There are just five methods that perform read/write in the trie:
- `mutableTrie.get`
- `mutableTrie.put`
- `mutableTrie.getValueLength`
- `mutableTrie.getValueHash`
- `mutableTrie.deleteRecursive`
- `mutableTrie.getStorageKeys`

Now each time we want to use any of the above methods, we need to call it by using the internal method (eg `internalGet`). This can be improved, that's a major refactor, and it goes out of scope.

#### MutableRepositoryTracked

This class extends `MutableRepository` to add tracking capabilities. As previously mentioned, "internal" methods have been overridden to track involved trie-nodes.

#### RentedNode

I added this class to calculate 
- the rent amount
- new timestamp
- rollback fee (if the tx failed)

NOTE: This can be simplified by moving the logic into `StorageRentManager`

#### MutableTrie

Added methods to update and get rent timestamps for a given trie-key.

#### Trie

Added support for new trie-nodes with rent:
- modified serialize/deserialize methods.
  - now, these methods will use `TrieSerializer` to serialize/deserialize trie elements.
- adding methods to get and set timestamps for a given key.


fedejinich:
pipeline:run

fedejinich:
pipeline:run

fedejinich:
pipeline:run

fedejinich:
pipeline:run

fedejinich:
pipeline:run

PR #1978, opened 2023-02-20
Do not create peer scoring REPEATED_MESSAGE for different senders

rmoreliovlabs:
## Description
Do not create peer scoring REPEATED_MESSAGE record in case duplicate messages arrive from different peers

## Motivation and Context
As of now, we do create a peer scoring record for duplicate messages (block or transactions types of messages) sent by different peers [here](https://github.com/rsksmart/rskj/blob/d10450b71fdc369aaffc04f0b3f2677fbf93c726/rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java#L213-L231)

We need to fix it probably by taking into consideration a sender of a message while deciding if it’s repeated or not.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -66,7 +63,7 @@ public class NodeMessageHandler implements MessageHandler, InternalService, Runn
     private final PeerScoringManager peerScoringManager;
 
     private final StatusResolver statusResolver;
-    private final Set<Keccak256> receivedMessages = Collections.synchronizedSet(new HashSet<>());
+    private final Map<Peer, Set<Keccak256>> receivedPeerMessages = Collections.synchronizedMap(new HashMap<>());
```
asoto-iov, line 66: I think ConcurrentHashMap could be used instead.

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -213,14 +210,19 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
 
         if (!contains) {
             if (message.getMessageType() == MessageType.BLOCK_MESSAGE || message.getMessageType() == MessageType.TRANSACTIONS) {
-                if (this.receivedMessages.size() >= MAX_NUMBER_OF_MESSAGES_CACHED) {
-                    this.receivedMessages.clear();
+                if (receivedMessages.size() >= MAX_NUMBER_OF_MESSAGES_CACHED) {
+                    receivedMessages.clear();
                 }
-                this.receivedMessages.add(encodedMessage);
+                if (this.receivedPeerMessages.isEmpty()) {
```
asoto-iov, line 221: If I'm getting it right using this condition only one sender would be added to the map.
I think we should check `this.receivedPeerMessages.containsKey(sender)` instead.

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -213,14 +210,19 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
     private boolean allowByMessageUniqueness(Peer sender, Message message) {
         Keccak256 encodedMessage = new Keccak256(HashUtil.keccak256(message.getEncoded()));
 
+        Set<Keccak256> receivedMessages = receivedPeerMessages.getOrDefault(sender, Collections.synchronizedSet(new HashSet<>()));
         boolean contains = receivedMessages.contains(encodedMessage);
 
         if (!contains) {
             if (message.getMessageType() == MessageType.BLOCK_MESSAGE || message.getMessageType() == MessageType.TRANSACTIONS) {
```
asoto-iov, line 217: A message is only added if this condition is reached, so maybe, we could check the MessageType at the beginning of the method and directly return false if the message is not a block or a transaction type.

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
rmoreliovlabs: I don't know the context why `synchronizedSet` was selected in the past, I just followed the pattern. I just assumed they had a valid reason.

Maybe @Vovchyk can give more context here?

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -211,23 +208,30 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
+        }
 
-        if (!contains) {
-            if (message.getMessageType() == MessageType.BLOCK_MESSAGE || message.getMessageType() == MessageType.TRANSACTIONS) {
-                if (this.receivedMessages.size() >= MAX_NUMBER_OF_MESSAGES_CACHED) {
-                    this.receivedMessages.clear();
-                }
-                this.receivedMessages.add(encodedMessage);
-            }
+        Set<Keccak256> receivedMessages = receivedPeerMessages.get(sender);
```
illuque-iov, line 221: You may want to use `Map.computeIfAbsent()` instead, it is even recommended by Sonar.

illuque-iov (changes requested):
We would be introducing a bug with this changes, please review comments.

asoto-iov (commented):
adding pending comments, sorry I didn't submit the review before

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -213,21 +212,22 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
-                if (this.receivedMessages.size() >= MAX_NUMBER_OF_MESSAGES_CACHED) {
-                    this.receivedMessages.clear();
-                }
-                this.receivedMessages.add(encodedMessage);
+        if (receivedMessages.contains(encodedMessage)) {
+            reportEventToPeerScoring(sender, EventType.REPEATED_MESSAGE, "Received repeated message on {}, not added to the queue");
+            return false;
+        }
+
+        if (Stream.of(MessageType.BLOCK_MESSAGE, MessageType.TRANSACTIONS).anyMatch(t -> message.getMessageType() == t)) {
```
asoto-iov, line 222: I would keep the `if (message.getMessageType() == MessageType.BLOCK_MESSAGE || message.getMessageType() == MessageType.TRANSACTIONS)` instead of the stream. Is more efficient and simply to read.

asoto-iov (dismissed):

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -206,23 +232,32 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
+        ReceivedPeerMessageKey receivedPeerMessageKey = new ReceivedPeerMessageKey(sender.getPeerNodeID(), encodedMessage);
 
-        } else {
+        if (receivedPeerMessages.containsKey(receivedPeerMessageKey)) {
             reportEventToPeerScoring(sender, EventType.REPEATED_MESSAGE, "Received repeated message on {}, not added to the queue");
+            return false;
         }
 
-        return !contains;
+        if (Stream.of(MessageType.BLOCK_MESSAGE, MessageType.TRANSACTIONS).anyMatch(t -> message.getMessageType() == t)) {
```
Vovchyk, line 256: I guess this check is not needed any more, as there's one at the very top

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -206,23 +232,32 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
+            this.receivedPeerMessages.put(receivedPeerMessageKey, Instant.now().getEpochSecond());
```
Vovchyk, line 257:
```suggestion
            this.receivedPeerMessages.put(receivedPeerMessageKey, System.currentTimeMillis());
```
I'd rather use milliseconds - they are more accurate

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -206,23 +232,32 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
             }
+        }
+
+        Keccak256 encodedMessage = new Keccak256(HashUtil.keccak256(message.getEncoded()));
+        ReceivedPeerMessageKey receivedPeerMessageKey = new ReceivedPeerMessageKey(sender.getPeerNodeID(), encodedMessage);
 
-        } else {
+        if (receivedPeerMessages.containsKey(receivedPeerMessageKey)) {
             reportEventToPeerScoring(sender, EventType.REPEATED_MESSAGE, "Received repeated message on {}, not added to the queue");
+            return false;
```
Vovchyk, line 253: we probably need to update `receivedPeerMessages` with the current timestamp before returning

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -206,23 +227,32 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
-                    this.receivedMessages.clear();
-                }
-                this.receivedMessages.add(encodedMessage);
-            }
+        synchronized (receivedPeerMessages) {
+            logger.trace("Cleaning {} messages from rlp queue", receivedPeerMessages.size());
+            receivedPeerMessages.entrySet()
+                    .stream()
+                    .filter(entry -> currentTime - entry.getValue() > RECEIVED_MESSAGES_CACHE_DURATION)
+                    .forEach(entry -> receivedPeerMessages.remove(entry.getKey()));
```
illuque-iov, line 241: If I'm not mistaken this will cause a `ConcurrentModificationException` as we are removing an item during the iteration.

You may want to try with `receivedPeerMessages.entrySet().removeIf`.

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -206,23 +227,32 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
 
-        } else {
+        if (receivedPeerMessages.containsKey(receivedPeerMessageKey)) {
             reportEventToPeerScoring(sender, EventType.REPEATED_MESSAGE, "Received repeated message on {}, not added to the queue");
+            this.receivedPeerMessages.put(receivedPeerMessageKey, currentTime);
+            return false;
         }
 
-        return !contains;
+        this.receivedPeerMessages.put(receivedPeerMessageKey, currentTime);
```
illuque-iov, line 253: This line is duplicated, you can probably call it just before the `if (receivedPeerMessages.containsKey(receivedPeerMessageKey))`

illuque-iov: Sorry, explained myself wrong probably, I should've mentioned creating a variable storing if already contained before doing the put.

illuque-iov (changes requested):

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
rmoreliovlabs: Addressed

rmoreliovlabs:
pipeline:run

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -206,23 +250,30 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
+        long currentTime = System.currentTimeMillis();
 
-        if (!contains) {
-            if (message.getMessageType() == MessageType.BLOCK_MESSAGE || message.getMessageType() == MessageType.TRANSACTIONS) {
-                if (this.receivedMessages.size() >= MAX_NUMBER_OF_MESSAGES_CACHED) {
-                    this.receivedMessages.clear();
-                }
-                this.receivedMessages.add(encodedMessage);
-            }
+        synchronized (receivedPeerMessages) {
```
Vovchyk, line 259: what do you think about placing this cleanup logic between the `if (receivedPeerMessages.containsKey() {...` check and `this.receivedPeerMessages.put(receivedPeerMessageKey, currentTime);`? In this case some useful entries may still be available for the check before the cleanup.

also you could place entire work with the `receivedPeerMessages` collection inside the `synchronized` block, so that in this case there's no need to wrap the map with `Collections.synchronizedMap()` and do double synchronisation.

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
rmoreliovlabs: Addressed

rmoreliovlabs:
pipeline:run

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -206,23 +201,25 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
     private boolean allowByMessageUniqueness(Peer sender, Message message) {
+        if (message.getMessageType() != MessageType.BLOCK_MESSAGE && message.getMessageType() != MessageType.TRANSACTIONS) {
+            return true;
+        }
+
         Keccak256 encodedMessage = new Keccak256(HashUtil.keccak256(message.getEncoded()));
 
-        boolean contains = receivedMessages.contains(encodedMessage);
+        synchronized (receivedPeerMessagesMap) {
+            Set<NodeID> nodeIds = receivedPeerMessagesMap.getOrDefault(encodedMessage, new HashSet<>());
```
Vovchyk, line 211:
```suggestion
        boolean noDuplicateDetected;
        synchronized (receivedPeerMessagesMap) {
            Set<NodeID> nodeIds = receivedPeerMessagesMap.computeIfAbsent(encodedMessage, k -> new HashSet<>());
```

minor thing, but with this optimisation a new instance of `HashSet<>` would be created only if no entry found in the map (the lambda itself is stateless / doesn't capture anything, no a singleton would be used behind scenes )

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -206,23 +201,25 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
-                }
-                this.receivedMessages.add(encodedMessage);
+            if (nodeIds.contains(sender.getPeerNodeID())) {
+                reportEventToPeerScoring(sender, EventType.REPEATED_MESSAGE, "Received repeated message on {}, not added to the queue");
+                return false;
             }
 
-        } else {
-            reportEventToPeerScoring(sender, EventType.REPEATED_MESSAGE, "Received repeated message on {}, not added to the queue");
+            nodeIds.add(sender.getPeerNodeID());
```
Vovchyk, line 218:
```suggestion
            noDuplicateDetected = nodeIds.isEmpty();
            nodeIds.add(sender.getPeerNodeID());
```

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -206,23 +201,25 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
...
+            receivedPeerMessagesMap.put(encodedMessage, nodeIds);
         }
 
-        return !contains;
+        return true;
```
Vovchyk, line 222:
```suggestion
        return noDuplicateDetected;
```

before your change this method was returning `!contains`, which means that false was being returned even if a duplicate message arrived from another peer, not necessary same peer. I'd try to preserve the previous logic of what was returned before.

Regarding the peer scoring punishment logic, it LGTM

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
```diff
@@ -206,23 +201,25 @@ private boolean allowByMinerNotBanned(Peer sender, Message message) {
      * record event if message is repeated
      */
     private boolean allowByMessageUniqueness(Peer sender, Message message) {
```
Vovchyk, line 203: the logic of this method has change a bit.. could you please make it visible for tests and cover different inputs and expected results?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/net/NodeMessageHandler.java
rmoreliovlabs: Addressed

Vovchyk (approved):

rmoreliovlabs:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk (approved):

illuque-iov (approved):

PR #1984, opened 2023-02-27
Return "0x0" type in transaction receipt and result

casiojapi:
To be backwards compatible with Ethereum we need to extend our transaction receipt DTO and transaction result DTO in the RPC response with new the new field “type”, and return “0x0” as its value. - following EIP-2718

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/rpc/dto/TransactionReceiptDTO.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/dto/TransactionReceiptDTO.java
@@ -72,7 +72,7 @@ public class TransactionReceiptDTO {
         transactionHash = receipt.getTransaction().getHash().toJsonString();
         transactionIndex = HexUtils.toQuantityJsonHex(txInfo.getIndex());
         logsBloom = HexUtils.toUnformattedJsonHex(txInfo.getReceipt().getBloomFilter().getData());
-        type = "0x00";
+        type = "0x0";
     }
 
     public String getTransactionHash() {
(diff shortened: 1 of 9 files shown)


rskj-core/src/main/java/org/ethereum/rpc/dto/TransactionReceiptDTO.java
```diff
@@ -71,6 +72,7 @@ public TransactionReceiptDTO(Block block, TransactionInfo txInfo, SignatureCache
         transactionHash = receipt.getTransaction().getHash().toJsonString();
         transactionIndex = HexUtils.toQuantityJsonHex(txInfo.getIndex());
         logsBloom = HexUtils.toUnformattedJsonHex(txInfo.getReceipt().getBloomFilter().getData());
+        type = "0x0";
```
asoto-iov, line 75: This value doesn't change so it is better to use a static constant instead of using a var for each object.

rskj-core/src/main/java/org/ethereum/rpc/dto/TransactionResultDTO.java
```diff
@@ -63,6 +64,8 @@ public TransactionResultDTO(Block b, Integer index, Transaction tx, boolean zero
 
         gasPrice = HexUtils.toQuantityJsonHex(tx.getGasPrice().getBytes());
 
+        type = "0x0";
```
asoto-iov, line 67: as mentioned above for this case it is better to use a constant instead.

asoto-iov:
Both TransactionResultDTO and TransactionReceiptDTO have its own test class. I would add a unit test there to check this new value there.

Vovchyk (dismissed):

rskj-core/src/main/java/org/ethereum/rpc/dto/TransactionReceiptDTO.java
casiojapi: Yep, makes total sense. thanks

casiojapi:
pipeline:run

PR #1985, opened 2023-02-27
eth_getLogs remove redundant queries

illuque-iov:
## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
@@ -446,8 +446,8 @@ public class BlockChainImpl implements Blockchain {
     }
 
     @Override
-    public TransactionInfo getTransactionInfoForMainChainBlock(Transaction tx, Block blockMainChain) {
-        TransactionInfo txInfo = receiptStore.get(tx.getHash().getBytes(), blockMainChain.getHash().getBytes()).orElse(null);
+    public TransactionInfo getTransactionInfoForBlock(Transaction tx, byte[] blockHash) {
+        TransactionInfo txInfo = receiptStore.get(tx.getHash().getBytes(), blockHash).orElse(null);
         if (txInfo == null) {
             return null;
         }
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
@@ -446,7 +446,7 @@ public class BlockChainImpl implements Blockchain {
     }
 
     @Override
-    public TransactionInfo getTransactionInfoForBlock(Transaction tx, byte[] blockHash) {
+    public TransactionInfo getTransactionInfoByBlock(Transaction tx, byte[] blockHash) {
         TransactionInfo txInfo = receiptStore.get(tx.getHash().getBytes(), blockHash).orElse(null);
         if (txInfo == null) {
             return null;
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
@@ -446,6 +446,18 @@ public class BlockChainImpl implements Blockchain {
     }
 
     @Override
+    public TransactionInfo getTransactionInfoForMainChainBlock(Transaction tx, Block blockMainChain) {
+        TransactionInfo txInfo = receiptStore.get(tx.getHash().getBytes(), blockMainChain.getHash().getBytes()).orElse(null);
+        if (txInfo == null) {
+            return null;
+        }
+
+        txInfo.setTransaction(tx);
+
+        return txInfo;
+    }
+
+    @Override
     public BlockDifficulty getTotalDifficulty() {
         return status.getTotalDifficulty();
     }
--- a/rskj-core/src/main/java/co/rsk/logfilter/BlocksBloom.java
+++ b/rskj-core/src/main/java/co/rsk/logfilter/BlocksBloom.java
@@ -29,19 +29,34 @@ public class BlocksBloom {
     private long fromBlock;
     private long toBlock;
     private boolean empty;
+    private final boolean backwardsAddition;
 
-    public BlocksBloom() {
+    private BlocksBloom(boolean backwardsAddition) {
         this.bloom = new Bloom();
         this.fromBlock = 0;
         this.toBlock = 0;
         this.empty = true;
+        this.backwardsAddition = backwardsAddition;
     }
 
-    public BlocksBloom(long fromBlock, long toBlock, Bloom bloom) {
+    private BlocksBloom(long fromBlock, long toBlock, Bloom bloom) {
         this.bloom = bloom;
         this.fromBlock = fromBlock;
         this.toBlock = toBlock;
         this.empty = false;
+        this.backwardsAddition = false;
+    }
+
+    public static BlocksBloom createEmpty() {
+        return new BlocksBloom(false);
+    }
+
+    public static BlocksBloom createEmptyWithBackwardsAddition() {
+        return new BlocksBloom(true);
+    }
+
+    public static BlocksBloom createForExisting(long fromBlock, long toBlock, Bloom bloom) {
+        return new BlocksBloom(fromBlock, toBlock, bloom);
     }
 
     public Bloom getBloom() { return this.bloom; }
@@ -67,15 +82,7 @@ public class BlocksBloom {
     }
 
     public void addBlockBloom(long blockNumber, Bloom blockBloom) {
-        addBlockBloomInternal(blockNumber, blockBloom, false);
-    }
-
-    public void addBlockBloomBackwards(long blockNumber, Bloom blockBloom) {
-        addBlockBloomInternal(blockNumber, blockBloom, true);
-    }
-
-    private void addBlockBloomInternal(long blockNumber, Bloom blockBloom, boolean reverseOrder) {
-        long nextBlock = reverseOrder ? toBlock - 1 : toBlock + 1;
+        long nextBlock = backwardsAddition ? toBlock - 1 : toBlock + 1;
 
         if (this.empty) {
             this.fromBlock = blockNumber;
--- a/rskj-core/src/main/java/co/rsk/logfilter/BlocksBloom.java
+++ b/rskj-core/src/main/java/co/rsk/logfilter/BlocksBloom.java
@@ -67,12 +67,22 @@ public class BlocksBloom {
     }
 
     public void addBlockBloom(long blockNumber, Bloom blockBloom) {
+        addBlockBloomInternal(blockNumber, blockBloom, false);
+    }
+
+    public void addBlockBloomBackwards(long blockNumber, Bloom blockBloom) {
+        addBlockBloomInternal(blockNumber, blockBloom, true);
+    }
+
+    private void addBlockBloomInternal(long blockNumber, Bloom blockBloom, boolean reverseOrder) {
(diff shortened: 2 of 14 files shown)


rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
```diff
@@ -445,6 +445,18 @@ public TransactionInfo getTransactionInfo(byte[] txHash) {
         return txInfo;
     }
 
+    @Override
+    public TransactionInfo getTransactionInfoForMainChainBlock(Transaction tx, Block blockMainChain) {
```
Vovchyk, line 449:
```suggestion
    public TransactionInfo getTransactionInfo(Transaction tx, byte[] blockHash) {
```
in this method we don't need entire block object. in some scenarios we might not have a block object to provide but only its hash

rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java
```diff
@@ -78,19 +73,48 @@ void onTransaction(Transaction tx, Block b, int txIndex) {
...
+                onLogMatch(logInfo, block, txIndex, receipt.getTransaction(), i);
             }
         }
     }
 
-    void onBlock(Block b) {
-        if (addressesTopicsFilter.matchBloom(new Bloom(b.getLogBloom()))) {
-            int txIdx = 0;
+    void onTransaction(Transaction tx, Block block, int txIndex) {
+        TransactionInfo txInfo = blockchain.getTransactionInfo(tx.getHash().getBytes());
```
Vovchyk, line 82:
```suggestion
        TransactionInfo txInfo = blockchain.getTransactionInfo(tx.getHash().getBytes(), block.getHash().getBytes());

```

seems we don't need extra methods here - `onBlockFromMainChain()`  and `onTransactionFromMainChain()`? Incorporating that new method `BlockChainImpl::getTransactionInfo(Transaction tx, byte[] blockHash)` you added should be sufficient?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloom.java
```diff
@@ -67,12 +67,22 @@ public boolean hasBlockBloom(long blockNumber) {
     }
 
     public void addBlockBloom(long blockNumber, Bloom blockBloom) {
+        addBlockBloomInternal(blockNumber, blockBloom, false);
+    }
+
+    public void addBlockBloomBackwards(long blockNumber, Bloom blockBloom) {
```
asoto-iov, line 73: Not sure if it is ok for a bloom to have the option of going in both directions (forward and backward). Could have sense to create a BackwardsBlooksBloom?

illuque-iov: Good point, this could lead to misuse of the bloom if sometimes called forward and sometimes backward. Will address it.

rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java
illuque-iov: Good point. Addressed.

rskj-core/src/main/java/co/rsk/core/bc/BlockChainImpl.java
illuque-iov: 👍 addressed

rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java
```diff
@@ -231,47 +232,75 @@ private static void processSingleBlockByHash(String blockHash, Blockchain blockc
...
+        }
+
+        // start accumulating blocks the first time a lastInRange block is reached to keep only complete blooms
+        if (bloomAccumulator == null) {
+            return null;
+        }
+
+        bloomAccumulator.addBlockBloom(block.getNumber(), new Bloom(block.getLogBloom()));
+
+        boolean firstInRange = blocksBloomStore.firstNumberInRange(block.getNumber()) == block.getNumber();
```
asoto-iov, line 287: is there a reason to create isLastInRange and not isFirstInRange?

rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java
```diff
@@ -231,47 +232,75 @@ private static void processSingleBlockByHash(String blockHash, Blockchain blockc
...
+            if (needsBloom) {
+                bloomAccumulator = addBlockBloom(block, bloomAccumulator, blocksBloomStore);
+            }
 
-                if (auxiliaryBlocksBloom != null) {
-                    auxiliaryBlocksBloom.addBlockBloom(blockNum, new Bloom(block.getLogBloom()));
-                }
+            // reverseTxOrder=true to process txs in reverse order for later complete list reverse
+            filter.onBlock(block, true);
+            block = blockchain.getBlockByHash(block.getParentHash().getBytes());
```
asoto-iov, line 262: What is the advantage of work with the block object instead of the block number into the loop?

rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java
```diff
@@ -231,47 +232,75 @@ private static void processSingleBlockByHash(String blockHash, Blockchain blockc
             return;
         }
 
-        long blockNumber = blockByHash.getNumber();
-        processBlocks(blockNumber, blockNumber, filter, blockchain, blocksBloomStore);
+        processBlocks(blockByHash, blockByHash, filter, blockchain, blocksBloomStore);
```
asoto-iov, line 235: If only one block is processed, is there a reason to not use filter.onBlock(..)?

rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java
```diff
@@ -231,47 +232,75 @@ private static void processSingleBlockByHash(String blockHash, Blockchain blockc
...
+            // reverseTxOrder=true to process txs in reverse order for later complete list reverse
+            filter.onBlock(block, true);
+            block = blockchain.getBlockByHash(block.getParentHash().getBytes());
+        }
 
-                if (auxiliaryBlocksBloom != null && blocksBloomStore.lastNumberInRange(blockNum) == blockNum) {
-                    blocksBloomStore.addBlocksBloom(auxiliaryBlocksBloom);
-                }
+        // sort in a from-to fashion after looping in reverse order
+        filter.reverseEvents();
```
asoto-iov, line 266: maybe the filter already have events, so maybe it could be better to only revert these events and not all the filter list.

rskj-core/src/test/java/co/rsk/logfilter/BlocksBloomTest.java
```diff
@@ -11,7 +11,7 @@ class BlocksBloomTest {
```
asoto-iov, line 11: Maybe it would be nice to create a test with the backward logic, and also to very it works the same just to prevent future problems.

illuque-iov: Actually I found a bug thanks to this.

rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java
illuque-iov: Agree, but I will leave it like this (as it was before also) just in case we want to create a bloom even for latest block (regtest mode, whatever).

rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java
illuque-iov: It doesn't, Filter was created specifically for this "backwards" round.

rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java
illuque-iov: Addressed.

rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java
illuque-iov: We do less queries compared to `blockStore#getChainBlockByNumber` and also when using hash and looping backwards we prevent a potential bug on forks. @Vovchyk spotted this 2 things.

asoto-iov (dismissed):

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloom.java
```diff
@@ -71,11 +86,11 @@ public void addBlockBloom(long blockNumber, Bloom blockBloom) {
             this.fromBlock = blockNumber;
             this.toBlock = blockNumber;
             this.empty = false;
-        }
-        else if (blockNumber == toBlock + 1) {
+        } else if (backwardsAddition && blockNumber == toBlock - 1) {
```
Vovchyk, line 89: shouldn't it be `&& blockNumber == fromBlock - 1` ?

Vovchyk: do we need this `backwardsAddition ` flag? I guess we could simply support addition op in both directions?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/logfilter/BlocksBloom.java
illuque-iov: > shouldn't it be && blockNumber == fromBlock - 1 ?

not sure if I get your point Vlad, but we are adding blocks backwards in this mode, so to ensure no gaps exist, we need to assert that received block is current one minus one.

illuque-iov: > do we need this backwardsAddition flag? I guess we could simply support addition op in both directions?

I had sth like that before, having just two add methods, but [this comment](https://github.com/rsksmart/rskj/pull/1985#discussion_r1122760985) from @asoto-iov made me realise it could be problematic (and misleader for next dev using it) to allow adding in both directions.

And since for now we don't have the use case of adding in both directions, IMO it's simpler for the dev to know what direction will be used for the whole BlocksBloom lifecycle.

WDYT? Do you see anything I might be missing?

Vovchyk (approved):

Vovchyk:
pipeline:run

rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java
```diff
@@ -68,40 +67,44 @@ void onLogMatch(LogInfo logInfo, Block b, int txIndex, Transaction tx, int logId
...
+            onTransaction(txs.get(txIdx), block, txIdx, reverseTxOrder);
         }
     }
 
     @Override
     public void newBlockReceived(Block b) {
         if (this.fromLatestBlock) {
             this.clearEvents();
-            onBlock(b);
+            onBlock(b, false);
```
VaWheel, line 104: isn't reverseTxOrder (true) should be specified here?

illuque-iov: I might be missing sth because the logic is a bit complex, but I think it's correct for it to be `false`, as when we receive a new block we always add it forward to the bloom.

Bear in mind that `org.ethereum.rpc.LogFilter#retrieveHistoricalData` is static and receives its own (and newly created) `LogFilter`. This filter is then added to `installedFilters` list on `org.ethereum.rpc.FilterManager#registerFilter`. This list is then looped and ends up calling `org.ethereum.rpc.FilterManager#newBlockReceived` that itself calls `org.ethereum.rpc.LogFilter#newBlockReceived` for each installed filter.

Furthermore, we are only adding items backwards if the bloom range does not exist already for the given filter.

illuque-iov:
pipeline:run

PR #1997, opened 2023-03-24
Adding max response size to JsonRPC requests. 

asoto-iov:
## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcCustomServer.java
```diff
@@ -68,14 +72,24 @@ protected JsonResponse handleJsonNodeRequest(final JsonNode node) throws JsonPar
             response = super.handleJsonNodeRequest(node);
             ExecTimeoutContext.checkIfExpired();
 
-            return response;
+        } else {
```
rmoreliovlabs, line 75: Why did we move this here inside an `else` block?

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Iterator;
+import java.util.Map;
+
+public class JsonResponseSizeLimiter {
+    private static final int BRACKETS_SIZE_IN_BYTES = 2;
+    private static final int COMMA_SIZE = 1;
+    private static final int JSON_FILED_MISSING_SYMBOLS = 3;
+    private static final long serialVersionUID = -4364188205571356867L;
+
+    private int acc = 0;
```
rmoreliovlabs, line 19: I feel changing this variable in multiple instanced methods is not a good idea. Not sure if sharing the state is a good idea.

I could be wrong.

What do you think @Vovchyk  @illuque-iov ?

asoto-iov: I'm not sure if I got your point. 
This value is not static and each object has its own value to sum the total length in bytes. From other parts of the code we call the static method, getSizeInBytesWithLimit and this method is the one in charge of instantiate the object and return the result.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,106 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import org.ethereum.rpc.exception.RskJsonRpcRequestException;
+
+import java.util.Iterator;
+import java.util.Map;
+
+public class JsonResponseSizeLimiter {
```
asoto-iov, line 13: I don't like too much the way the symbols sizes are calculated when traversing through the different elements of the JSON object. Any suggestion about how to improve is well received.

rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcCustomServer.java
asoto-iov: In order to measure the size of the object I need to get the response and calculate it. 
So if I understood it correctly, the condition in the line 71 `timeout <= 0` is the one used to decide what to do, if create the new timeout or not, so I added the if else to generate the response and the timeout and them calculate the size after this. 

The time to calculate the size is not being considered for these timeouts but I consider it irrelevant. 

In any case the timeout is not working as expected for batch request so It will be refactored.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,106 @@
...
+        return acc;
+    }
+
+    private int getFiledLength(Map.Entry<String, JsonNode> field) throws JsonProcessingException {
+        JsonNode value = field.getValue();
+        if (value.isObject() || value.isArray()) {
+            int objectSize = getLength(value);
+            return (field.getKey()).getBytes().length + JSON_FILED_MISSING_SYMBOLS + objectSize;
+        }
+        return acc + objectMapper.writeValueAsBytes(field).length - BRACKETS_SIZE_IN_BYTES;
```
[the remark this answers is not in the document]
illuque-iov: Also, I think it is not responsibility of `getFiledLength` to know/use `acc`, we could probably just return `objectMapper.writeValueAsBytes(field).length - BRACKETS_SIZE_IN_BYTES;` and sum it in the caller.

But again, I think this point is never reached.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,106 @@
...
+        while (elements.hasNext()) {
+            acc = getLength(elements.next());
+            eleNo++;
+        }
+        return acc + eleNo + BRACKETS_SIZE_IN_BYTES;
+    }
+
+    private int handleObject(JsonNode rootNode) throws JsonProcessingException {
+        Iterator<Map.Entry<String, JsonNode>> fields = rootNode.fields();
+        int fieldsNo = fields.hasNext() ? -1 : 0;
```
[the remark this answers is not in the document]
illuque-iov: I was thinking that maybe it's simpler to add the comma on each iteration and have a flag "itemsAdded" to then subtract the extra one at the end of the method. Up to you.

illuque-iov (commented):
Not finished yet neither did the QG, but dropping some comments already for later call.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
illuque-iov: Probably Reynold's doubt can be avoided if you create a explicit creator method outside of `getSizeInBytesWithLimit` and then invoke `getSizeInBytesWithLimit` from it.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,106 @@
...
+        JsonResponseSizeLimiter limiter = new JsonResponseSizeLimiter(limit, node);
+        try {
+            return limiter.getJsonLength();
+        } catch (JsonProcessingException jsonProcessingException) {
+            throw RskJsonRpcRequestException.invalidParamError("Couldn't process Json Object:  " + jsonProcessingException.getMessage());
+        }
+    }
+
+    private int getJsonLength() throws JsonProcessingException {
+        acc = 0;
```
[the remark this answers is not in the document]
illuque-iov: Ok, I guess it is intended to ensure restarting from 0 if called multiple times.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public static int getSizeInBytesWithLimit(JsonNode node, int limit) {
+        if (node == null) {
+            return 0;
+        }
+        JsonResponseSizeLimiter limiter = new JsonResponseSizeLimiter(limit, node);
+        try {
+            return limiter.getJsonLength();
+        } catch (JsonProcessingException jsonProcessingException) {
```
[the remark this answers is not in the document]
illuque-iov: Ok, I got confused with another exception, said nothing.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,106 @@
...
+            throw RskJsonRpcRequestException.invalidParamError("Couldn't process Json Object:  " + jsonProcessingException.getMessage());
+        }
+    }
+
+    private int getJsonLength() throws JsonProcessingException {
```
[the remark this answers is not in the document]
illuque-iov: Clarified, you will use this to get total size as well.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    private int handleArray(ArrayNode arrayNode) throws JsonProcessingException {
+        Iterator<JsonNode> elements = arrayNode.elements();
+        int eleNo = elements.hasNext() ? -1 : 0;
+        while (elements.hasNext()) {
+            acc = getLength(elements.next());
+            eleNo++;
+        }
+        return acc + eleNo + BRACKETS_SIZE_IN_BYTES;
```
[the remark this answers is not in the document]
illuque-iov: Ok, it is representing the commas. Said nothing.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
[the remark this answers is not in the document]
asoto-iov: It is the one used for getting the length of each field of a object and it is always reached. The if else is avoiding it when the field is an inner object and not a single filed.

rskj-core/src/main/java/co/rsk/rpc/netty/ResponseSizeLimitContext.java
```diff
@@ -0,0 +1,62 @@
...
+        accumulatedResponseSize.remove();
+    }
+
+    public static ResponseSizeLimitContext createResponseSizeContext(int limit) {
+        ResponseSizeLimitContext ctx = new ResponseSizeLimitContext(limit);
+        accumulatedResponseSize.set(ctx);
+        return ctx;
+    }
+
+    public static void addResponseSize(int size) {
```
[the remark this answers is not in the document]
asoto-iov: Some of fields methods were not deleted after a small refactor. Good points.

rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcWeb3ServerHandler.java
```diff
@@ -54,26 +52,26 @@ public class JsonRpcWeb3ServerHandler extends SimpleChannelInboundHandler<ByteBu
...
     @Override
     protected void channelRead0(ChannelHandlerContext ctx, ByteBufHolder request) throws Exception {
         ByteBuf responseContent = Unpooled.buffer();
         int responseCode;
-
         try (ByteBufOutputStream os = new ByteBufOutputStream(responseContent);
-             ByteBufInputStream is = new ByteBufInputStream(request.content().retain())) {
-
+             ByteBufInputStream is = new ByteBufInputStream(request.content().retain());
+             ResponseSizeLimitContext rslCtx = ResponseSizeLimitContext.createResponseSizeContext(maxResponseSize)) {
```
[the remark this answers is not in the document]
asoto-iov: It is an auto-closeable object so it is declared there to always close it.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,106 @@
...
+            throw RskJsonRpcRequestException.invalidParamError("Couldn't process Json Object:  " + jsonProcessingException.getMessage());
+        }
+    }
+
+    private int getJsonLength() throws JsonProcessingException {
+        acc = 0;
+        return getLength(root);
+    }
+
+    private int getLength(JsonNode jsonNode) throws JsonProcessingException {
```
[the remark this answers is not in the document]
asoto-iov: I think with int value is enough.

rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcWeb3ServerHandler.java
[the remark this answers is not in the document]
illuque-iov: Ah true, did not realise it was on the `try` line 👍

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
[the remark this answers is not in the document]
illuque-iov: Yep, missed that the other day.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,97 @@
...
+        Iterator<Map.Entry<String, JsonNode>> fields = rootNode.fields();
+        int fieldSeparatorCount = fields.hasNext() ? -1 : 0;
+        while (fields.hasNext()) {
+            fieldSeparatorCount++;
+            getFieldLength(fields.next());
+        }
+        sumAndCheck(BRACKETS_SIZE_IN_BYTES + (COMMA_SIZE * fieldSeparatorCount));
+    }
+
+    private void getFieldLength(Map.Entry<String, JsonNode> field) throws JsonProcessingException {
```
illuque-iov, line 80: Minor thing, but I was thinking... WDYT about calling this one `handleField` to be consistent with the other methods?

rskj-core/src/test/java/co/rsk/rpc/netty/Web3HttpServerTest.java
```diff
@@ -298,6 +306,229 @@ void smokeTestProducesMethodInvalidException() throws Exception {
         smokeTest(APPLICATION_JSON, "localhost", filteredModules, null, mockResult, "web3sha3");
     }
 
+    @Test
+    void testMaxResponseSize() throws Exception {
+        Web3 web3Mock = Mockito.mock(Web3.class);
+        String mockResult = getMediumJsonResponse();
+        int responseSize = mockResult.getBytes().length * 2 - 1;
```
illuque-iov, line 313: Taking the string as source is not a good estimation if it includes spaces, new lines, etc. If you get the JsonNode from it and then do `writeValueAsBytes` you will get a size of `3655` vs `mockResult.getBytes().length` which is `4610`, not a negligible difference.

rskj-core/src/test/java/co/rsk/rpc/netty/Web3HttpServerTest.java
```diff
@@ -298,6 +306,229 @@ void smokeTestProducesMethodInvalidException() throws Exception {
...
+        Mockito.when(mockCorsConfiguration.hasHeader()).thenReturn(true);
+        Mockito.when(mockCorsConfiguration.getHeader()).thenReturn("*");
+
+        int randomPort = 9111;
+
+        List<ModuleDescription> filteredModules = Collections.singletonList(new ModuleDescription("web3", "1.0", true, Collections.emptyList(), Collections.emptyList(), 0, new HashMap<>()));
+        JsonRpcWeb3FilterHandler filterHandler = new JsonRpcWeb3FilterHandler("*", InetAddress.getLoopbackAddress(), new ArrayList<>());
+        JsonRpcWeb3ServerProperties properties = JsonRpcWeb3ServerProperties.builder()
+                .rpcModules(filteredModules)
+                .rpcTimeout(expectedTimeoutPerRequest / 2)
```
illuque-iov, line 495: Too much margin for the test to be accurate probably.

illuque-iov (approved):
Quality Gate done. Left some comments that I think could make tests even better.

rskj-core/src/test/java/co/rsk/rpc/netty/JsonRpcCustomServerTest.java
```diff
@@ -0,0 +1,215 @@
...
+        JsonResponse actualResponse = jsonRpcCustomServer.handleJsonNodeRequest(request);
+        assertEquals(response, actualResponse.getResponse().get("result").asText());
+    }
+
+    @Test
+    void testHandleJsonNodeRequest_WithResponseLimit() throws Exception {
+        String response = "test_method_response";
+        JsonNode request = objectMapper.readTree(FIRST_METHOD_REQUEST);
+        Web3Test handler = mock(Web3Test.class);
+        ResponseSizeLimitContext.createResponseSizeContext(response.getBytes(StandardCharsets.UTF_8).length / 2);
```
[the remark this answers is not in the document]
asoto-iov: The idea of this test is to be sure that the current exception is thrown when the limit is reached. So here I'm using that one. 
Maybe the accurate ones can be added in a test class for `ResponseSizeLimitContext`

rskj-core/src/test/java/co/rsk/rpc/netty/JsonRpcCustomServerTest.java
```diff
@@ -0,0 +1,215 @@
...
+    @Test
+    void testHandleJsonNodeRequest_WithMethodModule() throws Exception {
+        JsonNode request = objectMapper.readTree(SECOND_METHOD_REQUEST);
+        Web3Test handler = mock(Web3Test.class);
+        jsonRpcCustomServer = new JsonRpcCustomServer(handler, Web3Test.class, getModulesWithMethodTimeout(0,125));
+        String response = "test_method_response";
+
+        when(handler.test_second(anyString(), anyString())).thenReturn(response);
+        when(handler.test_second(anyString(), anyString())).thenAnswer(invocation -> {
+            Thread.sleep(250);
```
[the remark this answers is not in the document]
asoto-iov: I preferred to use Thread.sleep for simplicity. It should be a problem if we use sleep in order to wait for something to happen, like a method to be completed or a response to be served, which is something that we can not guarantee, but in this case we are just waiting to force the timeout exception and I don't see any problem with it.  Also it is just a test and not in our codebase.

rskj-core/src/test/java/co/rsk/rpc/netty/JsonRpcCustomServerTest.java
```diff
@@ -0,0 +1,215 @@
...
+        JsonResponse actualResponse = jsonRpcCustomServer.handleJsonNodeRequest(request);
+        assertEquals(response, actualResponse.getResponse().get("result").asText());    }
+
+
+    @Test //The timeout is applied per method and not per request
+    void testHandleJsonNodeRequest_WithMethodTimeout_BatchRequest_OK() throws Exception {
+        long timeoutPerMethod = 550;
+        long sleepTimePerRequest = 300;
+        //The request has 2 methods, each one will sleep for 300ms, so the total time will be 600ms.
+        // The timeout per method is 550ms, so the request will be executed successfully.
```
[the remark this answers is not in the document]
asoto-iov: I didn't though about it, any suggestion or alternative? 🤔

rskj-core/src/test/java/co/rsk/rpc/netty/JsonRpcCustomServerTest.java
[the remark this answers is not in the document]
asoto-iov: To avoid extra delays caused by thread management I'll implement your solution :)

rskj-core/src/main/java/co/rsk/rpc/netty/ResponseSizeLimitContext.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public void close() throws Exception {
+        accumulatedResponseSize.remove();
+    }
+
+    public static ResponseSizeLimitContext createResponseSizeContext(int limit) {
+        ResponseSizeLimitContext ctx = new ResponseSizeLimitContext(limit);
+        accumulatedResponseSize.set(ctx);
+        return ctx;
```
Vovchyk, line 44: do you think we could return some dummy value without setting it in `accumulatedResponseSize `, if limit is zero / disabled? eg.
```
    private static final ResponseSizeLimitContext DISABLED = new ResponseSizeLimitContext(0) {
        @Override
        public void close() throws Exception {
            // do nothing
        }
    };
```

rskj-core/src/main/resources/config/testnet.conf
```diff
@@ -81,6 +81,11 @@ wallet {
     accounts = []
 }
 
+rpc {
```
Vovchyk, line 84: do we need this config in testnet?

rskj-core/src/main/resources/reference.conf
```diff
@@ -314,6 +314,9 @@ rpc {
     callGasCap = 50000000
     # (experimental, OFF by default) RPC timeout in milliseconds. Disabled when it's zero.
     timeout = 0
+    # Maximum Response size allowed in bytes. If value is less than 0 then there is no limit.
+    maxResponseSize = -1
```
Vovchyk, line 318: I guess we should be consistent with other configs (eg. timeout) and use zero value to disable the feature, wdyt? and probably do not allow negative values?..

rskj-core/src/main/java/co/rsk/rpc/exception/JsonRpcResponseLimitError.java
```diff
@@ -0,0 +1,12 @@
+package co.rsk.rpc.exception;
+
+public class JsonRpcResponseLimitError extends Error {
+    public static final int ERROR_CODE = -32011;
```
Vovchyk, line 4: do you think it'd make sense to define it in `JsonError` ?

rskj-core/src/main/java/co/rsk/rpc/exception/JsonRpcTimeoutError.java
```diff
@@ -0,0 +1,12 @@
+package co.rsk.rpc.exception;
+
+import static com.googlecode.jsonrpc4j.ErrorResolver.JsonError.INTERNAL_ERROR;
+
+public class JsonRpcTimeoutError extends Error {
+    public static final int ERROR_CODE = INTERNAL_ERROR.code;
```
Vovchyk, line 6: I guess it'd make sense to create a custom error code for this too, wdyt? same as for `JsonRpcResponseLimitError`

rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcWeb3ServerProperties.java
```diff
@@ -0,0 +1,91 @@
...
+    private List<ModuleDescription> rpcModules;
+    private int rpcMaxResponseSize;
+    private long rpcTimeout;
+
+    //Default values
+    private JsonRpcWeb3ServerProperties() {
+        this.maxBatchRequestsSize = 100;
+        this.rpcModules = new ArrayList<>();
+        this.rpcMaxResponseSize = -1;
+        this.rpcTimeout = -1;
```
Vovchyk, line 19: I guess we should use zeros as defaults for the consistency reason

Vovchyk (changes requested):

rskj-core/src/main/resources/reference.conf
asoto-iov: Ok it has sense. I suppose no one would like to add a limit = 0 😓

rskj-core/src/main/java/co/rsk/rpc/exception/JsonRpcResponseLimitError.java
asoto-iov: Updated how is done. Now the error code is stored in JsonRpcError class, and each exception returns a JsonRpcError with the corresponding code.

asoto-iov:
pipeline:run

asoto-iov:
pipeline:run

rskj-core/src/main/java/co/rsk/rpc/netty/ResponseSizeLimitContext.java
```diff
@@ -0,0 +1,92 @@
...
+        accumulatedResponseSize.remove();
+    }
+
+    public static ResponseSizeLimitContext createResponseSizeContext(int limit) {
+        if (limit <= 0) {
+            return createEmptyContext();
+        }
+        ResponseSizeLimitContext existingContext = accumulatedResponseSize.get();
+        if(existingContext != null) {
+            throw new JsonRpcInternalError("ResponseSizeLimitContext already exists");
```
Vovchyk, line 67: I guess we could simply throw RuntimeException, or something derived from that. As I see, this can only happen if there's some bug

rskj-core/src/main/java/co/rsk/rpc/exception/JsonRpcThrowableError.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package co.rsk.rpc.exception;
+
+import co.rsk.jsonrpc.JsonRpcError;
+
+public abstract class JsonRpcThrowableError extends Error {
```
Vovchyk, line 23: just an idea, `JsonRpcLimitError` ? - to emphasise that it's limit related

rskj-core/src/main/java/co/rsk/jsonrpc/JsonRpcError.java
```diff
@@ -28,6 +29,8 @@ public class JsonRpcError implements JsonRpcResultOrError {
 	public static final int METHOD_NOT_FOUND = -32601;
 	public static final int INVALID_PARAMS = -32602;
 	public static final int INTERNAL_ERROR = -32603;
+	public static final int RESPONSE_LIMIT_ERROR = -32011;
```
Vovchyk, line 32:
```suggestion
	public static final int RPC_LIMIT_ERROR = -32011;
```

... as this is used for both JsonRpcResponseLimitError & JsonRpcTimeoutError?

rskj-core/src/main/resources/config/testnet.conf
Vovchyk: so can we remove it? not sure why you added it, the config is `testnet` specific

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,116 @@
...
+        this.max = max;
+        this.root = root;
+        this.objectMapper = new ObjectMapper();
+    }
+
+    public static int getSizeInBytesWithLimit(JsonNode node, int limit) {
+        if (node == null) {
+            return 0;
+        }
+        JsonResponseSizeLimiter limiter = new JsonResponseSizeLimiter(limit, node);
```
Vovchyk, line 52: ideally, the JsonResponseSizeLimiter should be as lightweight as possible. Meaning that no new objects are being created on the way. Eg. `objectMapper.writeValueAsBytes(node)` makes a new object in heap and basically duplicates that. Imagine that we have a result like this: `"result": "0x<10mb of data>"`, then `objectMapper.writeValueAsBytes()` would generate extra 10mb of data in ram.

I guess we could approximately calculate size of a node object. For instance, at runtime we could check if value is a string and simply add its length to a sum. Same for numbers and bools. In JSON, values must be one of the following data types:
```
a string
a number
an object (JSON object)
an array
a boolean
null
```
.. and for composite types like arrays/objects this can be done recursively. Obviously, this would be an approximated size, but at the very end when a response is "flushed" to OutputStream you can make a final "strict" check.

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
asoto-iov: It is only happening with fields so I though it was not a big deal and I preferred to make it simple and not to repeat the logic of check each type etc, but I will review it.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,107 @@
...
+            handleArray((ArrayNode) jsonNode);
+        } else if (jsonNode.isValueNode()) {
+            handleValueNode(jsonNode);
+        } else {
+            handleObject(jsonNode);
+        }
+    }
+
+    private void handleValueNode(JsonNode node) {
+        sumAndCheck(node.toString().getBytes().length);
```
Vovchyk, line 67:
```suggestion
        sumAndCheck(node.asText().length());
```
`JsonNode.toString()` involves extra memory allocation by using `ObjectWriter`, I'd rather use `asText()` one. Also `.getBytes()` allocates extra copy of it, probably `String.length()` should be sufficient, wdyt?

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,107 @@
...
+        sumAndCheck(BRACKETS_SIZE_IN_BYTES + (COMMA_SIZE * fieldSeparatorCount));
+    }
+
+    private void handleField(Map.Entry<String, JsonNode> field) {
+        JsonNode value = field.getValue();
+        if (value.isObject() || value.isArray()) {
+            addLength(value);
+            sumAndCheck(field.getKey().getBytes().length + JSON_FIELD_MISSING_SYMBOLS);
+        } else {
+            sumAndCheck(field.getValue().toString().getBytes().length + field.getKey().getBytes().length + JSON_FIELD_MISSING_SYMBOLS);
```
Vovchyk, line 96:
```suggestion
            sumAndCheck(field.getValue().asText().length() + field.getKey().length() + JSON_FIELD_MISSING_SYMBOLS);
```
same here

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private void addLength(JsonNode jsonNode) {
+        if (jsonNode.isArray()) {
+            handleArray((ArrayNode) jsonNode);
+        } else if (jsonNode.isValueNode()) {
+            handleValueNode(jsonNode);
+        } else {
+            handleObject(jsonNode);
+        }
```
Vovchyk, line 63:
```suggestion
        } else if (jsonNode.isObject()) {
            handleObject(jsonNode);
        } else if (jsonNode.isValueNode()) {
            handleValueNode(jsonNode);
        }
```

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/rpc/exception/JsonRpcTimeoutError.java
```diff
@@ -0,0 +1,35 @@
...
+public class JsonRpcTimeoutError extends JsonRpcThrowableError {
+    private static final long serialVersionUID = 950001095761882084L;
+
+    public JsonRpcTimeoutError(String  msg) {
+        super(msg);
+    }
+
+    @Override
+    public JsonRpcError getErrorResponse() {
+        return new JsonRpcError(JsonRpcError.RPC_LIMIT_ERROR, getMessage());
```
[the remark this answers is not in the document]
asoto-iov: I think it could be useful for the user to know the configured limit when using the API.

rskj-core/src/main/java/co/rsk/rpc/json/JsonResponseSizeLimiter.java
asoto-iov: asText() doesn't have the same result as toString(),  using it will require extra checks. 
Using `String.length()` instead of `getBytes().length` will work for ASCII characters which will be the most common so it can work.

rskj-core/src/main/java/co/rsk/rpc/netty/ResponseSizeLimitContext.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    private void add(JsonNode response) {
+        if (limit <= 0) {
+            return;
+        }
+
+        if (response != null) {
+            add(JsonResponseSizeLimiter.getSizeInBytesWithLimit(response, limit));
```
Vovchyk, line 53: it looks like that both `ResponseSizeLimitContext` and `JsonResponseSizeLimiter` accumulate response size. Can we do this only in one place? Eg. `ResponseSizeLimitContext` could do accumulation work and `JsonResponseSizeLimiter` (what about calling it something like `JsonResponseUtils`?) could simply return an approximated size?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/rpc/netty/ResponseSizeLimitContext.java
asoto-iov: The `JsonResponseSizeLimiter` only stores the size during the calculation and then the result is being accumulated into the `ResponseSizeLimitContext`.

rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcCustomServer.java
```diff
@@ -47,35 +47,43 @@ protected JsonResponse handleJsonNodeRequest(final JsonNode node) throws JsonPar
...
+                .findFirst()
+                .orElse(null);
+        if(moduleDescription == null) {
+            return 0;
         }
-
-        try (ExecTimeoutContext ignored = ExecTimeoutContext.create(timeout)) {
-            response = super.handleJsonNodeRequest(node);
-            ExecTimeoutContext.checkIfExpired();
+        Long methodTimeout = moduleDescription.getMethodTimeout(methodName);
```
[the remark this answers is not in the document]
asoto-iov: huh, good point. Not sure the reason, but as I update the logic in some points, just didn't realize about the `getTimeout(String) ` method and actually I'm repeating part of the logic. I'll update it.

rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcWeb3ServerHandler.java
```diff
@@ -54,26 +52,26 @@ public class JsonRpcWeb3ServerHandler extends SimpleChannelInboundHandler<ByteBu
...
         jsonRpcServer.setErrorResolver(new MultipleErrorResolver(new RskErrorResolver(), AnnotationsErrorResolver.INSTANCE, DefaultErrorResolver.INSTANCE));
-
-        this.defaultTimeout = rskSystemProperties.getRpcTimeout();
+        this.defaultTimeout = jsonRpcWeb3ServerProperties.getRpcTimeout();
+        this.maxResponseSize = jsonRpcWeb3ServerProperties.getRpcMaxResponseSize();
     }
 
     @Override
     protected void channelRead0(ChannelHandlerContext ctx, ByteBufHolder request) throws Exception {
         ByteBuf responseContent = Unpooled.buffer();
```
Vovchyk, line 70: we might need to make a "final" check of a response size, as response limiter makes an approximate calcs. I believe we could make use of `responseContent.readableBytes()` (similar as it is used [here](https://github.com/rsksmart/rskj/blob/91af660633a23660c58b7146e1da2498a3d8bfbf/rskj-core/src/main/java/co/rsk/rpc/netty/Web3ResultHttpResponseHandler.java#L52)), wdyt?

Vovchyk (changes requested):

Vovchyk (dismissed):

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk (approved):

PR #1998, opened 2023-03-24
Allow specifying VM options when running RSKj in docker container

rmoreliovlabs:
## Description
Allow JAVA_OPTS env variable when running docker container.

## Motivation and Context
In the rskj repo we have a GitHub workflow that builds and pushes a docker image to DockerHub. Here’s a docker [file](https://github.com/rsksmart/rskj/blob/9c14a23e49de1b4729c8ab226c9e68445a8437a3/Dockerfile#L31-L36) for that

As of now, only a specific set of sys args is specified in the entrypoint statement and there’s no way to override or add extra vm args, eg. -Xmx or -Xss.

The idea is to define an env. var. in the Dockerfile, which would allow us to do that.

## How Has This Been Tested?
Running the follwoing commands:
```
docker build -t rskj-node -f Dockerfile . 
docker run --name rskj-node -e DEFAULT_JVM_OPTS="-Xms3G -Xmx5G" rskj-node
```

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,16 +28,14 @@ USER rsk
 WORKDIR /var/lib/rsk
 COPY --from=build --chown=rsk:rsk /home/rsk/rsk.jar ./
 
-ENV JAVA_OPTS=""
+ENV DEFAULT_JVM_OPTS="-Xms3G -Xmx5G"
+ENV RSKJ_SYS_PROPS="-Drpc.providers.web.http.bind_address=0.0.0.0 -Drpc.providers.web.http.hosts.0=localhost -Drpc.providers.web.http.hosts.1=127.0.0.1 -Drpc.providers.web.http.hosts.2=::1"
 ENV RSKJ_CLASS=co.rsk.Start
 ENV RSKJ_OPTS=""
 
 ENTRYPOINT java \
-    $JAVA_OPTS \
-    -Drpc.providers.web.http.bind_address=0.0.0.0 \
-    -Drpc.providers.web.http.hosts.0=localhost \
-    -Drpc.providers.web.http.hosts.1=127.0.0.1 \
-    -Drpc.providers.web.http.hosts.2=::1 \
+    $DEFAULT_JVM_OPTS \
+    $RSKJ_SYS_PROPS \
     -cp rsk.jar \
     $RSKJ_CLASS \
     $RSKJ_OPTS
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ USER rsk
 WORKDIR /var/lib/rsk
 COPY --from=build --chown=rsk:rsk /home/rsk/rsk.jar ./
 
-ENV DEFAULT_JVM_OPTS="-Xms3G -Xmx5G"
+ENV DEFAULT_JVM_OPTS="-Xms4G"
 ENV RSKJ_SYS_PROPS="-Drpc.providers.web.http.bind_address=0.0.0.0 -Drpc.providers.web.http.hosts.0=localhost -Drpc.providers.web.http.hosts.1=127.0.0.1 -Drpc.providers.web.http.hosts.2=::1"
 ENV RSKJ_CLASS=co.rsk.Start
 ENV RSKJ_OPTS=""
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,9 +28,16 @@ USER rsk
 WORKDIR /var/lib/rsk
 COPY --from=build --chown=rsk:rsk /home/rsk/rsk.jar ./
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -28,9 +28,16 @@ USER rsk
...
+
+ENTRYPOINT java \
+    $JAVA_OPTS \
+    -Drpc.providers.web.http.bind_address=0.0.0.0 \
+    -Drpc.providers.web.http.hosts.0=localhost \
+    -Drpc.providers.web.http.hosts.1=127.0.0.1 \
+    -Drpc.providers.web.http.hosts.2=::1 \
+    -cp rsk.jar \
+    $RSKJ_CLASS \
+    $RSKJ_OPTS
```
Vovchyk, line 43: I'd rather use `DEFAULT_JVM_OPTS` name like [here](https://github.com/rsksmart/rskj/blob/6ebaacfaa486b5086b5f2a3e70adb4cb0057675f/gradlew#L89), as `JAVA_OPTS` seems to be a semi-standard jvm env. variable name and may collide in some scenarious.

Also I believe that the default amount of ram being provided to a java process may not be enough for an RSKj node to operate - so what about defining it something like this?

```suggestion
ENV DEFAULT_JVM_OPTS='"-Xms3G" "-Xmx5G"'
ENV RSKJ_SYS_PROPS='"-Drpc.providers.web.http.bind_address=0.0.0.0" "-Drpc.providers.web.http.hosts.0=localhost" "-Drpc.providers.web.http.hosts.1=127.0.0.1" "-Drpc.providers.web.http.hosts.2=::1"'
ENV RSKJ_CLASS=co.rsk.Start
ENV RSKJ_OPTS=""

ENTRYPOINT java \
    $DEFAULT_JVM_OPTS \
    $RSKJ_SYS_PROPS \
    -cp rsk.jar \
    $RSKJ_CLASS \
    $RSKJ_OPTS
```

note: I haven't tried that - it's just pseudocode, so needs to be tested

Vovchyk (changes requested):

Dockerfile
rmoreliovlabs: Addressed

Dockerfile
```diff
@@ -28,9 +28,14 @@ USER rsk
...
+ENV RSKJ_SYS_PROPS="-Drpc.providers.web.http.bind_address=0.0.0.0 -Drpc.providers.web.http.hosts.0=localhost -Drpc.providers.web.http.hosts.1=127.0.0.1 -Drpc.providers.web.http.hosts.2=::1"
+ENV RSKJ_CLASS=co.rsk.Start
+ENV RSKJ_OPTS=""
+
+ENTRYPOINT java \
+    $DEFAULT_JVM_OPTS \
+    $RSKJ_SYS_PROPS \
+    -cp rsk.jar \
+    $RSKJ_CLASS \
+    $RSKJ_OPTS
```
Vovchyk, line 41: one problem I've noticed with this setup is that it doesn't allow to specify a CMD/extra flags, eg. the following command ignores `--reset` flag:
```bash
docker run -d --rm -e RSKJ_OPTS="--regtest" rskj-node --reset
```
It seems that the following trick allows to work that around:
```
ENTRYPOINT ["/bin/sh", "-c", "java $DEFAULT_JVM_OPTS $RSKJ_SYS_PROPS -cp rsk.jar $RSKJ_CLASS $RSKJ_OPTS \"${@}\"", "--"]
```
Let me know what you think on that

rmoreliovlabs: Done

Vovchyk (dismissed):

Dockerfile
```diff
@@ -28,9 +28,10 @@ USER rsk
 WORKDIR /var/lib/rsk
 COPY --from=build --chown=rsk:rsk /home/rsk/rsk.jar ./
 
-ENTRYPOINT ["java", \
-    "-Drpc.providers.web.http.bind_address=0.0.0.0", \
-    "-Drpc.providers.web.http.hosts.0=localhost", \
-    "-Drpc.providers.web.http.hosts.1=127.0.0.1", \
-    "-Drpc.providers.web.http.hosts.2=::1", \
-    "-cp", "rsk.jar", "co.rsk.Start"]
+ENV DEFAULT_JVM_OPTS="-Xms3G -Xmx5G"
```
Vovchyk, line 31:
```suggestion
ENV DEFAULT_JVM_OPTS="-Xms4G"
```

Vovchyk (changes requested):

Vovchyk (approved):

rmoreliovlabs:
pipeline:run

Vovchyk (approved):

PR #2005, opened 2023-03-31
Changes in RSKIP144 activation config verification

julianlen:
Apart from testing, when RSKIP144 is deactivated, the txExecutionSubl…istsEdges is null, otherwise is initially an empty array or it has the correct content.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionSublistsEdgesRule.java
+++ b/rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionSublistsEdgesRule.java
@@ -1,6 +1,8 @@
 package co.rsk.validators;
 
 import org.ethereum.config.Constants;
+import org.ethereum.config.blockchain.upgrades.ActivationConfig;
+import org.ethereum.config.blockchain.upgrades.ConsensusRule;
 import org.ethereum.core.Block;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionSublistsEdgesRule.java
```diff
@@ -15,12 +17,22 @@
...
+
+    public ValidTxExecutionSublistsEdgesRule(ActivationConfig activationConfig) {
+        this.activationConfig = activationConfig;
+    }
+
 
     @Override
     public boolean isValid(Block block) {
+        if (!activationConfig.isActive(ConsensusRule.RSKIP144, block.getHeader().getNumber())) {
+            return true;
```
ilanolkies, line 30: Even if execution is done by checking activation config I would add

```suggestion
            return edges == null;
```

ilanolkies (changes requested):

rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionSublistsEdgesRule.java
julianlen: Not needed, it doesn't matter what comes in the txExecutionSublistsEdge because before RSKIP144, the code shouldn't take it into account.

javicid (approved):

PR #2014, opened 2023-04-11
RemascConfigFactory refactoring

rmoreliovlabs:
## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
+++ b/rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
@@ -20,21 +20,12 @@ package co.rsk.config;
 
 import co.rsk.remasc.RemascException;
 import co.rsk.util.JacksonParserUtil;
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.BufferedReader;
 import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.nio.charset.StandardCharsets;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Optional;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 
 /**
  * Created by mario on 12/12/16.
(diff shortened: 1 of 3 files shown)


rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
```diff
@@ -44,9 +48,24 @@ public RemascConfigFactory(String remascConfigFile) {
     public RemascConfig createRemascConfig(String config) {
         RemascConfig remascConfig;
 
+        logger.info("createRemascConfig config = {} ", config);
+        logger.info("createRemascConfig configPath = {} ", configPath);
+
         try (InputStream is = RemascConfigFactory.class.getClassLoader().getResourceAsStream(this.configPath)) {
-            JsonNode node = JacksonParserUtil.readTree(mapper, is);
+            String sis = new BufferedReader(new InputStreamReader(is, StandardCharsets.UTF_8))
```
Vovchyk, line 55: the config is constant, isn't it? I don't see any reason to read it from `resources` on each `createRemascConfig()` method call. Can this be done once in the ctor?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
rmoreliovlabs: This code is just to add more logs so I can debug better using grafana

rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
```diff
@@ -34,19 +38,39 @@ public class RemascConfigFactory {
...
 
-        try (InputStream is = RemascConfigFactory.class.getClassLoader().getResourceAsStream(this.configPath)) {
-            JsonNode node = JacksonParserUtil.readTree(mapper, is);
-            remascConfig = JacksonParserUtil.treeToValue(mapper, node.get(config), RemascConfig.class);
+        logger.info("createRemascConfig config = {} ", config);
+
+        try {
+            logger.info("createRemascConfig node.get(config) = {} ", configNode.get(config));
+
+            remascConfig = JacksonParserUtil.treeToValue(mapper, configNode.get(config), RemascConfig.class);
```
Vovchyk, line 71: It seems that `RemascConfig` is immutable and permanent. What about parsing this only once and storing results, for instance, in `ConcurrentHashMap<String, RemascConfig>` for later usage?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
rmoreliovlabs: Addressed

rmoreliovlabs:
pipeline:run

rmoreliovlabs:
pipeline:run

rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
```diff
@@ -34,24 +42,57 @@ public class RemascConfigFactory {
     private static final Logger logger = LoggerFactory.getLogger("RemascConfigFactory");
 
     private ObjectMapper mapper;
-    private String configPath;
+    private final Map<String, RemascConfig> remascConfigMap = new ConcurrentHashMap<>();
 
     public RemascConfigFactory(String remascConfigFile) {
```
Vovchyk, line 47: the initial idea was to fail earlier, and it seems that in the current style of its usage `RemascConfigFactory` is instantiated for each call of `getContractForAddress()`. See `org.ethereum.vm.PrecompiledContracts#getContractForAddress`. 

Do you think we could create an instance of `RemascConfigFactory` in the `PrecompiledContracts`'s ctor?

rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
```diff
@@ -34,24 +42,57 @@ public class RemascConfigFactory {
...
         this.mapper = new ObjectMapper();
-        this.configPath = remascConfigFile;
+        logger.info("RemascConfigFactory remascConfigFile = {} ", remascConfigFile);
+
+        try (InputStream is = RemascConfigFactory.class.getClassLoader().getResourceAsStream(remascConfigFile)) {
+            String sis = new BufferedReader(new InputStreamReader(is, StandardCharsets.UTF_8))
+                    .lines()
+                    .collect(Collectors.joining("\n"));
+
+            logger.info("RemascConfigFactory sis = {} ", sis);
```
Vovchyk, line 56: Do we need this new precise logging? Before we were logging only one error.

rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
```diff
@@ -34,24 +42,57 @@ public class RemascConfigFactory {
...
-            remascConfig = JacksonParserUtil.treeToValue(mapper, node.get(config), RemascConfig.class);
+        logger.info("fetchRemascConfigMap config = {} ", config);
+
+        try {
+            logger.info("fetchRemascConfigMap node.get(config) = {} ", configNode.get(config));
+
+            remascConfig = JacksonParserUtil.treeToValue(mapper, configNode.get(config), RemascConfig.class);
+
+            logger.info("fetchRemascConfigMap remascConfig = {} ", remascConfig);
         } catch (Exception ex) {
```
Vovchyk, line 86: do we really need to catch this exception in this private method? It's being done at the upper level too.

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
rmoreliovlabs: Addessed

rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
rmoreliovlabs: I want to log as much as possible to track the scenario where the UnzipException occurred. If it does not happen for some time, we can remove these logs and assume this refactor fixes it, but yeah we can just remove it if you think this will not happen again after this change.

rskj-core/src/main/java/co/rsk/config/RemascConfigFactory.java
rmoreliovlabs: Addressed

Vovchyk (approved):

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

PR #2017, opened 2023-04-14
Store standard redeemscript

nathanieliov:


rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -1295,6 +1295,358 @@ void testTxIsProcessableInLegacyVersion() {
...
     }
 
+    @Test
+    void testIsMigrationTx_sending_funds_from_retired_p2sh_fed_to_active_p2sh_fed() {
+        NetworkParameters networkParameters = bridgeConstantsMainnet.getBtcParams();
+        Context btcContext = new Context(networkParameters);
+        when(activations.isActive(ConsensusRule.RSKIP201)).thenReturn(true);
+
+        List<BtcECKey> retiredFederationKeys = Stream.of(
+            BtcECKey.fromPrivate(Hex.decode("fc01")),
```
jeremy-then, line 1305: Isn't it better to actually use a random hex string with the length of an actual private key? I don't think 'fc01' is a good example of a private key here.

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -1295,6 +1295,358 @@ void testTxIsProcessableInLegacyVersion() {
...
+            1L,
+            bridgeConstantsMainnet.getBtcParams(),
+            bridgeConstantsMainnet.getErpFedPubKeysList(),
+            bridgeConstantsMainnet.getErpFedActivationDelay(),
+            activations
+        );
+
+        List<BtcECKey> activeFederationKeys = Stream.of(
+            BtcECKey.fromPrivate(Hex.decode("fa01")),
+            BtcECKey.fromPrivate(Hex.decode("fa02"))
```
jeremy-then, line 1321: The same here, an actual private key-like hex would be better. Or is this some kind of a seed?

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -1295,6 +1295,358 @@ void testTxIsProcessableInLegacyVersion() {
...
```
jeremy-then, line 1391: Same as above.

jeremy-then (approved):
LGTM.

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
nathanieliov: I would say yes if this test were related to testing something with signing/or with signature. But, this is not the case. It's just testing that for a given btc transaction sending funds to the fed standardP2shScript of each type of Federation, the tx is processed correctly by the bridge.

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
nathanieliov: Same answered.

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -1295,6 +1295,358 @@ void testTxIsProcessableInLegacyVersion() {
...
+        TransactionInput migrationTxInput = new TransactionInput(
+            networkParameters,
+            migrationTx,
+            new byte[]{},
+            new TransactionOutPoint(networkParameters, 0, Sha256Hash.ZERO_HASH)
+        );
+        migrationTx.addInput(migrationTxInput);
+        signWithNecessaryKeys(retiredFederation, retiredFederationKeys, migrationTxInput, migrationTx);
+
+        assertTrue(BridgeUtils.isMigrationTx(
```
marcos-iov, line 1474: We should also assert that it works with the P2SH script, not just the standard one

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -1295,6 +1295,358 @@ void testTxIsProcessableInLegacyVersion() {
...
+            new TransactionOutPoint(networkParameters, 0, Sha256Hash.ZERO_HASH)
+        );
+        migrationTx.addInput(migrationTxInput);
+        signWithNecessaryKeys(retiringFederation, retiringFederationKeys, migrationTxInput, migrationTx);
+
+        assertTrue(BridgeUtils.isMigrationTx(
+            migrationTx,
+            activeFederation,
+            null,
+            retiringFederation.getStandardP2SHScript(),
```
marcos-iov, line 1535:
```suggestion
            retiringFederation,
            null,
```
You should be passing the retiring federation here, not the p2sh script

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -1295,6 +1295,358 @@ void testTxIsProcessableInLegacyVersion() {
...
+            new TransactionOutPoint(networkParameters, 0, Sha256Hash.ZERO_HASH)
+        );
+        migrationTx.addInput(migrationTxInput);
+        signWithNecessaryKeys(retiringFederation, retiringFederationKeys, migrationTxInput, migrationTx);
+
+        assertTrue(BridgeUtils.isMigrationTx(
+            migrationTx,
+            activeFederation,
+            retiringFederation,
+            null,
```
marcos-iov, line 1643: Same as above

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigsForTest.java
```diff
@@ -160,6 +160,17 @@ private static List<ConsensusRule> getHop401Rskips() {
         return rskips;
     }
 
+    private static List<ConsensusRule> getFingerroot500Rskips() {
+        List<ConsensusRule> rskips = new ArrayList<>();
+        rskips.addAll(Arrays.asList(
+            ConsensusRule.RSKIP374,
+            ConsensusRule.RSKIP375,
+            ConsensusRule.RSKIP377
```
marcos-iov, line 168: Are those all the RSKIPs we have in fingerroot?

rskj-core/src/test/java/co/rsk/peg/PowpegMigrationTest.java
```diff
@@ -1403,6 +1396,54 @@ void test_change_powpeg_from_p2shErpFederation_with_mainnet_powpeg_post_RSKIP_35
         );
     }
 
+    @Test
+    void test_change_powpeg_from_p2shErpFederation_with_mainnet_powpeg_post_RSKIP_377_stores_standard_redeemScript() throws Exception {
```
marcos-iov, line 1400: Seems like we are lacking an assertion that the standard p2sh redeem script was stored as expected.

Perhaps getting the value from the storage should be enough to validate it.

We should also check that pre-fingerrot, the non-standard p2sh script is stored. Perhaps this last assertion can be added as part of the other tests, so it's not necessary to create a new test just for this

marcos-iov (changes requested):
Nice job, left a few comments regarding the tests

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
nathanieliov: Done.

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigsForTest.java
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/PowpegMigrationTest.java
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
marcos-iov: That is correct, my bad

rskj-core/src/test/java/co/rsk/peg/PowpegMigrationTest.java
```diff
@@ -543,6 +545,29 @@ private void testChangePowpeg(
             false,
             true
         );
+
+        if (oldPowPegFederationType == FederationType.erp || newPowPegFederationType == FederationType.erp){
+            return;
+        }
```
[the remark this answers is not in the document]
nathanieliov: Changed

marcos-iov (approved):

PR #2025, opened 2023-04-24
Rskip351+rskip144+gaslimit

julianlen:
Changes in the gas limit logic include:
- Each tx is checked whether its gas limit + sequentialsublist.gasUsed < sequentialSublist.gasLimit instead of using the block gas used and gas limit.
- The sublist gas limit is block.gasLimit/ threads+1 (si gasUsed is always below gas limit)
- All the tests were adapted.


rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -76,7 +77,7 @@ public class TransactionExecutor {
     private final PrecompiledContracts precompiledContracts;
     private final boolean enableRemasc;
     private String executionError = "";
-    private final long gasUsedInTheBlock;
+    private final long gasUsedInTheContainer;
```
javicid, line 80: Maybe name it just `gasUsed`?

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -173,9 +163,9 @@ private boolean init() {
         }
 
         long txGasLimit = GasCost.toGas(tx.getGasLimit());
-        long curBlockGasLimit = GasCost.toGas(executionBlock.getGasLimit());
+        long containerGasLimit = activations.isActive(RSKIP144)? sublistGasLimit : GasCost.toGas(executionBlock.getGasLimit());
```
javicid, line 166: `gasLimit`?

javicid (changes requested):

javicid (approved):

PR #2028, opened 2023-05-02
Increase the new federation activation age to 40,320

nathanieliov:
Since RSKIP383 after a commit_federation event is emitted we expect the new federation to activate after 40,320 blocks instead of the current 18,500 blocks.

Changed made:

- Created new federationActivationAgeLegacy bridge constants variable and set it with the old value 18,500.
- Set federationActivationAge bridge constants with the new value: 40320.
- Refactored method BridgeConstants.getFederationActivationAge.
- Added new tests and updated existence tests based on the changes.

*fed:FINGERROOT-5.0.0-rc*
*fit:FINGERROOT-5.0.0-rc*
*ci:FINGERROOT-5.0.0-rc*


marcos-iov:
pipeline:run

marcos-iov:
Moved to https://github.com/rsksmart/rskj/pull/2037

PR #2045, opened 2023-05-23
Using temp directories in unit tests instead of user home directory

Vovchyk:
## Description
This PR updates unit tests in a way that user's home directory is not being used in tests anymore

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/test/java/co/rsk/NodeRunnerSmokeTest.java
+++ b/rskj-core/src/test/java/co/rsk/NodeRunnerSmokeTest.java
@@ -20,44 +20,56 @@ package co.rsk;
 
 import org.ethereum.util.RskTestContext;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.nio.file.Path;
+
 import static org.hamcrest.MatcherAssert.assertThat;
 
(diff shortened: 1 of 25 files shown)


asoto-iov (dismissed):
In general everything is ok for me. Just a possible suggestion, what do you think about implement the logic of the TempDir into the RskTestContext class?
Additionally e could use a factory or default constructor to create it and avoid others to use it in a wrong way.
Also we could create default factory methods for testnet, regtest, etc.

nagarev:
Nice! WDYT of github complaining about ignoring error statuses? maybe we could throw an exception if error is returned, it would be extremely rare tho.

nagarev (approved):
Nice! WDYT of github complaining about ignoring error statuses? maybe we could throw an exception if error is returned, it would be extremely rare tho.

Vovchyk:
> In general everything is ok for me. Just a possible suggestion, what do you think about implement the logic of the TempDir into the RskTestContext class? Additionally e could use a factory or default constructor to create it and avoid others to use it in a wrong way. Also we could create default factory methods for testnet, regtest, etc.

the thing here is that `@TempDir` has to be a part of a unit test class

nagarev (approved):
LGTM!

rmoreliovlabs (approved):

PR #2075, opened 2023-07-13
Add input to CallArguments class

rmoreliovlabs:
## Description

- Added `input` field to CallArguments
- Added new test for RPC methods that acccept a CallArguments object as argument.

## Motivation and Context
Following the investigation done in [CORE-2939](https://rsklabs.atlassian.net/browse/CORE-2939), regarding an issue with some of the RPC methods we use in rskj, we found that both Web3js and Geth now have a parameter called `input`  which currently is not supported by our code, this parameter seems to behave similar to the `data` parameter given that Web3js automatically add `input` in the payload if it is missing add sets its value to a copy from `data`'s. 

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/config/Constants.java
+++ b/rskj-core/src/main/java/org/ethereum/config/Constants.java
@@ -39,7 +39,6 @@ public class Constants {
     public static final byte TESTNET_CHAIN_ID = (byte) 31;
     public static final byte DEVNET_CHAIN_ID = (byte) 32;
     public static final byte REGTEST_CHAIN_ID = (byte) 33;
-    public static final String TEST_DATA = "0x603d80600c6000396000f3007c01000000000000000000000000000000000000000000000000000000006000350463c6888fa18114602d57005b6007600435028060005260206000f3";
 
     private static final byte[] FALLBACKMINING_PUBKEY_0 = Hex.decode("041e2b148c024770e19c4f31db2233cac791583df95b4d14a5e9fd4b38dc8254b3048f937f169446b19d2eca40db1dd93fab34c0cd8a310afd6e6211f9a89e4bca");
     private static final byte[] FALLBACKMINING_PUBKEY_1 = Hex.decode("04b55031870df5de88bdb84f65bd1c6f8331c633e759caa5ac7cad3fa4f8a36791e995804bba1558ddcf330a67ff5bfa253fa1d8789735f97a97e849686527976e");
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/org/ethereum/config/Constants.java
```diff
@@ -39,6 +39,7 @@ public class Constants {
     public static final byte TESTNET_CHAIN_ID = (byte) 31;
     public static final byte DEVNET_CHAIN_ID = (byte) 32;
     public static final byte REGTEST_CHAIN_ID = (byte) 33;
+    public static final String TEST_DATA = "0x603d80600c6000396000f3007c01000000000000000000000000000000000000000000000000000000006000350463c6888fa18114602d57005b6007600435028060005260206000f3";
```
Vovchyk, line 42: it looks like this is tests specific field and only used in `EthModuleTest`. I'd rather move it to tests

rskj-core/src/main/java/org/ethereum/rpc/CallArguments.java
```diff
@@ -36,6 +38,7 @@ public class CallArguments {
     private String nonce;
     private String chainId;
     private String type; // ignore, see https://github.com/rsksmart/rskj/pull/1601
+    private String input;
```
Vovchyk, line 41: do we really need this extra field? Can we reuse `data` and have something like this?:
```
    public void setInput(String input) {
        this.data = input;
    }
```

rskj-core/src/test/java/co/rsk/rpc/modules/eth/EthModuleTest.java
```diff
@@ -353,4 +355,359 @@ void chainId() {
         );
         assertThat(eth.chainId(), is("0x21"));
     }
+
+    @Test
+    void whenExecuteCallWithDataParameter_callExecutorWithData() {
```
Vovchyk, line 360: nice coverage 👍 could we also test `CallArguments` as an unit? Eg as in here? - `web3_CallArguments_toString()`

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/config/Constants.java
rmoreliovlabs: Done, I moved it to the test file.

rskj-core/src/main/java/org/ethereum/rpc/CallArguments.java
rmoreliovlabs: True, I've removed it and implemented your suggested approach.

rskj-core/src/test/java/co/rsk/rpc/modules/eth/EthModuleTest.java
rmoreliovlabs: Done

Vovchyk (approved):

PR #2080, opened 2023-07-18
Adding missing blockNumber or blockID  argument to eth_estimateGas

asoto-iov:
Adding missing parameter to eth_estimateGas JsonRPC method. 
## Description

To continue working in the compatibility with Ethereum, we are adding the blockNumber parameter missing in the eth_estimateGas JsonRPC method following the rules defined in the Ethereum documentation. 

An example curl code containing the blockNumber as an argument like this:
`curl --location 'http://localhost:4444' \
--header 'Content-Type: application/json' \
--data '{
    "jsonrpc":"2.0",
    "method":"eth_estimateGas",
    "params":
    [{
        "from": "0xfc9f095353a1fef1ba5d2e7b4a16702791c26e8f",
        "to": "0xE7B8E91401bF4d1669f54Dc5f98109D7EfBC4EEa",
        "gas": "0x76c0",
        "gasPrice": "0x9184e72a000",
        "value": "0x9184e72a",
        "data": "0xd46e8dd67c5d32be8d46e8dd67c5d32be8058bb8eb970870f072445675058bb8eb970870f072445675"
    },"0x255"
    ],
    "id":1
}'`

is returning 
`{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32602,
        "message": "method parameters invalid"
    }
}`

Which is not compatible with the current Ethereum specification. This PR solves the issue

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -1065,6 +1066,30 @@ public boolean personal_lockAccount(String address) {
...
 
+    @Override
+    public String eth_estimateGas(CallArguments args) {
+        return eth_estimateGas(args, null);
+    }
+
+    @Override
+    public String eth_estimateGas(CallArguments args, String bnOrId) {
+        Block block = null;
+        if (null == bnOrId) {
```
rmoreliovlabs, line 1077: Nit: bnOrId == null

asoto-iov: Hey Reynold, what do you mean?

casiojapi: to switch to: "if (bnOrId == null)" because its easier to read like that

rmoreliovlabs: Yes, sorry, that was too vague.

rskj-core/src/main/java/org/ethereum/rpc/validation/BnTagOrNumberValidator.java
```diff
@@ -0,0 +1,29 @@
...
+import co.rsk.util.HexUtils;
+import org.ethereum.core.genesis.BlockTag;
+import org.ethereum.rpc.exception.RskJsonRpcRequestException;
+
+public class BnTagOrNumberValidator {
+    private BnTagOrNumberValidator() {
+
+    }
+
+    public static boolean isValid(String parameter) {
```
casiojapi, line 12: (just a suggestion) As we are not using return value false, can we make the function just void? Or return false, and handle the exception outside?

Vovchyk: yes, this is actually what sonar is complaining about

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -175,6 +175,12 @@ public String estimateGas(CallArguments args) {
         }
     }
 
+    public String estimateGas(CallArguments args) {
+        Block bestBlock = blockchain.getBestBlock();
+        return estimateGas(args, bestBlock);
+    }
```
Vovchyk, line 181: not in use?

rskj-core/src/main/java/co/rsk/rpc/Web3InformationRetriever.java
```diff
@@ -44,8 +44,6 @@
  */
 public class Web3InformationRetriever {
 
-    private static final String EARLIEST = "earliest";
-    private static final String LATEST = "latest";
     private static final String PENDING = "pending";
```
Vovchyk, line 47: `PENDING` constant can also be replaced with the new enum you introduced?

rskj-core/src/main/java/org/ethereum/core/genesis/BlockTag.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.ethereum.core.genesis;
+
+public enum BlockTag {
+    PENDING("pending"), LATEST("latest"), EARLIEST("earliest"), FINALIZED("finalized"), SAFE("safe");
+
+    private String tag;
```
Vovchyk, line 23:
```suggestion
    private final String tag;
```

rskj-core/src/main/java/org/ethereum/core/genesis/BlockTag.java
```diff
@@ -0,0 +1,45 @@
...
+
+    BlockTag(String tag) {
+        this.tag = tag;
+    }
+
+    public static BlockTag fromString(String text) {
```
Vovchyk, line 29:
```suggestion
    @Nullable
    public static BlockTag fromString(String text) {
```

rskj-core/src/main/java/org/ethereum/rpc/JsonRpcArgumentValidator.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.ethereum.rpc;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class JsonRpcArgumentValidator {
```
Vovchyk, line 23: not used?

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/rpc/JsonRpcArgumentValidator.java
asoto-iov: Forget to delete it

rskj-core/src/main/java/co/rsk/rpc/Web3InformationRetriever.java
asoto-iov: In this case this pending status was used with the transaction pool so I didn't want to mix it. is it ok to do it?

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
asoto-iov: It is being used in test, but I didn't removed it in case it could be used as a lib by powpeg or others and don't break compatibility

rskj-core/src/main/java/org/ethereum/rpc/validation/BnTagOrNumberValidator.java
asoto-iov: just wanted to do it that way for test to have a positive result but it could be switched to void

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
Vovchyk: seems it's not used in powpeg. also the int. tests will would fail

rskj-core/src/main/java/co/rsk/rpc/Web3InformationRetriever.java
Vovchyk: I guess so. As I see it's treated as a block tag, so why not?

Vovchyk:
pipeline:run

rskj-core/src/main/java/org/ethereum/rpc/validation/BnTagOrNumberValidator.java
```diff
@@ -0,0 +1,40 @@
...
+import co.rsk.util.HexUtils;
+import org.ethereum.core.genesis.BlockTag;
+import org.ethereum.rpc.exception.RskJsonRpcRequestException;
+
+public class BnTagOrNumberValidator {
+    private BnTagOrNumberValidator() {
+
+    }
+
+    public static void isValid(String parameter) {
```
Vovchyk, line 29:
```suggestion
    public static void validate(String parameter) {
```

rskj-core/src/main/java/org/ethereum/rpc/exception/RskJsonRpcRequestException.java
```diff
@@ -57,4 +57,12 @@ public static RskJsonRpcRequestException blockNotFound(String message) {
     public static RskJsonRpcRequestException stateNotFound(String message) {
         return new RskJsonRpcRequestException(-32600, message);
     }
+
+    public static RskJsonRpcRequestException unknownBLockException(){
```
Vovchyk, line 61:
```suggestion
    public static RskJsonRpcRequestException unknownBlockException() {
```

rskj-core/src/main/java/org/ethereum/rpc/exception/RskJsonRpcRequestException.java
```diff
@@ -57,4 +57,12 @@ public static RskJsonRpcRequestException blockNotFound(String message) {
...
+        return new RskJsonRpcRequestException(-39001, "Unknown block");
+    }
+
+    public static RskJsonRpcRequestException headerNotFound(){
```
Vovchyk, line 65: just to be consistent with `eth_call`, can we reuse `RskJsonRpcRequestException::blockNotFound()`?

rskj-core/src/main/java/org/ethereum/rpc/exception/RskJsonRpcRequestException.java
asoto-iov: uh didn't see that 😓

rskj-core/src/main/java/org/ethereum/rpc/exception/RskJsonRpcRequestException.java
asoto-iov: I just aligned the error codes with Geth, maybe we should check eth_call and verify it.

rskj-core/src/main/java/co/rsk/rpc/Web3InformationRetriever.java
asoto-iov: I though maybe there were "transaction tags" as well

rskj-core/src/main/java/org/ethereum/rpc/exception/RskJsonRpcRequestException.java
Vovchyk: well, I believe the problem wasn't with message itself but rather with the fact that internal server error was being returned. I guess it's fine if we make use / re-use the `eth_call`'s message

asoto-iov: they have different error codes, it is ok with that?

Vovchyk: I guess so, error code standardisation is out of scope, plus error codes themselves seem to be out of spec and client specific

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -150,15 +151,14 @@ public String call(CallArguments args, String bnOrId) {
         }
     }
 
-    public String estimateGas(CallArguments args) {
+    public String estimateGas(CallArguments args, @Nonnull Block block) {
```
Vovchyk, line 154: in general logic looks good to me, but the approach used doesn't look consistent with other module classes. Eg. `co.rsk.rpc.modules.eth.EthModule#call(CallArguments args, String bnOrId)` accepts "raw" string `bnOrId` value, or `co.rsk.rpc.modules.eth.EthModuleWallet#sign(String addr, String data)` also accepts `addr` in raw format. Parsing itself happens in a specific module implementation class.

And in your case the work is being done in the "facade" class - `Web3Impl`

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -175,6 +175,12 @@ public String estimateGas(CallArguments args) {
         }
     }
 
+    public String estimateGas(CallArguments args) {
```
Vovchyk, line 178: not used

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
asoto-iov: It has sense, code was updated!

Vovchyk (approved):

Vovchyk (approved):

PR #2084, opened 2023-07-25
Align error handling with GETH

rmoreliovlabs:
## Description

- Added checks to validate hex prefix (0x), odd length and data types.
- Added an invocation listener to JsonRpcWeb3ServerHandler to validate arguments data types.

## Motivation and Context
As of now, there’s a discrepancy between RSKj and Eth Spec on how failed RPC requests are being presented to end users. In some cases, RSKj node can return a successful result with null value whether Eth node returns an error object. And in other cases, Eth node may return some successful result while RSKj returns an error (for instance, due to invalid arg).

We need to aline RSKj behaviour with the Eth Spec.

Note: here’s the [link](https://www.notion.so/iovlabs/Assessment-of-JSON-RPC-compatibility-with-Ethereum-64aaf9d4299643aba5e77a7f952d37d2) with compatibility assesment.

Known problematic RPC methods:

eth_getBlockTransactionCountByHash

eth_getBlockByHash

eth_getTransactionByHash

eth_getTransactionByBlockHashAndIndex

eth_getUncleByBlockHashAndIndex

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/crypto/Keccak256.java
```diff
@@ -55,6 +55,11 @@ public String toJsonString() {
         return HexUtils.toUnformattedJsonHex(this.bytes);
     }
 
+    public String toHexString(boolean addPrefix) {
```
asoto-iov, line 58: If only it is going to be used on test, not sure if we should have this method here, maybe can be used wrongly. 
Also there is an method in HexUtils to add the prefix if needed `co.rsk.util.HexUtils.ToJsonHex(String param)`

rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcWeb3ServerHandlerInvocationListener.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.ethereum.rpc.exception.RskJsonRpcRequestException.invalidParamError;
+
+public class JsonRpcWeb3ServerHandlerInvocationListener implements InvocationListener {
+    private final Map<String, Boolean> methodsToCheckMap = new HashMap<>();
```
asoto-iov, line 14: What it is the intention of this map? It seems that all the items has the value set to true.

rskj-core/src/main/java/co/rsk/util/HexUtils.java
```diff
@@ -223,7 +236,7 @@ public static long jsonHexToLong(String x) {
      * if the parameter has the hex prefix 
      */
     public static boolean hasHexPrefix(final String data) {
-        return data != null && data.startsWith(HEX_PREFIX);
+        return data != null && data.regionMatches(true, 0, HEX_PREFIX, 0, 2);
```
asoto-iov, line 239: why this change is being done? I think the current implementation is doing what it should do.

rskj-core/src/main/java/co/rsk/util/HexUtils.java
```diff
@@ -354,7 +367,7 @@ public static byte[] decode(byte[] dataBytes) {
      */
     public static int jsonHexToInt(final String param) {
         if (!hasHexPrefix(param)) {
-            throw invalidParamError(INCORRECT_HEX_SYNTAX);
+            throw invalidParamError("invalid argument 1: json: cannot unmarshal hex string without 0x prefix into value of type int");
```
asoto-iov, line 370: Not sure about the error message. In this case you are supposing that the invalid argument will be always in the position 1 (I don't have the certainty of it). 

And as commented above I think we shouldn't return directly an RPC related error here.

rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -676,9 +680,19 @@ public BlockResultDTO eth_getBlockByNumber(String bnOrId, Boolean fullTransactio
 
     @Override
     public TransactionResultDTO eth_getTransactionByHash(String transactionHash) {
+        if(!HexUtils.hasHexPrefix(transactionHash)) {
```
asoto-iov, line 683: I think the aim of this task is to make rskj compatible with GETH, but if RSKJ were supporting hashes without the `0x` prefix, then we shouldn't be more restrictive now. and allow then.

rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -783,10 +796,25 @@ public TransactionReceiptDTO eth_getTransactionReceipt(String transactionHash) {
 
     @Override
     public BlockResultDTO eth_getUncleByBlockHashAndIndex(String blockHash, String uncleIdx) {
+        if(!HexUtils.hasHexPrefix(blockHash)) {
```
asoto-iov, line 799: same comment as above, we are making this more restrictive.

rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -783,10 +796,25 @@ public TransactionReceiptDTO eth_getTransactionReceipt(String transactionHash) {
...
+            throw invalidParamError("invalid argument 0: cannot unmarshal hex string without 0x prefix into hash");
+        }
+
         BlockResultDTO s = null;
 
         try {
-            Block block = blockchain.getBlockByHash(stringHexToByteArray(blockHash));
+            byte[] bhash = stringEvenHexToByteArray(blockHash);
```
asoto-iov, line 806: same comment as in the `eth_getTransactionByHash` method. 
Additionally,  the validation of a transaction hash would be the same for almost any method. So I think we should extract the code to a validator class and create related unit tests instead of repeat it.

asoto-iov:
I think we should add some unit tests related with the added validations.

rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
rmoreliovlabs: I see, got it. I did it like this because the ticket provided a document that shows this difference so I assumed they should make the same validation.

rskj-core/src/main/java/co/rsk/util/HexUtils.java
rmoreliovlabs: Similar to [this case](https://github.com/rsksmart/rskj/pull/2084#discussion_r1275370029), my aim was to make RSKJ do the same validations as GETH. In this case I saw that GETH is case insensitive when check for the `0x`, this change makes it so RSKJ behaves the same way.

rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
rmoreliovlabs: Makes sense, I'll refactor this PR and implement the validator class approach Max uses.

rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcWeb3ServerHandlerInvocationListener.java
rmoreliovlabs: My reasoning here is that there may be methods that do not require a check for its argument's data types, this map constains the methods that should be checked, so far I added the one I worked on for this ticket.

rskj-core/src/main/java/co/rsk/rpc/netty/JsonRpcWeb3ServerHandler.java
```diff
@@ -101,6 +103,11 @@ protected void channelRead0(ChannelHandlerContext ctx, ByteBufHolder request) th
             int errorCode = error.getCode();
             responseContent = buildErrorContent(errorCode, error.getMessage());
             responseCode = errorCode;
+        } catch (RskJsonRpcRequestException e) {
```
asoto-iov, line 106: what is the use case catch here? In most of then the exception is handled into the `JsonRPCBasicServer`

rmoreliovlabs: That will catch the errors thrown by the `JsonRpcWeb3ServerHandlerInvocationListener` we added above.

rmoreliovlabs:
pipeline:run

PR #2109, opened 2023-08-24
Add get sighash util method

nathanieliov:


rskj-core/src/main/java/co/rsk/peg/bitcoin/BitcoinUtils.java
```diff
@@ -0,0 +1,31 @@
...
+        }
+        TransactionInput txInput = btcTx.getInput(0);
+        Optional<Script> redeemScriptOptional = BridgeUtils.extractRedeemScriptFromInput(txInput);
+        if (!redeemScriptOptional.isPresent()) {
+            return Optional.empty();
+        }
+
+        return Optional.of(btcTx.hashForSignature(
+            0,
+            redeemScriptOptional.get(),
```
kelvinator07, line 26:
```suggestion
        Optional<Script> redeemScript = BridgeUtils.extractRedeemScriptFromInput(txInput);
        if (!redeemScript.isPresent()) {
            return Optional.empty();
        }

        return Optional.of(btcTx.hashForSignature(
            0,
            redeemScript.get(),
```

rskj-core/src/test/java/co/rsk/peg/PegTestUtils.java
```diff
@@ -92,15 +92,13 @@ public static Sha256Hash createHash(int nHash) {
...
-        RedeemData redeemData = RedeemData.of(federation.getBtcPublicKeys(), redeemScript);
+        RedeemData redeemData = RedeemData.of(federation.getBtcPublicKeys(), federation.getRedeemScript());
         Script inputScript = scriptPubKey.createEmptyInputScript(redeemData.keys.get(0), redeemData.redeemScript);
         return inputScript;
     }
 
     public static Script createBaseRedeemScriptThatSpendsFromTheFederation(Federation federation) {
-        Script redeemScript = ScriptBuilder.createRedeemScript(federation.getNumberOfSignaturesRequired(), federation.getBtcPublicKeys());
-        return redeemScript;
+        return federation.getRedeemScript();
```
kelvinator07, line 101: It Seems This Method Isn't Doing Much With Your New Update, I'll Suggest We Remove It And Update Its Usages To `federation.getRedeemScript();` ?

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+        BtcTransaction btcTx = new BtcTransaction(btcMainnetParams);
+
+        for (int i = 1; i <= 50; i++) {
+            btcTx.addInput(BitcoinTestUtils.createHash(1), 0, new Script(new byte[]{}));
+            RedeemData redeemData = RedeemData.of(pubKeys, redeemScript);
+
+            Script inputScript = p2SHScript.createEmptyInputScript(redeemData.keys.get(0), redeemData.redeemScript);
+            btcTx.getInput(i-1).setScriptSig(inputScript);
+        }
```
kelvinator07, line 277:
```suggestion
           RedeemData redeemData = RedeemData.of(pubKeys, redeemScript);

            Script inputScript = p2SHScript.createEmptyInputScript(redeemData.keys.get(0), redeemData.redeemScript);
        for (int i = 1; i <= 50; i++) {
            btcTx.addInput(BitcoinTestUtils.createHash(1), 0, new Script(new byte[]{}));
            btcTx.getInput(i - 1).setScriptSig(inputScript);
        }
```

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+    @Test
+    void test_getFirstInputSigHash_from_fed() {
+        // Arrange
+        List<BtcECKey> signers = Arrays.asList(
+            BtcECKey.fromPrivate(Hex.decode("fa01")),
+            BtcECKey.fromPrivate(Hex.decode("fa02")),
+            BtcECKey.fromPrivate(Hex.decode("fa03"))
+        );
+        signers.sort(BtcECKey.PUBKEY_COMPARATOR);
```
kelvinator07, line 87: Lots Of Tests Are Using This, So We Can Put This Into An @Before Method

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+        Federation federation = new Federation(
+            FederationTestUtils.getFederationMembersWithBtcKeys(signers),
+            Instant.ofEpochMilli(1000L),
+            0L,
+            btcMainnetParams
+        );
+
+        Address destinationAddress = BitcoinTestUtils.createP2PKHAddress(btcMainnetParams, "destinationAddress");
```
kelvinator07, line 96: @Before Method

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+        BtcTransaction btcTx = new BtcTransaction(btcMainnetParams);
```
kelvinator07, line 98: @Before Method

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,294 @@
...
+        btcTx.addInput(BitcoinTestUtils.createHash(1), 0, new Script(new byte[]{}));
+        btcTx.addOutput(Coin.COIN, destinationAddress);
+
+        Script fedInputScript = PegTestUtils.createBaseInputScriptThatSpendsFromTheFederation(federation);
+        btcTx.getInput(0).setScriptSig(fedInputScript);
+
+        Sha256Hash expectedSigHash = btcTx.hashForSignature(
+            0,
```
kelvinator07, line 106:
```suggestion
            FIRST_INPUT_INDEX,
```
We Can Make This A Constant And Use It Everywhere

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,294 @@
...
+        Optional<Sha256Hash> sigHash = BitcoinUtils.getFirstInputSigHash(btcTx);
+
+        // Assert
+        Assertions.assertTrue(sigHash.isPresent());
+        Assertions.assertEquals(expectedSigHash, sigHash.get());
+    }
+
+    @Test
+    void test_getFirstInputSigHash_from_p2shFed() {
+        ActivationConfig.ForBlock activations = ActivationConfigsForTest.tbd600().forBlock(0);
```
kelvinator07, line 122: @Before Method

rskj-core/src/test/java/co/rsk/peg/bitcoin/PocSighashTest.java
```diff
@@ -473,6 +473,67 @@ void test_each_input_sighash_is_unique_for_a_signed_erp_tx_testnet() {
...
+        ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
+        when(activations.isActive(ConsensusRule.RSKIP284)).thenReturn(true);
+        when(activations.isActive(ConsensusRule.RSKIP293)).thenReturn(true);
+
+        P2shErpFederation fed = new P2shErpFederation(
+            fedMembers.stream().map(FedSigner::getFed).collect(Collectors.toList()),
+            Instant.now(),
+            0L,
+            networkParameters,
+            erpFedMembers.stream().map(FedSigner::getFed).map(FederationMember::getBtcPublicKey).collect(Collectors.toList()),
```
kelvinator07, line 494:
```suggestion
            erpFedMembers.stream().map(FedSigner::getFed).map(FederationMember::getBtcPublicKey).collect(Collectors.toList()),
```
Lets Extract This To A List

rskj-core/src/test/java/co/rsk/peg/bitcoin/PocSighashTest.java
```diff
@@ -473,6 +473,67 @@ void test_each_input_sighash_is_unique_for_a_signed_erp_tx_testnet() {
...
+
+        BtcTransaction pegoutTx = new BtcTransaction(networkParameters);
+        pegoutTx.addOutput(pegoutTxAmount, destinationAddress);
+        pegoutTx.setVersion(2);
+
+        pegoutTx.getInputs().forEach((transactionInput -> {
+            Script redeemScriptFromInput = getRedeemScriptFromInput(transactionInput);
+            assertEquals(fed.getRedeemScript(), redeemScriptFromInput);
+        }));
+    }
```
kelvinator07, line 535: Do We Also Need To Test The New Method Here: **`getFirstInputSigHash`** ?

rskj-core/src/test/java/co/rsk/peg/bitcoin/PocSighashTest.java
```diff
@@ -473,6 +473,67 @@ void test_each_input_sighash_is_unique_for_a_signed_erp_tx_testnet() {
...
+
+        List<FedSigner> fedMembers = FedSigner.listOf("federator1", "federator2", "federator6");
+        List<FedSigner> erpFedMembers = FedSigner.listOf("erp-fed-01", "erp-fed-02", "erp-fed-03");
+
+        ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
+        when(activations.isActive(ConsensusRule.RSKIP284)).thenReturn(true);
+        when(activations.isActive(ConsensusRule.RSKIP293)).thenReturn(true);
+
+        P2shErpFederation fed = new P2shErpFederation(
+            fedMembers.stream().map(FedSigner::getFed).collect(Collectors.toList()),
```
kelvinator07, line 490: Lets Extract This To A List

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: This is specific for this test. It's not the same for all tests.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: I believe each input should have its own instance.

rskj-core/src/main/java/co/rsk/peg/bitcoin/BitcoinUtils.java
```diff
@@ -0,0 +1,33 @@
...
+            return Optional.empty();
+        }
+        TransactionInput txInput = btcTx.getInput(FIRST_INPUT_INDEX);
+        Optional<Script> redeemScript = BridgeUtils.extractRedeemScriptFromInput(txInput);
+        if (!redeemScript.isPresent()) {
+            return Optional.empty();
+        }
+
+        return Optional.of(btcTx.hashForSignature(
+            0,
```
marcos-iov, line 27:
```suggestion
            FIRST_INPUT_INDEX,
```

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+class BitcoinUtilsTest {
+    private static final BridgeConstants bridgeMainnetConstants = BridgeMainNetConstants.getInstance();
+    private static final NetworkParameters btcMainnetParams = bridgeMainnetConstants.getBtcParams();
+    private static final int FIRST_OUTPUT_INDEX = 0;
+    private static final int FIRST_INPUT_INDEX = 0;
+
+    List<BtcECKey> signers;
+    List<BtcECKey> pubKeys;
+    Address destinationAddress;
```
marcos-iov, line 38: private?

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+        Address destinationAddress = BitcoinTestUtils.createP2PKHAddress(btcMainnetParams, "destinationAddress");
+
+        BtcTransaction btcTx = new BtcTransaction(btcMainnetParams);
+        btcTx.addInput(BitcoinTestUtils.createHash(1), FIRST_OUTPUT_INDEX, new Script(new byte[]{}));
+        btcTx.addOutput(Coin.COIN, destinationAddress);
+
+        RedeemData redeemData = RedeemData.of(pubKeys, redeemScript);
+        Script p2SHScript = ScriptBuilder.createP2SHOutputScript(requiredSignatures, pubKeys);
+        Script inputScript = p2SHScript.createEmptyInputScript(redeemData.keys.get(0), redeemData.redeemScript);
```
marcos-iov, line 67:
```suggestion
        Script p2SHScript = ScriptBuilder.createP2SHOutputScript(redeemScript);
        Script inputScript = p2SHScript.createEmptyInputScript(null, redeemData.redeemScript);
        btcTx.getInput(FIRST_INPUT_INDEX).setScriptSig(inputScript);
```
See the implementation for createEmptyInputScript, you either need the key if you are creating a script sig that pays to a key, or the redeem script if you are creating a script sig that pays to a script. Not both.

And I guess this will be quite commonly used so it could be extracted to it's own method. Is it only used in tests? Or would it make sense to have it in BitcoinUtils class?

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,262 @@
...
+        btcTx.getInput(FIRST_INPUT_INDEX).setScriptSig(inputScript);
+
+        Sha256Hash expectedSigHash = btcTx.hashForSignature(
```
marcos-iov, line 70: This is the frequent issue we have where we are testing something against the very same implementation. We do the same in the function as in the test, so the result will always be the same, if the function implementation is wrong the test will not catch it.

Is it possible to somehow get the sighash of existing signed transactions and compare against it? I think those make for more valuable tests

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,262 @@
...
+            Instant.ofEpochMilli(1000L),
+            0L,
+            btcMainnetParams
+        );
+
+        BtcTransaction btcTx = new BtcTransaction(btcMainnetParams);
+        btcTx.addInput(BitcoinTestUtils.createHash(1), FIRST_OUTPUT_INDEX, new Script(new byte[]{}));
+        btcTx.addOutput(Coin.COIN, destinationAddress);
+
+        Script fedInputScript = PegTestUtils.createBaseInputScriptThatSpendsFromTheFederation(federation);
```
marcos-iov, line 99: It's the same logic as the one used in the test above, no need to call this method from PegTestUtils. Just use ScriptBuilder

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,262 @@
...
+        int requiredSignatures = totalSigners / 2 + 1;
+
+        Script redeemScript = ScriptBuilder.createRedeemScript(requiredSignatures, signers);
+        Script p2SHScript = ScriptBuilder.createP2SHOutputScript(requiredSignatures, pubKeys);
+
+        Address destinationAddress = BitcoinTestUtils.createP2PKHAddress(btcMainnetParams, "destinationAddress");
+
+        BtcTransaction btcTx = new BtcTransaction(btcMainnetParams);
+
+        for (int i = 1; i <= 50; i++) {
```
marcos-iov, line 239: why not start at 0? to avoid `btcTx.getInput(i-1)`

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,262 @@
...
+            btcTx.addInput(BitcoinTestUtils.createHash(1), 0, new Script(new byte[]{}));
+            RedeemData redeemData = RedeemData.of(pubKeys, redeemScript);
+
+            Script inputScript = p2SHScript.createEmptyInputScript(redeemData.keys.get(0), redeemData.redeemScript);
+            btcTx.getInput(i-1).setScriptSig(inputScript);
+        }
+        btcTx.addOutput(Coin.COIN, destinationAddress);
+
+        Sha256Hash expectedSigHash = btcTx.hashForSignature(
```
marcos-iov, line 248: We could also assert that the sighash for different inputs is different

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
kelvinator07: The Objects Are Different For Each Loop?

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: Done

rskj-core/src/main/java/co/rsk/peg/bitcoin/BitcoinUtils.java
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: I did some research, assuming you mean getting the sighash without using the hashForSignature method, I couldn't find a way of doing that from the hex tx. We can do it manually, but it will be basically the same as what the hashForSignature does internally.

This is what `hashForSignature` method does:

1. Clear all input scripts.
2.  Set the input redeem script to the given input index.
3.  Add the sighash: SIGHASH_ALL
4. Serialize the tx.
5. Hashes the message(the transaction data serialized) using double Sha256Hash.
6. Returns the resulting hash.

[Source](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/core/Transaction.java#L1613)

However, an idea came to my mind, I tested it, and it worked. It consists of extracting the signature and verifying if the extracted signature was generated from the firstInputSigHash(got using the Bitcoin utils method). Signatures are unique. So, if the assertion is true. It means the extracted sighash is correct.

Take a look at this PoC test added, and let me know what you think:

https://github.com/rsksmart/rskj/pull/2109/commits/558340ba444a48a9fca7bb9115476d39ce05a747#diff-a7df218cc85a955d453d0693359919da88fc6a8f1483b1ca5a63bba1460aa081R58

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
marcos-iov: I'd check usages in the PowPeg node also

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,305 @@
...
+import co.rsk.bitcoinj.script.Script;
+import co.rsk.bitcoinj.script.ScriptBuilder;
+import co.rsk.bitcoinj.wallet.RedeemData;
+import co.rsk.config.BridgeConstants;
+import co.rsk.config.BridgeMainNetConstants;
+import co.rsk.peg.BridgeUtils;
+import co.rsk.peg.Federation;
+import co.rsk.peg.FederationTestUtils;
+import co.rsk.peg.P2shErpFederation;
+import co.rsk.peg.PegTestUtils;
```
marcos-iov, line 21: unused import

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,305 @@
...
+    private List<BtcECKey> signers;
+    private List<BtcECKey> pubKeys;
+    private Address destinationAddress;
+
+    @BeforeEach
+    void init() {
+        signers = BitcoinTestUtils.getBtcEcKeysFromSeeds(
+            new String[]{"fa01", "fa02", "fa03"}, true
+        );
+        pubKeys = signers.stream().map(BtcECKey::getPubKeyPoint).map(BtcECKey::fromPublicOnly)
```
marcos-iov, line 51: These are basically the same thing, the signers are BtcEcKey objects that contain both private and public keys. Not 100% sure, but I think that any place you need to pass a public key you can pass the object that has the private and public key and should work

rskj-core/src/test/java/co/rsk/peg/bitcoin/PocSighashTest.java
nathanieliov: Tests were refactored and the way I assert the sighash is valid was improved. Please check it out.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: Tests were refactored, and the way I assert the sighash is valid was improved. Please check it out.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: This was removed as well.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: New approach for assert sigHash was implemented.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: Removed.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: Resolved.

rskj-core/src/test/java/co/rsk/peg/bitcoin/PocSighashTest.java
```diff
@@ -473,6 +473,73 @@ void test_each_input_sighash_is_unique_for_a_signed_erp_tx_testnet() {
         }
     }
 
+    @Test
+    void test_redeemScript_can_be_obtained_from_input() {
+        // Arrange
+        NetworkParameters networkParameters = BridgeTestNetConstants.getInstance().getBtcParams();
+        int erpFedActivationDelay = 720;
```
kelvinator07, line 480:
```suggestion
        int erpFedActivationDelay = bridgeTestnetConstants.getErpFedActivationDelay();
```

nathanieliov: 720 is the correct value, BrigTestNetContstants has a different value(52_560). Remember here I'm using a real testnet tx. The federation spending funds was created with that erpFedActivationDelay.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -48,4 +55,33 @@ public static Sha256Hash createHash(int nHash) {
...
         return Sha256Hash.wrap(bytes);
     }
+
+    public static List<BtcECKey.ECDSASignature> extractSignatures(TransactionInput txInput) {
+        Script scriptSig = txInput.getScriptSig();
+        List<ScriptChunk> chunks = scriptSig.getChunks();
+        Script redeemScript = new Script(chunks.get(chunks.size() - 1).data);
+        RedeemScriptParser parser = RedeemScriptParserFactory.get(redeemScript.getChunks());
+
+        int lastChunk;
```
kelvinator07, line 65:
```suggestion
        int signatureChunks; ???
```
This Represents All The Chunks Except The Last Chunk, Which Is The RedeemScript? Right?

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -48,4 +55,33 @@ public static Sha256Hash createHash(int nHash) {
...
+        }
+
+        List<BtcECKey.ECDSASignature> signatures = new ArrayList<>();
+        for (int i = 1; i < lastChunk; i++) {
+            ScriptChunk chunk = chunks.get(i);
+            if (!chunk.isOpCode() && chunk.data.length == 0) {
+                continue;
+            } else {
+                signatures.add(TransactionSignature.decodeFromDER(chunk.data));
+            }
```
kelvinator07, line 83:
```suggestion
            if (chunk.isOpCode() && chunk.data.length > 0) {
                signatures.add(TransactionSignature.decodeFromDER(chunk.data));
            }
```

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
nathanieliov: The last chunk before the redeem script. Basically, it is the index of the last signature of the script sig chunks.

nathanieliov: I'll rename the variable.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -48,4 +55,33 @@ public static Sha256Hash createHash(int nHash) {
 
         return Sha256Hash.wrap(bytes);
     }
+
+    public static List<BtcECKey.ECDSASignature> extractSignatures(TransactionInput txInput) {
```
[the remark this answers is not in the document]
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,157 @@
...
+    @MethodSource("peginArgProvider")
+    void test_getFirstInputSigHash_pegin(String btcTxHex) {
+        // Arrange
+        BtcTransaction btcTx = new BtcTransaction(btcMainnetParams, Hex.decode(btcTxHex));
+
+        // Act
+        Optional<Sha256Hash> firstInputSigHash = BitcoinUtils.getFirstInputSigHash(btcTx);
+
+        // Assert
+        Assertions.assertFalse(firstInputSigHash.isPresent());
```
marcos-iov, line 126: Why is there no sig hash?

nathanieliov: There are three tx being processed for this test:

1 - The first one is a PEGIN BECH32. ScriptSig is empty. No chunks. If no redeemScript is found, no sighash is returned. This is a limitation of `BitcoinUtils#getFirstInputSigHash`. It only works for txs where the first input has a redeem script.
2 - Second, is a PEGIN legacy. It failed to extract the redeem script. Since no redeem script is extracted, no sighash can be generated.
3- Third one, the same as the second.

Do you think we should change the name of the method to be explicit that it works only for bridge pegout transactions?

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,157 @@
...
...0bf6b140e2df8770025e333869fa6f54a6cf152e6666c1054eb3424022023f255af0855464f26165a026ede6d6db7475c7442353a3dcd17dc4bc61fb14201004dc901645521020ace50bab1230f8002a0bfe619482af74b338cc9e4c956add228df47e6adae1c21025093f439fb8006fd29ab56605ffec9cdc840d16d2361004e1337a2f86d8bd2db21026b472f7d59d201ff1f540f111b6eb329e071c30a9d23e3d2bcd128fe73dc254c210275d473555de2733c47125f9702b0f870df1d817379f5587f09b6c40ed2c6c9492102a95f095d0ce8cb3b9bf70cc837e3ebe1d107959b1fa3f9b2d8f33446f9c8cbdb2103250c11be0561b1d7ae168b1f59e39cbc1fd1ba3cf4d2140c1a365b2723a2bf9321034851379ec6b8a701bd3eef8a0e2b119abb4bdde7532a3d6bcbff291b0daf3f252103b58a5da144f5abab2e03e414ad044b732300de52fa25c672a7f7b358887719062103e05bf6002b62651378b1954820539c36ca405cbb778c225395dd9ebff678029959ae670350cd00b275532102370a9838e4d15708ad14a104ee5606b36caaaaf739d833e67770ce9fd9b3ec80210257c293086c4d4fe8943deda5f890a37d11bebd140e220faa76258a41d077b4d42103c2660a46aa73078ee6016dee953488566426cf55fc8011edd0085634d75395f92103cd3e383ec6e12719a6c69515e5559bcbe037d0aa24c187e1e26ce932e22ad7b354ae68ffffffff0281c00e00000000001976a9140ab5592fc4c28f2301014da9981a3f1d7ad3d4af88ac597ebc2d4900000017a914056d0d9c5b14dd720d9f61fdb3f557c074f95cef8700000000"),
```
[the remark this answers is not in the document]
nathanieliov: Yes, but in that case, I will need to create the BtcTransaction object here. I see no benefit in doing that for these simple tests.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -48,4 +55,32 @@ public static Sha256Hash createHash(int nHash) {
...
+        ) {
+            flowOpCodeChunk = chunks.size() - 1;
+        } else {
+            flowOpCodeChunk = chunks.size() - 2;
+        }
+
+        List<BtcECKey.ECDSASignature> signatures = new ArrayList<>();
+        for (int i = 1; i < flowOpCodeChunk; i++) {
+            ScriptChunk chunk = chunks.get(i);
+            if (chunk.isOpCode() || chunk.data.length > 0) {
```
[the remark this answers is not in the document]
nathanieliov: I just recreated `countMissingSignatures` method. See [co.rsk.peg.BridgeUtils#countMissingSignatures](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java#L570).

Before the condition was:

```
if (!chunk.isOpCode() && chunk.data.length == 0) {
                unsigned++;
            }
```

Since Kelvin left a comment asking to have the condition as positive, I changed it to how it is now.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -48,4 +55,32 @@ public static Sha256Hash createHash(int nHash) {
...
+        RedeemScriptParser.MultiSigType multiSigType = parser.getMultiSigType();
+
+        if (multiSigType == RedeemScriptParser.MultiSigType.STANDARD_MULTISIG
+                || multiSigType == RedeemScriptParser.MultiSigType.FAST_BRIDGE_MULTISIG
+        ) {
+            flowOpCodeChunk = chunks.size() - 1;
+        } else {
+            flowOpCodeChunk = chunks.size() - 2;
+        }
+
```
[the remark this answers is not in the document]
nathanieliov: I created this method based on [countMissingSignatures](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java#L570) and [hasEnoughSignatures](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java#L610C27-L610C46).

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,157 @@
...
+        TransactionInput txInput = btcTx.getInput(FIRST_INPUT_INDEX);
+
+        Optional<Script> redeemScriptFromInput = BridgeUtils.extractRedeemScriptFromInput(txInput);
+        Assertions.assertTrue(redeemScriptFromInput.isPresent());
+        RedeemScriptParser redeemScriptParser = RedeemScriptParserFactory.get(redeemScriptFromInput.get().getChunks());
+        List<BtcECKey> pubKeys = redeemScriptParser.getPubKeys();
+        Assertions.assertFalse(pubKeys.isEmpty());
+
+        List<BtcECKey.ECDSASignature> signatures = BitcoinTestUtils.extractSignaturesFromTxInput(txInput);
+        Assertions.assertFalse(signatures.isEmpty());
```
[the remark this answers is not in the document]
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,157 @@
...
+            BtcECKey.ECDSASignature signature = sigs.getFirst();
+            if(pubKey.verify(sigHash, signature)){
+                sigs.pollFirst();
+            }
+            if (sigs.size() > keys.size()) {
+                valid = false;
+                break;
+            }
+        }
+        return valid;
```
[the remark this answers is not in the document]
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,157 @@
...
+        boolean valid = true;
+        while (sigs.size() > 0){
+            BtcECKey pubKey = keys.pollFirst();
+            BtcECKey.ECDSASignature signature = sigs.getFirst();
+            if(pubKey.verify(sigHash, signature)){
+                sigs.pollFirst();
+            }
+            if (sigs.size() > keys.size()) {
+                valid = false;
+                break;
```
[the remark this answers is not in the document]
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -0,0 +1,157 @@
...
+        Assertions.assertFalse(signatures.isEmpty());
+
+        // Act
+        Optional<Sha256Hash> firstInputSigHash = BitcoinUtils.getFirstInputSigHash(btcTx);
+        Assertions.assertTrue(firstInputSigHash.isPresent());
+
+        // Assert
+        boolean valid = isSigHashValid(firstInputSigHash.get(), pubKeys, signatures);
+        // It will be false if the extracted inputSigHash is different from the inputSigHash used for generating the signature.
+        Assertions.assertTrue(valid);
```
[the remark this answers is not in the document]
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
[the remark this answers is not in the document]
marcos-iov: I don't think those 2 are equivalent. Did you try doing a truth table?

Also careful with copy pasting, what is the content of a chunk with a signature? Should it be an op code?

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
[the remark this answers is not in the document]
marcos-iov: Careful with copy pasting here as well. When using standard multisig there is no flow op code, that is part of the ERP

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
marcos-iov: Sounds like three different test scenarios, not sure if it makes sense to have them all in the same method.

First case, makes sense once you explain it but not when reading the test. Maybe it should be part of the test name.

Second case, no direct relationship between being a pegin legacy and not having a redeem script. It doesn't have a redeem script because the sender is a P2PKH address and not a P2SH. It is true that legacy peg-ins can only be done from P2PKH addresses but still, sounds a bit stretched. I'd simply test that it will fail to find the sighash of a P2PKH transaction. Not related to peg-ins. If we want to test peg-ins, why not test peg-in v1 where the sender can be P2SH?

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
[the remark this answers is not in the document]
nathanieliov: I don't get it. So, are those methods wrong? `BridgeUtils.hasEnoughSignatures` is being used at `BridgeSupport.processSigning` for all kinds of pegout, from multig, ERP, and P2sh. Is it not right to assume this logic is correct? And based on it be able to create a method that does almost the same but in this case returns the signatures, what's wrong with it?

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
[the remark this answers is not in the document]
nathanieliov: Same here.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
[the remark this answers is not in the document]
nathanieliov: Updated

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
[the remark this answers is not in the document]
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -48,4 +56,36 @@ public static Sha256Hash createHash(int nHash) {
...
+            lastSigChunk = chunks.size() - 1;
+        } else {
+            lastSigChunk = chunks.size() - 2;
+        }
+
+        List<BtcECKey.ECDSASignature> signatures = new ArrayList<>();
+        for (int i = 1; i < lastSigChunk; i++) {
+            ScriptChunk chunk = chunks.get(i);
+            if (!chunk.isOpCode() && chunk.data.length > 0) {
+                // Bitcoin signatures always starts with the prefix ‘30’.
```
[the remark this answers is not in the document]
nathanieliov: I think the same. I just added the if checking the prefix as a double check. And yes, according to the source I researched, that prefix is for all networks. Here is one of the sources: https://github.com/libbitcoin/libbitcoin-system/wiki/ECDSA-and-DER-Signatures.

But since it's not mandatory and it is enough with the above if, I'll remove it.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -48,4 +56,36 @@ public static Sha256Hash createHash(int nHash) {
...
+        if (multiSigType == RedeemScriptParser.MultiSigType.STANDARD_MULTISIG
+                || multiSigType == RedeemScriptParser.MultiSigType.FAST_BRIDGE_MULTISIG
+        ) {
+            lastSigChunk = chunks.size() - 1;
+        } else {
+            lastSigChunk = chunks.size() - 2;
+        }
+
+        List<BtcECKey.ECDSASignature> signatures = new ArrayList<>();
+        for (int i = 1; i < lastSigChunk; i++) {
```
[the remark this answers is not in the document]
nathanieliov: Nice. I like this idea.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -48,4 +56,36 @@ public static Sha256Hash createHash(int nHash) {
...
     }
+
+    public static List<BtcECKey.ECDSASignature> extractSignaturesFromTxInput(TransactionInput txInput) {
+        String btcHexSignaturesPrefix = "30";
+        Script scriptSig = txInput.getScriptSig();
+        List<ScriptChunk> chunks = scriptSig.getChunks();
+        Script redeemScript = new Script(chunks.get(chunks.size() - 1).data);
+        RedeemScriptParser parser = RedeemScriptParserFactory.get(redeemScript.getChunks());
+
+        // This is the last chunk after the signatures.
```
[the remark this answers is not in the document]
nathanieliov: Lines were removed due to method refactored.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -48,4 +56,36 @@ public static Sha256Hash createHash(int nHash) {
...
+        int lastSigChunk;
+        RedeemScriptParser.MultiSigType multiSigType = parser.getMultiSigType();
+
+        if (multiSigType == RedeemScriptParser.MultiSigType.STANDARD_MULTISIG
+                || multiSigType == RedeemScriptParser.MultiSigType.FAST_BRIDGE_MULTISIG
+        ) {
+            lastSigChunk = chunks.size() - 1;
```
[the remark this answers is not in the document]
nathanieliov: Lines were removed due to method refactored.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -48,4 +56,36 @@ public static Sha256Hash createHash(int nHash) {
...
+        } else {
+            lastSigChunk = chunks.size() - 2;
```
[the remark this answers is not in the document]
nathanieliov: Lines were removed due to method refactored.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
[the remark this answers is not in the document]
nathanieliov: Implemented.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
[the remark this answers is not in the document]
nathanieliov: Removed.

marcos-iov (approved):
Well done! Very complete now

PR #2113, opened 2023-08-25
Release witness transaction builder

julia-zack:


rskj-core/src/test/java/co/rsk/peg/P2shP2wshErpFederationNewRedeemTest.java
```diff
@@ -0,0 +1,273 @@
+/*
```
marcos-iov, line 1: Should this be all commented out?

julia-zack: i actually deleted this and other test related to segwit (not the ReleaseWitnessTransactionBuilderTest). they are now included in bitcoinj-thin final-segwit-poc branch!

PR #2143, opened 2023-09-27
Refactor federation

julia-zack:
Make Federation class abstract
Create a new class StandardMultisigFederationthat inherits from Federation
Create a new abstract class ErpFederation that inherits from Federation but adds the behavior related to the ERP
Rename current ERPFederation class to LegacyErpFederation and tag it as deprecated. This is the class that generates the faulty non standard redeem script
Make P2SHErpFederation class inherit from ERPFederation

Eventually, we will add a new class P2shP2wshErpFederation class that also inherits from ERPFederation so the file is created with commented (and wrong) code


rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -91,18 +91,7 @@ public long getCreationBlockNumber() {
...
-    }
-
-    public Script getStandardRedeemScript() {
-        return getRedeemScript();
-    }
-
+    public abstract Script getRedeemScript();
     public Script getP2SHScript() {
         if (p2shScript == null) {
             p2shScript = ScriptBuilder.createP2SHOutputScript(getNumberOfSignaturesRequired(), getBtcPublicKeys());
```
marcos-iov, line 97:
```suggestion
            p2shScript = ScriptBuilder.createP2SHOutputScript(getRedeemScript());
```
The P2SH script is built from the redeem script `OP_HASH160 redeemScriptHash OP_EQUAL` if I'm not mistaken. See the implementation for `createP2SHOutputScript` that receives the threshold and the public keys. It will build a default redeem script with those values. This will not work for federations that use different redeem script formats. So, it's best to pass the redeem script directly

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -152,33 +140,12 @@ public boolean isMember(FederationMember federationMember){
         return this.members.contains(federationMember);
     }
 
-    @Override
```
marcos-iov, line 155: It's ok to override this method, it's part of the `Object` java class from which all clases inherit

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -152,33 +140,12 @@ public boolean isMember(FederationMember federationMember){
...
     public String toString() {
         return String.format("%d of %d signatures federation", getNumberOfSignaturesRequired(), members.size());
     }
 
-    @Override
```
marcos-iov, line 160: Same as the toString

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -152,33 +140,12 @@ public boolean isMember(FederationMember federationMember){
         return this.members.contains(federationMember);
     }
 
-    @Override
     public String toString() {
         return String.format("%d of %d signatures federation", getNumberOfSignaturesRequired(), members.size());
```
marcos-iov, line 144: We should reconsider this

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -152,33 +140,12 @@ public boolean isMember(FederationMember federationMember){
...
-
-        return this.getNumberOfSignaturesRequired() == otherFederation.getNumberOfSignaturesRequired() &&
-            this.getSize() == otherFederation.getSize() &&
-            this.getCreationTime().equals(otherFederation.getCreationTime()) &&
-            this.creationBlockNumber == otherFederation.creationBlockNumber &&
-            this.btcParams.equals(otherFederation.btcParams) &&
-            this.members.equals(otherFederation.members) &&
-            this.getRedeemScript().equals(otherFederation.getRedeemScript());
-    }
+    public abstract boolean equals(Object other);
```
marcos-iov, line 147: No need to add it actually if we won't implement it. It's part of the Object class so all classes that extend from this one will still have it even if it's not declared here

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -152,33 +140,12 @@ public boolean isMember(FederationMember federationMember){
...
 
-    @Override
```
marcos-iov, line 181: Same as toString

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -49,7 +49,7 @@ public class Federation {
     protected Script p2shScript;
     protected Address address;
 
-    public Federation(List<FederationMember> members, Instant creationTime, long creationBlockNumber,  NetworkParameters btcParams) {
+    public Federation(List<FederationMember> members, Instant creationTime, long creationBlockNumber, NetworkParameters btcParams) {
```
marcos-iov, line 52: We should have a look at the method `getNumberOfSignaturesRequired`. I don't think that implementation is correct. We should get the required signers from the OP_M contained in the redeem script, that's the source of truth.

The RedeemScriptParser in bitcoinj-thin has an `getM` method, but I think that implementation might be flawed also. Let's have a look at that

rskj-core/src/main/java/co/rsk/peg/StandardMultisigFederation.java
```diff
@@ -0,0 +1,75 @@
...
+import co.rsk.bitcoinj.script.ScriptBuilder;
+
+import java.time.Instant;
+import java.util.List;
+
+/**
+ * Immutable representation of an RSK Federation in the context of
+ * a specific BTC network.
+ *
+ * @author Ariel Mendelzon
```
marcos-iov, line 32: Ariel? ;)

rskj-core/src/main/java/co/rsk/peg/StandardMultisigFederation.java
```diff
@@ -0,0 +1,75 @@
...
+            return true;
+        }
+
+        if (other == null || this.getClass() != other.getClass()) {
+            return false;
+        }
+
+        StandardMultisigFederation otherFederation = (StandardMultisigFederation) other;
+
+        return this.getNumberOfSignaturesRequired() == otherFederation.getNumberOfSignaturesRequired() &&
```
marcos-iov, line 67: We could also discuss this. What makes 2 federations be equal? If both have the same redeem script but different creation time, are they different? I think not, the way I see it the redeem script is basically what defines the federation.

Maybe same redeem script but different networks we could consider them different, at least the address will be different. What do you think?

rskj-core/src/main/java/co/rsk/peg/P2shP2wshErpFederation.java
```diff
@@ -0,0 +1,86 @@
+/*
```
marcos-iov, line 1: Let's remove this one

rskj-core/src/main/java/co/rsk/peg/ErpFederation.java
```diff
@@ -17,16 +15,18 @@
...
+public abstract class ErpFederation extends Federation {
     private static final Logger logger = LoggerFactory.getLogger(ErpFederation.class);
 
+    protected static final byte[] ERP_TESTNET_REDEEM_SCRIPT_BYTES = Hex.decode("6453210208f40073a9e43b3e9103acec79767a6de9b0409749884e989960fee578012fce210225e892391625854128c5c4ea4340de0c2a70570f33db53426fc9c746597a03f42102afc230c2d355b1a577682b07bc2646041b5d0177af0f98395a46018da699b6da210344a3c38cd59afcba3edcebe143e025574594b001700dec41e59409bdbd0f2a0921039a060badbeb24bee49eb2063f616c0f0f0765d4ca646b20a88ce828f259fcdb955670300cd50b27552210216c23b2ea8e4f11c3f9e22711addb1d16a93964796913830856b568cc3ea21d3210275562901dd8faae20de0a4166362a4f82188db77dbed4ca887422ea1ec185f1421034db69f2112f4fb1bb6141bf6e2bd6631f0484d0bd95b16767902c9fe219d4a6f5368ae");
```
marcos-iov, line 21: This is a specific flawed federation that existed in testnet, we need to keep it for consensus but I would move it to LegacyErpFederation since it's of that type

rskj-core/src/main/java/co/rsk/peg/ErpFederation.java
```diff
@@ -17,16 +15,18 @@
...
+
     protected final List<BtcECKey> erpPubKeys;
     protected final long activationDelay;
     protected final ActivationConfig.ForBlock activations;
-
     protected Script standardRedeemScript;
+    // add erpRedeemScript? is it necessary?
```
marcos-iov, line 27: It's also inherited from Federation class, the `redeemScript` field

rskj-core/src/main/java/co/rsk/peg/ErpFederation.java
```diff
@@ -17,16 +15,18 @@
...
     protected Script standardP2SHScript;
+    // add erpP2SHScript?
```
marcos-iov, line 29: Same as above

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -152,33 +140,12 @@ public boolean isMember(FederationMember federationMember){
...
-            this.getCreationTime().equals(otherFederation.getCreationTime()) &&
-            this.creationBlockNumber == otherFederation.creationBlockNumber &&
-            this.btcParams.equals(otherFederation.btcParams) &&
-            this.members.equals(otherFederation.members) &&
-            this.getRedeemScript().equals(otherFederation.getRedeemScript());
-    }
+    public abstract boolean equals(Object other);
 
-    @Override
     public int hashCode() {
```
marcos-iov, line 149: Maybe this should be part of StandarFederation, right? Because ERPFederation is defining it's own hashCode

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        return redeemScript;
+    }
+
+    @Override
+    public Script getStandardRedeemScript() {
+        if (standardRedeemScript == null) {
+            // i think is not quite right to just "fix" the standardRedeemScript in the buggy fed.
+            // the real case is that the OP_CHECKMULTISIG is not before the OP_ELSE... as the method adds.
```
marcos-iov, line 55: Didn't quite get this

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
```diff
@@ -0,0 +1,86 @@
...
+            standardRedeemScript = ErpFederationRedeemScriptParser.extractStandardRedeemScript(
+                getRedeemScript().getChunks()
+            );
+        }
+        return standardRedeemScript;
+    }
+
+    @Override
+    public boolean equals(Object other) {
```
marcos-iov, line 64: The equals defined in ErpFederation class should be enough right?

rskj-core/src/main/java/co/rsk/peg/P2shErpFederation.java
```diff
@@ -28,6 +29,13 @@ public P2shErpFederation(
 
     @Override
     public final Script getRedeemScript() {
+        // not sure if i should add this here or not. I think not but just in case
```
marcos-iov, line 32: Nope, no need

rskj-core/src/main/java/co/rsk/peg/P2shErpFederation.java
```diff
@@ -49,4 +57,27 @@ public final Script getStandardRedeemScript() {
         }
         return standardRedeemScript;
     }
+
+    @Override
+    public boolean equals(Object other) {
```
marcos-iov, line 62: Any difference between this and the one in ErpFederation and the one in LegacyErpFederation?

rskj-core/src/main/java/co/rsk/peg/ErpFederation.java
julia-zack: okey, perfect. I will remove it from the abstract class and just implement it in LegacyErpFederation.

rskj-core/src/main/java/co/rsk/peg/Federation.java
julia-zack: Yes, i noticed that when doing changes in bitcoinj-thin. it won't be that easy since the threshold is represented as an OP_PUSHNUM_M, pretty sure it is not an opcode.

rskj-core/src/main/java/co/rsk/peg/Federation.java
julia-zack: You are absolutely right. In my mind it was getting the redeem script directly. Thanks!

rskj-core/src/main/java/co/rsk/peg/Federation.java
julia-zack: Yes, of course. Absolutely right

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
julia-zack: My concern is that in the ErpFederationRedeemScriptParser.extractStandardRedeemScript, we are adding the OP_CHECKMULTISIG opcode at the end of the redeem script's chunks

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
julia-zack: Mmm, not exactly, because the way of extracting the standard redeem scripts are different: in the ErpFederationRedeemScriptParser we add the OP_CHECKMULTISIG opcode at the end. And we don't do that in the P2shErpFederationRedeemScriptParser.
The methods belong to two different classes

rskj-core/src/main/java/co/rsk/peg/P2shErpFederation.java
julia-zack: yes, the parsing of the federation type. because the methods can be different (for example the extractStandardRedeemScript one)

rskj-core/src/main/java/co/rsk/peg/StandardMultisigFederation.java
julia-zack: Im pretty sure that i agree on what you say. But is an interesting discussion to have and we should definitely have it.

rskj-core/src/main/java/co/rsk/peg/Federation.java
julia-zack: Yes, sounds good.

rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
```diff
@@ -24,13 +24,15 @@
...
+
+import co.rsk.peg.StandardMultisigFederation;
 import org.ethereum.config.blockchain.upgrades.ActivationConfig;
 import org.ethereum.config.blockchain.upgrades.ConsensusRule;
 
 public abstract class BridgeConstants {
     protected String btcParamsString;
 
-    protected Federation genesisFederation;
+    protected StandardMultisigFederation genesisFederation;
```
marcos-iov, line 35: Unless there's a good reason, I would always try to keep the declaration as the base object. `Federation` in this case

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
```diff
@@ -273,7 +273,7 @@ private static Federation deserializeFederationWithDeserializer(
             federationMembers.add(member);
         }
 
-        return new Federation(federationMembers, creationTime, creationBlockNumber, networkParameters);
+        return new StandardMultisigFederation(federationMembers, creationTime, creationBlockNumber, networkParameters);
```
marcos-iov, line 276: Now this method changes. The way I see it, it's serializing a StandardMultisigFederation, not any  federation type. So I'd change the signature to receive an object of `StandardMultisigFederation` type

marcos-iov: Just in case, check where it's used. Make sure that it's always called with a SatandarMultisigFederation, if it's not we might need to reconsider. And I would also change the name of the method to make it clear that it is not serializing any federation

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
```diff
@@ -326,7 +326,7 @@ public static ErpFederation deserializeErpFederation(
             BridgeSerializationUtils::deserializeFederationMember
         );
 
-        return new ErpFederation(
+        return new LegacyErpFederation(
```
marcos-iov, line 329: Same here, I guess this method is in charge of deserializing a LegacyErpFederation right? In that case, I'd change the name, parameter and probably the return type also

marcos-iov: But please check first if that is the case

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -2079,7 +2079,7 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash) throws IOExce
             // Preserve federation change info
             long nextFederationCreationBlockHeight = rskExecutionBlock.getNumber();
             provider.setNextFederationCreationBlockHeight(nextFederationCreationBlockHeight);
-            Script oldFederationP2SHScript = activations.isActive(RSKIP377)? oldFederation.getStandardP2SHScript(): oldFederation.getP2SHScript();
+            Script oldFederationP2SHScript = activations.isActive(RSKIP377)? ((ErpFederation) oldFederation).getStandardP2SHScript(): oldFederation.getP2SHScript();
```
marcos-iov, line 2082: `oldFederation` object is of type `Federation`, right? So how can it be calling `getP2SHScript` method?

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -402,7 +402,7 @@ public static boolean isValidPegInTx(
                         throw new ScriptException(message);
                     }
                     Script inputStandardRedeemScript = redeemScriptParser.extractStandardRedeemScript();
-                    if (activeFederations.stream().anyMatch(federation -> federation.getStandardRedeemScript().equals(inputStandardRedeemScript))) {
+                    if (activeFederations.stream().anyMatch(federation -> (federation instanceof ErpFederation ? ((ErpFederation) federation).getStandardRedeemScript() : federation.getRedeemScript()).equals(inputStandardRedeemScript))) {
```
marcos-iov, line 405: That's a long line, perhaps we can reduce it
```suggestion
if (activeFederations.stream().anyMatch(federation -> {
                        Script fedStandardRedeemScript = federation instanceof ErpFederation ?
                            ((ErpFederation) federation).getStandardRedeemScript() :
                            federation.getRedeemScript();
                         return fedStandardRedeemScript.equals(inputStandardRedeemScript);
                    })) {
```

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -490,7 +490,7 @@ private static boolean isPegOutTx(BtcTransaction tx, Federation federation, Acti
     }
 
     public static boolean isPegOutTx(BtcTransaction tx, List<Federation> federations, ActivationConfig.ForBlock activations) {
-        return isPegOutTx(tx, activations, federations.stream().filter(Objects::nonNull).map(Federation::getStandardP2SHScript).toArray(Script[]::new));
+        return isPegOutTx(tx, activations, federations.stream().filter(Objects::nonNull).map(fed -> fed instanceof ErpFederation ? ((ErpFederation) fed).getStandardRedeemScript() : fed.getRedeemScript()).toArray(Script[]::new));
```
marcos-iov, line 493: Same here

rskj-core/src/main/java/co/rsk/peg/ErpFederation.java
```diff
@@ -112,16 +58,17 @@
...
-            address = Address.fromP2SHScript(btcParams, getP2SHScript());
-        }
+    public int getNumberOfSignaturesRequired() {
+        List<ScriptChunk> standardRedeemScriptChunks = getStandardRedeemScript().getChunks();
 
-        return address;
+        // the threshold of a multisig is the first chunk of the redeemScript
+        // and the standardRedeemScript represents a multisig
+        ScriptChunk thresholdChunk = standardRedeemScriptChunks.get(0);
+        return Integer.parseInt(thresholdChunk.toString());
```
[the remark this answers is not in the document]
marcos-iov: See this warning. What if it's not a number that's in the first chunk? What should happen then?

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -152,41 +134,15 @@ public boolean isMember(FederationMember federationMember){
...
-
-        return this.getNumberOfSignaturesRequired() == otherFederation.getNumberOfSignaturesRequired() &&
-            this.getSize() == otherFederation.getSize() &&
-            this.getCreationTime().equals(otherFederation.getCreationTime()) &&
-            this.creationBlockNumber == otherFederation.creationBlockNumber &&
-            this.btcParams.equals(otherFederation.btcParams) &&
-            this.members.equals(otherFederation.members) &&
-            this.getRedeemScript().equals(otherFederation.getRedeemScript());
-    }
+    public abstract boolean equals(Object other);
```
marcos-iov, line 144: No need to add it here, it comes by default

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -152,41 +134,15 @@ public boolean isMember(FederationMember federationMember){
...
-        // Can use java.util.Objects.hash since all of Instant, int and List<BtcECKey> have
-        // well-defined hashCode()s
-        return Objects.hash(
-            getCreationTime(),
-            this.creationBlockNumber,
-            getNumberOfSignaturesRequired(),
-            getBtcPublicKeys()
-        );
-    }
+    public abstract int hashCode();
```
marcos-iov, line 147: Same as with equals

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
marcos-iov: The buggy redeem script (used in this class) has the following format approx, I may be missing some op code

```
OP_NOTIF
  OP_M
  pubkey1
  pubkey2
  ...
  pubkeyN
  OP_N
OP_ELSE
OP_CSV
OP_DROP
  OP_M
  emergencyPubkey1
  emergencyPubkey2
  ....
  emergencyPubkeyN
  OP_N
OP_ENDIF
OP_CHECKMULTISIG
```

When extracting the standard redeem script, it gets the chunks between OP_NOTIF and OP_ELSE. So it gets
```
OP_M
  pubkey1
  pubkey2
  ...
  pubkeyN
  OP_N
```
It's missing the OP_CHECKMULTISIG, that's why it needs to add it. Not related to the fact that's it's buggy, that's another story

rskj-core/src/main/java/co/rsk/peg/StandardMultisigFederation.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public int getNumberOfSignaturesRequired() {
+        List<ScriptChunk> standardRedeemScriptChunks = getRedeemScript().getChunks();
+
+        // the threshold of a multisig is the first chunk of the redeemScript
+        // and this fed is a standard multisig
+        ScriptChunk thresholdChunk = standardRedeemScriptChunks.get(0);
+        return Integer.parseInt(thresholdChunk.toString());
```
marcos-iov, line 63: Can't all this be solved by simply calling `getRedeemScript().getNumberOfSignaturesRequiredToSpend`?

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderFederationTests.java
```diff
@@ -614,7 +614,7 @@ private Federation createFederation(int version) {
                     activations
                 );
             case ERP_FEDERATION_FORMAT_VERSION:
```
marcos-iov, line 616: Should this be renamed to `LEGACY_ERP_FEDERATION_FORMAT_VERSION`?

rskj-core/src/test/java/co/rsk/peg/PowpegMigrationTest.java
```diff
@@ -591,14 +591,14 @@ private void testChangePowpeg(
             if (oldPowPegFederationType == FederationType.erp || oldPowPegFederationType == FederationType.p2sh){
                 assertNotEquals(lastRetiredFederationP2SHScript, originalPowpeg.getP2SHScript());
             }
-            assertEquals(lastRetiredFederationP2SHScript, originalPowpeg.getStandardP2SHScript());
+            assertEquals(lastRetiredFederationP2SHScript, originalPowpeg instanceof ErpFederation ? ((ErpFederation) originalPowpeg).getStandardP2SHScript() : originalPowpeg.getP2SHScript());
```
marcos-iov, line 594: Let's spit this and the ones below in multiple lines to make it easier to read

rskj-core/src/test/java/co/rsk/peg/FederationSupportTest.java
```diff
@@ -74,7 +74,7 @@ void whenNewFederationIsNullThenActiveFederationIsGenesisFederation() {
         when(provider.getNewFederation())
             .thenReturn(null);
         when(bridgeConstants.getGenesisFederation())
-            .thenReturn(genesisFederation);
+            .thenReturn((StandardMultisigFederation) genesisFederation);
```
marcos-iov, line 77: Shouldn't need to parse this

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
julia-zack: oookidoki. yeah i understood what it does, but i just didnt know why. thanks

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
julia-zack: yes, you are right. there are some names to change

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderFederationTests.java
julia-zack: yes, lot of renames must be done. i didnt realize that

rskj-core/src/main/java/co/rsk/peg/StandardMultisigFederation.java
julia-zack: not really. that also leads us to an stackOverflow error, since the `getRedeemScript` method uses the `getNumberOfSignaturesRequired` one

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
julia-zack: I think so because the methods `deserializeErpFederation` and `deserializeP2shErpFederation` exist too.

rskj-core/src/main/java/co/rsk/peg/ErpFederation.java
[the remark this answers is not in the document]
julia-zack: i'll resolve this conversation since this method was rollback to what it was

rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
```diff
@@ -94,7 +96,7 @@ public String getBtcParamsString() {
         return btcParamsString;
     }
 
-    public Federation getGenesisFederation() { return genesisFederation; }
+    public StandardMultisigFederation getGenesisFederation() { return (StandardMultisigFederation) genesisFederation; }
```
marcos-iov, line 99: Why return a `StandardMultisigFederation` instead of `Federation`?

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
marcos-iov: Should we rename it then to `deserializeStandardMultisigFederationWithDeserializer` and change the signature to return a `StandardMultisigFederation`?

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -1029,7 +1029,7 @@ private Federation deserializeFederationAccordingToVersion(
     ) {
         switch (version) {
             case ERP_FEDERATION_FORMAT_VERSION:
```
marcos-iov, line 1031: Should this constant be renamed also?

rskj-core/src/main/java/co/rsk/peg/BridgeUtils.java
```diff
@@ -515,7 +521,9 @@ public static boolean isPegOutTx(BtcTransaction tx, ActivationConfig.ForBlock ac
             }
 
             Script outputScript = ScriptBuilder.createP2SHOutputScript(redeemScript);
-            if (Stream.of(p2shScript).anyMatch(federationPayScript -> federationPayScript.equals(outputScript))) {
+            if (Stream.of(p2shScript).anyMatch(federationPayScript ->
+                federationPayScript.equals(outputScript))) {
+                System.out.println();
```
marcos-iov, line 526: Remove logging messages please

rskj-core/src/main/java/co/rsk/peg/ErpFederation.java
```diff
@@ -131,7 +62,6 @@ public boolean equals(Object other) {
         }
 
         ErpFederation otherErpFederation = (ErpFederation) other;
-
         return this.getNumberOfSignaturesRequired() == otherErpFederation.getNumberOfSignaturesRequired() &&
```
marcos-iov, line 65: Didn't we agree on checking the address?

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -76,7 +74,7 @@ public List<BtcECKey> getBtcPublicKeys() {
     }
 
     public int getNumberOfSignaturesRequired() {
-        return members.size() / 2 + 1;
+        return getMembers().size() / 2 +1 ;
```
marcos-iov, line 77: Did you choose to call getMembers instead of members for any particular reason? Just asking because I see a mix of both on this class so we should probably settle for one option

marcos-iov:
Have a look at the code smells, there ara many unused imports and other stuff of simple solution

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
julia-zack: yes good catch. should we rename all the "deserialize / serialize Federation..." to "deserialize / serialize StandardMultisigFederation" also?

rskj-core/src/test/java/co/rsk/peg/FederationSupportTest.java
julia-zack: it throws and error if not

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
julia-zack: yes, thanks

rskj-core/src/main/java/co/rsk/peg/Federation.java
julia-zack: nope, my bad. thanks!

rskj-core/src/main/java/co/rsk/peg/ErpFederation.java
julia-zack: ok!! wasn't sure

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
marcos-iov: If that's what they do, yes

marcos-iov: I would also change the return type of this function, to return a StandarMultisigFederation. You are expecting exactly that, not any generic Federation

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -168,25 +152,15 @@ public boolean equals(Object other) {
...
-            this.btcParams.equals(otherFederation.btcParams) &&
-            this.members.equals(otherFederation.members) &&
-            this.getRedeemScript().equals(otherFederation.getRedeemScript());
+        return this.getAddress().equals(otherFederation.getAddress());
     }
 
     @Override
     public int hashCode() {
         // Can use java.util.Objects.hash since all of Instant, int and List<BtcECKey> have
         // well-defined hashCode()s
```
marcos-iov, line 161: I think this comment is no longer relevant

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -168,25 +152,15 @@ public boolean equals(Object other) {
...
         return Objects.hash(
```
marcos-iov, line 162:
```suggestion
        return getAddress().hashCode();
```
Would that make sense?

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * @deprecated This class represents a Legacy ERP Federation that is non-standard.
+ *             It has been deprecated but it must be kept because of backwards compatibility.
+ */
+
+@Deprecated
+public class LegacyErpFederation extends ErpFederation {
+    private static final Logger logger = LoggerFactory.getLogger(LegacyErpFederation.class);
+    protected static final byte[] ERP_TESTNET_REDEEM_SCRIPT_BYTES = Hex.decode("6453210208f40073a9e43b3e9103acec79767a6de9b0409749884e989960fee578012fce210225e892391625854128c5c4ea4340de0c2a70570f33db53426fc9c746597a03f42102afc230c2d355b1a577682b07bc2646041b5d0177af0f98395a46018da699b6da210344a3c38cd59afcba3edcebe143e025574594b001700dec41e59409bdbd0f2a0921039a060badbeb24bee49eb2063f616c0f0f0765d4ca646b20a88ce828f259fcdb955670300cd50b27552210216c23b2ea8e4f11c3f9e22711addb1d16a93964796913830856b568cc3ea21d3210275562901dd8faae20de0a4166362a4f82188db77dbed4ca887422ea1ec185f1421034db69f2112f4fb1bb6141bf6e2bd6631f0484d0bd95b16767902c9fe219d4a6f5368ae");
```
marcos-iov, line 25: Is this protected for any reason? Shouldn't it be private?

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        return redeemScript;
+    }
+
+    @Override
+    public Script getStandardRedeemScript() {
+        if (standardRedeemScript == null) {
+            // i think is not quite right to just "fix" the standardRedeemScript in the buggy fed.
+            // the real case is that the OP_CHECKMULTISIG is not before the OP_ELSE... as the method adds.
```
marcos-iov, line 75: Let's remove this comment

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
```diff
@@ -0,0 +1,94 @@
...
+            standardRedeemScript = ErpFederationRedeemScriptParser.extractStandardRedeemScript(
+                getRedeemScript().getChunks()
+            );
+        }
+        return standardRedeemScript;
+    }
+
+    private void validateRedeemScript() {
+        if (activations.isActive(ConsensusRule.RSKIP293) &&
+            this.redeemScript.equals(new Script(ERP_TESTNET_REDEEM_SCRIPT_BYTES))) {
```
marcos-iov, line 85:
```suggestion
            this.getRedeemScript().equals(new Script(ERP_TESTNET_REDEEM_SCRIPT_BYTES))) {
```
This should do it right? And we shouldn't need to call `getRedeemScript` in the constructor that looks a bit weird

rskj-core/src/test/java/co/rsk/peg/ErpFederationTest.java
```diff
@@ -406,39 +406,9 @@ void testEquals_differentNumberOfMembers() {
...
-            federation.getCreationTime(),
-            federation.getCreationBlockNumber() + 1,
-            federation.getBtcParams(),
-            federation.getErpPubKeys(),
-            federation.getActivationDelay(),
-            activations
-        );
-
-        Assertions.assertNotEquals(federation, otherFederation);
-    }
```
marcos-iov, line 437: Should we remove these? We could assert that with different creation time and block number but same address they are still equal

rskj-core/src/test/java/co/rsk/peg/ErpFederationTest.java
```diff
@@ -827,7 +797,7 @@ private void validateErpRedeemScript(
         assertEquals(ScriptOpCodes.OP_NOTIF, script[index++]);
 
         // Next byte should equal M, from an M/N multisig
-        int m = defaultMultisigKeys.size() / 2 + 1;
+        int m = defaultMultisigKeys.size() / 2 + 1; // da 8
```
marcos-iov, line 800: Remove the comment please

rskj-core/src/test/java/co/rsk/peg/StandardMultisigFederationTest.java
```diff
@@ -22,55 +22,55 @@
...
-                BtcECKey.fromPrivate(BigInteger.valueOf(200)),
-                BtcECKey.fromPrivate(BigInteger.valueOf(300)),
-                BtcECKey.fromPrivate(BigInteger.valueOf(400)),
-                BtcECKey.fromPrivate(BigInteger.valueOf(500)),
-                BtcECKey.fromPrivate(BigInteger.valueOf(600)),
-        }).sorted(BtcECKey.PUBKEY_COMPARATOR).collect(Collectors.toList());
-
-        rskPubKeys = Stream.of(101, 201, 301, 401, 501, 601)
-                .map(i -> ECKey.fromPrivate(BigInteger.valueOf(i)))
+        sortedPublicKeys = btcPublicKeys.stream()
```
marcos-iov, line 69: Do we need this? The ones above are already sorted so it's basically the same

rskj-core/src/test/java/co/rsk/peg/StandardMultisigFederationTest.java
```diff
@@ -121,52 +121,20 @@ void redeemScript() {
...
-        }
+    void testArrayEquals_P2SHScript() {
+        ScriptBuilder p2shScriptBuilder = new ScriptBuilder();
+        p2shScriptBuilder.op(ScriptOpCodes.OP_HASH160);
+        p2shScriptBuilder.data(Hex.decode("57f76bf3ab818811c740929ac7a5e3ef8c7a34b9"));
+        p2shScriptBuilder.op(ScriptOpCodes.OP_EQUAL);
+
+        Script p2shScript = p2shScriptBuilder.build();
+
+        Assertions.assertArrayEquals(federation.getP2SHScript().getProgram(), p2shScript.getProgram());
```
marcos-iov, line 132:
```suggestion
        Assertions.assertEquals(federation.getP2SHScript(), p2shScript);
```
That should do it, right?

rskj-core/src/test/java/co/rsk/peg/StandardMultisigFederationTest.java
```diff
@@ -121,52 +121,20 @@ void redeemScript() {
...
-                    Assertions.assertArrayEquals(sortedPublicKeys.get(i).getPubKey(),
-                            publicKeys.get(i).getPubKey());
-                }
-                return new Script(new byte[]{(byte) 0xaa});
-            });
-            Assertions.assertArrayEquals(federation.getP2SHScript().getProgram(), new byte[]{(byte) 0xaa});
-            // Make sure the script creation happens only once
-            Assertions.assertEquals(1, calls.size());
-        }
+    void testArrayEquals_P2SHScript() {
```
marcos-iov, line 124:
```suggestion
    void getP2SHScript() {
```

marcos-iov: I wouldn't add the assertion as part of the test name

rskj-core/src/test/java/co/rsk/peg/ErpFederationTest.java
julia-zack: yes they are already removed

rskj-core/src/test/java/co/rsk/peg/StandardMultisigFederationTest.java
julia-zack: yesss i agree! but it was done that way and i didn't focus on refactoring the tests so much in this pr :)

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
julia-zack: that breaks some things so i should take a better look at it

julia-zack: for example, this code inside testGetNewFederationPostMultiKey function from BridgeStorageProviderTest throws an exception:

```
        try (MockedStatic<BridgeSerializationUtils> bridgeSerializationUtilsMocked = mockStatic(BridgeSerializationUtils.class)) {
            bridgeSerializationUtilsMocked.when(() -> BridgeSerializationUtils.deserializeStandardMultisigFederation(
                any(byte[].class),
                any(NetworkParameters.class))
            ).then((InvocationOnMock invocation) -> {
                deserializeCalls.add(0);
                byte[] data = invocation.getArgument(0);
                NetworkParameters networkParameters = invocation.getArgument(1);
                // Make sure we're deserializing what just came from the repo with the correct BTC context
                assertArrayEquals(new byte[]{(byte) 0xaa}, data);
                assertEquals(networkParameters, bridgeConstants.getBtcParams());
                return federation;
            });
```

because the _deserializeStandardMultisigFederation_ returns a StandardMultisigFederation, that cannot be cast to a ErpFederation

julia-zack: we may reconsider changing the tests

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -333,10 +333,10 @@ public Federation getNewFederation() {
                     return null;
                 }
                 if (storageVersion.isPresent()) {
-                    return deserializeFederationAccordingToVersion(data, storageVersion.get(), bridgeConstants);
+                    return deserializeFederationAccordingToVersion(data, storageVersion.get(), bridgeConstants); // ACA ES EL ERROR
```
marcos-iov, line 336: Remove this comment please

rskj-core/src/main/java/co/rsk/peg/ErpFederation.java
```diff
@@ -111,60 +54,20 @@ public Script getStandardP2SHScript() {
...
-            logger.debug("[validateRedeemScript] {}", message);
-            throw new FederationCreationException(message);
+        long maxCsvValue = ErpFederationRedeemScriptParser.MAX_CSV_VALUE;
+        if (activationDelay <= 0 || activationDelay > maxCsvValue) {
+            String message = String.format(
+                "Provided csv value %d must be larger than 0 and lower than %d",
+                activationDelay,
+                maxCsvValue
+            );
+            throw new VerificationException(message);
```
marcos-iov, line 70:
```suggestion
            throw new FederationCreationException(message);
```
That one seems more appropiate, `VerificationException` is a bitcoinj-thin exception

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -58,6 +56,9 @@ public Federation(List<FederationMember> members, Instant creationTime, long cre
         this.creationTime = creationTime.truncatedTo(ChronoUnit.MILLIS);
         this.creationBlockNumber = creationBlockNumber;
         this.btcParams = btcParams;
+
+        // TODO
+        // validateFederationValues();
```
marcos-iov, line 61: Let's remove this and create a ticket in Jira to handle this

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -168,25 +154,22 @@ public boolean equals(Object other) {
...
-        return Objects.hash(
-            getCreationTime(),
-            this.creationBlockNumber,
-            getNumberOfSignaturesRequired(),
-            getBtcPublicKeys()
-        );
+        return getAddress().hashCode();
     }
+
+/*  TODO implement this check
```
marcos-iov, line 165: Please remove

rskj-core/src/test/java/co/rsk/peg/FederationSupportTest.java
marcos-iov: Shouldn't throw anymore, now that it was changed in BridgeConstants

rskj-core/src/test/java/co/rsk/peg/StandardMultisigFederationTest.java
marcos-iov: Did you try this?

rskj-core/src/test/java/co/rsk/peg/StandardMultisigFederationTest.java
marcos-iov: Did you see this one?

marcos-iov (approved):

kelvinator07 (approved):

PR #2146, opened 2023-10-09
Make getBtcPublicKeyIndex return an Optional value

marcos-iov:


rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
```diff
@@ -35,6 +35,7 @@ public LegacyErpFederation(
 
         super(members, creationTime, creationBlockNumber, btcParams, erpPubKeys, activationDelay, activations);
 
+        validateRedeemScriptSize();
         validateRedeemScript();
```
marcos-iov, line 39: These 2 methods sound a bit redundant. I'd put everything inside `validateRedeemScript()`

rskj-core/src/test/java/co/rsk/peg/BridgeSupportGetTransactionTypeTest.java
```diff
@@ -49,9 +50,25 @@ class BridgeSupportGetTransactionTypeTest {
     void getTransactionType_sentFromP2SHErpFed() {
         ActivationConfig.ForBlock activations = ActivationConfigsForTest.hop400().forBlock(0);
 
+        // should hardcode the pkeys because the redeem script size validation threw error
```
marcos-iov, line 53: I'd remove this comment, does add much value to the tests it's only explaining the change done here

rskj-core/src/test/java/co/rsk/peg/LegacyErpFederationTest.java
```diff
@@ -30,10 +28,12 @@
 import java.nio.file.Paths;
 import java.time.Instant;
 import java.time.ZonedDateTime;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
+import com.google.common.collect.Lists;
```
marcos-iov, line 36: Is this used? In general we try to stay away from using external dependencies

rskj-core/src/test/java/co/rsk/peg/LegacyErpFederationTest.java
```diff
@@ -54,16 +55,102 @@ class LegacyErpFederationTest {
...
+
+    @Test
+    void createValidLegacyErpFederation_exactMaxCsvValue() {
+        activationDelayValue = ErpFederationRedeemScriptParser.MAX_CSV_VALUE;
+        assertDoesNotThrow(this::createDefaultLegacyErpFederation);
+    }
+
+    @Test
+    void createInvalidLegacyErpFederation_aboveMaxScriptSigSize() {
+        BtcECKey federator9PublicKey = BtcECKey.fromPublicOnly(
```
marcos-iov, line 141: Let's have the default one using 9 memebers, so we test with the maximum allowed. And here add only one new member and show that it fails

rskj-core/src/test/java/co/rsk/peg/LegacyErpFederationTest.java
```diff
@@ -75,89 +162,126 @@ void getActivationDelay() {
...
-        assertEquals(expectedAddress, fedAddress);
+        ErpFederation otherFederation = createDefaultLegacyErpFederation();
+        Assertions.assertNotEquals(federation, otherFederation);
     }
 
     @Test
-    void getErpPubKeys_compressed_public_keys() {
-        assertEquals(emergencyKeys, federation.getErpPubKeys());
+    void getP2SHScriptAndAddress() {
+        // standard and emergency keys from real legacy erp fed in testnet
```
marcos-iov, line 227: Add the address here as well to have for reference

rskj-core/src/test/java/co/rsk/peg/LegacyErpFederationTest.java
```diff
@@ -170,52 +294,57 @@ void getErpRedeemScript_compareOtherImplementation() throws IOException {
...
-                generatedScript.emergencyFed,
-                generatedScript.timelock,
-                activations
-            );
-
-            Script rskjScript = erpFederation.getRedeemScript();
-            Script alternativeScript = generatedScript.script;
-
-            assertEquals(alternativeScript, rskjScript);
+            if (generatedScript.script.getProgram().length <= 520) {
```
marcos-iov, line 297:
```suggestion
            // Skip test cases where the redeem script exceeds the maximum size
            if (generatedScript.script.getProgram().length <= 520) {
```

rskj-core/src/test/java/co/rsk/peg/LegacyErpFederationTest.java
```diff
@@ -54,16 +55,102 @@ class LegacyErpFederationTest {
...
+        BtcECKey federator2PublicKey = BtcECKey.fromPublicOnly(Hex.decode("0355a2e9bf100c00fc0a214afd1bf272647c7824eb9cb055480962f0c382596a70"));
+        BtcECKey federator3PublicKey = BtcECKey.fromPublicOnly(Hex.decode("02566d5ded7c7db1aa7ee4ef6f76989fb42527fcfdcddcd447d6793b7d869e46f7"));
+        BtcECKey federator4PublicKey = BtcECKey.fromPublicOnly(Hex.decode("0294c817150f78607566e961b3c71df53a22022a80acbb982f83c0c8baac040adc"));
+        BtcECKey federator5PublicKey = BtcECKey.fromPublicOnly(Hex.decode("0372cd46831f3b6afd4c044d160b7667e8ebf659d6cb51a825a3104df6ee0638c6"));
+        BtcECKey federator6PublicKey = BtcECKey.fromPublicOnly(Hex.decode("0340df69f28d69eef60845da7d81ff60a9060d4da35c767f017b0dd4e20448fb44"));
+        BtcECKey federator7PublicKey = BtcECKey.fromPublicOnly(Hex.decode("02ac1901b6fba2c1dbd47d894d2bd76c8ba1d296d65f6ab47f1c6b22afb53e73eb"));
+        BtcECKey federator8PublicKey = BtcECKey.fromPublicOnly(Hex.decode("031aabbeb9b27258f98c2bf21f36677ae7bae09eb2d8c958ef41a20a6e88626d26"));
+        standardKeys = Lists.newArrayList(
```
marcos-iov, line 67:
```suggestion
        standardKeys = Arrays.asList(
```

rskj-core/src/test/java/co/rsk/peg/P2shErpFederationTest.java
```diff
@@ -1,68 +1,246 @@
...
+
+    @Test
+    void createValidP2shErpFederation_exactMaxCsvValue()  {
+        activationDelayValue = ErpFederationRedeemScriptParser.MAX_CSV_VALUE;
+        assertDoesNotThrow(this::createDefaultP2shErpFederation);
+    }
+
+    @Test
+    void createInvalidFederation_aboveMaxScriptSigSize() {
+        BtcECKey federator9PublicKey = BtcECKey.fromPublicOnly(
```
marcos-iov, line 136: Same as in LegacyErpFederation, let's use the max by default and make it fail by adding only one extra member

rskj-core/src/test/java/co/rsk/peg/P2shErpFederationTest.java
```diff
@@ -142,22 +326,68 @@ void getPowPegAddress_mainnet() {
...
+        try {
+            rawRedeemScripts = Files.readAllBytes(Paths.get("src/test/resources/redeemScripts_p2shERP.json"));
+        } catch (IOException e) {
+            System.out.println("redeemScripts_p2shERP.json file not found");
+            throw(e);
+        }
+
+        RawGeneratedRedeemScript[] generatedScripts = new ObjectMapper().readValue(rawRedeemScripts, RawGeneratedRedeemScript[].class);
+        for (RawGeneratedRedeemScript generatedScript : generatedScripts) {
+            if (generatedScript.script.getProgram().length <= 520) {
```
marcos-iov, line 375: Same comment here as in LegacyErpFederation

rskj-core/src/test/java/co/rsk/peg/StandardMultisigFederationTest.java
```diff
@@ -18,65 +18,96 @@
 
 package co.rsk.peg;
 
+import static org.ethereum.TestUtils.assertThrows;
```
marcos-iov, line 21: Probably the wrong import this one

rskj-core/src/test/java/co/rsk/peg/StandardMultisigFederationTest.java
```diff
@@ -97,6 +93,23 @@ void setUp() {
             .collect(Collectors.toList());
     }
 
+    @Test
+    void createInvalidFederation_aboveMaxScriptSigSize() {
```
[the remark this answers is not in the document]
julia-zack: i don't think is necessary since we are instantiating the genesis fed that has 15 members (so it works)

rskj-core/src/test/java/co/rsk/peg/LegacyErpFederationTest.java
julia-zack: is used but can be replaced. thanks

rskj-core/src/test/java/co/rsk/peg/LegacyErpFederationTest.java
```diff
@@ -170,52 +318,58 @@ void getErpRedeemScript_compareOtherImplementation() throws IOException {
...
-                generatedScript.timelock,
-                activations
-            );
-
-            Script rskjScript = erpFederation.getRedeemScript();
-            Script alternativeScript = generatedScript.script;
-
-            assertEquals(alternativeScript, rskjScript);
+            // Skip test cases where the redeem script exceeds the maximum size
+            if (generatedScript.script.getProgram().length <= 520) {
```
marcos-iov, line 322:
```suggestion
            if (generatedScript.script.getProgram().length <= Standardness.MAX_SCRIPT_ELEMENT_SIZE) {
```

rskj-core/src/test/java/co/rsk/peg/P2shErpFederationTest.java
```diff
@@ -142,22 +320,69 @@ void getPowPegAddress_mainnet() {
...
+            rawRedeemScripts = Files.readAllBytes(Paths.get("src/test/resources/redeemScripts_p2shERP.json"));
+        } catch (IOException e) {
+            System.out.println("redeemScripts_p2shERP.json file not found");
+            throw(e);
+        }
+
+        RawGeneratedRedeemScript[] generatedScripts = new ObjectMapper().readValue(rawRedeemScripts, RawGeneratedRedeemScript[].class);
+        for (RawGeneratedRedeemScript generatedScript : generatedScripts) {
+            // Skip test cases where the redeem script exceeds the maximum size
+            if (generatedScript.script.getProgram().length <= 520) {
```
marcos-iov, line 370:
```suggestion
            if (generatedScript.script.getProgram().length <= Standardness.MAX_SCRIPT_ELEMENT_SIZE) {
```

rskj-core/src/main/java/co/rsk/rules/Standardness.java
```diff
@@ -0,0 +1,8 @@
+package co.rsk.rules;
+
+public class Standardness {
+
+    private Standardness() {
+    }
+    public static final int MAX_SCRIPT_ELEMENT_SIZE = 520;
```
kelvinator07, line 7:
```suggestion
    public static final int MAX_SCRIPT_ELEMENT_SIZE = 520;
    private Standardness() {
    }
```

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
```diff
@@ -83,6 +83,17 @@ private void validateRedeemScript() {
...
+        // we have to check if the size of every script inside the scriptSig is not above the maximum
+        // this scriptSig contains the signatures, the redeem script and some other bytes
+        // so it is ok to just check the redeem script size
+
+        int bytesFromRedeemScript = getRedeemScript().getProgram().length;
+
+        if (bytesFromRedeemScript > Standardness.MAX_SCRIPT_ELEMENT_SIZE) {
+            String message = "Unable to create LegacyErpFederation. The redeem script size is above the maximum allowed.";
+            throw new FederationCreationException(message);
+        }
```
kelvinator07, line 96: We Are Doing This Same Validation For All Federation Types.. Maybe Move This To A Utils File Or The Base Class?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1368,7 +1375,26 @@ public void addSignature(BtcECKey federatorPublicKey, List<byte[]> signatures, b
...
+
+        if (activeFederation.hasBtcPublicKey(federatorPublicKey)) {
+            return Optional.of(activeFederation);
+        }
+        if (retiringFederation != null && retiringFederation.hasBtcPublicKey(federatorPublicKey)) {
+            return Optional.of(retiringFederation);
+        }
+
+        return Optional.empty();
+    }
```
nathanieliov, line 1390: This method can be an util method. This way, we reduce the size of `BridgeSupport` class and get it less overloaded.

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
nathanieliov: I agree on moving it to a util file where you have the fed validation methods because you may have some federations that have validations in commons, thus you don't need to repeat the same method if you have it in a util class.

rskj-core/src/test/java/co/rsk/peg/LegacyErpFederationTest.java
```diff
@@ -54,16 +56,101 @@ class LegacyErpFederationTest {
...
+            emergencyKeys,
+            activationDelayValue,
+            activations
+        );
+    }
+
+    @Test
+    void createInvalidLegacyErpFederation_nullErpKeys() {
+        emergencyKeys = null;
+        assertThrows(FederationCreationException.class, this::createDefaultLegacyErpFederation);
```
nathanieliov, line 103: Can you please also assert the message returned is correct? see:  https://www.baeldung.com/junit-assert-exception

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
julia-zack: the reason why it is not in the base class is because this validation is no accurate for all federations but for legacy ones.
actually, its purpose is to have the size of the scriptsig validated, not the redeem script itself. but in legacy we have the redeemscript inside the scriptsig.
segwit (compatible or native) will be another erp federation that has a fixed size for the scriptsig, so this validation wont be right for it.

but i think that moving it into a utils file is good though, thanks!

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1337,27 +1341,30 @@ private void adjustBalancesIfChangeOutputWasDust(BtcTransaction btcTx, Coin sent
...
-        Federation federation =
-                activeFederation.hasBtcPublicKey(federatorPublicKey) ?
-                        activeFederation :
-                        (retiringFederation != null && retiringFederation.hasBtcPublicKey(federatorPublicKey) ?
-                                retiringFederation:
-                                null);
-
-        if (federation == null) {
-            logger.warn("Supplied federator public key {} does not belong to any of the federators.", federatorPublicKey);
+        Optional<Federation> optionalFederation = getFederationFromPublicKey(federatorPublicKey);
```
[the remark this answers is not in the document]
julia-zack: this one is all marcos!

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: i think this class deserves a whole refactor

nathanieliov: Yes, it does. But the only way possible so far is by small steps at once. Consider moving it to a fed utils class

rskj-core/src/main/java/co/rsk/peg/ErpFederation.java
```diff
@@ -2,19 +2,20 @@
...
 import java.util.Collections;
 import java.util.List;
 import org.ethereum.config.blockchain.upgrades.ActivationConfig;
+import static co.rsk.peg.FederationCreationException.Reason.INVALID_CSV_VALUE;
+import static co.rsk.peg.FederationCreationException.Reason.NULL_OR_EMPTY_EMERGENCY_KEYS;
 
 public abstract class ErpFederation extends Federation {
     protected final List<BtcECKey> erpPubKeys;
     protected final long activationDelay;
+    public static final long MAX_CSV_VALUE = 65_535L; // 2^16 - 1, since bitcoin will interpret up to 16 bits as the CSV value
```
marcos-iov, line 18: Could it be protected instead of public?

rskj-core/src/main/java/co/rsk/peg/utils/FederationUtils.java
```diff
@@ -0,0 +1,16 @@
+package co.rsk.peg.utils;
+
+import co.rsk.bitcoinj.script.Script;
+import co.rsk.peg.bitcoin.Standardness;
+
+public class FederationUtils {
+
+    public static boolean isScriptSizeValid(Script script) {
+        // we have to check if the size of the script does not exceed the maximum size allowed
```
nathanieliov, line 9:
```suggestion
        // Check if the size of the script does not exceed the maximum size allowed
```

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1389,7 +1415,12 @@ private void processSigning(BtcECKey federatorPublicKey, List<byte[]> signatures
             try {
                 sig = BtcECKey.ECDSASignature.decodeFromDER(signatures.get(i));
             } catch (RuntimeException e) {
```
nathanieliov, line 1417: Now, this exception is not being used either for logging the error message or rethrowing the exception. What about logging the exception message and the stack trace?

rskj-core/src/main/java/co/rsk/peg/utils/FederationUtils.java
```diff
@@ -0,0 +1,16 @@
+package co.rsk.peg.utils;
```
marcos-iov, line 1: I wouldn't put this into a `utils` package. It's not a good idea to group classes under `utils` or `exception` packages, classes that are related to each other should be together. Ideally we'd have a `federation` package with all federation related classes in it, but in the time being let's just keep it under co.rsk.peg package

rskj-core/src/main/java/co/rsk/peg/utils/FederationUtils.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class FederationUtils {
+
+    public static boolean isScriptSizeValid(Script script) {
+        // we have to check if the size of the script does not exceed the maximum size allowed
+        int bytesFromScript = script.getProgram().length;
+        return bytesFromScript <= Standardness.MAX_SCRIPT_ELEMENT_SIZE;
+    }
+
+    private FederationUtils() {
```
marcos-iov, line 14: Check the code smells, this constructor should go at the top right after the class definition

rskj-core/src/main/java/co/rsk/peg/utils/FederationUtils.java
```diff
@@ -0,0 +1,16 @@
+package co.rsk.peg.utils;
+
+import co.rsk.bitcoinj.script.Script;
+import co.rsk.peg.bitcoin.Standardness;
+
+public class FederationUtils {
+
+    public static boolean isScriptSizeValid(Script script) {
```
marcos-iov, line 8: It's better to have this method void and throw an exception (FederationCreationException could be) if the validation is not successful. That way we have less `if` statements in the code

rskj-core/src/test/java/co/rsk/peg/BridgeSupportGetTransactionTypeTest.java
```diff
@@ -14,6 +14,7 @@
 import co.rsk.config.BridgeRegTestConstants;
 import co.rsk.peg.bitcoin.BitcoinTestUtils;
 import co.rsk.test.builders.BridgeSupportBuilder;
+import com.google.common.collect.Lists;
```
marcos-iov, line 17: Preferable to use Arrays.asList instead of adding an external dependency

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
marcos-iov: Now that we have a FederationUtils class it could probably go in there. Having said that, I think that's another story since it will involve adding tests for this particular method

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
```diff
@@ -75,13 +76,26 @@ public Script getStandardRedeemScript() {
...
+            redeemScript.equals(new Script(LEGACY_ERP_TESTNET_REDEEM_SCRIPT_BYTES))) {
 
             String message = "Unable to create ERP Federation. The obtained redeem script matches the one hardcoded for testnet. "
                 + "This would cause bitcoinj-thin to identify it as invalid";
             logger.debug("[validateRedeemScript] {}", message);
-            throw new FederationCreationException(message);
+            throw new FederationCreationException(message, HARDCODED_LEGACY_ERP_TESTNET_REDEEM_SCRIPT);
+        }
+
+        // we need to check if every stack inside the scriptsig has a valid size.
```
nathanieliov, line 90: Below validation is checking the size of the whole scriptsig, not chunks by chunks.

rskj-core/src/test/java/co/rsk/peg/LegacyErpFederationTest.java
```diff
@@ -75,89 +180,151 @@ void getActivationDelay() {
...
-            redeemScript,
-            activationDelayValue,
-            false
-        );
+    void testEquals_basic() {
+        assertEquals(federation, federation);
+
+        assertNotEquals(null, federation);
+        assertNotEquals(federation, new Object());
+        assertNotEquals("something else", federation);
```
nathanieliov, line 188: This is redundant.

rskj-core/src/test/java/co/rsk/peg/StandardMultisigFederationTest.java
```diff
@@ -99,193 +110,212 @@ void membersImmutable() {
...
-        p2shScriptBuilder.op(ScriptOpCodes.OP_EQUAL);
-
-        Script p2shScript = p2shScriptBuilder.build();
+    void testEquals_basic() {
+        assertEquals(federation, federation);
 
-        Assertions.assertEquals(federation.getP2SHScript(), p2shScript);
+        assertNotEquals(null, federation);
+        assertNotEquals(federation, new Object());
+        assertNotEquals("something else", federation);
```
nathanieliov, line 118: Same here. This is redundant.

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
julia-zack: actually is just checking the redeem script size since is the only stack element that could exceed the limit

julia-zack:
```
// this scriptSig contains the signatures, the redeem script and some other bytes
// so it is ok to just check the redeem script size
```

rskj-core/src/test/java/co/rsk/peg/LegacyErpFederationTest.java
```diff
@@ -75,89 +180,151 @@ void getActivationDelay() {
...
+            p2shScript.getToAddress(networkParameters)
+        );
+        assertEquals(expectedAddress, address);
     }
 
     @Test
     void getErpPubKeys_uncompressed_public_keys() {
         // Public keys used for creating federation, but uncompressed format now
-        List<BtcECKey> uncompressedErpKeys = emergencyKeys
+        emergencyKeys = emergencyKeys
```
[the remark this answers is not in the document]
julia-zack: we have a BeforeEach that creates a LegacyErpFederation that is used in almost every test.
in the ones that we want to change some of the attributes of this federation (like this one), we change it directly.
this way, we can directly call the `createDefault...Federation` with the needed attributes for the test

rskj-core/src/test/java/co/rsk/peg/StandardMultisigFederationTest.java
julia-zack: yes, i agree. but this is a test that was there before and we decided to "just let it there"

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: okidoki

rskj-core/src/main/java/co/rsk/peg/LegacyErpFederation.java
```diff
@@ -37,6 +38,7 @@ public LegacyErpFederation(
         super(members, creationTime, creationBlockNumber, btcParams, erpPubKeys, activationDelay, activations);
 
         validateRedeemScript();
+        isScriptSizeValid(this.getRedeemScript());
```
marcos-iov, line 41: Maybe put this inside `validateRedeemScript`?

rskj-core/src/main/java/co/rsk/peg/FederationUtils.java
```diff
@@ -0,0 +1,22 @@
...
+import co.rsk.bitcoinj.script.Script;
+import co.rsk.peg.bitcoin.Standardness;
+
+import static co.rsk.peg.FederationCreationException.Reason.ABOVE_MAX_SCRIPT_ELEMENT_SIZE;
+
+public class FederationUtils {
+    private FederationUtils() {
+    }
+
+    public static void isScriptSizeValid(Script script) throws FederationCreationException {
```
marcos-iov, line 12: Nit picky, but since it's no longer a boolean method I would not call `is...`. That gives me the impression that it will return true or false. I'd simply call it `validateScriptSize` or something like that

rskj-core/src/main/java/co/rsk/peg/P2shErpFederation.java
```diff
@@ -53,4 +52,8 @@ public final Script getStandardRedeemScript() {
         }
         return standardRedeemScript;
     }
+
+    public void validateRedeemScriptSize() {
```
marcos-iov, line 56: Should be private

rskj-core/src/main/java/co/rsk/peg/StandardMultisigFederation.java
```diff
@@ -53,4 +52,8 @@ public Script getRedeemScript() {
 
         return redeemScript;
     }
+
+    public void validateRedeemScriptSize() {
```
marcos-iov, line 56: Should be private

PR #2147, opened 2023-10-10
Fix security issue in BridgeStorageProvider

julia-zack:


rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -1046,7 +1046,10 @@ private Federation deserializeFederationAccordingToVersion(
                     networkParameters
                 );
             default:
-                throw new IllegalArgumentException("Unknown Federation version: " + version);
+                return BridgeSerializationUtils.deserializeStandardMultisigFederation(
+                    data,
+                    networkParameters
+                );
```
marcos-iov, line 1052:
```suggestion
                // To keep backwards compatibility
                return BridgeSerializationUtils.deserializeStandardMultisigFederation(
                    data,
                    networkParameters
                );
```

marcos-iov (approved):

PR #2150, opened 2023-10-19
Add constants for peg-out tx index

nathanieliov:


rskj-core/src/main/java/co/rsk/config/BridgeDevNetConstants.java
```diff
@@ -168,5 +168,8 @@ public BridgeDevNetConstants(List<BtcECKey> federationPublicKeys) {
         maxInputsPerPegoutTransaction = 10;
 
         numberOfBlocksBetweenPegouts = 360; // 3 hours of RSK blocks (considering 1 block every 30 seconds)
+
+        estimatedPegoutTxIndexBtcActivationHeight = 1; // TODO: TBD. This is an estimation of the btc block number once RSKIP379 is activated.
+        pegoutTxIndexGracePeriodInBtcBlocks = 500; // TODO: TBD. Period of grace in btc blocks to continue using the legacy mechanism to identify btc tx.
```
kelvinator07, line 173: Will These Values Be Different Or Same For Each Networks?

nathanieliov: They will be different for each network.

rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
```diff
@@ -86,6 +86,10 @@ public abstract class BridgeConstants {
 
     protected int numberOfBlocksBetweenPegouts;
 
+    protected long estimatedPegoutTxIndexBtcActivationHeight;
```
marcos-iov, line 89:
```suggestion
    protected long btcHeightWhenPegoutTxIndexActivates; 
```
Just to keep it consistent with the other constant here `btcHeightWhenBlockIndexActivates`

rskj-core/src/main/java/co/rsk/config/BridgeDevNetConstants.java
marcos-iov, line 173: Just an idea, maybe we could define this value as something like 
```
btcToRskMinimumConfirmations * 5
```
What do you think? If that's the case we wouldn't need to have it as a Bridge constant since it would only depend on `btcToRskMinimumConfirmations` that already has it's own value for each network

nathanieliov: I like it. I believe a calculated value using the peg-out minimum is better than creating a constant.

rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
nathanieliov: Done

rskj-core/src/main/java/co/rsk/config/BridgeDevNetConstants.java
nathanieliov: Removed.

rskj-core/src/test/java/co/rsk/config/BridgeConstantsTest.java
```diff
@@ -122,4 +123,22 @@ void test_getMinimumPeginTxValue(BridgeConstants bridgeConstants, boolean isRSKI
...
+    }
+
+    @ParameterizedTest()
+    @MethodSource("bridgeConstantsArgProvider")
+    void test_getBtcHeightWhenPegoutTxIndexActivates(BridgeConstants bridgeConstants){
+        // Act
+        long btcHeightWhenPegoutTxIndexActivates = bridgeConstants.getBtcHeightWhenPegoutTxIndexActivates();
+
+        // assert
+        Assertions.assertTrue(btcHeightWhenPegoutTxIndexActivates > 0);
```
marcos-iov, line 142: Shouldn't you be checking the expected value for each network? This assertion is a bit weak, and we risk forgetting about changing it. Even if you end up changing the value for each network, we'd still have the test done and it will fail after the values are changed so that will also work as a remainder that the test needs to be updated.

rskj-core/src/main/java/co/rsk/config/BridgeConstants.java
```diff
@@ -86,6 +86,8 @@ public abstract class BridgeConstants {
 
     protected int numberOfBlocksBetweenPegouts;
 
+    protected long btcHeightWhenPegoutTxIndexActivates;
```
marcos-iov, line 89: Not a big deal at all, just nitpicking. But an `int` value should probably do it considering the current bitcoin heights. What do you think?

rskj-core/src/test/java/co/rsk/config/BridgeConstantsTest.java
nathanieliov: Cool. I assigned random values to each network and asserted against those random values. Once we set the final real value, the test will fail as expected.

rskj-core/src/test/java/co/rsk/config/BridgeConstantsTest.java
```diff
@@ -122,4 +123,22 @@ void test_getMinimumPeginTxValue(BridgeConstants bridgeConstants, boolean isRSKI
...
+    }
+
+    @ParameterizedTest()
+    @MethodSource("getBtcHeightWhenPegoutTxIndexActivatesArgProvider")
+    void test_getBtcHeightWhenPegoutTxIndexActivates(BridgeConstants bridgeConstants, int expectedValue){
+        // Act
+        int btcHeightWhenPegoutTxIndexActivates = bridgeConstants.getBtcHeightWhenPegoutTxIndexActivates();
+
+        // assert
+        Assertions.assertEquals(btcHeightWhenPegoutTxIndexActivates, expectedValue);
```
marcos-iov, line 142:
```suggestion
        Assertions.assertEquals(expectedValue, btcHeightWhenPegoutTxIndexActivates);
```
expected value goes first :)

nathanieliov: Done.

marcos-iov (approved):

PR #2170, opened 2023-11-07
Add environment contract

rmoreliovlabs:
## Description

Added Environment class to PrecompiledContracts.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/vm/PrecompiledContractArgs.java
+++ b/rskj-core/src/main/java/org/ethereum/vm/PrecompiledContractArgs.java
@@ -1,3 +1,21 @@
+/*
+ * This file is part of RskJ
+ * Copyright (C) 2017 RSK Labs Ltd.
+ * (derived from ethereumJ library, Copyright (c) 2016 <ether.camp>)
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
 package org.ethereum.vm;
 
 import org.ethereum.core.Block;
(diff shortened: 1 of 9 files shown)


rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
```diff
@@ -218,6 +224,17 @@ public abstract static class PrecompiledContract {
         public void init(Transaction tx, Block executionBlock, Repository repository, BlockStore blockStore, ReceiptStore receiptStore, List<LogInfo> logs) {
         }
 
+        public void init(ProgramInvoke programInvoke) {
```
Vovchyk, line 227: do we really need two init() methods in this abstract class? You could override the one below in `Environment`

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
rmoreliovlabs: We could do that, but it would make the code a little less readable and flexible in case we want to add a similar contract.

Vovchyk: could you please elaborate a bit more on how this adds flexible? I was just thinking about re-using the approach used for other precompiles..

rmoreliovlabs: Sure, it is more of a readability issue, imagine we add another contract that requires a different set of arguments, I think it cleaner and easy to read to just add a new init method with just the new arguments. This will also allow us to call the new init from classes that might not have the rest of the old arguments available by just passing the required arguments instead of also passing a lot of null values for the arguments that are not needed, we could also use the init as a factory method manage just the initialization data of an specific precompiled contract, you can think of it as a factory pattern for each init method.

One option would be to group all arguments in a class and use said class as the argument, that way we only need one init method.

Vovchyk: I like more the idea of args grouping, but this will require quite a good portion of refactoring I suppose. In fact, precompiled contracts are not being added that often.

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContractArgs.java
```diff
@@ -0,0 +1,58 @@
+package org.ethereum.vm;
```
Vovchyk, line 1: could you please add a cc header?

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContractArgsBuilder.java
```diff
@@ -0,0 +1,65 @@
+package org.ethereum.vm;
```
Vovchyk, line 1: could you please add a cc header?

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
```diff
@@ -215,7 +221,22 @@ public abstract static class PrecompiledContract {
 
         public abstract long getGasForData(byte[] data);
 
+        @Deprecated
         public void init(Transaction tx, Block executionBlock, Repository repository, BlockStore blockStore, ReceiptStore receiptStore, List<LogInfo> logs) {
```
Vovchyk, line 225:
```suggestion
        public final void init(Transaction tx, Block executionBlock, Repository repository, BlockStore blockStore, ReceiptStore receiptStore, List<LogInfo> logs) {
```
- I think we should mark it as `final` just to prohibit child classes to override this method
- we should probably call the newer version of `init()` in the `TransactionExecutor` and `Program` classes and provide an instance of `ProgramInvoke` in there?

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
```diff
@@ -215,7 +221,22 @@ public abstract static class PrecompiledContract {
 
         public abstract long getGasForData(byte[] data);
 
+        @Deprecated
```
Vovchyk, line 224: could you add a javadoc here which recommends to use new method you added?

Vovchyk (changes requested):

Vovchyk (approved):

PR #2173, opened 2023-11-09
Add Environment contract check and init call to Program

rmoreliovlabs:
## Description

Update the `Program#callToPrecompiledAddress` method to check if the contract beign called is `Environment` and if so, call its `init` method by passing a `PrecompiledContractArgs` instance with just its `programInvoke` variable set.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/vm/program/Program.java
+++ b/rskj-core/src/main/java/org/ethereum/vm/program/Program.java
@@ -1086,6 +1086,10 @@ public class Program {
         return result;
     }
 
+    public ProgramInvoke getInvoke() {
+        return this.invoke;
+    }
+
     public void setRuntimeFailure(RuntimeException e) {
         getResult().setException(e);
     }
@@ -1408,6 +1412,14 @@ public class Program {
             );
 
             contract.init(internalTx, executionBlock, track, this.invoke.getBlockStore(), null, result.getLogInfoList());
+        }
+
+        if (contract instanceof PrecompiledContracts.Environment) {
+            PrecompiledContractArgs args = PrecompiledContractArgsBuilder.builder()
+                    .programInvoke(this.invoke)
+                    .build();
+
+            contract.init(args);
         }
 
         long requiredGas = contract.getGasForData(data);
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/org/ethereum/vm/program/Program.java
```diff
@@ -1410,6 +1414,14 @@ public void callToPrecompiledAddress(MessageCall msg, PrecompiledContract contra
             contract.init(internalTx, executionBlock, track, this.invoke.getBlockStore(), null, result.getLogInfoList());
         }
 
+        if (contract instanceof PrecompiledContracts.Environment) {
+            PrecompiledContractArgs args = PrecompiledContractArgsBuilder.builder()
+                    .programInvoke(this.invoke)
+                    .build();
+
+            contract.init(args);
```
Vovchyk, line 1422: could you please explain why we need this second `init()` call? Can't we just provide the `invoke` instance in the [one](https://github.com/rsksmart/rskj/pull/2173/files#diff-c0381bea471a7d1522a7283019e890ec5e41f74e8eae4ee07eac99d432690d93R1414) above?

rskj-core/src/main/java/org/ethereum/vm/program/Program.java
```diff
@@ -1086,6 +1086,10 @@ public ProgramResult getResult() {
         return result;
     }
 
+    public ProgramInvoke getInvoke() {
```
Vovchyk, line 1089: this one is only needed for tests, isn't it? Can it be market with the visible for testing annotation? Also can it be package-level visibility?

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/vm/program/Program.java
rmoreliovlabs: We will also be using it in TransactionExecutor to call the `init` for the `Environment` contract there too.

Vovchyk (approved):

PR #2176, opened 2023-11-15
Updating how eth_gasPrice is calculated

asoto-iov:
We are updating the logic about how the gas price data returned by the eth_gasPrice rpc method is calculated.

## Description

The new behaviour is: 

1. We need to check [isFeeMarketWorking()](https://github.com/rsksmart/rskj/blob/5c5fe3db8b1ff07c87a7a79cf4a0f7db0f3c6b6d/rskj-core/src/main/java/org/ethereum/listener/GasPriceTracker.java#L128C33-L128C51) in the GasPriceTracker class to figure out if recent blocks are full enough. If the method returns true then go to the item 2 below. Otherwise, eth_gasPrice should return lastBestBlock.getMinimumGasPrice() * minGasPriceMultiplier , where minGasPriceMultiplier is a configuration value and should be 1.1 by default ( 10% increment of min gas price).

2.If a market is working (isFeeMarketWorking() returns true) , then eth_gasPrice should return same results as it returns right now (no changes in this case).

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/facade/EthereumImpl.java
+++ b/rskj-core/src/main/java/org/ethereum/facade/EthereumImpl.java
@@ -28,6 +28,7 @@ import org.ethereum.listener.GasPriceTracker;
 import org.ethereum.net.server.ChannelManager;
 
 import javax.annotation.Nonnull;
+import java.math.BigDecimal;
 
 public class EthereumImpl implements Ethereum {
 
@@ -36,13 +37,15 @@ public class EthereumImpl implements Ethereum {
     private final CompositeEthereumListener compositeEthereumListener;
     private final Blockchain blockchain;
     private final GasPriceTracker gasPriceTracker;
+    private final Double minGasPriceMultiplier;
 
     public EthereumImpl(
             ChannelManager channelManager,
             TransactionGateway transactionGateway,
             CompositeEthereumListener compositeEthereumListener,
             Blockchain blockchain,
-            GasPriceTracker gasPriceTracker) {
+            GasPriceTracker gasPriceTracker,
+            Double minGasPriceMultiplier) {
         this.channelManager = channelManager;
         this.transactionGateway = transactionGateway;
 
@@ -51,6 +54,7 @@ public class EthereumImpl implements Ethereum {
 
         this.gasPriceTracker = gasPriceTracker;
         compositeEthereumListener.addListener(gasPriceTracker);
+        this.minGasPriceMultiplier = minGasPriceMultiplier;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


asoto-iov:
pipeline:run

rmoreliovlabs (approved):
LGTM

rskj-core/src/test/java/org/ethereum/facade/EthereumImplTest.java
```diff
@@ -0,0 +1,46 @@
+package org.ethereum.facade;
```
Vovchyk, line 1: pls add a cc header

rskj-core/src/main/java/org/ethereum/facade/EthereumImpl.java
```diff
@@ -36,13 +37,15 @@ public class EthereumImpl implements Ethereum {
...
+    private final Double minGasPriceMultiplier;
 
     public EthereumImpl(
             ChannelManager channelManager,
             TransactionGateway transactionGateway,
             CompositeEthereumListener compositeEthereumListener,
             Blockchain blockchain,
-            GasPriceTracker gasPriceTracker) {
+            GasPriceTracker gasPriceTracker,
+            Double minGasPriceMultiplier) {
```
Vovchyk, line 48: does it really need to be a `Double`? this could lead to `NullPointerException`'s later when used..

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/facade/EthereumImpl.java
asoto-iov: I didn't want to make it mandatory, so if you want to use the implementation for testing purposes it is not needed to set a random value there. Also the other members are objects that can lead to a NPE as well and the object it is only created once. In any case if you think is better to use the primitive value I'll update it.

Vovchyk: well, `null`'s were a billion-dollar mistake, so that not sure it's a good idea to make that number even bigger ;) . 

Nonetheless, while using the scalar type instead of the class, in tests you can always provide some default / dummy value, eg. `1.0`. In this case your object won't be in a `inconsistent` state, which can crash on some occasions.

Also if the `null`'s are really needed and do not lead to objects being in the "inconsistent" state, then you can always mark them, for instance, with the `@Nullable` annotation, so that static analysis tools can later help with detecting unexpected code execution flows.

asoto-iov: I just wanted to be practical but you are right, so code updated!

asoto-iov:
pipeline:run

asoto-iov:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk (approved):

PR #2177, opened 2023-11-15
Create evaluate and process pegin methods

nathanieliov:


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -441,30 +461,111 @@ BtcBlockStoreWithCache getBtcBlockStore() {
         return btcBlockStore;
     }
 
-    protected PegTxType getTransactionType(BtcTransaction btcTx) {
+    private PeginEvaluationResult evaluatePegin(BtcTransaction btcTx, PeginInformation peginInformation) {
+        if(!activations.isActive(ConsensusRule.RSKIP379)){
+            throw new IllegalStateException("Method cannot be called before RSKIP379.");
+        }
```
marcos-iov, line 467: Why not? Nothing particular about RSKIP379 here

rskj-core/src/main/java/co/rsk/peg/utils/RejectedPeginReason.java
```diff
@@ -4,7 +4,9 @@ public enum RejectedPeginReason {
     PEGIN_CAP_SURPASSED(1),
     LEGACY_PEGIN_MULTISIG_SENDER(2),
     LEGACY_PEGIN_UNDETERMINED_SENDER(3),
-    PEGIN_V1_INVALID_PAYLOAD(4);
+    PEGIN_V1_INVALID_PAYLOAD(4),
+    NO_UTXO_OR_UTXO_BELOW_MINIMUM(5),
```
marcos-iov, line 8: Having an `OR` in the name sounds a bit bad to me. It's like it's used for 2 different things (which it is actually). But maybe we could think of a different name

INVALID_AMOUNT?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -441,30 +461,111 @@ BtcBlockStoreWithCache getBtcBlockStore() {
...
+        if (peginInformation.getProtocolVersion() == 0) {
+            switch (peginInformation.getSenderBtcAddressType()) {
+                case P2PKH:
+                case P2SHP2WPKH:
+                    return new PeginEvaluationResult(PeginProcessAction.CAN_BE_REGISTERED);
+                case P2SHMULTISIG:
+                case P2SHP2WSH:
+                    return new PeginEvaluationResult(PeginProcessAction.CAN_BE_REFUNDED, LEGACY_PEGIN_MULTISIG_SENDER);
+                default:
+                    return new PeginEvaluationResult(PeginProcessAction.CANNOT_BE_PROCESSED, LEGACY_PEGIN_UNDETERMINED_SENDER);
```
marcos-iov, line 500: If the sender could not be determined then the pegin information parser would have thrown an exception, right?

In that case we are not expecting to reach this point. I guess we should throw an exception if this happens

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -441,30 +461,111 @@ BtcBlockStoreWithCache getBtcBlockStore() {
...
+            }
+        } else if (peginInformation.getProtocolVersion() == 1) {
+            return new PeginEvaluationResult(PeginProcessAction.CAN_BE_REGISTERED);
+        }
```
marcos-iov, line 504: I'd move this on top like this
```suggestion
        if (peginInformation.getProtocolVersion() == 1) {
            return new PeginEvaluationResult(PeginProcessAction.CAN_BE_REGISTERED);
        }
        if (peginInformation.getProtocolVersion() == 0) {
        ....
```
Less `else` and possible paths in the code.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -441,30 +461,111 @@ BtcBlockStoreWithCache getBtcBlockStore() {
...
+        switch (peginEvaluationResult.getPeginProcessAction()) {
+            case CAN_BE_REGISTERED:
+                executePegIn(btcTx, peginInformation, totalAmount);
+                markTxAsProcessed(btcTx);
+                break;
+            case CAN_BE_REFUNDED:
+                rejectedPeginReason.ifPresent(rejectedReason -> eventLogger.logRejectedPegin(btcTx, rejectedReason));
+                generateRejectionRelease(btcTx, peginInformation.getBtcRefundAddress(), rskTx, totalAmount);
+                markTxAsProcessed(btcTx);
+                throw new RegisterBtcTransactionException(String.format("Error while trying to parse peg-in information for tx %s", btcTx.getHash()));
```
marcos-iov, line 537: Why throw this exception here? The transaction was refunded, why is an exception? I would simply add a log for each case stating what happened

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -441,30 +461,111 @@ BtcBlockStoreWithCache getBtcBlockStore() {
...
+            case CANNOT_BE_PROCESSED:
+                handleUnprocessableBtcTx(btcTx, peginInformation, rejectedPeginReason);
+                break;
+            default:
+                if (rejectedPeginReason.isPresent()){
+                    throw new RegisterBtcTransactionException(String.format("Error while trying to process pegin. Error Reason: %s", rejectedPeginReason));
+                } else {
+                    throw new RegisterBtcTransactionException(String.format("Error while trying to process pegin."));
+                }
```
marcos-iov, line 546: This is also something that should never happen, so I'd simply throw an exception here. `evaluatePegin` should always respond with one of the above results.

Maybe the default case is that it's unprocessable?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -441,30 +461,111 @@ BtcBlockStoreWithCache getBtcBlockStore() {
...
+
+        PeginInformation peginInformation = new PeginInformation(
+            btcLockSenderProvider,
+            peginInstructionsProvider,
+            activations
         );
+
+        PeginEvaluationResult peginEvaluationResult = evaluatePegin(btcTx, peginInformation);
+
+        Optional<RejectedPeginReason> rejectedPeginReason = peginEvaluationResult.getRejectedPeginReason();
```
marcos-iov, line 527: This could go inside CAN_BE_REFUNDED case, it's only used there

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -441,30 +461,111 @@ BtcBlockStoreWithCache getBtcBlockStore() {
...
+            default:
+                if (rejectedPeginReason.isPresent()){
+                    throw new RegisterBtcTransactionException(String.format("Error while trying to process pegin. Error Reason: %s", rejectedPeginReason));
+                } else {
+                    throw new RegisterBtcTransactionException(String.format("Error while trying to process pegin."));
+                }
+        }
+    }
+
+    private void handleUnprocessableBtcTx(BtcTransaction btcTx, PeginInformation peginInformation, Optional<RejectedPeginReason> rejectedPeginReason) throws RegisterBtcTransactionException {
```
marcos-iov, line 550: Avoid passing an Optional as parameter. Only call the method if the value is present

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -441,30 +461,111 @@ BtcBlockStoreWithCache getBtcBlockStore() {
...
+        String message = String.format("[tryToProcessPegin] Can't refund. No btc refund address provided and couldn't get sender address either for tx %s.", btcTx.getHash());
+        if (rejectedPeginReason.isPresent()){
+            if (rejectedPeginReason.get() == PEGIN_V1_INVALID_PAYLOAD){
+                if (peginInformation.getProtocolVersion() == 1) {
```
marcos-iov, line 554: This should change if you update `evaluatePegin` to return the proper reason

rskj-core/src/main/java/co/rsk/peg/PegUtils.java
```diff
@@ -37,7 +42,81 @@ public static PegTxType getTransactionTypeUsingPegoutIndex(
...
+        List<Federation> liveFeds = new ArrayList<>();
+        liveFeds.add(activeFederation);
+        if (retiringFederation != null){
+            liveFeds.add(retiringFederation);
+        }
+        Wallet liveFederationsWallet = new BridgeBtcWallet(Context.getOrCreate(bridgeConstants.getBtcParams()), liveFeds);
+
+        // TODO: THIS VALUE IS PENDING TO BE DEFINE. PLEASE UPDATE THIS CONSTANT LATER WITH THE CORRECT VALUE.
+        int btcHeightWhenPegoutTxIndexActivates = bridgeConstants.getBtcHeightWhenPegoutTxIndexActivates();
+        int pegoutTxIndexGracePeriodInBtcBlocks = bridgeConstants.getBtc2RskMinimumAcceptableConfirmations() * 5;
```
marcos-iov, line 92: Let's discuss this, but I think we might consider this becoming a constant. 1000 blocks for example. Or even more

Imagine a pegout is created right before the index activates
- 4000 rsk blocks until it's signed plus 100 BTC blocks untils it's registered
- Consider also that the btc height when the index activates is an estimate, we might fall short

This constant could then be used in the tests as well

rskj-core/src/main/java/co/rsk/peg/pegin/PeginProcessAction.java
```diff
@@ -0,0 +1,7 @@
+package co.rsk.peg.pegin;
+
+public enum PeginProcessAction {
+    CAN_BE_REGISTERED,
+    CAN_BE_REFUNDED,
+    CANNOT_BE_PROCESSED
```
marcos-iov, line 6: Being nit picky here, but the names `CAN_BE` don't convince me. I feel it should be something more direct, like SHOULD_REFUND. For example

```suggestion
    REGISTER,
    REFUND,
    UNPROCESSABLE
```
I'm not convinced about the last one UNPROCESSABLE isn't really an action. Maybe `IGNORE`?

marcos-iov (commented):
Nice work! Left comments

rskj-core/src/main/java/co/rsk/peg/PegUtils.java
nathanieliov: Let's discuss it.

rskj-core/src/main/java/co/rsk/peg/pegin/PeginProcessAction.java
marcos-iov: Well, the enum is called PeginProcessAction so one might expect it to contain actions that are in fact verbs right?

Maybe use a different name for the enum?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
marcos-iov: Oh right. Thanks

rskj-core/src/test/java/co/rsk/peg/BridgeSupportRegisterBtcTransactionTest.java
```diff
@@ -0,0 +1,2553 @@
...
+            rskTx,
+            btcTransaction.bitcoinSerialize(),
+            height,
+            pmt.bitcoinSerialize()
+        );
+
+        // assert
+        verify(provider, times(1)).setHeightBtcTxhashAlreadyProcessed(btcTransaction.getHash(false), rskExecutionBlock.getNumber());
+        Assertions.assertEquals(1, activeFederationUtxos.size());
+        Assertions.assertTrue(retiringFederationUtxos.isEmpty());
```
jeremy-then, line 1747: Is it possible to assert that the `retiringFederationUtxos.size()` is 1 before calling ` bridgeSupport.registerBtcTransaction`?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportRegisterBtcTransactionTest.java
```diff
@@ -0,0 +1,2553 @@
...
+            rskTx,
+            btcTransaction.bitcoinSerialize(),
+            height,
+            pmt.bitcoinSerialize()
+        );
+
+        // assert
+        verify(provider, times(1)).setHeightBtcTxhashAlreadyProcessed(btcTransaction.getHash(false), rskExecutionBlock.getNumber());
+        Assertions.assertEquals(40, activeFederationUtxos.size());
+        Assertions.assertTrue(retiringFederationUtxos.isEmpty());
```
jeremy-then, line 1856: Same here... Is it possible to assert that the retiringFederationUtxos.size() is 40 before calling  bridgeSupport.registerBtcTransaction?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportRegisterBtcTransactionTest.java
```diff
@@ -0,0 +1,2553 @@
...
+            verify(bridgeEventLogger, never()).logPeginBtc(any(), any(), any(), anyInt());
+            verify(provider, never()).setHeightBtcTxhashAlreadyProcessed(any(), anyLong());
+            verify(bridgeEventLogger, times(1)).logRejectedPegin(btcTransaction, INVALID_AMOUNT);
+        }
+
+        // tbd600Activations and after grace period - unknown tx are just ignored
+        else {
+            verify(bridgeEventLogger, never()).logPeginBtc(any(), any(), any(), anyInt());
+            verify(bridgeEventLogger, never()).logRejectedPegin(any(), any());
+            verify(bridgeEventLogger, never()).logUnrefundablePegin(any(), any());
```
jeremy-then, line 2057: Should we do a `verify(provider, never()).setHeightBtcTxhashAlreadyProcessed(any(), anyLong());` here too or no need?

rskj-core/src/test/java/co/rsk/peg/pegininstructions/PeginInstructionsProviderTest.java
```diff
@@ -120,6 +123,20 @@ void buildPeginInstructions_v1_withBtcrefundAddress() throws Exception {
         Assertions.assertEquals(btcRefundAddress, peginInstructionsVersion1.getBtcRefundAddress().get());
     }
 
+    @Test
+    void buildPeginInstructions_bech32() throws Exception {
+        // Arrange
+        String rawTx = "02000000000101cf8b3b2baa22df50b1959d83b2f279ef231fb7cf2009ebfa35644d9e1f0184930200000000fdffffff02706408000000000017a9146e4b5ae85d86e4db0e6e5db09f8c276328cdbf3f87ec5cd40500000000160014d7aa00421cd50c8f282dc7d32992a5e2932a92f3024730440220313d11b58bd2861e4a2f8b2d1b644250569e35c946fcba426de94ed28974f12102207b9074796eea9f823a2d56239f49674c4ec34d40f519b3babd5eba44f94495eb012102acbad9efed3a451f646b93b5fd37a796c1758875cc33eed1626d4ed673d00a9b5abd2600";
```
jeremy-then, line 129: Where did you get this raw tx from? Testnet? Devnet?

rskj-core/src/test/java/co/rsk/peg/PowpegMigrationTest.java
```diff
@@ -858,7 +868,12 @@ private void testPegins(
                 .anyMatch(utxo -> utxo.getHash().equals(peginToFuturePowPegHash));
 
             // This assertion should change when we change peg-in verification
```
jeremy-then, line 870: Should this comment be removed now?

rskj-core/src/test/java/co/rsk/peg/PegUtilsAllUTXOsToFedAreAboveMinimumPeginValueTest.java
```diff
@@ -0,0 +1,559 @@
...
+
+        btcTx.addOutput(minimumPegInTxValue, activeFederation.getAddress()); // With minimum
+        btcTx.addOutput(minimumPegInTxValue.add(Coin.SATOSHI), activeFederation.getAddress()); // Above minimum
+        btcTx.addOutput(minimumPegInTxValue.minus(Coin.SATOSHI), activeFederation.getAddress()); // Below minimum
+
+        Wallet fedWallet = new BridgeBtcWallet(context, Collections.singletonList(activeFederation));
+
+        assertFalse(
+            PegUtils.allUTXOsToFedAreAboveMinimumPeginValue(btcTx, fedWallet, minimumPegInTxValue, activationsForTbd600),
+        "When RSKIP379 is active and there are many utxos to the fed but there is at least one of them below the minimum, it should return true."
```
jeremy-then, line 316:
```suggestion
        "When RSKIP379 is active and there are many utxos to the fed but there is at least one of them below the minimum, it should return false."
```

jeremy-then (commented):
Left some comments. Let's discuss.

PR #2185, opened 2023-11-22
New grace period and event when peg-in below the minimum

marcos-iov:
- Add pegoutTxIndexGracePeriodInBtcBlocks as a Bridge constant
- Emmit event when a peg-in below the minimum is registered in the Bridge

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/PegUtils.java
+++ b/rskj-core/src/main/java/co/rsk/peg/PegUtils.java
@@ -104,8 +104,9 @@ public class PegUtils {
 
         int btcHeightWhenPegoutTxIndexActivates = bridgeConstants.getBtcHeightWhenPegoutTxIndexActivates();
         int pegoutTxIndexGracePeriodInBtcBlocks = bridgeConstants.getPegoutTxIndexGracePeriodInBtcBlocks();
-        int heightToStartUsingPegoutIndex = btcHeightWhenPegoutTxIndexActivates + pegoutTxIndexGracePeriodInBtcBlocks;
-        boolean shouldUsePegoutTxIndex = activations.isActive(ConsensusRule.RSKIP379) && btcTransactionHeight >= heightToStartUsingPegoutIndex;
+        int heightAtWhichToStartUsingPegoutIndex = btcHeightWhenPegoutTxIndexActivates + pegoutTxIndexGracePeriodInBtcBlocks;
+        boolean shouldUsePegoutTxIndex = activations.isActive(ConsensusRule.RSKIP379) &&
+            btcTransactionHeight >= heightAtWhichToStartUsingPegoutIndex;
 
         if (shouldUsePegoutTxIndex){
             return getTransactionTypeUsingPegoutIndex(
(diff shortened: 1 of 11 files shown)


nathanieliov (approved):
LGTM

rskj-core/src/main/java/co/rsk/peg/PegUtils.java
```diff
@@ -102,10 +102,10 @@ static PegTxType getTransactionType(
         }
         Wallet liveFederationsWallet = new BridgeBtcWallet(Context.getOrCreate(bridgeConstants.getBtcParams()), liveFeds);
 
-        // TODO: THIS VALUE IS PENDING TO BE DEFINE. PLEASE UPDATE THIS CONSTANT LATER WITH THE CORRECT VALUE.
         int btcHeightWhenPegoutTxIndexActivates = bridgeConstants.getBtcHeightWhenPegoutTxIndexActivates();
-        int pegoutTxIndexGracePeriodInBtcBlocks = bridgeConstants.getBtc2RskMinimumAcceptableConfirmations() * 5;
-        boolean shouldUsePegoutTxIndex = activations.isActive(ConsensusRule.RSKIP379) && btcTransactionHeight >= btcHeightWhenPegoutTxIndexActivates + pegoutTxIndexGracePeriodInBtcBlocks;
+        int pegoutTxIndexGracePeriodInBtcBlocks = bridgeConstants.getPegoutTxIndexGracePeriodInBtcBlocks();
+        int heightToStartUsingPegoutIndex = btcHeightWhenPegoutTxIndexActivates + pegoutTxIndexGracePeriodInBtcBlocks;
```
jeremy-then, line 107: Nice. I was thinking of a name for this logic here when I saw Natha's pr but couldn't think of one. Now that I see this, it would suggest something like `heightAtWhichToStartUsingPegoutIndex`. No big deal though.
@nathanieliov .

rskj-core/src/test/java/co/rsk/config/BridgeConstantsTest.java
```diff
@@ -139,6 +137,24 @@ void test_getBtcHeightWhenPegoutTxIndexActivates(BridgeConstants bridgeConstants
...
 
         // assert
-        Assertions.assertEquals(expectedValue, btcHeightWhenPegoutTxIndexActivates);
+        assertEquals(expectedValue, btcHeightWhenPegoutTxIndexActivates);
+    }
+
+    private static Stream<Arguments> getPegoutTxIndexGracePeriodInBtcBlocksArgProvider() {
+        return Stream.of(
+            Arguments.of(BridgeMainNetConstants.getInstance(), 4_320),
+            Arguments.of(BridgeTestNetConstants.getInstance(), 4_320),
```
jeremy-then, line 146: Can these values `4_320`, etc., be used from the constants classes? I see they are defined in `BridgeDevNetConstants`. I think it would be more readable this way.
Or at least from a constant in this test file if this value would change for tests.

jeremy-then (approved):
LGTM. Just a couple of optional suggestions.

rskj-core/src/test/java/co/rsk/config/BridgeConstantsTest.java
jeremy-then: Gotcha.

rskj-core/src/main/java/co/rsk/peg/PegUtils.java
marcos-iov: Thanks for the suggestion!

jeremy-then (commented):
LGTM.

PR #2191, opened 2023-11-23
Adding new rsk_getStorageBytesAt JSON-RPC method

asoto-iov:
## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/rpc/Web3RskModule.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/Web3RskModule.java
@@ -19,6 +19,9 @@
 package co.rsk.rpc;
 
 import co.rsk.rpc.modules.rsk.RskModule;
+import org.ethereum.rpc.parameters.BlockRefParam;
+import org.ethereum.rpc.parameters.HexAddressParam;
+import org.ethereum.rpc.parameters.HexNumberParam;
 
 public interface Web3RskModule {
 
@@ -47,4 +50,7 @@ public interface Web3RskModule {
     }
 
     RskModule getRskModule();
+
+    String rsk_getStorageBytesAt(HexAddressParam address, HexNumberParam storageIdx, BlockRefParam blockRefParam);
+
 }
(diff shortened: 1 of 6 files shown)


Vovchyk (changes requested):
we'd probably need to disable this method in [here](https://github.com/rsksmart/rskj/blob/a63d4b8b14fb5a21cb02bad886ce8e69b843d3c9/rskj-core/src/main/resources/reference.conf#L428), as it might be resource intensive in some cases

asoto-iov:
> we'd probably need to disable this method in [here](https://github.com/rsksmart/rskj/blob/a63d4b8b14fb5a21cb02bad886ce8e69b843d3c9/rskj-core/src/main/resources/reference.conf#L428), as it might be resource intensive in some cases

sure, I didn't added as I was testing it locally and then I forgot to do it. Thanks for spotting it!

rskj-core/src/main/java/co/rsk/rpc/Web3RskModule.java
```diff
@@ -47,4 +50,7 @@ default void rsk_flush() {
     }
 
     RskModule getRskModule();
+
+    String rsk_getStorageBytesAt(HexAddressParam address, HexNumberParam storageIdx, BlockRefParam blockRefParam);
```
casiojapi, line 54: just to understand, is there a reason why the default keyword isn't declared?

asoto-iov: I'm only declaring the method and in this case, it is not needed to add default implementation. 

default methods are used when you want to implement a default method behavior within an interface to not break compatibility with older implementations. It is mainly used in high-exposure interfaces but here it is not needed.

Vovchyk (approved):

casiojapi (approved):

rskj-core/src/main/java/co/rsk/rpc/Web3RskModule.java
casiojapi: great, huge thanks angel.

PR #2203, opened 2023-12-04
Create fed format version

julia-zack:


rskj-core/src/main/java/co/rsk/peg/NonStandardErpFederationContextFactory.java
```diff
@@ -1,25 +1,24 @@
-package co.rsk.peg.bitcoin;
+package co.rsk.peg;
 
 import co.rsk.bitcoinj.core.NetworkParameters;
 import org.ethereum.config.blockchain.upgrades.ActivationConfig;
 import org.ethereum.config.blockchain.upgrades.ConsensusRule;
 
-public class NonStandardErpRedeemScriptBuilderFactory {
```
marcos-iov, line 7: I think we should roll back this and keep it as a redeem script builder factory. The format version for this type of fed will always be the same. So it's responsability of the caller, to create a Context object using the non standard format version and the redeem script returned by this factory

PR #2205, opened 2023-12-05
Return properly encoded value by eth_getStorageAt RPC method for non-existing keys

casiojapi:
## Description
Updated eth_getStorageAt to return a 32-byte zero value for non-existing keys, aligning with geth client behavior.

## Motivation and Context
Ensures compatibility with Ethers.js and other clients like Geth, which expect a 32-byte zero value for non-existing keys.

## How Has This Been Tested?
Tested in local development environment; unit tests updated to reflect new return value.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
+++ b/rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
@@ -465,9 +465,8 @@ public class Web3Impl implements Web3 {
     public String eth_getStorageAt(HexAddressParam address, HexNumberParam storageIdx, BlockRefParam blockRefParam) {
         if (blockRefParam.getIdentifier() != null) {
             return this.eth_getStorageAt(address, storageIdx, blockRefParam.getIdentifier());
-        } else {
-            return this.eth_getStorageAt(address, storageIdx, blockRefParam.getInputs());
         }
+        return this.eth_getStorageAt(address, storageIdx, blockRefParam.getInputs());
     }
 
     private String eth_getStorageAt(HexAddressParam address, HexNumberParam storageIdx, Map<String, String> blockRef) {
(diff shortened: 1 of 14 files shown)


rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -487,7 +487,7 @@ private String eth_getStorageAt(HexAddressParam address, HexNumberParam storageI
                     .getStorageValue(addr, DataWord.valueOf(HexUtils.strHexOrStrNumberToByteArray(storageIdx.getHexNumber())));
 
             if (sv == null) {
-                s = "0x0";
+                s = "0x" + String.join("", Collections.nCopies(64, "0"));
```
Vovchyk, line 490: just an idea, could it simply be in the form of the string literal (probably defined as a constant) - same as in the tests, like this? - `0x0000000000000000000000000000000000000000000000000000000000000000`

casiojapi: Yeah, maybe that makes it more readable. I'll update it.

casiojapi:
pipeline:run

casiojapi:
pipeline:run

rmoreliovlabs:
A small suggestion, for the test files you could create a constant like you did for [Web3Impl](https://github.com/rsksmart/rskj/pull/2205/files#diff-6559cff582e5f737e87f69c2dcae4591c06166f215784b542b23747e13c91cc4R89) so you don't have to write the string literal more than once, you could also make the constant in Web3Impl public and use that one. Aside form that nit, it looks good to me.

Vovchyk (approved):

jeremy-then (commented):
Left a couple of comments.

By the way, why is there no documentation for `rsk_` methods in `rskj/doc/rpc/methods`?

PR #2206, opened 2023-12-06
Create formatVersion field in Federation class. Create FederationFactory to create federations without inconsistency between redeem script and format version. Adapt to that change. Move Federation related files to new federation package

julia-zack:
**Context**
This is the second federations refactor made in order to have a segwit compatible powpeg.
Instead of having federations inheritance, we now have a special redeem script builder that has the responsibility to build the redeem script of each of the different federations (_composition over inheritance_).
That was done in the `federations-new-refactor-integration` branch.

But we still had a few places left where we needed to ask for "instances of" the different builders in order to understand about which fed we were talking. Specially the cases related to the storage.
So the proposal is to have a new field, `formatVersion`, for each federation, representing the integer we are saving in the storage.

We will end up having both the erpRedeemScriptBuilder and the formatVersion in ErpFederations (since the standard multisig one does not have a builder). This is not good, since could lead us to inconsistency between them (for example, creating a federation with _standard_ format but _p2sh redeem script builder_).
So the best solution found was to create a FederationFactory that will be in charge of creating the federations. 

**Changes**

1. Create the `formatVersion` field inside Federation class (and `FederationFormatVersion` class to have them organized as an enum) and the method `getFormatVersion()` to get it.
2. Create `/federation` package inside main and test packages
Since we want that the factory is the **only** way of creating them, we had to create a `/federation` package in order to have all of those classes together so the Federation constructor is not reachable from outside.
Inside that package we will have the different types of federations, the FederationFactory and the FederationMember class (this last one to avoid circular dependency with `/peg` package).
3. Create `FederationFactory` class
The `FederationFactory` will have three methods.
`buildStandardMultisigFederation` that will create a `STANDARD_MULTISIG`.
`buildNonStandardErpFederation` that will create a `NON_STANDARD_ERP`.
`buildP2shErpFederation` that will create a `P2SH_ERP`.
Each method receives the attributes needed to create each federation, and creates it with the correct format version. With the erp methods we also ensure that we are setting consistent redeem script builder and format.
It is also important to highlight that the method that builds non-standard feds has to receive the activations, since the non-standard feds have different redeem script builders but the same format version.

**Things to have in mind when reviewing**
0. Lots of changes are related to importing things. The important stuff is only in a few files.
1. Where it said `new StandardMultiSigFederation(arguments)` it will now say `FederationFactory.buildStandardMultisigFederation(arguments)`.
2. Where it said `new ErpFederation(arguments..., P2shRedeemScriptBuilder)` it will now say `FederationFactory.buildP2shErpFederation(arguments)`
3. And the hardest one:
Where it said 
```
builder = NonStandardRedeemScriptBuilderFactory.getNonStandardBuilder(activations, networkParams)
new ErpFederation(arguments..., builder)
```
it will now say `FederationFactory.buildNonStandardErpFederation(arguments..., activations)` (the networkParams are already in federation arguments)


rskj-core/src/main/java/co/rsk/peg/federation/FederationFactory.java
```diff
@@ -0,0 +1,75 @@
...
+import co.rsk.peg.bitcoin.NonStandardErpRedeemScriptBuilderFactory;
+import co.rsk.peg.bitcoin.P2shErpRedeemScriptBuilder;
+import org.ethereum.config.blockchain.upgrades.ActivationConfig;
+
+import java.time.Instant;
+import java.util.List;
+
+import static co.rsk.peg.federation.FederationFormatVersion.*;
+
+public class FederationFactory {
```
marcos-iov, line 16: Tests for this class?

rskj-core/src/main/java/co/rsk/peg/federation/StandardMultisigFederation.java
```diff
@@ -34,13 +35,14 @@
 
 public class StandardMultisigFederation extends Federation {
 
-    public StandardMultisigFederation(
+    protected StandardMultisigFederation(
```
marcos-iov, line 38: Test classes should also be moved into a `federation` package in `tests` package

rskj-core/src/main/java/co/rsk/peg/federation/FederationFactory.java
```diff
@@ -0,0 +1,75 @@
+package co.rsk.peg.federation;
+
+import co.rsk.bitcoinj.core.BtcECKey;
+import co.rsk.bitcoinj.core.NetworkParameters;
+import co.rsk.peg.FederationMember;
```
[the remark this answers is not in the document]
julia-zack: Thanks! you are right

rskj-core/src/main/java/co/rsk/peg/federation/FederationFactory.java
julia-zack: Created a task to take care of that

rskj-core/src/main/java/co/rsk/peg/PendingFederation.java
```diff
@@ -21,9 +21,9 @@
 import co.rsk.bitcoinj.core.BtcECKey;
 import co.rsk.config.BridgeConstants;
 import co.rsk.crypto.Keccak256;
-import co.rsk.peg.bitcoin.ErpRedeemScriptBuilder;
-import co.rsk.peg.bitcoin.NonStandardErpRedeemScriptBuilderFactory;
-import co.rsk.peg.bitcoin.P2shErpRedeemScriptBuilder;
+import co.rsk.peg.federation.Federation;
```
marcos-iov, line 24: Should this class also be moved to federation package? What do you think?

julia-zack: i thought that too! but it uses `BridgeSerializationUtils` class in `getHash` method

rskj-core/src/main/java/co/rsk/peg/federation/FederationMember.java
```diff
@@ -38,10 +38,53 @@
...
+     *               MST_PUB_KEY(M1) <lex MST_PUB_KEY(M2))
+     * 2) M1 == M2 iff BTC_PUB_KEY(M1) ==lex BTC_PUB_KEY(M2) AND
+     *                 RSK_PUB_KEY(M1) ==lex RSK_PUB_KEY(M2) AND
+     *                 MST_PUB_KEY(M1) ==lex MST_PUB_KEY(M2) AND
+     * 3) M1 > M2 otherwise
+     *
+     * where <lex and ==lex is given by negative and zero values (resp.) of the
+     * UnsignedBytes.lexicographicalComparator() comparator.
+     */
+    public static final Comparator<FederationMember> BTC_RSK_MST_PUBKEYS_COMPARATOR = new Comparator<FederationMember>() {
```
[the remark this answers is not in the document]
julia-zack: Ill create a task for that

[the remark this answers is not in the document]
julia-zack: done

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -79,13 +81,14 @@
...
-    private static final int ERP_FEDERATION_FORMAT_VERSION = 2000;
-    private static final int P2SH_ERP_FEDERATION_FORMAT_VERSION = 3000;
+    private static final int STANDARD_MULTISIG_FEDERATION_FORMAT_VERSION = STANDARD_MULTISIG_FEDERATION.getFormatVersion();
+    private static final int NON_STANDARD_ERP_FEDERATION_FORMAT_VERSION = NON_STANDARD_ERP_FEDERATION.getFormatVersion();
+    private static final int P2SH_ERP_FEDERATION_FORMAT_VERSION = P2SH_ERP_FEDERATION.getFormatVersion();
 
     private final TestSystemProperties config = new TestSystemProperties();
     private final ActivationConfig.ForBlock activationsBeforeFork = ActivationConfigsForTest.genesis().forBlock(0L);
     private final ActivationConfig.ForBlock activationsAllForks = ActivationConfigsForTest.all().forBlock(0);
+    private final ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
```
marcos-iov, line 91: I wouldn't declare this object here because it can get mixed with the other 2 activation objects. Since it's basically just used to build ErpFederation objects then we could simply create it when needed

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -866,44 +868,48 @@ void saveOldFederation_postMultikey() {
...
     }
 
     @Test
     void saveOldFederation_postMultikey_RSKIP_353_active_p2sh_erp_fed() {
         ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
         when(activations.isActive(ConsensusRule.RSKIP123)).thenReturn(true);
+        when(activations.isActive(ConsensusRule.RSKIP201)).thenReturn(true);
+        when(activations.isActive(ConsensusRule.RSKIP284)).thenReturn(true);
+        when(activations.isActive(ConsensusRule.RSKIP293)).thenReturn(true);
+        when(activations.isActive(ConsensusRule.RSKIP353)).thenReturn(true);
```
marcos-iov, line 902: For this type of things I would use ActivationConfigsForTest class. It gives you the activations object with the active RSKIPs for a given network upgrade. Simpler and easier to read

marcos-iov (approved):

jeremy-then (approved):
LGTM.

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -1090,30 +1067,30 @@ private Federation deserializeFederationAccordingToVersion(
...
+                data,
+                bridgeConstants
+            );
+        }
+        // To keep backwards compatibility
+        return BridgeSerializationUtils.deserializeStandardMultisigFederation(
+            data,
+            networkParameters
+        );
     }
```
nathanieliov, line 1094: I believe we should move this method out of the bridge storage provider to reduce the amount of code in this class. The same applies to `BridgeSerializationUtils`. We could have two one class for serialization and a second one for deserialization. Let me know what you think.

julia-zack: Agree but I think is better to create another task for that

rskj-core/src/main/java/co/rsk/peg/federation/FederationMemberPubKeysComparator.java
```diff
@@ -0,0 +1,40 @@
...
+     *                 MST_PUB_KEY(M1) ==lex MST_PUB_KEY(M2) AND
+     * 3) M1 > M2 otherwise
+     *
+     * where <lex and ==lex is given by negative and zero values (resp.) of the
+     * UnsignedBytes.lexicographicalComparator() comparator.
+     */
+    private final Comparator<byte[]> comparator = UnsignedBytes.lexicographicalComparator();
+
+    @Override
+    public int compare(FederationMember m1, FederationMember m2) {
```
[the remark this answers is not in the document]
julia-zack: its done

rskj-core/src/main/java/co/rsk/peg/federation/FederationMember.java
```diff
@@ -83,49 +90,6 @@ public static List<FederationMember> getFederationMembersFromKeys(List<BtcECKey>
         return pks.stream().map(pk -> getFederationMemberFromKey(pk)).collect(Collectors.toList());
```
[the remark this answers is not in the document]
julia-zack: done

PR #2215, opened 2023-12-15
Weighted percentile for gasPrice calculation

asoto-iov:
A option to calculate the gas price using weighted percentile considering the gas used is being added to the client. At the same time the legacy calculation is being kept and used as default for now.

## Description

Plain percentile calculation was extracted from the gasPriceTracker and a new GasUsed weighted percentile calculator was added in order to update how the gasPrice is being calculated. 
The calculator to use (legacy/weighted) can be configured. 

## Motivation and Context
We want to improve the current algorithm which is being used by eth_gasPrice method. Right now we only consider actual gas price of recently mined txs when we are looking for a price percentile. The suggestion is to besides gas price also consider tx.gasUsed.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/RskContext.java
+++ b/rskj-core/src/main/java/co/rsk/RskContext.java
@@ -105,6 +105,7 @@ import org.ethereum.db.ReceiptStoreImplV2;
 import org.ethereum.facade.Ethereum;
 import org.ethereum.facade.EthereumImpl;
 import org.ethereum.listener.CompositeEthereumListener;
+import org.ethereum.listener.GasPriceCalculator;
 import org.ethereum.listener.GasPriceTracker;
 import org.ethereum.net.EthereumChannelInitializerFactory;
 import org.ethereum.net.NodeManager;
@@ -558,7 +559,8 @@ public class RskContext implements NodeContext, NodeBootstrapper {
         double gasPriceMultiplier = getRskSystemProperties().gasPriceMultiplier();
 
         if (this.gasPriceTracker == null) {
-            this.gasPriceTracker = GasPriceTracker.create(getBlockStore(), gasPriceMultiplier);
+            GasPriceCalculator.GasCalculatorType calculatorType = getRskSystemProperties().getGasCalculatorType();
+            this.gasPriceTracker = GasPriceTracker.create(getBlockStore(), gasPriceMultiplier, calculatorType);
         }
         return this.gasPriceTracker;
     }
--- a/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
+++ b/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
@@ -30,6 +30,7 @@ import org.ethereum.config.SystemProperties;
 import org.ethereum.core.Account;
 import org.ethereum.crypto.ECKey;
 import org.ethereum.crypto.HashUtil;
+import org.ethereum.listener.GasPriceCalculator;
 
 import javax.annotation.Nullable;
 import java.nio.charset.StandardCharsets;
@@ -58,6 +59,8 @@ public class RskSystemProperties extends SystemProperties {
     private static final String RPC_MODULES_PATH = "rpc.modules";
     private static final String RPC_ETH_GET_LOGS_MAX_BLOCKS_TO_QUERY = "rpc.logs.maxBlocksToQuery";
     private static final String RPC_ETH_GET_LOGS_MAX_LOGS_TO_RETURN = "rpc.logs.maxLogsToReturn";
+    public static final String TX_GAS_PRICE_CALCULATOR_TYPE = "transaction.gasPriceCalculatorType";
+
     private static final String RPC_GAS_PRICE_MULTIPLIER_CONFIG = "rpc.gasPriceMultiplier";
     private static final String DISCOVERY_BUCKET_SIZE = "peer.discovery.bucketSize";
 
@@ -504,6 +507,18 @@ public class RskSystemProperties extends SystemProperties {
         }
 
         return value;
+    }
+
+    public GasPriceCalculator.GasCalculatorType getGasCalculatorType() {
+        String value = configFromFiles.getString(TX_GAS_PRICE_CALCULATOR_TYPE);
+        if (value == null || value.isEmpty()) {
+            return GasPriceCalculator.GasCalculatorType.PLAIN_PERCENTILE;
+        }
+        GasPriceCalculator.GasCalculatorType gasCalculatorType = GasPriceCalculator.GasCalculatorType.fromString(value);
+        if(gasCalculatorType == null) {
+            throw new RskConfigurationException("Invalid gasPriceCalculatorType: " + value);
+        }
+        return gasCalculatorType;
     }
 
     private void fetchMethodTimeout(Config configElement, Map<String, Long> methodTimeoutMap) {
--- a/rskj-core/src/main/java/org/ethereum/listener/GasPriceTracker.java
+++ b/rskj-core/src/main/java/org/ethereum/listener/GasPriceTracker.java
@@ -23,7 +23,6 @@ import co.rsk.core.Coin;
 import co.rsk.crypto.Keccak256;
 import co.rsk.remasc.RemascTransaction;
 import org.ethereum.core.Block;
-import org.ethereum.core.Transaction;
 import org.ethereum.core.TransactionReceipt;
 import org.ethereum.db.BlockStore;
 import org.slf4j.Logger;
@@ -35,10 +34,10 @@ import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Calculates a 'reasonable' Gas price based on statistics of the latest transaction's Gas prices
- *
+ * <p>
  * Normally the price returned should be sufficient to execute a transaction since ~25% of the latest
  * transactions were executed at this or lower price.
- *
+ * <p>
  * Created by Anton Nashatyrev on 22.09.2015.
  */
 public class GasPriceTracker extends EthereumListenerAdapter {
@@ -53,8 +52,6 @@ public class GasPriceTracker extends EthereumListenerAdapter {
 
     private static final double DEFAULT_GAS_PRICE_MULTIPLIER = 1.1;
 
-    private final Coin[] txWindow = new Coin[TX_WINDOW_SIZE];
-
     private final Double[] blockWindow = new Double[BLOCK_WINDOW_SIZE];
 
     private final AtomicReference<Coin> bestBlockPriceRef = new AtomicReference<>();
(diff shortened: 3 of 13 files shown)


rskj-core/src/main/java/org/ethereum/listener/GasPriceTracker.java
```diff
@@ -54,26 +53,46 @@ public class GasPriceTracker extends EthereumListenerAdapter {
...
+    }
+
+    /**
+     * @deprecated Use {@link #create(BlockStore, GasCalculator.GasCalculatorType)} instead.
+     */
+    @Deprecated
     public static GasPriceTracker create(BlockStore blockStore) {
-        GasPriceTracker gasPriceTracker = new GasPriceTracker(blockStore);
+        //Will be using the legacy gas calculator as default option
+        GasPriceTracker gasPriceTracker = new GasPriceTracker(blockStore, new LegacyGasCalculator());
```
rmoreliovlabs, line 95: Just a small suggestion, I think here you can just call the new create method and pass `GasCalculator.GasCalculatorType.LEGACY` as the `gasCalculatorType` argument, that way you can just do `return create(blockStore, GasCalculator.GasCalculatorType.LEGACY)` and avoid repeating the following 2 lines.

rskj-core/src/main/resources/reference.conf
```diff
@@ -165,7 +165,7 @@ peer {
 miner {
     # The default gas price
     minGasPrice = 0
-
+    gasPriceCalculatorType = LEGACY
```
Vovchyk, line 168: as this doesn't seem to be that specific to miners, I guess this config should go under the `rpc` section - same as `minGasPriceMultiplier`, wdyt?

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
```diff
@@ -482,6 +483,17 @@ public double getTopBest() {
         return value;
     }
 
+    public GasCalculator.GasCalculatorType getGasCalculatorType() {
+        String value = configFromFiles.getString("miner.gasPriceCalculatorType");
```
Vovchyk, line 487: could you extract the key string to constants?

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -554,9 +555,9 @@ public synchronized Ethereum getRsk() {
 
     public GasPriceTracker getGasPriceTracker() {
         checkIfNotClosed();
-
+        GasCalculator.GasCalculatorType calculatorType = getRskSystemProperties().getGasCalculatorType();
```
Vovchyk, line 558: minor thing: this line could be inside the if-block

rskj-core/src/main/java/org/ethereum/listener/GasCalculator.java
```diff
@@ -0,0 +1,32 @@
+package org.ethereum.listener;
```
Vovchyk, line 1: pls add cc header

rskj-core/src/main/java/org/ethereum/listener/GasCalculator.java
```diff
@@ -0,0 +1,32 @@
...
+                case "legacy":
+                    return LEGACY;
+                default:
+                    return null;
+            }
+        }
+    }
+
+    Optional<Coin> getGasPrice();
+    void  onBlock(Block block, List<TransactionReceipt> receipts);
```
Vovchyk, line 31:
```suggestion
    void onBlock(Block block, List<TransactionReceipt> receipts);
```

rskj-core/src/main/java/org/ethereum/listener/LegacyGasCalculator.java
```diff
@@ -0,0 +1,54 @@
+package org.ethereum.listener;
```
Vovchyk, line 1: pls add cc header

rskj-core/src/main/java/org/ethereum/listener/GasWeightedCalc.java
```diff
@@ -0,0 +1,105 @@
+/*
+ * This file is part of RskJ
+ * Copyright (C) 2017 RSK Labs Ltd.
```
Vovchyk, line 3:
```suggestion
 * Copyright (C) 2024 RSK Labs Ltd.
```

rskj-core/src/test/java/org/ethereum/listener/GasWeightedCalcTest.java
```diff
@@ -0,0 +1,143 @@
+package org.ethereum.listener;
```
Vovchyk, line 1: pls add cc header

rskj-core/src/main/java/org/ethereum/listener/GasWeightedCalc.java
```diff
@@ -0,0 +1,105 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.math.BigDecimal;
+import java.util.*;
+
+public class GasWeightedCalc implements GasCalculator {
+    private static final Logger logger = LoggerFactory.getLogger("gaspricetracker");
+    private static final int WINDOW_SIZE = 512;
+    private final Deque<GasEntry> gasWindow = new ArrayDeque<>(WINDOW_SIZE);
+    private final Map<Coin, Long> windowMap = new HashMap<>();
```
Vovchyk, line 37: didn't quite get the idea of using two collections. Isn't it enough to have let's say one `LinkedList<GasEntry>` (a queue with max size of 512 with entries corresponding to recent txs that were mined)? - so when a new tx arrives, a new entry of gasPrice/gasUsed is then get added to this collection and oldest one gets removed. Then you can use that list to calculate average value based on txs' costs (gasPrice * gasUsed). Am I missing something in this alg?

Vovchyk (changes requested):

rskj-core/src/main/resources/reference.conf
asoto-iov: I was not sure where  to place it, because it is not only affecting the rpc module. Just put it there because we had the min gas price property there, but maybe we could move it to a general section?

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
asoto-iov: oh sure, good point

rskj-core/src/main/java/org/ethereum/listener/GasCalculator.java
asoto-iov: I need the code style plugin :)

rskj-core/src/main/java/org/ethereum/listener/GasWeightedCalc.java
asoto-iov: I'm choose the map to accumulate the amount of gas per gas price. So as the price is being repeated along the TX the calculation is cheaper using the map. The list is used to know the amount of gas to remove from the map. 

In the first iteration I was calculating the value each time it was requested so it was cheaper to use the map but maybe as it is now it is not worthy. 
WDYT?

rskj-core/src/main/resources/reference.conf
Vovchyk: you mean that it's also used in mem pool? hmm.. what about putting it under the `transaction` section? similar to `transaction.gasPriceBump`

rskj-core/src/main/java/org/ethereum/listener/WeightedPercentileGasPriceCalculator.java
```diff
@@ -0,0 +1,146 @@
...
+
+        if (gasWindow.size() == WINDOW_SIZE) {
+            gasWindow.removeFirst();
+
+        }
+        gasWindow.add(new GasEntry(gasPrice, gasUsed));
+
+        if (txCount > WINDOW_SIZE) {
+            txCount = 0; // Reset the count
+            cachedGasPrice = calculateGasPrice();
```
Vovchyk, line 90: what about setting `null` instead here? If `getGasPrice()` is never called, then won't be a need to trigger `calculateGasPrice()`..

rskj-core/src/main/java/org/ethereum/listener/PercentileGasPriceCalculator.java
```diff
@@ -0,0 +1,77 @@
...
+                Coin[] values = Arrays.copyOf(txWindow, TX_WINDOW_SIZE);
+                Arrays.sort(values);
+                lastVal = values[values.length / 4];  // 25% percentile
+            }
+            return Optional.of(lastVal);
+        }
+    }
+
+    @Override
+    public void onBlock(Block block, List<TransactionReceipt> receipts) {
```
Vovchyk, line 52: looks like it'd make sense to make this method `synchronized`. Same as `getGasPrice()` above

rskj-core/src/main/java/org/ethereum/listener/WeightedPercentileCalc.java
```diff
@@ -0,0 +1,50 @@
...
+package org.ethereum.listener;
+
+import co.rsk.core.Coin;
+
+import java.util.Collections;
+import java.util.List;
+
+public class WeightedPercentileCalc {
+
+    public Coin calculateWeightedPercentile(float percentile, List<WeightedPercentileGasPriceCalculator.GasEntry> gasEntries) {
```
Vovchyk, line 28:
```suggestion
class WeightedPercentileCalc {

    Coin calculateWeightedPercentile(float percentile, List<WeightedPercentileGasPriceCalculator.GasEntry> gasEntries) {
```
does this method and `WeightedPercentileCalc ` class need to be `public`? This causes a warning - `Class 'WeightedPercentileGasPriceCalculator.GasEntry' is exposed outside its defined visibility scope`

rskj-core/src/main/java/org/ethereum/listener/PercentileGasPriceCalculator.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Optional;
+
+public class PercentileGasPriceCalculator implements GasPriceCalculator {
+    private static final int TX_WINDOW_SIZE = 512;
+
+    private final Coin[] txWindow = new Coin[TX_WINDOW_SIZE];
+    private int txIdx = TX_WINDOW_SIZE - 1;
+    private Coin lastVal;
+
+    public synchronized Optional<Coin> getGasPrice() {
```
Vovchyk, line 38:
```suggestion
    @Override
    public synchronized Optional<Coin> getGasPrice() {
```

rskj-core/src/main/java/org/ethereum/listener/WeightedPercentileGasPriceCalculator.java
```diff
@@ -0,0 +1,146 @@
...
+        this(new WeightedPercentileCalc());
+    }
+
+    public WeightedPercentileGasPriceCalculator(WeightedPercentileCalc weightedPercentileCalc) {
+        auxCalculator = weightedPercentileCalc;
+        gasWindow = new ArrayDeque<>(WINDOW_SIZE);
+    }
+
+    @Override
+    public Optional<Coin> getGasPrice() {
```
Vovchyk, line 52: looks like it'd make sense to make this method synchronized. Same as `onBlock()` below, wdyt?

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/listener/GasPriceTracker.java
```diff
@@ -91,38 +108,25 @@ public synchronized void onBlock(Block block, List<TransactionReceipt> receipts)
...
+        if(!gasPriceResult.isPresent()) {
             return defaultPrice;
         }
 
-        if (lastVal == null) {
-            Coin[] values = Arrays.copyOf(txWindow, TX_WINDOW_SIZE);
-            Arrays.sort(values);
-            lastVal = values[values.length / 4];  // 25% percentile
-        }
+        logger.debug("Gas provided by GasWindowCalc: {}", gasPriceCalculator.getGasPrice());
```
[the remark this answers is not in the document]
asoto-iov: yep, I updated this part during development to compare different things and forgot to revert it. Good catch.

Vovchyk:
pipeline:run

Vovchyk (approved):

Vovchyk:
@asoto-iov lgtm, pls make sure the pipeline is passed

PR #2220, opened 2024-01-03
Gas price config

rmoreliovlabs:
This PR adds a feature to the `eth_gasPrice` function that allows node operators or rsk-devs to manually set the returned gas price customize the gas price behavior based on specific requirements.

## Description
- Created an utility function to load properties from a `.conf` resource file.
- Updated `getGasPrice` in `EthereumImpl` to first check for the property and return its value if one is found, other wise return teh default value.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/test/resources/test-rskj.conf
```diff
@@ -161,6 +161,7 @@ vm.structured.storage.dictionary.enabled = false
 
 miner.client.enabled = false
 miner.server.enabled = false
+miner.gasPriceBuffer = 105
```
asoto-iov, line 164: Before this change the expected value would be 110. Having this new property here, wouldn't affect other tests using previous default values?

rskj-core/src/test/java/org/ethereum/facade/EthereumImplTest.java
```diff
@@ -22,14 +22,20 @@
...
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 class EthereumImplTest {
 
+    @AfterEach
+    void resetMocks() {
+        Mockito.reset();
+    }
```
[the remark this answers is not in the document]
rmoreliovlabs: That is from a previous version of this PR, I overlook it. I'll remove it righ away.

rskj-core/src/test/resources/test-rskj.conf
rmoreliovlabs: This value is used only by [RskSystemPropertiesTest](https://github.com/rsksmart/rskj/pull/2220/files#diff-f6561bc832eef1ae925e5bd576c6e88f42c7e6919f60d8f805fbb11b0ca09162R231) just to test teh new method added to that class so it should not affect other tests.

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
```diff
@@ -51,6 +52,7 @@ public class RskSystemProperties extends SystemProperties {
 
     private static final String MINER_REWARD_ADDRESS_CONFIG = "miner.reward.address";
     private static final String MINER_COINBASE_SECRET_CONFIG = "miner.coinbase.secret";
+    private static final String MINER_GAS_PRICE_BUFFER_CONFIG = "miner.gasPriceBuffer";
```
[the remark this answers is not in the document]
rmoreliovlabs: Sure, makes sense.

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
```diff
@@ -185,6 +186,16 @@ public boolean isWalletEnabled() {
         return getBoolean("wallet.enabled", false);
     }
 
+    public double gasPriceMultiplier() {
+        double gasPriceMultiplier = getDouble(RPC_GAS_PRICE_MULTIPLIER_CONFIG, 1.1);
+
+        if(gasPriceMultiplier > 1.0) {
```
Vovchyk, line 192: just thinking out loud.. probably it makes sense to restrict negative values. But should we really reject values lower than 1 (which basically would mean not increasing but decreasing by some percentage)?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
rmoreliovlabs: I was under the impresion that there would not be a case when the buffer would decrease de value. But if you think we should allow it I can make the change quickly.

rmoreliovlabs: Done

Vovchyk (approved):

PR #2225, opened 2024-01-11
Create getErpArgs method for ErpFederation. Create tests for both FederationArgs and ErpFederationArgs classes

julia-zack:
This pr creates tests for the new FederationArgs and ErpFederationArgs added classes.
Also, it includes the creation of the getErpArgs() method to return erpFederationArgs from erp federations.
The method getArgs() that returns federationArgs from federations was already created

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/federation/NonStandardErpFederationsTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/federation/NonStandardErpFederationsTest.java
@@ -259,25 +259,18 @@ class NonStandardErpFederationsTest {
     }
 
     @Test
-    void values_from_erpFederationArgs_equal_values_from_nonStandardErpFederation() {
-        ErpFederationArgs erpFederationArgs = nonStandardErpFederation.getErpArgs();
-        List<FederationMember> federationMembersFromErpArgs = erpFederationArgs.getMembers();
-        Instant creationTimeFromErpArgs = erpFederationArgs.getCreationTime();
-        long creationBlockNumberFromErpArgs = erpFederationArgs.getCreationBlockNumber();
-        NetworkParameters btcParamsFromErpArgs = erpFederationArgs.getBtcParams();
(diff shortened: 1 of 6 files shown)


rskj-core/src/test/java/co/rsk/peg/federation/NonStandardErpFederationsTest.java
```diff
@@ -258,6 +258,45 @@ void testEquals_same() {
         assertEquals(nonStandardErpFederation, otherFederation);
     }
 
+    @Test
+    void values_from_erpFederationArgs_equal_values_from_nonStandardErpFederation() {
+        ErpFederationArgs erpFederationArgs = nonStandardErpFederation.getErpArgs();
```
marcos-iov, line 263: What do you think about adding an `equals` function to FederationArgs? I think that would simplify these tests and it's always a good idea for java object to have an equals function

jeremy-then (approved):
LGTM.

rskj-core/src/test/java/co/rsk/peg/federation/NonStandardErpFederationsTest.java
julia-zack: good one! thanks

nathanieliov (approved):
LGTM

rskj-core/src/test/java/co/rsk/peg/federation/NonStandardErpFederationsTest.java
nathanieliov: Since we are implementing equals and hashcode properly, what about implementing the `toString` method. Not a big deal but it would be useful in case we want to print `federationArgs` object.

julia-zack: good idea as well :)
thanks guys

marcos-iov (approved):

PR #2226, opened 2024-01-12
Remove ErpFederationArgs

julia-zack:
The ErpFederationArgs class was created following the logic of having a FederationArgs one.
But after some analysis, this class started to look unnecessary, since the only purpose of creating FederationArgs was to reduce the amount of parameters we were passing to Federation constructors.
Therefore, the decision was to delete ErpFederationArgs and just maintain FederationArgs.
So the ErpFederation constructor no longer receives ErpFederationArgs (that wraps up FederationArgs and erp values) but the FederationArgs and the erp values.


rskj-core/src/main/java/co/rsk/peg/federation/ErpFederation.java
```diff
@@ -8,10 +8,10 @@
 import co.rsk.bitcoinj.script.ScriptChunk;
 import co.rsk.peg.bitcoin.ErpRedeemScriptBuilder;
 import co.rsk.peg.bitcoin.RedeemScriptCreationException;
-import co.rsk.peg.utils.EcKeyUtils;
 import java.util.Collections;
 import java.util.List;
 
+import static co.rsk.peg.federation.ErpFederationCreationException.Reason.NULL_OR_EMPTY_EMERGENCY_KEYS;
```
julia-zack, line 14: this logic had been moved to ErpFederationArgs. Now is back here

nathanieliov (approved):
LGTM

nathanieliov:
pipeline:run

marcos-iov (approved):

jeremy-then (approved):
LGTM.

PR #2227, opened 2024-01-15
add eth_pendingTransaction support

casiojapi:
## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -831,6 +835,23 @@ public TransactionReceiptDTO eth_getTransactionReceipt(TxHashParam transactionHa
         return new TransactionReceiptDTO(block, txInfo, signatureCache);
     }
 
+    @Override
+    public TransactionResultDTO[] eth_pendingTransactions() {
+        // get pending txs
+        List<Transaction> pendingTransactions = web3InformationRetriever.getTransactions("pending");
```
Vovchyk, line 841: we also could make `Web3InformationRetriever.PENDING` public and use it here as a constant

rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -831,6 +835,23 @@ public TransactionReceiptDTO eth_getTransactionReceipt(TxHashParam transactionHa
...
+
+        // get list of accounts managed by the node
+        Set<String> managedAccountSet = new HashSet<>(Arrays.asList(personalModule.listAccounts()));
```
Vovchyk, line 844: `personalModule.listAccounts()` throws an exception if personal wallet is disabled.

Alternatively, we could add a new `pendingTransactions()` method to `EthModuleWallet` interface and properly implement in derived classes, and then make use of it in this `eth_pendingTransactions()` like this `getEthModule().pendingTransactions()`, wdyt?

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
casiojapi: yep, makes sense. just pushed it

rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
Vovchyk: you are still using `personalModule.listAccounts()` in this context which would throw an error if the personal module is disabled - eg by default in Testnet and Mainnet. What about moving this stuff to `EthModuleWallet` as was suggested in he previous comment?

Vovchyk (changes requested):

casiojapi:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk (changes requested):

Vovchyk (approved):

PR #2230, opened 2024-01-18
Load active peers from previous session saved into file.

asoto-iov:
A client node would be able to re-connect to previously connected peers on restart.

## Description
Persist discovered peers across node restarts

## Motivation and Context
As of now, RSKj node keeps discovered peers in a so-called distance table in memory, so that after node restart those are being lost. Next time that the node starts over it has to start peer discovery process from scratch by having only list of boot nodes.

To improve, and most importantly speed the process up, we want to preserve a list of already discovered peers on a disk.

Note: it looks like we could make use of the org.ethereum.util.MapSnapshot<> class for that purpose.

Expected behaviour:

when node stops, it should save all discovered peers to a file

when node starts, it should check if the file with discovered peers exists in a database directory, and if so, load peer list from there. Otherwise, do nothing and proceed as before

Notes:

this functionality is part of the peer discovery protocol (see the PeerExplorer class)

the node should continue using bootstrap nodes the same way as it does now

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/net/server/ChannelManagerImpl.java
```diff
@@ -96,9 +103,28 @@ public void start() {
 
     @Override
     public void stop() {
+        saveActivePeersToFile();
```
Vovchyk, line 106: saving/loading of last discovered peers should probably be tackled in the `PeerExplorer` class. `ChannelManagerImpl` deals with the wire protocol, not discovery, and may include a subset of discovered peers

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/net/server/ChannelManagerImpl.java
asoto-iov: Updated!

rskj-core/src/test/java/co/rsk/config/RskSystemPropertiesTest.java
```diff
@@ -223,4 +226,21 @@ void testGetRpcModulesWithObject() {
         assertEquals(9000, netModule.getTimeout());
         assertEquals(30000, netModule.getMethodTimeout("eth_getBlockByHash"));
     }
+
+    @Test
+    void testPeerLastSession() throws IOException {
```
fmacleal, line 231: Woudn't be possible to load the actual configuration and check if the property it's enabled or not and if yes, check that the path to peersFile it's correctly configured? Yo

In my perspective, this should only validate the flags and not create and load the file, you could mock the File class. And the return from the filePath.

The test to generate the peer file and check if it was really generated could stay on the `DiscoveredPeerPersistenceServiceTest`  class.

rskj-core/src/test/java/co/rsk/net/discovery/DiscoveredPeersPersistenceServiceTest.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Test
+    void testStop() throws IOException {
+        List<String> knownPeers = Arrays.asList("peer1", "peer2", "peer3");
+        when(peerExplorer.getKnownHosts()).thenReturn(knownPeers);
+
+        service.stop();
+
+        verify(fileDataSaver, times(1)).saveDataIntoFile(anyString(), any(Path.class));
```
fmacleal, line 55: Wdyt if we verify the fields exactly instead of any input?
You have the filePath that it would try to save and you have the input peers mocked, you could check it instead of any:

`        verify(fileDataSaver, times(1)).saveDataIntoFile(eq("peer1\npeer2\npeer3\n"), eq(filePath));`

would do the job.

rskj-core/src/main/java/org/ethereum/util/SimpleFileWriter.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public static SimpleFileWriter getInstance() {
+        if (instance == null) {
+            instance = new SimpleFileWriter();
+        }
+        return instance;
+    }
+
+    public void saveDataIntoFile(String data, Path filePath) throws IOException {
+        File tempFile = File.createTempFile(filePath.toString(), TMP);
```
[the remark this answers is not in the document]
fmacleal: I am assuming that there is no sensitive data on this path and it's ok save it without permissions right? We can ignore this warning.

fmacleal (commented):
Good job! 

I just have some few recommendations. Let me know what you think about them.

rskj-core/src/test/java/co/rsk/net/discovery/DiscoveredPeersPersistenceServiceTest.java
asoto-iov: Yep it is a good idea!

rskj-core/src/test/java/co/rsk/config/RskSystemPropertiesTest.java
asoto-iov: Here it is a simply test to check that we are reading the file provided by the config and reading it line by line. I have never mocked the file but it would be quicker than write and read one.

In relation with the enable property I could create another test for that.

fmacleal: it's up to you, you could either create another test or refactor this one. :) 

In my point of view, I think that on this class would be good to validate that the flag and parameters necessary to write the file are working fine.

fmacleal (approved):
Good job! I see that all recommendations were handled, I am approving it. Well done! :)

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -106,6 +109,7 @@ public abstract class SystemProperties {
     public static final String PROPERTY_PERSIST_BLOOMS_CACHE_SNAPSHOT = "cache.blooms.persist-snapshot";
 
     /* Testing */
+    public static final String LAST_KNEW_PEERS_FILE = "lastKnewPeers.txt";
```
Vovchyk, line 112: what do you think about using some well-known / established data format, for instance, java `.properties` with `<nodeId>: <address>` mapping?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/NodeRunnerImpl.java
```diff
@@ -145,7 +145,11 @@ public synchronized void stop() {
         logger.info("Shutting down RSK node");
 
         for (int i = internalServices.size() - 1; i >= 0; i--) {
-            internalServices.get(i).stop();
+            try {
+                internalServices.get(i).stop();
```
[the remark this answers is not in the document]
a-soto: If the stop of any of the services fails throwing an exception the execution would stop and the missing services won't stop nicely.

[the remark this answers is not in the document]
Vovchyk: ok, I see - so this is not required by the feature itself

I guess it's a tradeoff: non-properly finished service may sometimes result to data inconsistency in next services. We should be very careful with such changes in shutdown process

rskj-core/src/main/java/co/rsk/net/discovery/DiscoveredPeersPersistenceService.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public void start() {
+        // Nothing to do during start
+    }
+
+    @Override
+    public void stop() {
+        logger.debug("Stop in progress.. Saving known peers list to file");
```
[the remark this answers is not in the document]
a-soto: I thought it was a good idea to encapsulate this logic separately, avoiding modifying other classes or services.
Maybe during the start, we can check the conditions to enable or disable the service instead of increase the PeerExplorer logic.

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
a-soto: As it was for internal usage only I didn't think about any specific format.

rskj-core/src/main/java/co/rsk/net/discovery/DiscoveredPeersPersistenceService.java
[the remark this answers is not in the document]
asoto-iov: I thought it was a good idea to encapsulate this logic separately, avoiding modifying other classes or services.
Maybe during the start, we can check the conditions to enable or disable the service instead of increase the PeerExplorer logic.

rskj-core/src/main/java/co/rsk/NodeRunnerImpl.java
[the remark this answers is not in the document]
asoto-iov: (previous deleted reply) If the stop of any of the services fails throwing an exception the execution would stop and the missing services won't stop nicely.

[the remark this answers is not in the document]
asoto-iov: > ok, I see - so this is not required by the feature itself
> 
> I guess it's a tradeoff: non-properly finished service may sometimes result to data inconsistency in next services. We should be very careful with such changes in shutdown process

Huh if it could happen I think it is something we should fix. A wrong/different ending of one service shouldn't affect to the others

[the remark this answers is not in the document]
Vovchyk: if that happens then there's a bug in the code which has to be fixed. We don't expect `service.stop()` to throw any checked exception. Only unchecked exceptions can happen here (eg. RuntimeException), and if that happens we should propagate it call stack up. Not sure it's a good idea to catch an exception and continue stopping other services as node is already in inconsistent state - best we can do here is to catch it at upper level and log it. So I'd rather not to add this `try-catch` in here

rskj-core/src/main/java/co/rsk/net/discovery/DiscoveredPeersPersistenceService.java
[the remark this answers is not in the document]
Vovchyk: agree, encapsulation is a good approach. But not sure we really need one more service for that.. What about having a new class with that logic with name something like `PeerStorageManager` and passing its instance to the PeerExplorer's ctor? Something similar to NodeChallengeManager or PeerExplorerCleaner

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
Vovchyk: yeah, we already use `.properties` in the db folder for other needs, so I just thought that it would be a good idea to continue using it. `.txt` seemed to me more like a textual document rather than a storage file

rskj-core/src/main/java/co/rsk/NodeRunnerImpl.java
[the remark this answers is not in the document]
asoto-iov: ok got it, I'll remove it. But In my opinion the services should be independent and in case there are dependencies we could create inner groups and close them together. Maybe is something we can try to do in the future. 
As an example, in the case of this service. If the service throws an error due to the file persist or any other thing processing an unexpected input data or so, the others services shouldn't be affected and should be stopped properly.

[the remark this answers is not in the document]
Vovchyk: yeah, that was my point. If a service for any reason throws an [unchecked](https://www.baeldung.com/java-checked-unchecked-exceptions#unchecked) exception, which means that the node most probably cannot recover from that state. If the service can recover from an exception that was thrown in the `stop()` method, then it should not be propagated up but rather handled in it.

Vovchyk:
pipeline:run

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
```diff
@@ -136,6 +146,10 @@ synchronized void start(boolean startConversation) {
     }
 
     public synchronized void dispose() {
+        if(knownPeersSaver != null) {
```
Vovchyk, line 149:
```suggestion
        if (knownPeersSaver != null) {
```

Vovchyk, line 149: the idea here is to trigger all dispose logic after the `state == ExecState.FINISHED` check. Similar to what `this.cleaner.dispose()` does

rskj-core/src/main/java/co/rsk/net/discovery/KnownPeersSaver.java
```diff
@@ -0,0 +1,58 @@
...
+        this(peerListFileDir, SimpleFileWriter.getInstance());
+    }
+
+    public KnownPeersSaver(Path peerListFileDir, SimpleFileWriter fileDataSaver) {
+        this.peerListFileDir = peerListFileDir;
+        this.fileDataSaver = fileDataSaver;
+    }
+
+    public void savePeers(List<String> knownPeers) {
+        logger.debug("Stop in progress.. Saving known peers list to file");
```
Vovchyk, line 43:
```suggestion
        logger.debug("Saving known peers list to file");
```
minor thing.. but still a stop process seems to be unrelated to this class and saving can in theory be triggered in the middle of node execution

rskj-core/src/main/java/co/rsk/net/discovery/KnownPeersSaver.java
```diff
@@ -0,0 +1,58 @@
...
+        if (knownPeers != null && !knownPeers.isEmpty()) {
+
+            StringBuilder sb = new StringBuilder();
+            for (String peerAddress : knownPeers) {
+                logger.debug("Saving knew peer: {}", peerAddress);
+                sb.append(peerAddress).append("\n");
```
Vovchyk, line 49: what about using a bulit-in java `Properties` class for saving and reading? For instance, like in nodeId generation in [here](https://github.com/rsksmart/rskj/blob/ada67715db96146e022de4dce712743348ff8d74/rskj-core/src/main/java/org/ethereum/config/SystemProperties.java#L401C21-L401C39)

As I mentioned in my previous comment, it could be `nodeId -> address` mapping.

rskj-core/src/main/java/org/ethereum/util/SimpleFileWriter.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.ethereum.util;
```
Vovchyk, line 19: what about having it in the `co.rsk.net.discovery.util` java package as this is rsk new stuff and not part of the forked project ?

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
```diff
@@ -601,4 +615,8 @@ private boolean isBanned(Node node) {
 
         return address != null && this.peerScoringManager.isAddressBanned(address) || this.peerScoringManager.isNodeIDBanned(node.getId());
     }
+
+    List<String> getKnownHosts(){
```
Vovchyk, line 619:
```suggestion
    @VisibleForTesting
    List<String> getKnownHosts() {
```

rskj-core/src/main/java/co/rsk/NodeRunnerImpl.java
```diff
@@ -145,7 +145,7 @@ public synchronized void stop() {
         logger.info("Shutting down RSK node");
 
         for (int i = internalServices.size() - 1; i >= 0; i--) {
-            internalServices.get(i).stop();
+                internalServices.get(i).stop();
```
Vovchyk, line 148:
```suggestion
            internalServices.get(i).stop();
```

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -1582,17 +1582,16 @@ protected PeerExplorer getPeerExplorer() {
...
-            if (activePeers != null) {
-                for (Node n : activePeers) {
-                    InetSocketAddress address = n.getAddress();
-                    initialBootNodes.add(address.getHostName() + ":" + address.getPort());
-                }
+
+            KnownPeersSaver knownPeersSaver = null;
+            if(rskSystemProperties.usePeersFromLastSession()) {
+                knownPeersSaver = new KnownPeersSaver(getRskSystemProperties().getLastKnewPeersFilePath());
+
```
Vovchyk, line 1589:
```suggestion
            if (rskSystemProperties.usePeersFromLastSession()) {
                knownPeersSaver = new KnownPeersSaver(getRskSystemProperties().getLastKnewPeersFilePath());
```

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
```diff
@@ -93,10 +93,20 @@ public class PeerExplorer {
 
     private UDPChannel udpChannel;
 
+    private KnownPeersSaver knownPeersSaver;
```
Vovchyk, line 96:
```suggestion
    private final KnownPeersSaver knownPeersSaver;
```

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -1602,13 +1601,41 @@ protected PeerExplorer getPeerExplorer() {
...
-                    rskSystemProperties.peerDiscoveryMaxBootRetries()
+                    rskSystemProperties.peerDiscoveryMaxBootRetries(),
+                    knownPeersSaver
             );
         }
 
         return peerExplorer;
     }
 
+    List<String> getInitialBootNodes() {
```
Vovchyk, line 1612:
```suggestion
    @VisibleForTesting
    List<String> getInitialBootNodes() {
```

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -1602,13 +1601,41 @@ protected PeerExplorer getPeerExplorer() {
...
+        List<String> initialBootNodes = new ArrayList<>();
```
Vovchyk, line 1613: do you think we could use `Set<String>` here just to prevent duplicates?

rskj-core/src/main/java/org/ethereum/net/server/ChannelManagerImpl.java
```diff
@@ -175,6 +175,7 @@ private boolean isRecent(Instant disconnectionTimeout, Instant currentTime) {
     }
 
     private void addToActives(Channel peer) {
+
```
Vovchyk, line 178:
```suggestion
```

rskj-core/src/main/resources/expected.conf
```diff
@@ -101,6 +101,7 @@ blockchain = {
 }
 
 peer = {
+
```
Vovchyk, line 104:
```suggestion
```

Vovchyk (changes requested):

rmoreliovlabs (approved):
LGTM

Vovchyk (approved):

PR #2238, opened 2024-01-30
Added getMemberByBtcPublicKey new method

nathanieliov:
- Created a new method to get federation member by btc public key
- Added new tests for the new method:
  - Passing a null should return an empty optional.
  - Passing a random BTC public key should return an empty optional.
  - Passing an existing BTC public key should return the matched federation member.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/Federation.java
+++ b/rskj-core/src/main/java/co/rsk/peg/Federation.java
@@ -64,6 +64,10 @@ public abstract class Federation {
         return members;
     }
 
+    public Optional<FederationMember> getMemberByBtcPublicKey(BtcECKey btcPublicKey) {
+        return members.stream().filter(federationMember -> federationMember.getBtcPublicKey().equals(btcPublicKey)).findFirst();
+    }
+
     public List<BtcECKey> getBtcPublicKeys() {
         // Copy instances since we don't control
         // immutability of BtcECKey instances
--- a/rskj-core/src/main/java/co/rsk/peg/Federation.java
+++ b/rskj-core/src/main/java/co/rsk/peg/Federation.java
@@ -65,7 +65,7 @@ public abstract class Federation {
     }
 
     public Optional<FederationMember> getMemberByBtcPublicKey(BtcECKey btcPublicKey) {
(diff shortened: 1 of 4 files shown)


marcos-iov (approved):

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -64,6 +64,10 @@ public List<FederationMember> getMembers() {
         return members;
     }
 
+    public Optional<FederationMember> getMemberByBtcPublicKey(BtcECKey btcPublicKey) {
+        return members.stream().filter(federationMember -> federationMember.getBtcPublicKey().equals(btcPublicKey)).findFirst();
```
julia-zack, line 68: i think is more accurate to use findAny instead of findFirst.
We should expect to never have two members with same btcPublicKey (that check is being done in BridgeSupport: https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java#L2111)

rskj-core/src/test/java/co/rsk/peg/LegacyErpFederationTest.java
julia-zack, line 1: i will use
**method_input_output** structure for tests naming

i.e., 

i'll rename **test_getMemberByBtcPublicKey_passing_existing_btcPublicKey**
to **getMemberByBtcPublicKey_existingBtcPublicKey_returnsExistingMember**,

**test_getMemberByBtcPublicKey_passing_non_existing_btcPublicKey**
to **getMemberByBtcPublicKey_nonExistingBtcPublicKey_returnsEmpty**

something like that.
Apart from that, LGTM!

rskj-core/src/main/java/co/rsk/peg/Federation.java
```diff
@@ -64,6 +64,10 @@ public List<FederationMember> getMembers() {
         return members;
     }
 
+    public Optional<FederationMember> getMemberByBtcPublicKey(BtcECKey btcPublicKey) {
```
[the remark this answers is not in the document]
julia-zack: ohhhh got it. Makes sense

julia-zack (approved):
lgtm!

PR #2243, opened 2024-01-31
Fix the web 3 retriever with block hash as identifier

fmacleal:
## Description
With the current logic implementation, the hash is working to retrieve the block. We had to change the way to get blockReference that wasn't considering the identifier to be a hash.

With the change, we try to get the bockReference as if the identifier was a number, if it still doesn't work, we try as a hash. Since both ways are accepted as a string identifier.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/rpc/Web3InformationRetriever.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/Web3InformationRetriever.java
@@ -21,6 +21,7 @@ package co.rsk.rpc;
 import co.rsk.core.bc.AccountInformationProvider;
 import co.rsk.db.RepositoryLocator;
 import co.rsk.util.HexUtils;
+import org.bouncycastle.util.encoders.DecoderException;
 import org.ethereum.core.Block;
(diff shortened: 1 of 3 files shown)


rskj-core/src/main/java/co/rsk/rpc/Web3InformationRetriever.java
```diff
@@ -70,14 +73,22 @@ public Web3InformationRetriever(TransactionPool transactionPool,
...
+                block = executionBlockRetriever.retrieveExecutionBlock(identifier).getBlock();
+                break;
+            case LATEST:
+                block = blockchain.getBestBlock();
+                break;
+            case EARLIEST:
+                block = blockchain.getBlockByNumber(0);
+                break;
+            default:
+                byte[] hash = getBlockHash(identifier);
```
asoto-iov, line 88: I would at some unit test to check this logic related with blockNumber and blockHash.

asoto-iov (approved):

Vovchyk:
pipeline:run

PR #2245, opened 2024-01-31
Changes RemascFederationProvider.getFederatorAddress logic and adds t…

jeremy-then:
Changes:

* Refactored `RemascFederationProvider` constructor to receive only `ActivationConfig.ForBlock activations` and `FederationSupport federationSupport`, since this class only needs these 2 dependencies. It was receiving many other dependencies only to create the  `FederationSupport` instance. Hence, I decided to simplify it by only receiving 2 dependencies instead of 4.
* Included an activations check in `RemascFederationProvider:: getFederatorAddress`. If `RSKIP415` is active, then use the rsk federator public key to generate the rsk address. Otherwise, keep using the btc public key.
* Refactored the files using the `RemascFederationProvider` to pass the new paramemters to the constructor.
* Created 2 new tests for `RemascFederationProvider` in `RemascFederationProviderTest` file. The `RemascFederationProvider` class is very simply. It only uses a `FederationSupport` support instance and the activations. So, in reality what we need to check is that `federationSupport.getFederatorBtcPublicKey` is still called when `RSKIP415` is not active, and when `RSKIP415` is active, then the one to call is `federationSupport.getFederatorPublicKeyOfType(n, FederationMember.KeyType.RSK)`. Due to the simplicity of the requirements, I didn't include more tests.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/remasc/Remasc.java
+++ b/rskj-core/src/main/java/co/rsk/remasc/Remasc.java
@@ -18,6 +18,7 @@
 
 package co.rsk.remasc;
 
+import co.rsk.config.BridgeConstants;
 import co.rsk.config.RemascConfig;
 import co.rsk.core.Coin;
 import co.rsk.core.RskAddress;
@@ -179,7 +180,7 @@ public class Remasc {
         Coin payToRskLabs = syntheticReward.divide(BigInteger.valueOf(remascConstants.getRskLabsDivisor()));
         feesPayer.payMiningFees(processingBlockHeader.getHash().getBytes(), payToRskLabs, rskLabsAddress, logs);
         syntheticReward = syntheticReward.subtract(payToRskLabs);
-        Coin payToFederation = payToFederation(constants, isRskip85Enabled, processingBlock, processingBlockHeader, syntheticReward);
+        Coin payToFederation = payToFederation(constants, processingBlock, processingBlockHeader, syntheticReward);
         syntheticReward = syntheticReward.subtract(payToFederation);
 
         if (!siblings.isEmpty()) {
@@ -201,18 +202,20 @@ public class Remasc {
         return isRskip218Enabled ? remascConstants.getRskLabsAddressRskip218() : remascConstants.getRskLabsAddress();
     }
 
-    private Coin payToFederation(Constants constants, boolean isRskip85Enabled, Block processingBlock, BlockHeader processingBlockHeader, Coin syntheticReward) {
+    private Coin payToFederation(Constants constants, Block processingBlock, BlockHeader processingBlockHeader, Coin syntheticReward) {
 
         ActivationConfig.ForBlock activations = activationConfig.forBlock(processingBlock.getNumber());
+
+        BridgeConstants bridgeConstants = constants.getBridgeConstants();
 
         BridgeStorageProvider bridgeStorageProvider = new BridgeStorageProvider(
                 repository,
                 PrecompiledContracts.BRIDGE_ADDR,
-                constants.getBridgeConstants(),
+                bridgeConstants,
                 activations
         );
 
-        FederationSupport federationSupport = new FederationSupport(constants.getBridgeConstants(), bridgeStorageProvider, processingBlock, activations);
+        FederationSupport federationSupport = new FederationSupport(bridgeConstants, bridgeStorageProvider, processingBlock, activations);
 
         RemascFederationProvider federationProvider = new RemascFederationProvider(activations, federationSupport);
         Coin federationReward = syntheticReward.divide(BigInteger.valueOf(remascConstants.getFederationDivisor()));
@@ -224,7 +227,7 @@ public class Remasc {
         Coin payToFederator = payAndRemainderToFederator[0];
         Coin restToLastFederator = payAndRemainderToFederator[1];
 
-        if (isRskip85Enabled) {
+        if (activations.isActive(ConsensusRule.RSKIP85)) {
             BigInteger minimumFederatorPayableGas = constants.getFederatorMinimumPayableGas();
             Coin minPayableFederatorFees = executionBlock.getMinimumGasPrice().multiply(minimumFederatorPayableGas);
             if (payToFederator.compareTo(minPayableFederatorFees) < 0) {
--- a/rskj-core/src/main/java/co/rsk/remasc/Remasc.java
+++ b/rskj-core/src/main/java/co/rsk/remasc/Remasc.java
@@ -22,6 +22,8 @@ import co.rsk.config.RemascConfig;
 import co.rsk.core.Coin;
 import co.rsk.core.RskAddress;
 import co.rsk.core.bc.SelectionRule;
+import co.rsk.peg.BridgeStorageProvider;
+import co.rsk.peg.FederationSupport;
 import co.rsk.rpc.modules.trace.ProgramSubtrace;
 import org.ethereum.config.Constants;
 import org.ethereum.config.blockchain.upgrades.ActivationConfig;
@@ -32,6 +34,7 @@ import org.ethereum.core.Repository;
 import org.ethereum.core.Transaction;
 import org.ethereum.db.BlockStore;
 import org.ethereum.vm.LogInfo;
+import org.ethereum.vm.PrecompiledContracts;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


rskj-core/src/main/java/co/rsk/remasc/Remasc.java
```diff
@@ -199,7 +202,19 @@ RskAddress getRskLabsAddress() {
     }
 
     private Coin payToFederation(Constants constants, boolean isRskip85Enabled, Block processingBlock, BlockHeader processingBlockHeader, Coin syntheticReward) {
-        RemascFederationProvider federationProvider = new RemascFederationProvider(activationConfig, constants.getBridgeConstants(), repository, processingBlock);
+
+        ActivationConfig.ForBlock activations = activationConfig.forBlock(processingBlock.getNumber());
```
marcos-iov, line 206: Since we already have the activations object, does it makes to receive a `boolean isRskip85Enabled`? 

Totally out of scope for this task but could you please have a look and see if it's something redundant or if there is a reason for this?

rskj-core/src/main/java/co/rsk/remasc/RemascFederationProvider.java
```diff
@@ -17,44 +17,38 @@
...
 
     public RskAddress getFederatorAddress(int n) {
-        byte[] publicKey = this.federationSupport.getFederatorBtcPublicKey(n);
-        return new RskAddress(ECKey.fromPublicOnly(publicKey).getAddress());
+        if(!activations.isActive(ConsensusRule.RSKIP415)) {
+            byte[] btcPublicKey = this.federationSupport.getFederatorBtcPublicKey(n);
+            return new RskAddress(ECKey.fromPublicOnly(btcPublicKey).getAddress());
+        }
+        byte[] rskPublicKey = this.federationSupport.getFederatorPublicKeyOfType(n, FederationMember.KeyType.RSK);
+        return new RskAddress(ECKey.fromPublicOnly(rskPublicKey).getAddress());
```
marcos-iov, line 52: I have a couple of cleany codey ideas here. What I see is that the whole `return new RskAddress(` statement is repeated. It only changes the public key you provide.

Some ideas
```
        KeyType keyType = KeyType.RSK;
        if(!activations.isActive(ConsensusRule.RSKIP415)) {
            keyType = KeyType.BTC;
        }
        byte[] publicKey = this.federationSupport.getFederatorPublicKeyOfType(n, keyType);
        return new RskAddress(ECKey.fromPublicOnly(publicKey).getAddress());
```
Or 
```
        if(!activations.isActive(ConsensusRule.RSKIP415)) {
            return getRskAddressFromBtcKey();
        }
        return getRskAddressFromRskKey();
```
Maybe the second one is easier tp read but it will result in more duplicated code surely

rskj-core/src/test/java/co/rsk/remasc/RemascFederationProviderTest.java
```diff
@@ -2,43 +2,110 @@
...
+    void getFederatorAddress_preRSKIP415_returnsRskAddressDerivedFromBtcPublicKey() {
+
+        int federatorIndex = 0;
+
+        ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
+        when(activations.isActive(ConsensusRule.RSKIP415)).thenReturn(false);
+
+        FederationSupport federationSupport =  mock(FederationSupport.class);
+
+        byte[] btcPublicKey = HexUtils.strHexOrStrNumberToByteArray(BTC_PUBLIC_KEY);
```
marcos-iov, line 45: Not sure if there's any difference or benefit, but we usually use `Hex.decode`. Maybe you could check if there's any difference between using one or the other

rskj-core/src/test/java/co/rsk/remasc/RemascFederationProviderTest.java
```diff
@@ -2,43 +2,110 @@
...
 
-        byte[] address = provider.getFederatorAddress(0).getBytes();
+        when(federationSupport.getFederatorBtcPublicKey(federatorIndex))
+                .thenReturn(btcPublicKey);
+
+        RemascFederationProvider provider = new RemascFederationProvider(activations, federationSupport);
+
+        RskAddress actualRskAddress = provider.getFederatorAddress(federatorIndex);
+        RskAddress expectedRskAddress = new RskAddress(ECKey.fromPublicOnly(btcPublicKey).getAddress());
```
marcos-iov, line 53: Since the public key is already hardcoded, I would also hardcode the expected address.

I think it's better to test against hardcoded values rather than repeating the same logic that's in the method under test

rskj-core/src/test/java/co/rsk/remasc/RemascFederationProviderTest.java
```diff
@@ -2,43 +2,110 @@
...
+
+        Assertions.assertEquals(expectedRskAddress, actualRskAddress);
+        verify(federationSupport, never()).getFederatorPublicKeyOfType(federatorIndex, FederationMember.KeyType.RSK);
+        verify(federationSupport, times(1)).getFederatorBtcPublicKey(federatorIndex);
+
+    }
+
+    @Test
```
marcos-iov, line 61: Think we could have these 2 as single parameterized test?

rskj-core/src/test/java/co/rsk/remasc/RemascFederationProviderTest.java
```diff
@@ -2,43 +2,110 @@
...
         Genesis genesisBlock = new BlockGenerator().getGenesisBlock();
         BlockChainBuilder builder = new BlockChainBuilder().setTesting(true).setGenesis(genesisBlock);
         Blockchain blockchain = builder.build();
 
-        return new RemascFederationProvider(
-                ActivationConfigsForTest.all(),
-                BridgeRegTestConstants.getInstance(),
+        ActivationConfig.ForBlock activations = ActivationConfigsForTest.all().forBlock(blockchain.getBestBlock().getNumber());
+
+        BridgeRegTestConstants bridgeRegTestConstants = BridgeRegTestConstants.getInstance();
```
marcos-iov, line 95: Let's use mainnet constants, so we test in a more realistic scenario

rskj-core/src/main/java/co/rsk/remasc/RemascFederationProvider.java
julia-zack: I know is more efficient and it actually makes more sense, but defining the key type as rsk first and then change it is a bit confusing for me.
and maybe the second option is too much... but i think i like it better

rskj-core/src/main/java/co/rsk/remasc/Remasc.java
```diff
@@ -199,7 +202,19 @@ RskAddress getRskLabsAddress() {
...
+
```
julia-zack, line 207: we could define 
`bridgeConstants = constants.getBridgeConstants()` and use the variable when needed

rskj-core/src/test/java/co/rsk/remasc/RemascFederationProviderTest.java
julia-zack: but its doing different checks right? 

```
verify(federationSupport, never()).getFederatorPublicKeyOfType(federatorIndex, FederationMember.KeyType.RSK);
verify(federationSupport, times(1)).getFederatorBtcPublicKey(federatorIndex);
```

versus

```
verify(federationSupport, times(1)).getFederatorPublicKeyOfType(federatorIndex, FederationMember.KeyType.RSK);
verify(federationSupport, never()).getFederatorBtcPublicKey(federatorIndex);
```

julia-zack: That being said, i would invert the order in the second test:

```
verify(federationSupport, never()).getFederatorBtcPublicKey(federatorIndex);
verify(federationSupport, times(1)).getFederatorPublicKeyOfType(federatorIndex, FederationMember.KeyType.RSK);
```

to see the difference between both tests more clear

rskj-core/src/main/java/co/rsk/remasc/RemascFederationProvider.java
```diff
@@ -17,44 +17,51 @@
...
+            FederationSupport federationSupport) {
+        this.federationSupport = federationSupport;
+        this.activations = activations;
     }
 
     public int getFederationSize() {
         return this.federationSupport.getFederationSize();
     }
 
     public RskAddress getFederatorAddress(int n) {
```
marcos-iov, line 46: Nit picky, but if we want to go full clean cody I'd change that `int n` variable name

marcos-iov (approved):

julia-zack (approved):

rskj-core/src/main/java/co/rsk/remasc/Remasc.java
jeremy-then: Updated!

PR #2250, opened 2024-02-07
Fixed misused of activation config passing wrong execution block number.

nathanieliov:
Background

- Fix the `ActivationConfig` instance being created using an incorrect block at the `processMinersFees` method. Because of this issue activations checks were wrong resulting in a fork when syncing previous code for a boton instance. The `ActivationConfig` object should always use the execution block number to check which rskip is active. So we refactored how the activations object is created, now it uses only the execution block number, and instead of using `ActivationConfig` object we changed it to be an `ActivationConfig.ForBlock`, so we only need to pass the block number once when we creating the object.
- Also, we abstract the RSKIP85 activation check into a separate method: `shouldRewardBeAboveMinimumPayableGas`. This way the method itself expresses what this rskip is for and what flow the code should use when the rskip is active.
- Create a new test class `RemascTest` for adding unit tests for the 'processMinersFees' and for the rest of the public methods later tasks.
- Updated `RemascRskAddressActivationTest.testActivation` method to comply with the refactor of the internal use of the Activations in `Remasc` class

- Finally, we add some unit tests before and after rskip85: 

- - We mock many of the required objects to create a `Remasc` instance and be able to test the `processMinersFees` method.
- - We set the execution block to be at height: 4100. So the next block to be rewarded should be found among: `start = maturity -  1 - uncleGenerationLimit`. Maturity is a constant equal to `4000` and `uncleGenerationLimit` equal to `8` where both are used in combination as the `from` block number of the range to find the candidates' blocks to reward. Regarding the `to` of the range, it uses the execution block number.
- - Thus, we set a mock blockchain with 8 blocks representing the `uncleGenerationLimit` constant. We set the same paid fee for each block. 
- - Before the rskip85, we set a paid fee lower than the minPayableFees, the test should passed since the validation of the paid fee above minPayableFees is not being used yet.
 - - And the opposite after the rskip85. We set a paid fee equal to the sum of the candidate's blocks fee to be rewarded (using the syntheticSpan constant), but below the minPayableFees to each block, the test should not passed since the reward is not equal or above the minimum payable fees.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
@@ -956,7 +956,7 @@ public class BridgeStorageProvider {
         pegoutTxSigHashes.add(sigHash);
     }
 
-    private void savePegoutTxSigHashes() {
+    protected void savePegoutTxSigHashes() {
         if (!activations.isActive(RSKIP379) || pegoutTxSigHashes == null) {
             return;
         }
--- a/rskj-core/src/main/java/co/rsk/remasc/Remasc.java
+++ b/rskj-core/src/main/java/co/rsk/remasc/Remasc.java
@@ -51,7 +51,7 @@ public class Remasc {
     private static final Logger logger = LoggerFactory.getLogger(Remasc.class);
 
     private final Constants constants;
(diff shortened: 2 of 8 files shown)


marcos-iov (approved):

jeremy-then (approved):
LGTM.

PR #2265, opened 2024-03-05
feat(genesis): allows external genesis file

jurajpiar:
## Description

First checks for files inside the resources folder to provide default backward compatibility and if not found it tries the file system.

## Motivation and Context

Needed for loading a custom genesis for pre-built image.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Requires Activation Code (Hard Fork)

* **Other information**:

PR #2266, opened 2024-03-05
feat(genesis): allows external genesis file

Vovchyk:
## Description

First checks for files inside the resources folder to provide default backward compatibility and if not found it tries the file system.

( in favor of #2265 )

## Types of changes

## Checklist:

* **Other information**:
Documented on devportal in [PR-1004](https://github.com/rsksmart/devportal/pull/1004).

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/genesis/GenesisLoaderImpl.java
+++ b/rskj-core/src/main/java/org/ethereum/core/genesis/GenesisLoaderImpl.java
@@ -44,6 +44,8 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.math.BigInteger;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -54,6 +56,7 @@ import java.util.Map;
  * 4. Registers the genesis state root in the state root handler
  */
 public class GenesisLoaderImpl implements GenesisLoader {
+
     private static final byte[] EMPTY_LIST_HASH = HashUtil.keccak256(RLP.encodeList());
 
     private static final Logger logger = LoggerFactory.getLogger(GenesisLoaderImpl.class);
@@ -81,7 +84,7 @@ public class GenesisLoaderImpl implements GenesisLoader {
                 activationConfig,
                 stateRootHandler,
                 trieStore,
-                GenesisLoaderImpl.class.getResourceAsStream("/genesis/" + genesisFile),
+                loadGenesisFile(genesisFile),
                 initialNonce,
                 isRsk,
                 useRskip92Encoding,
@@ -230,6 +233,21 @@ public class GenesisLoaderImpl implements GenesisLoader {
         }
     }
 
+    static InputStream loadGenesisFile(String fileName) {
+        InputStream inputStream = GenesisLoaderImpl.class.getResourceAsStream("/genesis/" + fileName);
+        if (inputStream != null) {
+            return inputStream;
+        }
+
+        try {
+            return Files.newInputStream(Paths.get(fileName));
+        } catch (Exception e) {
+            logger.error("Cannot read genesis json file");
+
+            throw new GenesisLoaderException("Cannot open genesis block configuration file", e);
+        }
+    }
+
     public static void loadGenesisInitalState(Repository repository, Genesis genesis) {
         // first we need to create the accounts, which creates also the associated ContractDetails
         for (RskAddress accounts : genesis.getAccounts().keySet()) {
@@ -259,7 +277,7 @@ public class GenesisLoaderImpl implements GenesisLoader {
         } catch (Exception e) {
             logger.error("Cannot read genesis json file");
 
-            throw new RuntimeException("Genesis block configuration is corrupted or not found ./resources/genesis/...", e);
+            throw new GenesisLoaderException("Genesis block configuration is corrupted or not found", e);
         } finally {
             closeStream(inputStream);
         }
--- a/rskj-core/src/test/java/org/ethereum/core/genesis/GenesisLoaderImplTest.java
+++ b/rskj-core/src/test/java/org/ethereum/core/genesis/GenesisLoaderImplTest.java
@@ -1,0 +1,73 @@
+package org.ethereum.core.genesis;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Path;
+import java.util.Objects;
+
+class GenesisLoaderImplTest {
+    private final String GENESIS_FILE_NAME = "temp_genesis.json";
+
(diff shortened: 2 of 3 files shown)


jurajpiar (dismissed):

jurajpiar:
how do you re-run the sonar check again?

jurajpiar (approved):
@Vovchyk I don't think I should be the one approving it .)

rskj-core/src/main/java/org/ethereum/core/genesis/GenesisLoaderImpl.java
```diff
@@ -230,6 +232,20 @@ private void setupPrecompiledContractsStorage(Repository repository) {
...
+        InputStream inputStream = GenesisLoaderImpl.class.getResourceAsStream("/genesis/" + fileName);
+        if (inputStream != null) {
+            return inputStream;
+        }
+
+        try {
+            return Files.newInputStream(Paths.get(fileName));
+        } catch (Exception e) {
+            logger.error("Could not load file due to:", e);
+            return null;
```
fmacleal, line 245: Why return null here if at the end, you will throw an exception on the method  `readGenesisJson`? Wdyt in throw the exception here straight away?

rskj-core/src/main/java/org/ethereum/core/genesis/GenesisLoaderImpl.java
```diff
@@ -230,6 +232,20 @@ private void setupPrecompiledContractsStorage(Repository repository) {
...
+        }
```
fmacleal, line 246: Since you are opening creating and opening the stream here, I would advise to close it in a `finally` clause, just in case an exception occurs.

rskj-core/src/main/java/org/ethereum/core/genesis/GenesisLoaderImpl.java
```diff
@@ -230,6 +232,20 @@ private void setupPrecompiledContractsStorage(Repository repository) {
         }
     }
 
+    public static InputStream loadGenesisFile(String fileName) {
```
fmacleal, line 235: Does it need to be public? Since you need only visible for testing, you can let it package private with the default access, wdyt?

rskj-core/src/test/java/org/ethereum/core/genesis/GenesisLoaderImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+        Assertions.assertNull(GenesisLoaderImpl.loadGenesisFile(genesisFile.getPath()));
+    }
+
+    @Test
+    void loadGenesisFile_fromSystem(@TempDir Path tempGenesisDir) throws IOException {
+        File genesisFile = new File(tempGenesisDir + "/" + GENESIS_FILE_NAME);
+        Assertions.assertTrue(genesisFile.createNewFile());
+        InputStream genesisFileStream = GenesisLoaderImpl.loadGenesisFile(genesisFile.getPath());
+        assert genesisFileStream != null;
```
fmacleal, line 62: I know it's silly, but it's just to keep the consistency in the way that we assert stuffs.

```suggestion
        Assertions.assertNotNull(genesisFileStream);
```

rskj-core/src/test/java/org/ethereum/core/genesis/GenesisLoaderImplTest.java
```diff
@@ -0,0 +1,74 @@
...
+    @AfterEach
+    public void cleanUpFiles() {
+        boolean deleted = (new File(RESOURCES_GENESIS_FILE_PATH)).delete();
+        Assertions.assertTrue(deleted);
+    }
+    
+    @Test
+    void loadGenesisFile_fromResourcesDir() {
+        InputStream genesisFileStream = GenesisLoaderImpl.loadGenesisFile(GENESIS_FILE_NAME);
+        assert genesisFileStream != null;
```
fmacleal, line 45: I know it's silly, but it's just to keep the consistency in the way that we assert stuffs.

```suggestion
        Assertions.assertNotNull(genesisFileStream);
```

fmacleal (changes requested):
Very good job, just small observations.

If you agree with the change of throw the exceptions straight away, we would have to change the tests to assert exceptions as well. Let me know what you think. :)

rskj-core/src/main/java/org/ethereum/core/genesis/GenesisLoaderImpl.java
jurajpiar: Erm... The stream is not open yet, so the exception can only happen before or whilst opening it. Therefore, there is nothing to close, is it?

rskj-core/src/main/java/org/ethereum/core/genesis/GenesisLoaderImpl.java
jurajpiar: oh yea cool, wasn't aware of default access. thanks!
That makes me think about all the other "unit" tests that could be simplified and cleaner if we used default access...🧐

rskj-core/src/test/java/org/ethereum/core/genesis/GenesisLoaderImplTest.java
jurajpiar: It's not silly, it's a fair point. However, with throwing exception early this will be removed altogether.

rskj-core/src/test/java/org/ethereum/core/genesis/GenesisLoaderImplTest.java
jurajpiar: It's not silly, it's a fair point. However, with throwing exception early this will be removed altogether.

rskj-core/src/main/java/org/ethereum/core/genesis/GenesisLoaderImpl.java
fmacleal: True, but if you pass the return from this `Files.newInputStream(Paths.get(fileName));` to the variable `inputStream` you will have. 

It's just on the case that some error happened during the opening of this stream,  It's probably very unlikely to happen and probably the GC would handle, but it was a suggestion just to play safe.

But you know what, this is really too much. And looking now, I noticed that the stream will be closed when we are reading the json if something weird happens. But now it occurred to me that you could just catch the Exception strictly necessary, Exception is too broad and we want to catch what is the real root cause and why is not covered if something happens. 

```
        try {
            return Files.newInputStream(Paths.get(fileName));
        } catch (IOException | RuntimeException e) {
            logger.error("Cannot read genesis json file");

            throw new RuntimeException("Cannot open genesis block configuration file", e);
        }```

File: rskj-core/src/main/java/org/ethereum/core/genesis/GenesisLoaderImpl.java [hunk 6]
```diff
@@ -54,6 +56,13 @@
  * 4. Registers the genesis state root in the state root handler
  */
 public class GenesisLoaderImpl implements GenesisLoader {
+
+    static class GenesisLoaderExceptopn extends RuntimeException {
```
jurajpiar, line 60: Response to sonar cloud 
Screenshot 2024-04-02 at 12 19 57

jurajpiar:
pipeline:run

rskj-core/src/main/java/org/ethereum/core/genesis/GenesisLoaderImpl.java
jurajpiar: outdated

fmacleal (approved):

PR #2276, opened 2024-03-26
Implementation of EthSwap mechanism

rmoreliovlabs:
## Description
This pull request is not intended to be merged, its purpose is to review the proof of contect of the EthSwap mechanism.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/util/ContractUtil.java
```diff
@@ -0,0 +1,125 @@
...
+            } else if(arg instanceof BigInteger) {
+                SolidityType uint256Type = SolidityType.getType(SolidityType.UINT);
+                encodedArg = uint256Type.encode(arg);
+            }
+
+            encodedArguments = ByteUtil.merge(encodedArguments, encodedArg);
+        }
+
+        return encodedArguments;
+    }
```
fmacleal, line 89: Some suggestion with a simplification.

```suggestion
    public static byte[] encodePacked(Object...arguments) {
        return Arrays.stream(arguments)
                .map(ContractUtil::encodeArgument)
                .reduce(ByteUtil::merge)
                .orElse(new byte[]{});
    }

    private static byte[] encodeArgument(Object arg) {
        if(arg instanceof byte[]) {
            return SolidityType.getType(SolidityType.BYTES32).encode(arg);
        } else if(arg instanceof RskAddress) {
            byte[] encodedAddress = SolidityType.getType(SolidityType.ADDRESS).encode(((RskAddress) arg).toHexString());
            return org.bouncycastle.util.Arrays.copyOfRange(encodedAddress, 12, encodedAddress.length);
        } else if(arg instanceof BigInteger) {
            return SolidityType.getType(SolidityType.UINT).encode(arg);
        }
        return new byte[]{};
    }
```

rskj-core/src/main/java/co/rsk/util/ContractUtil.java
```diff
@@ -0,0 +1,125 @@
...
+                BigInteger amount = (BigInteger) invocation.args[1];
+                Coin balanceWithClaim = accountInformationProvider.getBalance(newTx.getSender(signatureCache))
+                        .add(Coin.valueOf(amount.longValue()));
+
+                return txCost.compareTo(balanceWithClaim) <= 0;
+            }
+        }
+
+        return false;
+    }
```
fmacleal, line 124: Suggestion of simplification.

```suggestion
    public static boolean isClaimTxAndValid(Transaction newTx, Coin txCost, Constants constants, SignatureCache signatureCache, PropertyGetter propertyGetter) {
        if(isTxValid(newTx, constants)) {
            String swapHash = HexUtils.toUnformattedJsonHex(calculateSwapHash(newTx, signatureCache));
            byte[] key = HashUtil.keccak256(HexUtils.decode(HexUtils.stringToByteArray(swapHash + SWAPS_MAP_POSITION)));
            AccountInformationProvider accountInformationProvider = propertyGetter.getWeb3InfoRetriever().getInformationProvider("latest");
            DataWord swapRecord = accountInformationProvider.getStorageValue(newTx.getReceiveAddress(), DataWord.valueOf(key));
            if(swapRecord != null){
                BigInteger amount = (BigInteger) getTxInvocation(newTx).args[1];
                Coin balanceWithClaim = accountInformationProvider.getBalance(newTx.getSender(signatureCache)).add(Coin.valueOf(amount.longValue()));
                return txCost.compareTo(balanceWithClaim) <= 0;
            }
        }
        return false;
    }

    private static boolean isTxValid(Transaction newTx, Constants constants) {
        return newTx.getReceiveAddress() != null
                && newTx.getData() != null
                && newTx.getReceiveAddress().toHexString().equalsIgnoreCase(constants.getEtherSwapContractAddress())
                && Arrays.equals(Arrays.copyOfRange(newTx.getData(), 0, 4), Constants.CLAIM_FUNCTION_SIGNATURE);
    }
```

rskj-core/src/test/java/co/rsk/core/CallContractTest.java
```diff
@@ -82,7 +82,8 @@ private static ProgramResult callContract(World world, RskAddress receiveAddress
                     blockFactory,
                     new ProgramInvokeFactoryImpl(),
                     new PrecompiledContracts(config, bridgeSupportFactory, world.getBlockTxSignatureCache()),
-                    world.getBlockTxSignatureCache()
+                    world.getBlockTxSignatureCache(),
+                    null
```
fmacleal, line 86: Is this really necessary? If this field can be nullable, we could just create a constructor without it and call the old one with null no?

Something like this:

```
    public TransactionExecutorFactory(
            RskSystemProperties config,
            BlockStore blockStore,
            ReceiptStore receiptStore,
            BlockFactory blockFactory,
            ProgramInvokeFactory programInvokeFactory,
            PrecompiledContracts precompiledContracts,
            BlockTxSignatureCache blockTxSignatureCache) {
        this(config, blockStore, receiptStore, blockFactory, programInvokeFactory, precompiledContracts, blockTxSignatureCache, null);
    }
```

rskj-core/src/test/java/co/rsk/core/bc/BlockExecutorTest.java
```diff
@@ -867,7 +867,8 @@ private static BlockExecutor buildBlockExecutor(TrieStore store, RskSystemProper
                         BLOCK_FACTORY,
                         new ProgramInvokeFactoryImpl(),
                         new PrecompiledContracts(config, bridgeSupportFactory, signatureCache),
-                        signatureCache
+                        signatureCache,
+                        null
```
fmacleal, line 871: If we create the constructor suggested above, we wouldn't need this change.

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
```diff
@@ -98,7 +98,8 @@ protected RepositoryLocator buildRepositoryLocator() {
                 10,
                 100,
                 Mockito.mock(TxQuotaChecker.class),
-                Mockito.mock(GasPriceTracker.class));
+                Mockito.mock(GasPriceTracker.class),
+                null);
```
fmacleal, line 102: Same idea with an override constructor can be applied here making this change unnecessary.

```
    public TransactionPoolImpl(RskSystemProperties config, RepositoryLocator repositoryLocator, BlockStore blockStore, BlockFactory blockFactory, EthereumListener listener, TransactionExecutorFactory transactionExecutorFactory, SignatureCache signatureCache, int outdatedThreshold, int outdatedTimeout, TxQuotaChecker txQuotaChecker, GasPriceTracker gasPriceTracker) {
        this(config, repositoryLocator, blockStore, blockFactory, listener, transactionExecutorFactory, signatureCache, outdatedThreshold, outdatedTimeout, txQuotaChecker, gasPriceTracker, null);
    }
```

rskj-core/src/test/java/co/rsk/mine/MinerServerTest.java
```diff
@@ -124,7 +124,8 @@ protected RepositoryLocator buildRepositoryLocator() {
                 10,
                 100,
                 mock(TxQuotaChecker.class),
-                mock(GasPriceTracker.class));
+                mock(GasPriceTracker.class),
+                null);
```
fmacleal, line 128: If we create the other constructor, we wouldn't need this change.

rskj-core/src/test/java/co/rsk/net/SyncProcessorTest.java
```diff
@@ -797,7 +797,8 @@ void processBodyResponseWithTransactionAddsToBlockchain() {
                         blockFactory,
                         new ProgramInvokeFactoryImpl(),
                         new PrecompiledContracts(config, bridgeSupportFactory, signatureCache),
-                        new BlockTxSignatureCache(new ReceivedTxSignatureCache())
+                        new BlockTxSignatureCache(new ReceivedTxSignatureCache()),
+                        null
```
fmacleal, line 801: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/co/rsk/peg/RskForksBridgeTest.java
```diff
@@ -475,7 +475,8 @@ private BridgeState callGetStateForDebuggingTx() throws IOException {
                 new BlockFactory(beforeBambooProperties.getActivationConfig()),
                 new ProgramInvokeFactoryImpl(),
                 new PrecompiledContracts(beforeBambooProperties, world.getBridgeSupportFactory(), world.getBlockTxSignatureCache()),
-                world.getBlockTxSignatureCache()
+                world.getBlockTxSignatureCache(),
+                null
```
fmacleal, line 479: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/co/rsk/remasc/RemascProcessMinerFeesTest.java
```diff
@@ -1034,7 +1034,8 @@ private BlockExecutor buildBlockExecutor(RepositoryLocator repositoryLocator, Bl
                         blockFactory,
                         new ProgramInvokeFactoryImpl(),
                         new PrecompiledContracts(config, bridgeSupportFactory, signatureCache),
-                        new BlockTxSignatureCache(new ReceivedTxSignatureCache())
+                        new BlockTxSignatureCache(new ReceivedTxSignatureCache()),
+                        null
```
fmacleal, line 1038: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/co/rsk/remasc/RemascStorageProviderTest.java
```diff
@@ -491,7 +491,8 @@ void paysOnlyBlocksWithEnoughBalanceAccumulatedAfterRFS() throws IOException {
                         new BlockFactory(config.getActivationConfig()),
                         new ProgramInvokeFactoryImpl(),
                         new PrecompiledContracts(config, bridgeSupportFactory, signatureCache),
-                        new BlockTxSignatureCache(new ReceivedTxSignatureCache())
+                        new BlockTxSignatureCache(new ReceivedTxSignatureCache()),
+                        null
```
fmacleal, line 495: Also here

rskj-core/src/test/java/co/rsk/remasc/RemascTestRunner.java
```diff
@@ -146,7 +146,8 @@ public void start() {
                         blockFactory,
                         programInvokeFactory,
                         precompiledContracts,
-                        blockTxSignatureCache
+                        blockTxSignatureCache,
+                        null
```
fmacleal, line 150: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/co/rsk/test/World.java
```diff
@@ -183,7 +183,8 @@ public BlockExecutor getBlockExecutor() {
                             new BlockFactory(config.getActivationConfig()),
                             programInvokeFactory,
                             new PrecompiledContracts(config, bridgeSupportFactory, blockTxSignatureCache),
-                            blockTxSignatureCache
+                            blockTxSignatureCache,
+                            null
```
fmacleal, line 187: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/co/rsk/test/builders/BlockBuilder.java
```diff
@@ -115,7 +115,8 @@ public Block build() {
                             new BlockFactory(config.getActivationConfig()),
                             new ProgramInvokeFactoryImpl(),
                             new PrecompiledContracts(config, bridgeSupportFactory, new BlockTxSignatureCache(new ReceivedTxSignatureCache())),
-                            new BlockTxSignatureCache(new ReceivedTxSignatureCache())
+                            new BlockTxSignatureCache(new ReceivedTxSignatureCache()),
+                            null
```
fmacleal, line 119: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/co/rsk/test/builders/BlockChainBuilder.java
fmacleal, line 1: Same idea with an override constructor can be used here as well, for the two changes.

rskj-core/src/test/java/co/rsk/test/dsl/WorldDslProcessor.java
```diff
@@ -323,7 +323,8 @@ private void processBlockChainCommand(DslCommand cmd) {
                             new BlockFactory(config.getActivationConfig()),
                             programInvokeFactory,
                             null,
-                            world.getBlockTxSignatureCache()
+                            world.getBlockTxSignatureCache(),
+                            null
```
fmacleal, line 327: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/co/rsk/vm/MinerHelper.java
```diff
@@ -103,7 +103,7 @@ public void processBlock( Block block, Block parent) {
                     null,
                     blockFactory,
                     null,
-                    new PrecompiledContracts(config, bridgeSupportFactory, signatureCache), blockTxSignatureCache);
+                    new PrecompiledContracts(config, bridgeSupportFactory, signatureCache), blockTxSignatureCache, null);
```
fmacleal, line 106: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/org/ethereum/core/ImportLightTest.java
fmacleal, line 1: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/org/ethereum/core/TransactionExecutorTest.java
fmacleal, line 1: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/org/ethereum/core/TransactionTest.java
fmacleal, line 1: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/org/ethereum/jsontestsuite/runners/StateTestRunner.java
fmacleal, line 1: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/org/ethereum/jsontestsuite/TestRunner.java
fmacleal, line 1: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/org/ethereum/rpc/Web3ImplTest.java
fmacleal, line 1: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/org/ethereum/util/EthModuleTestUtils.java
fmacleal, line 1: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/org/ethereum/vm/program/NestedContractsTest.java
```diff
@@ -205,6 +205,7 @@ private EthModule buildEthModule(World world) {
                 null,
                 new ProgramInvokeFactoryImpl(),
                 new PrecompiledContracts(config, world.getBridgeSupportFactory(), new BlockTxSignatureCache(new ReceivedTxSignatureCache())),
+                null,
```
fmacleal, line 208: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/co/rsk/mine/TransactionModuleTest.java
fmacleal, line 1: Same idea with an override constructor can be used here as well.

rskj-core/src/test/java/co/rsk/util/ContractUtilTest.java
```diff
@@ -0,0 +1,92 @@
...
+        when(mockedAccountInformationProvider.getStorageValue(any(RskAddress.class), any(DataWord.class)))
+                .thenReturn(DataWord.valueOf(1));
+
+        PropertyGetter propertyGetter = () -> mockedWeb3InformationRetriever;
+
+        boolean result = ContractUtil.isClaimTxAndValid(mockedTx, txCost, testConstans, signatureCache, propertyGetter);
+
+        assertTrue(result);
+    }
+
```
fmacleal, line 69: What do you think in add a scenario of test where the Tx doesn't cover the gas and the function `isClaimTxAndValid` return false?

We could also have scenarios where the received address from the contract isn't the expected one and other where the signature isn't valid.

fmacleal (commented):
Very good job, I have a few suggestions to simplify some functions and ideas to avoid some changes that could be skipped.

Let me know your thoughts. 🙂

rmoreliovlabs:
pipeline:run

rskj-core/src/main/java/co/rsk/util/ContractUtil.java
rmoreliovlabs: I did something similar to what was suggested for this [method](https://github.com/rsksmart/rskj/pull/2276/files#diff-fa9928b27fe4111c5e6d17d930a4e40c2370206982c394242dd7b3f15f05a4f9R139).

rskj-core/src/main/java/co/rsk/util/ContractUtil.java
```diff
@@ -0,0 +1,125 @@
...
+import org.ethereum.core.Transaction;
+import org.ethereum.crypto.HashUtil;
+import org.ethereum.solidity.SolidityType;
+import org.ethereum.util.ByteUtil;
+import org.ethereum.vm.DataWord;
+
+import java.math.BigInteger;
+import java.util.Arrays;
+
+public class ContractUtil {
```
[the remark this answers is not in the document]
rmoreliovlabs: Yes, ContractUtil was a confusing name, I changed it to `EthSwapUtil` instead, let me know what you think.

rskj-core/src/main/java/org/ethereum/config/Constants.java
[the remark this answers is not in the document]
rmoreliovlabs: Since this is a proof of concept I only installed the contract in the `regtest` network but the idea is to install it in the other ones and have the address be dynamic according to the network.

rskj-core/src/test/java/co/rsk/core/CallContractTest.java
rmoreliovlabs: You are right, I'll update it.

rskj-core/src/integrationTest/java/co/rsk/EtherSwapTest.java
```diff
@@ -0,0 +1,263 @@
...
+                proc -> {
+                    try {
+                        Response getBalanceResponse = getBalanceRequest(claimAddress);
+                        JsonNode jsonRpcResponse = objectMapper.readTree(getBalanceResponse.body().string());
+                        JsonNode currentBalance = jsonRpcResponse.get(0).get("result");
+                        BigInteger balanceBigInt = new BigInteger(HexUtils.removeHexPrefix(currentBalance.asText()), 16);
+                        Assertions.assertEquals(0, balanceBigInt.compareTo(BigInteger.ZERO));
+
+                        lockTxRequest(refundAddress, lockData, amount);
+                        TimeUnit.SECONDS.sleep(5);
```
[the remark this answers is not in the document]
rmoreliovlabs: I'll try to implement it that way.

rskj-core/src/test/java/co/rsk/core/CallContractTest.java
rmoreliovlabs: In this [commit](https://github.com/rsksmart/rskj/pull/2276/commits/1b367e3e10d961f54bd67936c4833d3c364c724b) I reverted this and similar changes so this is nor longer an issue.

rskj-core/src/main/java/co/rsk/util/ContractUtil.java
fmacleal: I think it's debatable, for big arrays the stream sometimes wins on performance. I prefer because makes the code way more readable. But you could enclose the if-else chain in a method with a meaningful name stating your intention. This would help in the readability of this if-else chain. Something like:

```
private static byte[] encodeArgumentAccordingInstanceType(Object arg) {
        if(arg instanceof byte[]) {
            return SolidityType.getType(SolidityType.BYTES32).encode(arg);
        } else if(arg instanceof RskAddress) {
            byte[] encodedAddress = SolidityType.getType(SolidityType.ADDRESS).encode(((RskAddress) arg).toHexString());
            return org.bouncycastle.util.Arrays.copyOfRange(encodedAddress, 12, encodedAddress.length);
        } else if(arg instanceof BigInteger) {
            return SolidityType.getType(SolidityType.UINT).encode(arg);
        }
        return new byte[]{};
}
```

rskj-core/src/test/java/co/rsk/core/CallContractTest.java
fmacleal: I saw this change, I will make some comments there on the class created.

rskj-core/src/main/java/co/rsk/util/ContractUtil.java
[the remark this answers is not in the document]
fmacleal: Yep, it's better.

rskj-core/src/main/java/org/ethereum/config/Constants.java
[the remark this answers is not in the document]
fmacleal: Since you changed the approach and code, I will check the new one.

rskj-core/src/integrationTest/java/co/rsk/EtherSwapTest.java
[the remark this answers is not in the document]
fmacleal: The tests still fails when I run on intellij. Is there some configuration needed?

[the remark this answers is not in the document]
fmacleal: Also failed running on command line. 😢 

image

rskj-core/src/main/java/co/rsk/net/handler/txvalidator/TxValidatorStep.java
```diff
@@ -32,6 +33,7 @@
  */
 public interface TxValidatorStep {
 
-    TransactionValidationResult validate(Transaction tx, @Nullable AccountState state, BigInteger gasLimit, Coin minimumGasPrice, long bestBlockNumber, boolean isFreeTx);
+    TransactionValidationResult validate(Transaction tx, @Nullable AccountState state, BigInteger gasLimit, Coin minimumGasPrice, long bestBlockNumber, boolean isFreeTx, ClaimTransactionInfoHolder claimTransactionInfoHolder);
```
fmacleal, line 36: I think add this method to be implemented to all other ones just because of a parameter it's too cumbersome. You had to change all validators just to change the logic of one one them.

Wouldn't be possible `TxValidatorAccountBalanceValidator` to receive the parameters necessary to create a copy of the ClaimTransactionInfoHolder? This way, you wouldn't need to pass the object every time you are calling `shouldAcceptTx`?

rskj-core/src/main/java/org/ethereum/config/Constants.java
fmacleal, line 1: Good refactor. 👍

rskj-core/src/integrationTest/java/co/rsk/EtherSwapTest.java
```diff
@@ -0,0 +1,318 @@
...
+                "        \"from\": \"" + refundAddress + "\",\n" +
+                "        \"to\": \"0x" + Constants.regtest().getEtherSwapContractAddress() + "\",\n" +
+                "        \"data\": \"0x" + ByteUtil.toHexString(lockData) + "\",\n" +
+                "        \"value\": \"" + HexUtils.toQuantityJsonHex(amount.longValue()) + "\",\n" +
+                "        \"gas\": \"0xc350\",\n" +
+                "        \"gasPrice\": \"0x1\"\n" +
+                "    }],\n" +
+                "    \"id\": 1,\n" +
+                "    \"jsonrpc\": \"2.0\"\n" +
+                "}]";
```
jurajpiar, line 103: Could we use some cleaner pattern for JSON construction, something like a builder? This is quite diff to read.
If none is available at hand, it could be as simple as:
```
public class JsonBuilder {
        private final ObjectMapper mapper = new ObjectMapper();
        private final ObjectNode rootNode = mapper.createObjectNode();

        public JsonBuilder set(String key, String value) {
            rootNode.put(key, value);
            return this;
        }

        public JsonBuilder set(String key, int value) {
            rootNode.put(key, value);
            return this;
        }

        public JsonBuilder set(String key, ObjectNode value) {
            rootNode.set(key, value);
            return this;
        }

        public JsonBuilder set(String key, ArrayNode value) {
            rootNode.set(key, value);
            return this;
        }

        public String build() {
            return rootNode.toString();
        }

        public String buildArray() {
            return "[" + rootNode.toString() + "]";
        }
    }
```

so then we can use it as so:

```suggestion
String lockTxRequestContent = new JsonBuilder().set("method", "eth_sendTransaction")
                .set("params", new JsonBuilder()
                        .set("from", refundAddress)
                        .set("to", "0x" + Constants.regtest().getEtherSwapContractAddress())
                        .set("data", "0x" + ByteUtil.toHexString(lockData))
                        .set("value", HexUtils.toQuantityJsonHex(amount.longValue()))
                        .set("gas", "0xc350")
                        .set("gasPrice", "0x1")
                        .buildArray())
                .set("id", 1)
                .set("jsonrpc", "2.0").buildArray();
```

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
```diff
@@ -0,0 +1,209 @@
...
+        return false;
+    }
+
+    private static boolean isSameTx(Transaction pendingTx, Transaction newTx, SignatureCache signatureCache) {
+        return pendingTx.getSender(signatureCache).toHexString().equals(newTx.getSender(signatureCache).toHexString())
+                && pendingTx.getReceiveAddress().toHexString().equals(newTx.getReceiveAddress().toHexString())
+                && Arrays.equals(pendingTx.getData(), newTx.getData());
+    }
+
+    public static boolean isClaimTxAndSenderCanPayAlongPendingTx(ClaimTransactionInfoHolder claimTransactionInfoHolder, List<Transaction> pendingTransactions) {
```
jurajpiar, line 168: Java naming conventions are brutal `isClaimTxAndSenderCanPayAlongPendingTx` 😮🔫

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
```diff
@@ -0,0 +1,209 @@
...
+import org.ethereum.solidity.SolidityType;
+import org.ethereum.util.ByteUtil;
+import org.ethereum.vm.DataWord;
+import org.ethereum.vm.GasCost;
+
+import java.math.BigInteger;
+import java.util.Arrays;
+import java.util.List;
+
+public class EthSwapUtil {
```
jurajpiar, line 40: Why do we call it `EthSwap`? I know that's the orig contract this is based on, but the use case is so different that this name is just confusing. Even Sergio's `ClaimTimelock` would have been better, even though that's still not descriptive enough, but at least it doesn't add confusion by referring to ETH and swap. This goes for all fields and methods, too. Unless I'm missing something?
I'd suggest something in the lines of gas deferring, or self-paying.
Perhaps:
- DefferedGasTransaction
- SelfPayingTransaction
- DefferedGasVault
- DefferedGasLocker
- GasDeferredWithdraw
- SelfPayingWithdraw

I'm sure we can think of a better name together.

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
```diff
@@ -0,0 +1,209 @@
...
+            }
+
+            encodedArguments = ByteUtil.merge(encodedArguments, encodedArg);
+        }
+
+        return encodedArguments;
+    }
+
+    public static CallTransaction.Invocation getTxInvocation(Transaction newTx) {
+        String abi = "[{\"constant\":false,\"inputs\":[{\"name\":\"preimage\",\"type\":\"bytes32\"}, {\"name\":\"amount\",\"type\":\"uint256\"}, {\"name\":\"address\",\"type\":\"address\"}, {\"name\":\"timelock\",\"type\":\"uint256\"}],\"name\":\"claim\",\"outputs\":[],\"payable\":false,\"type\":\"function\"}]";
```
jurajpiar, line 95: again, why not make it more readable?

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    private static Coin getTxCost(Transaction tx,
+                                  ActivationConfig.ForBlock activation,
+                                  Constants constants,
+                                  SignatureCache signatureCache) {
+        Coin txCost = tx.getValue();
+        if (activation == null || tx.transactionCost(constants, activation, signatureCache) > 0) {
+            BigInteger gasLimit = BigInteger.valueOf(GasCost.toGas(tx.getGasLimit()));
+            txCost = txCost.add(tx.getGasPrice().multiply(gasLimit));
```
asoto-iov, line 131: you can refactor this method by doing returning directly `tx.getValue().add(tx.getGasPrice().multiply(gasLimit));`

and if the if conditionals are false then `return tx.getValue();`

There is no need for the local `Coin txCost = tx.getValue();`

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    private static Coin getLockedAmount(Transaction tx) {
+        CallTransaction.Invocation invocation = getTxInvocation(tx);
+        BigInteger amount = (BigInteger) invocation.args[1];
+
+        return Coin.valueOf(amount.longValue());
+    }
+
+    private static Coin getTxCost(Transaction tx,
```
asoto-iov, line 124: By this name the fuction returns the _transaction cost_ but then we have a method `tx.transactionCost(constants, activation, signatureCache)`
What is the difference between these two costs?

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.List;
+
+public class EthSwapUtil {
+    private static final String SWAPS_MAP_POSITION = "0000000000000000000000000000000000000000000000000000000000000000";
+
+    private EthSwapUtil() {
+        throw new IllegalStateException("Utility class");
+    }
+
+    public static byte[] calculateSwapHash(Transaction newTx, SignatureCache signatureCache) {
```
asoto-iov, line 47: This method is being used only once inside this class, is it needed to be public?

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
asoto-iov, line 40: Hey guys I'm not sure about this utility class. There is only 4 methods called from other classes and we have 7 public methods. 

When methods are moved to utility clases, the reason usually is because they are going to be used in different contexts and there is no need to instantiate a new object for doing this kind of operations.

But in this case it seems that there is only one specific usage related with specific kind of Transaction. 

These kind of classes could be dangerous because it could be easy to lost the context of it when working with then. Also the own class or some of the  methods could grow or being modified ending in situations hard to maintain or leading to errors.

rskj-core/src/main/java/co/rsk/core/bc/ClaimTransactionInfoHolder.java
```diff
@@ -0,0 +1,113 @@
...
+        Coin txCost = tx.getGasPrice().multiply(tx.getGasLimitAsInteger());
+
+        Coin senderB = getSenderBalance();
+        if (senderB.compareTo(txCost) >= 0) {
+            return true;
+        }
+
+        Coin txUserB = getTxUserBalance();
+        return txUserB.add(senderB).compareTo(txCost) >= 0;
+    }
```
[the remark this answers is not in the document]
asoto-iov: +1 I would remove un used methods.

asoto-iov (commented):
Added some comments, most of then related with the utility class but I'm still reviewing the PR.

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
rmoreliovlabs: Fair point. I can move this logic to a instantiable class, maybe combine this with the ClaimTransactionInfoHolder (changing the name of course) since I am creating an instance for that anyways. Wdyt? @asoto-iov @jurajpiar @fmacleal.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -456,7 +461,15 @@ private Block createFakePendingBlock(Block best) {
...
     private TransactionValidationResult shouldAcceptTx(Transaction tx, RepositorySnapshot currentRepository) {
-        return validator.isValid(tx, bestBlock, currentRepository.getAccountState(tx.getSender(signatureCache)));
+        ClaimTransactionInfoHolder claimTransactionInfoHolder = new ClaimTransactionInfoHolder(
+                tx,
+                currentRepository,
+                signatureCache,
+                config.getNetworkConstants(),
+                bestBlock == null ? null : config.getActivationConfig().forBlock(bestBlock.getNumber())
+        );
+
```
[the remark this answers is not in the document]
rmoreliovlabs: True, I'll create a method to get the `forBlock` object.

rskj-core/src/main/java/co/rsk/net/handler/txvalidator/TxValidatorStep.java
rmoreliovlabs: At first I created a constructor to send the parameters like you mentioned but this would work in the end because some parameter like RepositorySnapshot might change after the `TxValidatorAccountBalanceValidator` instance is created. That is why I suggested overloading the validate method to add the ClaimTransactionInfoHolder wrapper, that way I can pass the most recent values of the arguments that might change.

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
```diff
@@ -0,0 +1,209 @@
...
+        CallTransaction.Contract contract = new CallTransaction.Contract(abi);
+        return contract.parseInvocation(newTx.getData());
+    }
+
+    public static boolean isClaimTx(Transaction tx, Constants constants) {
+        return tx.getReceiveAddress() != null
+                && !tx.getReceiveAddress().toHexString().isEmpty()
+                && tx.getData() != null
+                && tx.getReceiveAddress().toHexString().equalsIgnoreCase(constants.getEtherSwapContractAddress())
+                && Arrays.equals(Arrays.copyOfRange(tx.getData(), 0, 4), Constants.CLAIM_FUNCTION_SIGNATURE);
```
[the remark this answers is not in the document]
rmoreliovlabs: I overlooked that, I'll add that check.

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
rmoreliovlabs: What about `ClaimTransactionValidator`? Since the class is used to validate transactions that are sent specifically to claim funds from a contract.

asoto-iov: Have you think about create a new type of Transaction?
As an example we have the RemascTransaction maybe we can have a ClaimTransaction which extends Transaction to contain the logic related to it.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -282,6 +291,14 @@ private BlockResult executeInternal(
...
             logger.trace("apply block: [{}] tx: [{}] ", block.getNumber(), i);
 
+            if (EthSwapUtil.isClaimTx(tx, constants)
+                    && !EthSwapUtil.hasLockedFunds(tx,signatureCache, track)) {
+                logger.warn("block: [{}] discarded claim tx: [{}], because the funds it tries to claim no longer exist in contract",
+                        block.getNumber(),
+                        tx.getHash());
+                continue;
+            }
+
```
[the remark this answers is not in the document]
rmoreliovlabs: Yes, this logic covers the cases when the funds are claimed by another mean after the claim transaction is added to the mempool but before it is executed. This first confirm the transaction is a claim transaction and then checks if its corresponding funds are still locked in the contract and if not it discards the transaction.

rskj-core/src/integrationTest/java/co/rsk/EtherSwapTest.java
rmoreliovlabs: I think we can do this in another ticket since the change will just affect the readability of the code and we could also refactor other instances were there are other Json variables with the same format.

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
```diff
@@ -0,0 +1,209 @@
...
+
+            encodedArguments = ByteUtil.merge(encodedArguments, encodedArg);
+        }
+
+        return encodedArguments;
+    }
+
+    public static CallTransaction.Invocation getTxInvocation(Transaction newTx) {
+        String abi = "[{\"constant\":false,\"inputs\":[{\"name\":\"preimage\",\"type\":\"bytes32\"}, {\"name\":\"amount\",\"type\":\"uint256\"}, {\"name\":\"address\",\"type\":\"address\"}, {\"name\":\"timelock\",\"type\":\"uint256\"}],\"name\":\"claim\",\"outputs\":[],\"payable\":false,\"type\":\"function\"}]";
+        CallTransaction.Contract contract = new CallTransaction.Contract(abi);
```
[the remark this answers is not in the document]
rmoreliovlabs: Since this object does not change I moved it and declare it as a class vairable.

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
rmoreliovlabs: I though I needed the if check this method is used in to get the correct tx cost value but it looks like I don't. I've removed the check so this method is not necessary here.

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
rmoreliovlabs: Not really, at least not right now. My reasoning was to make it public in cases we need it somewhere out side the class in the future but for now I've changed it to private.

rmoreliovlabs:
pipeline:run

rmoreliovlabs:
pipeline:run

rskj-core/src/integrationTest/java/co/rsk/snapshotsync/SnapshotSyncIntegrationTest.java
fmacleal, line 1: Remember to remove this class, it shouldn't go to master.

Please, do a rebase as well with master to get the latest changes on the integration tests.

rmoreliovlabs: Will do

rskj-core/src/main/java/co/rsk/net/handler/TxPendingValidator.java
```diff
@@ -86,7 +86,7 @@ public TransactionValidationResult isValid(Transaction tx, Block executionBlock,
         }
 
         for (TxValidatorStep step : validatorSteps) {
-            TransactionValidationResult validationResult = step.validate(tx, state, blockGasLimit, minimumGasPrice, bestBlockNumber, basicTxCost == 0);
+            TransactionValidationResult validationResult = step.validate(tx, state, blockGasLimit, minimumGasPrice, bestBlockNumber, basicTxCost == 0, claimTransactionInfoHolder);
```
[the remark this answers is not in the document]
rmoreliovlabs: The original `validate` method was updated to recieve the new arguments so I removed the other one I created.

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
```diff
@@ -0,0 +1,209 @@
...
+        return contract.parseInvocation(newTx.getData());
+    }
+
+    public static boolean isClaimTx(Transaction tx, Constants constants) {
+        return tx.getReceiveAddress() != null
+                && !tx.getReceiveAddress().toHexString().isEmpty()
+                && tx.getData() != null
+                && tx.getReceiveAddress().toHexString().equalsIgnoreCase(constants.getEtherSwapContractAddress())
```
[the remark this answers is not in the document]
rmoreliovlabs: Done

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
rmoreliovlabs: We can discuss this idea with the team.

rskj-core/src/test/java/co/rsk/util/ClaimTransactionValidatorTest.java
```diff
@@ -0,0 +1,146 @@
...
+//    @Test
+//    public void whenCalculateSwapHashIsCalled_shouldReturnExpectedHash() {
+//        String expectedHash = "0x3dc21e0a710489c951f29205f9961b2c311d48fdf5a35545469d7b43e88f7624";
+//
+//        Transaction mockedClaimTx = createClaimTx(10, "preimage".getBytes(StandardCharsets.UTF_8), 0);
+//
+//        byte[] result = claimTransactionValidator.calculateSwapHash(mockedClaimTx, new ReceivedTxSignatureCache());
+//
+//        assertEquals(expectedHash, HexUtils.toJsonHex(result));
+//    }
```
fmacleal, line 63: Check this test, either delete or uncomment it.

fmacleal (approved):
Good job mate, it's a complex issue and I know you had to change it lots of time. 
Congrats!
:)

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -282,6 +288,16 @@ private BlockResult executeInternal(
         for (Transaction tx : block.getTransactionsList()) {
             logger.trace("apply block: [{}] tx: [{}] ", block.getNumber(), i);
 
+            if (claimTransactionValidator.isFeatureActive(activationConfig.forBlock(block.getNumber()))) {
+                if(claimTransactionValidator.isClaimTx(tx)
+                        && !claimTransactionValidator.hasLockedFunds(tx, track)) {
+                    logger.warn("block: [{}] discarded claim tx: [{}], because the funds it tries to claim no longer exist in contract",
+                            block.getNumber(),
+                            tx.getHash());
+                    continue;
```
VaWheel, line 297: may cause an issue if just skipped. should be analyzed in more detail.

rskj-core/src/main/java/co/rsk/util/EthSwapUtil.java
rmoreliovlabs: addressed, changed the name to `ClaimTransactionValidator`

rskj-core/src/integrationTest/java/co/rsk/EtherSwapTest.java
```diff
@@ -0,0 +1,263 @@
...
+                "    \"method\": \"eth_getTransactionReceipt\",\n" +
+                "    \"params\": [\n" +
+                "        \"" + txHash + "\"\n" +
+                "    ],\n" +
+                "    \"id\": 1,\n" +
+                "    \"jsonrpc\": \"2.0\"\n" +
+                "}]";
+
+        return OkHttpClientTestFixture.sendJsonRpcMessage(getReceiptRequestContent, LOCAL_PORT);
+    }
```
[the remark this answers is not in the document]
rmoreliovlabs: Addressed, moved it to it's own class, `ClaimTransactionTest`

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
rmoreliovlabs: The reason I used this approach is that it aligns with the existing logic in the `BlockExecutor`. Just like how invalid transactions are managed there, invalid `Claim` transactions are discarded if `discardInvalidTxs` is set to true. If not, the execution is stopped

For General Transactions, the handling is similar:
```
if (!acceptInvalidTransactions && !transactionExecuted) {
    if (discardInvalidTxs) {
        logger.warn("block: [{}] discarded tx: [{}]", block.getNumber(), tx.getHash());
        continue;
    } else {
        logger.warn("block: [{}] execution interrupted because of invalid tx: [{}]",
                    block.getNumber(), tx.getHash());
        profiler.stop(metric);
        return BlockResult.INTERRUPTED_EXECUTION_BLOCK_RESULT;
    }
}
```
Just as we discard invalid general transactions, we also discard invalid Claim transactions based on specific validation criteria. This ensures we maintain consistency in handling invalid transactions across the code
Let me know if you have any further questions!

rskj-core/src/test/java/co/rsk/util/ContractUtilTest.java
rmoreliovlabs: solved, added `whenIsClaimTxAndValidIsCalled_withInsufficientFunds_shouldReturnFalse()` test

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
VaWheel: But `Claim` transactions that don't have locked funds `!claimTransactionValidator.hasLockedFunds(tx, track))` would be just discarded without interrupting the execution regardless of `discardInvalidTxs `?

fmacleal (approved):

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
rmoreliovlabs: Yes, invalid claim transactions (those without locked funds) are indeed discarded without interrupting block execution, regardless of the `discardInvalidTxs` value. This behavior is intentional and aligned with the design outlined in the **[TDD](https://www.notion.so/iovlabs/EthSwap-PoC-Technical-Design-Document-f0adbaac6fd848ab8f33d9b5bc56151c)** and Sergio's proposal

The logic was implemented to ensure that claim transactions, which don't meet the requirement of having locked funds, are simply discarded to prevent them from affecting the overall block processing. This ensures that the chain execution continues smoothly **without** being interrupted by these invalid claim transactions. The proposal specifies that invalid claims should not stop the block execution but rather be **skipped** to maintain stability and performance. You can find the detailed reasoning behind this in the initial Sergio's proposal document **[here](https://docs.google.com/document/d/1OA6neNxmA4rprmGjETWnhV3O2GFKjK1Bep21TNeeRAk/edit?pli=1#heading=h.ap10gtlttkb6)**

Happy to explain further if needed!

PR #2279, opened 2024-04-01
fix(discovery): makes pong send only IP without hostname

jurajpiar:
As RSK user I want IP addresses (instead of DNS names) to be sent to other peers in response to FIND_NODE peer discovery message

## Description

Instead of DNS hostnames a node always sends IP addresses in response to the FIND_NODE peer discovery message

## Motivation and Context

As of now, in the peer discovery protocol, when a peer asks for neighbours by sending FIND_NODE message in response it gets a sub list of other discovered peers in a network. Sometimes, instead of IP addresses of remote peers that list might include DNS names resolved via reverse DNS lookup in a context of the peer that returns that response. In some cases this might lead to issues when the reverse DNS resolution happens in different network contexts, for instance, when inside AWS VPC an IP can be resolved to ip-10-10-133-54.us-west-2.compute.internal, but outside that host name would make no sense.

## How Has This Been Tested?

Unit test included.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
+++ b/rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
@@ -259,7 +259,7 @@ public class PeerExplorer {
             this.pendingPingRequests.remove(message.getMessageId());
             NodeChallenge challenge = this.challengeManager.removeChallenge(message.getMessageId());
             if (challenge == null) {
-                this.addConnection(message, request.getAddress().getHostString(), request.getAddress().getPort());
+                this.addConnection(message, request.getAddress().getAddress().getHostAddress(), request.getAddress().getPort());
             }
         } else {
             logger.debug("handlePong - Peer discovery request with id [{}] is either null or invalid", message.getMessageId());
--- a/rskj-core/src/test/java/co/rsk/net/discovery/PeerExplorerTest.java
+++ b/rskj-core/src/test/java/co/rsk/net/discovery/PeerExplorerTest.java
@@ -144,7 +144,7 @@ class PeerExplorerTest {
 
         ECKey key1 = ECKey.fromPrivate(Hex.decode(KEY_1)).decompress();
         String check = UUID.randomUUID().toString();
-        PingPeerMessage pingPeerMessageWithDifferentNetwork = PingPeerMessage.create(HOST_1, PORT_1, check, key1, this.NETWORK_ID2);
+        PingPeerMessage pingPeerMessageWithDifferentNetwork = PingPeerMessage.create(HOST_1, PORT_1, check, key1, NETWORK_ID2);
         DiscoveryEvent incomingPingEvent = new DiscoveryEvent(pingPeerMessageWithDifferentNetwork, new InetSocketAddress(HOST_1, PORT_1));
 
         //A message is received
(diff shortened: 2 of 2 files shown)


asoto-iov (approved):

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
```diff
@@ -242,7 +242,7 @@ private void handlePong(InetSocketAddress pongAddress, PongPeerMessage message)
             this.pendingPingRequests.remove(message.getMessageId());
             NodeChallenge challenge = this.challengeManager.removeChallenge(message.getMessageId());
             if (challenge == null) {
-                this.addConnection(message, request.getAddress().getHostString(), request.getAddress().getPort());
+                this.addConnection(message, request.getAddress().getAddress().getHostAddress(), request.getAddress().getPort());
```
asoto-iov, line 245: By doing this change `Node.getHost()` will store the same value as the one provided by `Node.getAddress().getAddress().getHostAddress()` and we will be loosing the current content of the host var. Instead of update this logic to store a repeated value, wouldn't be better to update the place where the Node.getHost() is used to be be replaced by the current `...getAddress().getHostAddress() `option? 

For me it seems that the problem is where the getHost() is being used instead of the value being provided by node.getHost()

rskj-core/src/test/java/co/rsk/net/discovery/PeerExplorerTest.java
fmacleal, line 1: Shouldn't we add the validation to not contain Host name also on the tests `handleFindNodeMessageWithExtraNodes` and `testCleanPeriod` for example? They also handle pong/find_nodes discovery messages as well. Wdyt?

fmacleal (approved):
Good job. I don't have any specific remark and I think that what @asoto-iov commented already cover an important part.

I am just wondering if makes sense some more small tests on the places I have commented below. Let me know your opinion, it doesn't prevent the approval.

rskj-core/src/main/java/co/rsk/net/discovery/PeerExplorer.java
Vovchyk: the thing is that we do not use any DNS name received from outside. To get an address of a peer that sent us a message in the `handleMessage()` method we use `event.getAddress()`, which returns an instance of `InetSocketAddress` class. In essence, this instance combines two values: actual IP address of a sender and host name resolved locally by reverse dns lookup (see how `InetAddress.getByName()` works). But that host name might make no sense in other network contexts. It can only lead to issues related to specific network setups, cause what we care about is a source IP address of a peer that sends us a message

rskj-core/src/test/java/co/rsk/net/discovery/PeerExplorerTest.java
jurajpiar: I don't think so personally, as they don't add any more cases to the problem.
In theory, this should be its own test (well, 2 tests), but the existing tests are already covering more than a single case, so it doesn't seem to make sense to now start writing clean tests.

jurajpiar:
pipeline:run

fmacleal (approved):

asoto-iov (dismissed):

rskj-core/src/test/java/co/rsk/net/discovery/PeerExplorerTest.java
```diff
@@ -82,6 +81,19 @@ class PeerExplorerTest {
     private static final int NETWORK_ID1 = 1;
     private static final int NETWORK_ID2 = 2;
 
+    private static void assertNoHostnameFromDiscovery(PeerExplorer peerExplorer) {
```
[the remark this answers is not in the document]
jurajpiar: done

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

Vovchyk (approved):

PR #2290, opened 2024-04-16
feat(ci): migrates circleci to github actions

jurajpiar:
As RSK dev I want to migrate CircleCI pipeline in RSKj to Github Actions

Depends on: https://github.com/rsksmart/mining-integration-tests/pull/32


build.gradle
```diff
@@ -7,3 +7,13 @@ subprojects {
...
+
+sonarqube {
+    properties {
+        property "sonar.java.binaries", "rskj-core/build/classes/java/main"
+        property "sonar.scm.provider", "git"
+        property "sonar.projectBaseDir", project.projectDir
+        property "sonar.projectKey", "rskj"
+        property "sonar.organization", "rsksmart"
+    }
+}
```
jurajpiar, line 19: didn't work without these. I added them one by one as required by the failing process.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,259 @@
...
+          else
+            extra_flags="-Dsonar.branch.name=master"
+          fi
+          pwd
+          ls -al
+          ./gradlew sonarqube --no-daemon -x build -x test \
+              $extra_flags \
+              -Dsonar.organization=rsksmart \
+              -Dsonar.host.url="https://sonarcloud.io" \
+              -Dsonar.login="${{ secrets.SONAR_TOKEN }}"
```
jurajpiar, line 111: didn't work with `-Dsonar.token=`. Can't remember why, but I did also find a stackoverflow comment suggesting to use `.login` for now.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Get Gradle jar
+        run: |
+          if [ ! -f ./gradle/wrapper/gradle-wrapper.jar ]; then
+            ./configure.sh
+          fi
+
```
lucasvuotto, line 89: Is this step needed? I'd think that it isn't, as this job `needs` build which saves it to the cache.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+      - name: Run SonarQube analysis
+        run: |
+          sudo apt-get update && sudo apt-get install -yqq git
```
lucasvuotto, line 92: There is `git` already.

```suggestion
```

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+          extra_flags=""
+          if [ -n "$CIRCLE_PULL_REQUEST" ]; then
```
lucasvuotto, line 94: This particular environment variable isn't going to fly. The `github` context does have some utilities like the ones you used below, so it's better to check for those in here.

.github/workflows/build_and_test.yml
lucasvuotto: Weird. Sonar itself recommend using `sonar.token`. I'll take your word for it.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+            ./configure.sh
+          fi
+
+      - name: Unit tests
+        run: |
+          chmod +x ./gradlew
+          ./gradlew --no-daemon --stacktrace test
+      - name: Integration tests
+        run: |
+          chmod +x ./gradlew
```
lucasvuotto, line 156: There is really no need for doing this in 2 steps.

```suggestion
```

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+          ./gradlew --no-daemon --stacktrace integrationTest
+
+      - name: Save test results
+        run: |
+          mkdir -p ~/junit/
+          find rskj-core/build/test-results -type f -name "*.xml" -exec cp {} ~/junit/ \;
+        if: always()
```
lucasvuotto, line 163: I don't believe this accomplishes anything. `~/junit/` isn't consumed in anyway and GitHub doesn't offer a way to inspect iirc. Is it missing an step?

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Get Gradle jar
+        run: |
+          if [ ! -f ./gradle/wrapper/gradle-wrapper.jar ]; then
+            ./configure.sh
+          fi
+
```
lucasvuotto, line 149: Same comment as previous job.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+      - name: Setup System Tools
+        run: |
+          apt update -y
+          apt install -y curl
+      - name: Install Node.js
+        run: |
+          curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
+          apt-get install -y nodejs
+      - name: Check Node.js version
+        run: node --version
```
lucasvuotto, line 132: Same comment as in `build` job.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+
+      - name: Setup Java JDK
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+
+      - name: Verify Java installation
+        run: java -version
+
```
lucasvuotto, line 225: Unneeded.

```suggestion
```

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Get Gradle jar
+        run: |
+          if [ ! -f ./gradle/wrapper/gradle-wrapper.jar ]; then
+            ./configure.sh
+          fi
+
```
lucasvuotto, line 242: Same observation as in previous job.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+      - name: Install Mining Integration Tests dependencies
+        working-directory: mining-integration-tests
+        run: |
+          npm ci
+
+      - name: Change mining-integration-tests bitcoind url in config.json to localhost
+        working-directory: mining-integration-tests
+        run: |
+          sed -i 's/bitcoind01/localhost/g' config.json
```
lucasvuotto, line 210: I'm not a big fan of this kind of patterns. Can we add an `rskj-ci.json` to rsksmart/mining-integration-tests and here run a `cp -f rskj-ci.json config.json` instead?

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+        run: |
+          version=$(tr -d "'\"" < ../rskj-core/src/main/resources/version.properties \
+            | cut -d = -f 2- | paste -sd - -)
+          ls -la ../rskj-core/build/libs/rskj-core-"$version"-all.jar
+          java -Drsk.conf.file=./rsk-integration-test.conf -cp ../rskj-core/build/libs/rskj-core-"$version"-all.jar co.rsk.Start --regtest & rskpid=$!
+          until nc -z 127.0.0.1 4444
+            do
+              echo "Waiting for RskJ..."
+              sleep 1
+            done
```
lucasvuotto, line 254: The original block is over-indented. Besides that, this is "dangerous": if for any reason RskJ doesn't manage to be detected as up in port 4444, it'll loop forever. I suggest the following pattern instead:

```suggestion
          MAX_TRIES=30 tries=0
          while [ $tries -lt $MAX_TRIES ]; do
            sleep 1
            if nc -z 127.0.0.1 4444; then
              break
            fi
            tries=$(($tries + 1))
          done
          if [ $tries -eq $MAX_TRIES ]; then
            echo "Couldn't contact RskJ after $MAX_TRIES attempts." >&2
            exit 1
          fi
          done
```

lucasvuotto (commented):
There are some more general comments to address:

- There is a big mixture of versions:
  - `checkout` action is used at least as `v2` and `v4`. Please, use `v4` everywhere.
  - `build` job builds RskJ with Java 8. `mining-tests` execute with Java 17. Do we want that difference?
  - Might be interesting to take the chance and build and test for Java 8, 11 and 17.
- Is there any reason to split `build` and `tests` jobs? Other than copying over what was present in CircleCI, that is. In particular, does it make sense to run Sonarqube even in the case of failing tests?

.github/workflows/build_and_test.yml
jurajpiar: I'm sure I had that there at some point. ah well, I'll put it back.

.github/workflows/build_and_test.yml
jurajpiar: I mean, sure ...

.github/workflows/build_and_test.yml
jurajpiar: same answer

.github/workflows/build_and_test.yml
jurajpiar: same answer

.github/workflows/build_and_test.yml
jurajpiar: same answer

.github/workflows/build_and_test.yml
jurajpiar: Oh yea, I just put it there to discuss how we want to handle this (and forgot about), as there was such a mechanism in circleci.

.github/workflows/build_and_test.yml
jurajpiar: yea I guess not. not sure why I have it there. will remove

.github/workflows/build_and_test.yml
jurajpiar: ah man, how did that sneak back in. no, I know it isn't, mate (as you can see a couple of lines below). will change

.github/workflows/build_and_test.yml
jurajpiar: I know, but it isn't lol. Which is why I had to install it. But I'll test it again, just in case.

.github/workflows/build_and_test.yml
jurajpiar: yea, that was my assumption, but the jar wasn't always present, and so I decided to at least wrap it in a condition. 🤷

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,262 @@
...
+      - name: Setup System Tools
+        run: |
+          apt update -y
+          apt install -y gnupg2 curl
+      - name: Install Node.js
+        run: |
+          curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
+          apt-get install -y nodejs
+      - name: Check Node.js version
+        run: node --version
```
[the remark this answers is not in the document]
jurajpiar: Pretty sure it's there because the cache needed it. But I'll check again now that things work.

jurajpiar:
> There are some more general comments to address:
> 
> * There is a big mixture of versions:
>   
>   * `checkout` action is used at least as `v2` and `v4`. Please, use `v4` everywhere.
>   * `build` job builds RskJ with Java 8. `mining-tests` execute with Java 17. Do we want that difference?
>   * Might be interesting to take the chance and build and test for Java 8, 11 and 17.
> * Is there any reason to split `build` and `tests` jobs? Other than copying over what was present in CircleCI, that is. In particular, does it make sense to run Sonarqube even in the case of failing tests?

Thanks for the review. It seems taht it had annoyed you somewhat, but I might be reading too much into it. Well, in any case, sorry if it did :P
1. ah yea I tried using only v4, there were many versions of this as you can see from the history.
2. it's just that the mining tests (and I guess sonar too) would not work with v8, but we probably do want to run the unit and int tests with it since we clam to be supporting it
3. well, it would increase the dev time somewhat, I reckon. The tests are already taking ages.
4. tbh I didn't think too much about it, just wanted to migrate it as it was. but I think that's up for a discussion. I'd argue that sonarqube and mining tests should be done first as they take a fraction of time of that of the tests job. but I'd also do it via dependencies not budling diff jobs together. If a build fails as a dev you want to immediately know it is the build that failed, I'd say.

.github/workflows/build_and_test.yml
jurajpiar: Or perhaps you'd be ok with `jq`-ing it instead?

lucasvuotto: My worries with `sed` and `jq` is that:

- the final content isn't super clear
- it's error prone
- it's hard to maintain

By pushing the CI config into MIT repo, my worries are solved and the behaviour is more predictable.

Put into another way, treat both repositories as objects of different classes. Using `sed` to set private properties of the other object breaks encapsulation.

jurajpiar: I don't think this analogy works tbh because config files would be public methods, since they are meant to be used by the consumer; otherwise they could have been hardcoded. Which, in the case of the MIT, is the repo that uses it to test itself.
But I'm happy to create a PR in the MIT project for one config file that doesn't change anything for the repo itself. ;)

jurajpiar: also in MIT README file is:

> 3. Modify `config.json` to match the corresponding url, port and credentials for the bitcoind and rsk node instances

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
jurajpiar: not needed now. removed

.github/workflows/build_and_test.yml
jurajpiar: 🤯

.github/workflows/build_and_test.yml
jurajpiar: let me try again now that it all works

.github/workflows/build_and_test.yml
jurajpiar: removed

.github/workflows/build_and_test.yml
jurajpiar: as to the list of worries. I get that for `sed`, but with `jq` we can be very precise in what it's done to the json file. It's no different from you manually changing the config were you to run it locally.

.github/workflows/build_and_test.yml
jurajpiar: well must have been something else because this too now works :P

.github/workflows/build_and_test.yml
lucasvuotto: Ok, go with `jq`.

jurajpiar: Nevertheless, here you go: https://github.com/rsksmart/mining-integration-tests/pull/32. After you(?) merge this MIT PR, I can adapt the behaviour of this gha job.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,251 @@
...
+            rskj-core/build
+
+      - name: List the built artefacts
+        run: |
+          tree rskj-core/build -L 2  
+
+      - name: Run SonarQube analysis
+        run: |
+          extra_flags=""
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
```
bcodesido, line 102: why not using a github action condition instead of an if on bash?

The block should look like:

```
- name: If pull request
  if: github.event_name == 'pull_request'
  run: |
    git branch -f master origin/master
    pr_number==${{ github.event.pull_request.number }}
    extra_flags="-Dsonar.pullrequest.base=master
      -Dsonar.pullrequest.branch=${{ github.head_ref }}
       -Dsonar.pullrequest.key=$pr_number"

- name: if not pull request
  if: github.event_name != 'pull_request'
  run: extra_flags="-Dsonar.branch.name=master"

- name: Run SonarQube analysis
  run: |
              ./gradlew sonarqube --no-daemon -x build -x test \
              $extra_flags \
              -Dsonar.organization=rsksmart \
              -Dsonar.host.url="https://sonarcloud.io" \
              -Dsonar.token="${{ secrets.SONAR_TOKEN }}"
              
```

bcodesido (commented):
Looks good, only one small comment to make it simpler.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,251 @@
...
+        name: Restore gradle cache
+        with:
+          path: |
+            .gradle/caches
+            gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Unit tests
```
bcodesido, line 143: I don't know how much value is added by changing this, but I would split this step into two different ones, with one running the unit tests and the other running the integration tests only.

With the second dependent on the first. Besides that, at the end is the same (and lasts for the same time) visually is better than having an action running for 30 minutes.

jurajpiar: Fair enough. Although I'm leaning more towards having them run in parallel to get a faster result for either.

.github/workflows/build_and_test.yml
jurajpiar: done

.github/workflows/build_and_test.yml
jurajpiar: done

bcodesido (approved):
LGTM

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,287 @@
...
+
+      - name: Install Mining Integration Tests dependencies
+        working-directory: mining-integration-tests
+        run: |
+          npm ci
+
+      - name: Change mining-integration-tests bitcoind url in config.json to localhost
+        working-directory: mining-integration-tests
+        run: |
+          jq '.bitcoind.url = "localhost"' config.json > config.json.tmp && mv config.json.tmp config.json
```
lucasvuotto, line 227: This construction is error prone. I prefer the following, way more complex expression:

```suggestion
          jq '.bitcoind? | if has("url") then (.url = "localhost") else error(".bitcoind.url not found") end' config.json > config.json.tmp && mv config.json.tmp config.json
```

The reason is that your proposal will create the keys in the object if they don't exist. My proposal will blow up right here if the format of `config.json` ends up being modified, making it easier to detect the issue.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,287 @@
...
+            sleep 1
+          done
+          
+          if ! kill -0 $rskpid 2>/dev/null; then
+            echo "RskJ process has died"
+            exit 1
+          fi
+          
+          npm test
+          kill -9 $rskpid
```
lucasvuotto, line 287: The `export` aren't needed. Checking for the PID of RskJ doesn't make much sense, as it'll connectivity test will also fail in such case. If max attempts are reached, this will abort the run only if the RskJ died. Otherwise it'll try to run nevertheless. Despite being tests and a throwaway environment, `kill -9` is bad.

For all this reasons, I suggest the following instead:
```suggestion
          
          tries=0
          MAX_TRIES=10
          while [ $tries -lt $MAX_TRIES ]; do
            nc -z 127.0.0.1 && break
            echo "Waiting for RskJ..."
            tries=$((tries+1))
            sleep 1
          done
          
          if [ $tries -eq $MAX_TRIES ]; then
            echo "RskJ unreachable after $MAX_TRIES attempts. Aborting." >&2
            exit 1
          fi
          
          npm test
          kill $rskpid
```

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,287 @@
...
+          if [ ! -f gradle/wrapper/gradle-wrapper.jar ]; then
+            ./configure.sh
+            ./gradlew --no-daemon dependencies
+          fi
+
+      - name: Run tests
+        run: |
+          ./gradlew --no-daemon --stacktrace integrationTest
+
+  mining-tests:
```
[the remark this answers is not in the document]
jurajpiar: yea that's how it is done in circle, but I wanted to avoid that, since I didn't see a compelling reason to run an action in repo A to trigger an action in repo B that clones repo A at the branch of the PR.
If it is done that way to allow other repos to use it as well, then we should convert that repo to a github action and run it in their workflow like
```yaml
jobs:
  mit:
    steps:
      - name: mit
        uses: rsksmart/mining-integration-tests
```
shouldn't we?

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,287 @@
...
+            -Dsonar.token="${{ secrets.SONAR_TOKEN }}"
+
+      - name: If not
+        if: github.event_name != 'pull_request'
+        run: |
+          ./gradlew sonarqube --no-daemon -x build -x test \
+            -Dsonar.branch.name=master \
+            -Dsonar.organization=rsksmart \
+            -Dsonar.host.url="https://sonarcloud.io" \
+            -Dsonar.token="${{ secrets.SONAR_TOKEN }}"
```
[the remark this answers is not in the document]
jurajpiar: actually it is in the GHA UI where it makes sense as you always see the skipped steps.
but fair enough, I can change it to something more descriptive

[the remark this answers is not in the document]
jurajpiar: > Is there a particular reason for spliting it at the step level, instead of doing everything in a single step and using shell conditionals?

yes, https://github.com/rsksmart/rskj/pull/2290#discussion_r1626308422

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,287 @@
...
+      - name: Build
+        run: |
+          ./gradlew --no-daemon --stacktrace build -x test
+
+      - name: Archive build artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: build-files
+          path: |
+            rskj-core/build
```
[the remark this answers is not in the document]
jurajpiar: Splitting into smaller jobs and running in parallel helps developers to get response faster.

`rskj-core/build` is needed by both `smell-test` and `mining-test` jobs.
`unit-tests` and `integration-tests` do not use the build, but they do use the Gradle dependencies, that are also done here. I could split the `prepare-rskj` if you think that's preferable?

.github/workflows/build_and_test.yml
jurajpiar: the exports were needed for some reason. dunno why but it wouldn't recognise it otherwise. I mean, I can try again...

fmacleal: suggestion:

Add this logic to a script inside a folder on the `.github` directory. Then you could call this script here instead of having all this logic inside the .yml file.  All content inside the .github folder is accessible inside the github actions steps.

If we update the script, we don't need to update the .yml, would be an abstraction and easy to validate

fmacleal (approved):
LGTM, good job.

I just have a suggestion to migrate some logic to a script, don't know if the script will work fine though. But you could verify the output from the script and fail the step if necessary.

.github/workflows/build_and_test.yml
jurajpiar: done @lucasvuotto 

Unless it gets more votes I'll keep it in this file @fmacleal if that's alright?

.github/workflows/build_and_test.yml
jurajpiar: done

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,288 @@
...
+          fi
+
+      - name: Download build artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: build-files
+          path: |
+            rskj-core/build
+
+      - name: Sonarquebe for PRs
```
ed-iov, line 95: ``
```suggestion
      - name: Sonarqube for PRs
```

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,288 @@
...
+          extra_flags="-Dsonar.pullrequest.base=${{ github.base_ref }} \
+          -Dsonar.pullrequest.branch=${{ github.head_ref }} \
+          -Dsonar.pullrequest.key=$pr_number"
+          ./gradlew sonarqube --no-daemon -x build -x test \
+          $extra_flags \
+          -Dsonar.organization=rsksmart \
+          -Dsonar.host.url="https://sonarcloud.io" \
+          -Dsonar.token="${{ secrets.SONAR_TOKEN }}"
+
+      - name: Sonarquebe for master
```
ed-iov, line 108:
```suggestion
      - name: Sonarqube for master
```

.github/workflows/build_and_test.yml
jurajpiar: nice catch, thanks!

fmacleal (dismissed):

Vovchyk:
pipeline:run

Vovchyk:
pipeline:run

lucasvuotto:
pipeline: run

Vovchyk:
pipeline:run

Vovchyk (approved):

PR #2294, opened 2024-04-22
Modify ABICallElection

julia-zack:
There is an ongoing effort to organize Bridge classes in their package and avoid circular dependencies.
One of the classes we have identified that could be moved to their package are those related to ABI voting.
To capitalize this effort, there are some improvements that can be done in each class.

This PR modifies the ABICallElection:
- Make getWinner return Optional instead of null. 
- Add logs.
- Make validate method throw a new UnauthorizedVoterException.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/UnauthorizedVoterException.java
+++ b/rskj-core/src/main/java/co/rsk/peg/UnauthorizedVoterException.java
@@ -1,0 +1,7 @@
+package co.rsk.peg;
+
+public class UnauthorizedVoterException extends RuntimeException {
+
+    public UnauthorizedVoterException(String message) { super(message); }
+
+}
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -2302,8 +2302,9 @@ public class BridgeSupport {
         }
 
         // If enough votes have been reached, then actually execute the function
-        ABICallSpec winnerSpec = election.getWinner();
-        if (winnerSpec != null) {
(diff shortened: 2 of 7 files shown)


rskj-core/src/main/java/co/rsk/peg/UnauthorizedVoterException.java
```diff
@@ -0,0 +1,7 @@
+package co.rsk.peg;
+
+public class UnauthorizedVoterException extends RuntimeException {
```
marcos-iov, line 3: Maybe this should go in `abi` package?

rskj-core/src/main/java/co/rsk/peg/abi/ABICallElection.java
```diff
@@ -61,46 +66,54 @@ public void clear() {
      */
     public boolean vote(ABICallSpec callSpec, RskAddress voter) {
```
marcos-iov, line 67: We should probably log the voter also. Maybe at the beginning of the method

rskj-core/src/main/java/co/rsk/peg/UnauthorizedVoterException.java
julia-zack: i suggest we create an `/authorizer` package inside `/peg`, to have this class and AdressBasedAuthorizer. 
Wdyt?
This way we avoid the circular dependency between `/peg` and `/peg/abi` too

rskj-core/src/main/java/co/rsk/peg/abi/ABICallElection.java
julia-zack: Done

marcos-iov (approved):

PR #2295, opened 2024-04-22
Create /abi package.

julia-zack:
There is an ongoing effort to organize Bridge classes in their package and avoid circular dependencies.
One of the classes we have identified that could be moved to their package are those related to ABI voting (ABICallElection, ABICallSpec, ABICallVoteResult, AddressBasedAuthorizer).

This PR moves the ABI related classes to a new /vote package.


rskj-core/src/main/java/co/rsk/peg/abi/ABICallElection.java
```diff
@@ -16,9 +16,10 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-package co.rsk.peg;
+package co.rsk.peg.abi;
 
 import co.rsk.core.RskAddress;
+import co.rsk.peg.AddressBasedAuthorizer;
```
marcos-iov, line 22: AddressBasedAuthorizer will also be moved to abi package right? Is that part of a separate PR?

rskj-core/src/test/java/co/rsk/peg/abi/ABICallElectionTest.java
```diff
@@ -16,9 +16,12 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-package co.rsk.peg;
+package co.rsk.peg.abi;
 
 import co.rsk.core.RskAddress;
+import co.rsk.peg.AddressBasedAuthorizer;
+import co.rsk.peg.abi.ABICallElection;
+import co.rsk.peg.abi.ABICallSpec;
```
marcos-iov, line 24: Do you need these imports? It's in the same package as this class

Same in the other test classes

rskj-core/src/main/java/co/rsk/peg/abi/ABICallElection.java
julia-zack: I'm not sure about that, I think we should discuss it before. Either way I'd create a separate PR to keep them organized.

rskj-core/src/test/java/co/rsk/peg/abi/ABICallElectionTest.java
julia-zack: oh, I did a rebase and I thought I had removed them! Thanks

julia-zack: Done

marcos-iov (approved):

PR #2303, opened 2024-04-26
Make BridgeStorageAccessor an interface

julia-zack:
Make BridgeStorageAccessor an interface and create BridgeStorageAccessorImpl to implement it. Apply same logic for FeePerKbSupport and FeePerKbStorageProvider.
Move FeePerKbStorageProvider to /feeperkb package and make some methods protected instead of public.


rskj-core/src/main/java/co/rsk/peg/storage/BridgeStorageAccessor.java
```diff
@@ -1,55 +1,24 @@
...
 import org.ethereum.vm.PrecompiledContracts;
 
 import java.io.IOException;
 
-public class BridgeStorageAccessor {
+public interface BridgeStorageAccessor {
 
-    private final Repository repository;
-    private static final RskAddress contractAddress = PrecompiledContracts.BRIDGE_ADDR;
+    RskAddress contractAddress = PrecompiledContracts.BRIDGE_ADDR;
```
marcos-iov, line 11: I wouldn't put variables in an interface. It's a contract that exposes certain methods that allow access to the storage.

What the contract address is, I think that is related to the concrete implementation

julia-zack: good point. Maybe we should rename it to `StorageAccessor`, since it won't have references to the bridge

rskj-core/src/main/java/co/rsk/peg/storage/StorageAccessor.java
```diff
@@ -0,0 +1,25 @@
...
+
+import co.rsk.core.RskAddress;
+import org.ethereum.vm.DataWord;
+import org.ethereum.vm.PrecompiledContracts;
+
+import java.io.IOException;
+
+public interface StorageAccessor {
+
+    RskAddress contractAddress = PrecompiledContracts.BRIDGE_ADDR;
```
marcos-iov, line 11: Remember to remove this

rskj-core/src/main/java/co/rsk/peg/storage/StorageAccessor.java
```diff
@@ -8,8 +8,6 @@
 
 public interface StorageAccessor {
 
-    RskAddress contractAddress = PrecompiledContracts.BRIDGE_ADDR;
```
marcos-iov: Some of the imports might be unused now

marcos-iov (approved):

PR #2310, opened 2024-05-03
Configurable Fiat-Based Stable MinGasPrice

asoto-iov:
Enhancing Miner Configuration for Stable Transaction Costs

## Description
Develop and integrate a feature that allows RSK miners to configure the minimum gas price (MinGasPrice) in fiat currency instead of the current cryptocurrency (WEI) linked to Bitcoin. This feature aims to stabilize the transaction cost in fiat terms, improving predictability and user experience.

## Motivation and Context

Currently, RSK miners can set the minimum gas price (MinGasPrice) using the miner.minGasPrice configuration, which is denominated in WEI and inherently linked to the Bitcoin price. This linkage causes transaction costs to fluctuate with Bitcoin's value, potentially leading to increased costs when Bitcoin appreciates, thereby adversely affecting user experience. 
To address this issue, the proposed feature aims to empower miners with the ability to specify and configure the minimum gas price in fiat currency. This change would stabilize transaction costs in fiat terms, regardless of cryptocurrency market variations, and is intended to enhance predictability and user satisfaction. Currently, the system supports only one configuration parameter for setting the MinGasPrice.

The goal is to give miners a capability to specify / configure min gas price value denominated in a fiat currency.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/config/mining/HttpGetStableMinGasSystemConfig.java
```diff
@@ -0,0 +1,39 @@
...
+    private final String jsonPath;
+    private final int timeout;
+    private final String apiKey;
+
+    public HttpGetStableMinGasSystemConfig(Config config) {
+        this.url = config.getString(URL_PROPERTY);
+        this.jsonPath = config.getString(JSON_PATH_PROPERTY);
+        this.timeout = config.getInt(TIMEOUT_PROPERTY);
+        this.apiKey = config.getString(API_KEY_PROPERTY);
+    }
```
fmacleal, line 22: If you agree with the suggestion I have added on the Jira ticket, I would say that this object here could receive the map with the parameters and then initialise the properties according it. If it's missing some property, an exception could be thrown.

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -406,7 +409,9 @@ private String getGeneratedNodePrivateKey() {
                 file.getParentFile().mkdirs();
                 try (FileWriter writer = new FileWriter(file)) {
                     props.store(writer, "Generated NodeID. To use your own nodeId please refer to 'peer.privateKey' config option.");
-                    logger.info("New nodeID generated: {}", props.getProperty("nodeId"));
+                    if(logger.isInfoEnabled()) {
+                        logger.info("New nodeID generated: {}", props.getProperty("nodeId"));
+                    }
                     logger.info("Generated nodeID and its private key stored in {}", file);
```
fmacleal, line 415: Isn't this info supposed to be added to the ìf` as well, since you refactored to add the verification of this before log?

rskj-core/src/main/java/co/rsk/mine/gas/MinGasPriceProviderType.java
```diff
@@ -0,0 +1,7 @@
+package co.rsk.mine.gas;
+
+public enum MinGasPriceProviderType {
+    DEFAULT,
+    HTTP_GET,
+    ETH_CALL
```
fmacleal, line 6: I have a suggestion to use the enum to our favor and mapping directly according the method configured. We could even set to always return `DEFAULT` in case is anything different from the other two types.

My suggestion is the following:
```suggestion
    DEFAULT("DEFAULT"),
    HTTP_GET("HTTP_GET"),
    ETH_CALL("ETH_CALL");

    private String type;

    MinGasPriceProviderType(final String type){
        this.type = type;
    }

    public static MinGasPriceProviderType fromMethodType(String type) {
        return Arrays.stream(MinGasPriceProviderType.values())
                .filter( mgp -> mgp.type.equals(type))
                .findFirst()
                .orElse(MinGasPriceProviderType.DEFAULT);
    }
```

You can rename anything if you approve the suggestion and apply it.

I have created a test for it and it works perfectly:

```
class MinGasPriceProviderTypeTest {

    @Test
    void fromMethodTypeMapping() {
        //given
        String httpGet = "HTTP_GET";
        String ethCall = "ETH_CALL";
        String defaultType = "DEFAULT";
        String anything = "anything";
        String nullType = null;
        //when
        MinGasPriceProviderType httpGetType = MinGasPriceProviderType.fromMethodType(httpGet);
        MinGasPriceProviderType ethCallType = MinGasPriceProviderType.fromMethodType(ethCall);
        MinGasPriceProviderType defaultTypeType = MinGasPriceProviderType.fromMethodType(defaultType);
        MinGasPriceProviderType anythingType = MinGasPriceProviderType.fromMethodType(anything);
        MinGasPriceProviderType nullTypeType = MinGasPriceProviderType.fromMethodType(nullType);
        //then
        assertEquals(MinGasPriceProviderType.HTTP_GET, httpGetType);
        assertEquals(MinGasPriceProviderType.ETH_CALL, ethCallType);
        assertEquals(MinGasPriceProviderType.DEFAULT, defaultTypeType);
        assertEquals(MinGasPriceProviderType.DEFAULT, anythingType);
        assertEquals(MinGasPriceProviderType.DEFAULT, nullTypeType);
    }
}
```

rskj-core/src/main/java/co/rsk/mine/gas/MinGasPriceProvider.java
fmacleal, line 1: Very good interface, I liked, it comes perfectly for the use of factory design pattern.

rskj-core/src/main/java/co/rsk/mine/gas/StableMinGasPriceProvider.java
```diff
@@ -0,0 +1,32 @@
+package co.rsk.mine.gas;
+
+import co.rsk.core.Coin;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class StableMinGasPriceProvider implements MinGasPriceProvider {
+    private static final Logger logger = LoggerFactory.getLogger("StableMinGasPrice");
+    private final DefaultMinGasPriceProvider fallBackProvider;
+
```
fmacleal, line 10:
```suggestion
```

rskj-core/src/main/java/co/rsk/mine/gas/StableMinGasPriceProvider.java
fmacleal, line 1: Nice abstraction. Good job. 🙂

rskj-core/src/main/java/co/rsk/mine/gas/StableMinGasPriceProviderFactory.java
```diff
@@ -0,0 +1,43 @@
...
+import org.slf4j.LoggerFactory;
+
+
+public class StableMinGasPriceProviderFactory {
+
+    private static final Logger logger = LoggerFactory.getLogger("StableMinGasPrice");
+
+    private StableMinGasPriceProviderFactory() {}
+
+    public static MinGasPriceProvider create(StableMinGasPriceSystemConfig config, DefaultMinGasPriceProvider fallBackProvider) {
```
fmacleal, line 15: Do we really need to receive this default `DefaultMinGasPriceProvider`? 
If we are already receiving the configurations, we can easily initialise it with the `minGasPrice` from the config no?

rskj-core/src/main/java/co/rsk/mine/gas/StableMinGasPriceProviderFactory.java
```diff
@@ -0,0 +1,43 @@
...
+        MinGasPriceProviderType method = config.getMethod();
+        if (method == null) {
+            logger.error("Could not finde valid mehtod in config, using fallback provider: {}", fallBackProvider.getType().name());
```
fmacleal, line 18: Minor typos
```suggestion
            logger.error("Could not find a valid method on the config, using fallback provider: {}", fallBackProvider.getType().name());
```

rskj-core/src/main/java/co/rsk/mine/gas/StableMinGasPriceProviderFactory.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class StableMinGasPriceProviderFactory {
+
+    private static final Logger logger = LoggerFactory.getLogger("StableMinGasPrice");
+
+    private StableMinGasPriceProviderFactory() {}
+
+    public static MinGasPriceProvider create(StableMinGasPriceSystemConfig config, DefaultMinGasPriceProvider fallBackProvider) {
+        MinGasPriceProviderType method = config.getMethod();
+        if (method == null) {
```
fmacleal, line 17: Here it's a point that could be improved if we use the enum as I suggested, it would return the default always when the method isn't of the type `HTTP_GET` and `ETH_CALL`.

rskj-core/src/main/java/co/rsk/mine/gas/StableMinGasPriceProviderFactory.java
```diff
@@ -0,0 +1,43 @@
...
+                return createHttpGetStableGasPriceProvider(config, fallBackProvider);
+            case ETH_CALL:
+                throw new UnsupportedOperationException("ETH_CALL method is not supported yet");
+            default:
+                logger.debug("Could not find a valid implementation for the method {}. Returning fallback provider {}", method, fallBackProvider.getType().name());
+                return fallBackProvider;
+        }
+    }
+
+    private static HttpGetStableMinGasPriceProvider createHttpGetStableGasPriceProvider(StableMinGasPriceSystemConfig config, DefaultMinGasPriceProvider fallBackProvider) {
```
fmacleal, line 32: If we have a default config with a map, we could extract the key, value parameters from the map here and initialise according the need. Here we could also throw an exception if some parameter is missing from the config.

rskj-core/src/main/java/co/rsk/mine/gas/StableMinGasPriceProviderFactory.java
```diff
@@ -0,0 +1,43 @@
+package co.rsk.mine.gas;
+
+import co.rsk.config.mining.HttpGetStableMinGasSystemConfig;
+import co.rsk.config.mining.StableMinGasPriceSystemConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class StableMinGasPriceProviderFactory {
```
fmacleal, line 9: I have a few comments on the factory, but I really liked how you approach the problem. This factory would be exactly one of the things that crossed my mind to apply when I saw the ticket. Good job!

rskj-core/src/main/java/co/rsk/config/mining/StableMinGasPriceSystemConfig.java
```diff
@@ -0,0 +1,50 @@
...
+    private final Integer minStableGasPrice;
+    private final boolean enabled;
+    private final MinGasPriceProviderType method;
+    private final Config config;
+
+    public StableMinGasPriceSystemConfig(Config config) {
+        enabled = config.getBoolean(ENABLED_PROPERTY);
+        refreshRate = config.getInt(REFRESH_RATE_PROPERTY);
+        minStableGasPrice = config.getInt(MIN_STABLE_GAS_PRICE_PROPERTY);
+        method = config.getEnum(MinGasPriceProviderType.class, METHOD_PROPERTY);
```
fmacleal, line 23: If you agree with the change on structure of the config file, you would have to change a bit this logic and read the source object, then convert it to an object with the method and the key, value map. Something like this:

```
public class StableSourceGasPrice {

    private final MinGasPriceProviderType method;
    private final Map<String,String> params;
...
}
```

Then this method would be filled using the enum function converion, like this:

```
        method = MinGasPriceProviderType.fromMethodType(config.getString(METHOD_PROPERTY));
```

But I am not sure if it's possible to read the config like the object and extract the map. I am assuming it is, let me know your thoughts. 🙂

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -1851,6 +1857,17 @@ private BlockToMineBuilder getBlockToMineBuilder() {
         return blockToMineBuilder;
     }
 
+    private MinGasPriceProvider getMinGasPriceProvider() {
+        if (minGasPriceProvider == null) {
+            long minGasPrice = getRskSystemProperties().minerMinGasPrice();
+            DefaultMinGasPriceProvider defaultMinGasPriceProvider = new DefaultMinGasPriceProvider(minGasPrice);
```
fmacleal, line 1863: As I said, I am not sure if we need create this object here, I thought we had access to the config during the provider creation using the factory. Reflect about this, and let me know you opinion. 👍

fmacleal (commented):
I really liked the approach and the way you tackled the problem. Good job.

Generally, I don't have big changes to recommend. But some small things that IMO could make it even more flexible than the approach you have followed. Let me know your opinion, if you want to discuss it on slack, ping me. 😄

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
asoto-iov: Yep, I did it automatically to fix a sonar issue and didn't realize about it.

jurajpiar (changes requested):
This seems unnecessarily complicated. Why should a strategy know it's type? Imo there is no need for a factory and all those different abstractions. Also there is no need to change the MinimumGasPriceCalculator. I'd simplify it to the following, given that all the information needed is known on boot-up:

```mermaid
classDiagram
direction RL

class MinimumGasPriceCalculator {
    Unchanged - only to illustrate use
    +construcor(long)
}
class BlockToMineBuilder {
    Unchanged - only to illustrate use
    +constuctor(..., MinimumGasPriceCalculator)
}
RskContext --* BlockToMineBuilder
BlockToMineBuilder --o MinimumGasPriceCalculator
RskContext --* MinimumGasPriceCalculator

class RskSystemProperties {
    +getMinerMinGasPriceConfig(): MinGasPriceConfig
}

class RskContext {
    -getMiningConfig(): MiningConfig
    +getBlockToMineBuilder(): BlockToMineBuilder
}

class MinGasPriceConfig {
    +providerType: "onchain" | "http" | "default"
    +url: URL
    +minGasPriceTarget: long
    +otherStuffIfAny
    +createMinGasPriceProvider(MinGasPriceConfig): MinGasPriceProvider
}

class MiningConfig {
    -minGasPriceProvider: MinGasPriceProvider
    +constructor(MinGasPriceConfig)
    +getMinGasPriceTarget(): long
}

class MinGasPriceProvider {
    <<<Interface>>>
    +getMinGasPrice()
}

class DefaultMinGasPriceProvider {
    +constructor()
}

class OffChainMinGasPriceProvider {
    +constructor(OffChainMinGasPriceProviderConfig)
}

class OnChainMinGasPriceProvider {
    +constructor(OnChainMinGasPriceProviderConfig)
}


MinGasPriceConfig *-- RskSystemProperties
RskSystemProperties *-- RskContext

MinGasPriceProvider *-- MinGasPriceConfig
MinGasPriceConfig <-- MiningConfig
MinGasPriceProvider o-- MiningConfig
MiningConfig *-- RskContext


MinGasPriceProvider <|-- DefaultMinGasPriceProvider
MinGasPriceProvider <|-- OffChainMinGasPriceProvider
MinGasPriceProvider <|-- OnChainMinGasPriceProvider
```

rskj-core/src/main/java/co/rsk/mine/gas/MinGasPriceProviderType.java
asoto-iov: I like your suggestion to work with enums but for this case it is preferred to not return default automatically. As it is now if an user adds a non valid method or wrong naming an exception will be thrown by the config class automatically and the user will be forced to fix it accordingly.

rskj-core/src/main/java/co/rsk/mine/gas/StableMinGasPriceProviderFactory.java
asoto-iov: This should never happen as an error is thrown by the config class but I wanted to cover this edge case if the feature is modified in the future.

rskj-core/src/main/java/co/rsk/config/mining/HttpGetStableMinGasSystemConfig.java
asoto-iov: As mentioned I do not see the advantage of having a map instead of this.

rskj-core/src/main/java/co/rsk/mine/gas/MinGasPriceProviderType.java
fmacleal: Then, just throw an exception on the method instead return the DEFAULT. The enum implementation is unrelated with the default.

rskj-core/src/main/java/co/rsk/mine/gas/provider/MinGasPriceProviderFactory.java
```diff
@@ -0,0 +1,44 @@
...
+    private MinGasPriceProviderFactory() {
+    }
+
+    public static MinGasPriceProvider create(long fixedMinGasPrice, StableMinGasPriceSystemConfig stableMinGasPriceSystemConfig) {
+        FixedMinGasPriceProvider fixedMinGasPriceProvider = new FixedMinGasPriceProvider(fixedMinGasPrice);
+        if (stableMinGasPriceSystemConfig == null) {
+            logger.warn("Could not find stable min gas price system config, using {} provider", fixedMinGasPriceProvider.getType().name());
+            return fixedMinGasPriceProvider;
+        }
+        if (!stableMinGasPriceSystemConfig.isEnabled()) {
```
fmacleal, line 23: thought: maybe would be good to have logged here as well that the feature isn't enabled.

rskj-core/src/main/java/co/rsk/mine/gas/provider/MinGasPriceProviderFactory.java
```diff
@@ -0,0 +1,44 @@
...
+            logger.error("Could not find valid method in config, using fallback provider: {}", fixedMinGasPriceProvider.getType().name());
+            return fixedMinGasPriceProvider;
+        }
+
+
+        switch (method) {
+            case WEB:
+                return WebStableMinGasPriceProviderFactory.create(stableMinGasPriceSystemConfig, fixedMinGasPriceProvider);
+            case ON_CHAIN:
+                return OnChainMinGasPriceProviderFactory.create(stableMinGasPriceSystemConfig, fixedMinGasPriceProvider);
```
fmacleal, line 37: suggestion: as commented previously, I would create the `MinGasPriceProvider` implementation here directly according the method. You already have all the configurations here, you could have methods to get the config according each type passed to extract the configuration complexity creation. 

The concept of a factory is exactly hold these logics necessary to return different implementations, if start to become too complex, we can abstract more and create other classes. But for now, the first level of abstraction is more than enough.

fmacleal (commented):
Good job. I just added some comments with suggestions and thoughts. Let me know if you have some doubt, you can reply here or reach me on slack.

rskj-core/src/main/java/co/rsk/mine/gas/provider/onchain/OnChainMinGasPriceProvider.java
```diff
@@ -0,0 +1,29 @@
+package co.rsk.mine.gas.provider.onchain;
+
+import co.rsk.config.mining.OnChainMinGasPriceSystemConfig;
+import co.rsk.mine.gas.provider.MinGasPriceProvider;
+import co.rsk.mine.gas.provider.MinGasPriceProviderType;
+import co.rsk.mine.gas.provider.StableMinGasPriceProvider;
+
+public class OnChainMinGasPriceProvider extends StableMinGasPriceProvider {
```
jurajpiar, line 8: We should be careful with mixing problem domains. I suspect that the RBTC/USD exchange rate != min gas price!
And so all these classes should to be renamed. Furthermore, I don't see anywhere where we make this conversion from XR to min gas price.
What am I missing?

rskj-core/src/main/java/co/rsk/mine/gas/provider/MinGasPriceProviderType.java
```diff
@@ -0,0 +1,7 @@
+package co.rsk.mine.gas.provider;
+
+public enum MinGasPriceProviderType {
+    FIXED,
+    WEB,
+    ON_CHAIN
```
jurajpiar, line 6: I know I suggested this, but that was when I thought that the on chain call would be done directly in the vm rather than through eth module. And so, we should call it ETH_CALL.

Vovchyk (dismissed):

Vovchyk (approved):

PR #2311, opened 2024-05-06
Rollback response codes change

julia-zack:


rskj-core/src/main/java/co/rsk/peg/feeperkb/FeePerKbResponseCode.java
```diff
@@ -4,8 +4,8 @@ public enum FeePerKbResponseCode {
     SUCCESSFUL(1),
     UNSUCCESSFUL(-1),
     EXCESSIVE(-2),
-    NEGATIVE(-3),
-    UNAUTHORIZED(-4),
+    NEGATIVE(-1),
+    UNAUTHORIZED(-10),
     GENERIC(-10);
```
marcos-iov, line 9:
```suggestion
    GENERIC_ERROR(-10);
```
I think that sounds better, because these are response codes not error codes. So just to be clear that is a generic error not a generic response

PR #2319, opened 2024-05-10
Use original values for genesisFederationCreationTime

marcos-iov:
Update Bridge constants to use a ZonedDateTime object to indicate the `genesisFederationCreationTime` value instead of an Instant object.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/constants/BridgeDevNetConstants.java
+++ b/rskj-core/src/main/java/co/rsk/peg/constants/BridgeDevNetConstants.java
@@ -49,9 +49,8 @@ public class BridgeDevNetConstants extends BridgeConstants {
     public BridgeDevNetConstants(List<BtcECKey> federationPublicKeys) {
         btcParamsString = NetworkParameters.ID_TESTNET;
 
-        this.genesisFederationPublicKeys = federationPublicKeys;
-
-        genesisFederationCreationTime = ZonedDateTime.parse("2017-11-14T00:00:00Z").toInstant();
+        genesisFederationPublicKeys = federationPublicKeys;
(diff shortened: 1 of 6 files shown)


wilmerrootstock:
Could you please add a description to the PR?

julia-zack (approved):
LGTM

rskj-core/src/main/java/co/rsk/peg/constants/BridgeDevNetConstants.java
```diff
@@ -49,9 +49,8 @@ public class BridgeDevNetConstants extends BridgeConstants {
     public BridgeDevNetConstants(List<BtcECKey> federationPublicKeys) {
         btcParamsString = NetworkParameters.ID_TESTNET;
 
-        this.genesisFederationPublicKeys = federationPublicKeys;
-
-        genesisFederationCreationTime = ZonedDateTime.parse("2017-11-14T00:00:00Z").toInstant();
+        genesisFederationPublicKeys = federationPublicKeys;
+        genesisFederationCreationTime = ZonedDateTime.parse("1970-01-18T11:36:57.600Z").toInstant();
```
jeremy-then, line 53: Should we now leave a comment saying why we are using this date? Probably a bad idea?

julia-zack: i think is crear enough, we are just defining what is the genesis fed creation time

wilmerrootstock (approved):
LGTM!

jeremy-then (approved):
LGTM.

rskj-core/src/main/java/co/rsk/peg/constants/BridgeDevNetConstants.java
marcos-iov: That might add some confusion I think

PR #2323, opened 2024-05-14
Create pegout_transaction_created event

nathanieliov:
We are creating a new event named pegout_transaction_created, where the peg-out transaction hash and the list of outpoint values will be stored. This will enable the PowPeg to retrieve these values from the rsk transaction receipt where the peg-out was created, allowing it to provide them to the PowHsm.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeEvents.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeEvents.java
@@ -99,6 +99,12 @@ public enum BridgeEvents {
                     new CallTransaction.Param(true, Fields.BTC_TX_HASH, SolidityType.getType(SolidityType.BYTES32)),
                     new CallTransaction.Param(false, "pegoutCreationRskBlockNumber", SolidityType.getType(SolidityType.UINT256))
             }
+    ),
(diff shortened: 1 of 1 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeEvents.java
```diff
@@ -99,6 +99,12 @@ public enum BridgeEvents {
...
             }
+    ),
+    PEGOUT_TRANSACTION_CREATED("pegout_transaction_created",
+        new CallTransaction.Param[]{
+            new CallTransaction.Param(true, Fields.BTC_TX_HASH, SolidityType.getType(SolidityType.BYTES32)),
+            new CallTransaction.Param(false, Fields.UTXO_OUTPOINT_VALUES, SolidityType.getType(SolidityType.BYTES))
+        }
     );
 
     private String eventName;
```
marcos-iov, line 110: A couple of ideas, to do in another task.

- Rename this enum to `BridgeEvent`. Singular, as enums should be
- Move this `Fields` class to an enum. Guess the same fields could be used for bridge events and bridge methods, right?
- Make the attributes of the class `final`

nathanieliov: Task created.

PR #2328, opened 2024-05-15
Create utils methods to decode and encode outpoint values list

nathanieliov:
Create a UtxoUtils class and two new methods, one for encoding a list of outpoint values and another for decoding byte array outpoint values.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/bitcoin/InvalidOutpointValueException.java
+++ b/rskj-core/src/main/java/co/rsk/peg/bitcoin/InvalidOutpointValueException.java
@@ -1,0 +1,12 @@
+package co.rsk.peg.bitcoin;
+
+public class InvalidOutpointValueException extends RuntimeException{
+
+    public InvalidOutpointValueException(String message) {
+        super(message);
+    }
+
+    public InvalidOutpointValueException(String message, Throwable cause) {
+        super(message, cause);
+    }
+}
--- a/rskj-core/src/main/java/co/rsk/peg/bitcoin/InvalidOutpointValueException.java
+++ b/rskj-core/src/main/java/co/rsk/peg/bitcoin/InvalidOutpointValueException.java
@@ -1,6 +1,6 @@
 package co.rsk.peg.bitcoin;
 
-public class InvalidOutpointValueException extends RuntimeException{
+public class InvalidOutpointValueException extends RuntimeException {
 
     public InvalidOutpointValueException(String message) {
         super(message);
--- a/rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
+++ b/rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
@@ -6,25 +6,55 @@ import com.google.common.primitives.Bytes;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import org.spongycastle.util.encoders.Hex;
 
 public class UtxoUtils {
     private UtxoUtils() {}
 
+    /**
+     * Decode a {@code byte[]} of encoded outpoint values.
+     *
+     * @param encodedOutpointValues
+     * @return {@code List<Coin>} the list of outpoint values decoded in the same ordered they were before encoding.
+     * Or an {@code Collections.EMPTY_LIST} when {@code encodedOutpointValues} is {@code null} or {@code empty byte[]}.
+     */
     public static List<Coin> decodeOutpointValues(byte[] encodedOutpointValues){
         if (encodedOutpointValues == null || encodedOutpointValues.length == 0) {
-            return Collections.EMPTY_LIST;
+            return Collections.emptyList();
         }
-
         int offset = 0;
         List<Coin> outpointValues = new ArrayList<>();
+
         while (encodedOutpointValues.length > offset){
-            VarInt entryEncoded = new VarInt(encodedOutpointValues, offset);
-            offset += entryEncoded.getSizeInBytes();
-            outpointValues.add(Coin.valueOf(entryEncoded.value));
+            VarInt valueAsVarInt;
+            try {
+                valueAsVarInt = new VarInt(encodedOutpointValues, offset);
+            } catch (Exception ex){
+                throw new InvalidOutpointValueException(
+                    String.format("Invalid value with invalid VarInt format: %s",
+                        Hex.toHexString(encodedOutpointValues).toUpperCase()
+                    ),
+                    ex
+                );
+            }
+
+            offset += valueAsVarInt.getSizeInBytes();
+            Coin outpointValue = Coin.valueOf(valueAsVarInt.value);
+            assertValidOutpointValue(outpointValue);
+
+            outpointValues.add(outpointValue);
         }
         return outpointValues;
+
     }
 
+    /**
+     * Encode a {@code List<Coin} of outpoint values.
+     *
+     * @param outpointValues
+     * @return {@code byte[]} the list of outpoint values encoded preserving the order of the entries.
+     * Or an {@code empty byte[]} when {@code outpointValues} is {@code null} or {@code empty}.
+     */
     public static byte[] encodeOutpointValues(List<Coin> outpointValues) {
         if (outpointValues == null || outpointValues.isEmpty()) {
             return new byte[]{};
@@ -32,9 +62,17 @@ public class UtxoUtils {
 
         List<byte[]> encodedOutpointValues = new ArrayList<>();
         for (Coin outpointValue : outpointValues) {
+            assertValidOutpointValue(outpointValue);
             VarInt varIntOutpointValue = new VarInt(outpointValue.getValue());
             encodedOutpointValues.add(varIntOutpointValue.encode());
         }
         return Bytes.concat(encodedOutpointValues.toArray(new byte[][]{{}}));
+    }
+
+    private static void assertValidOutpointValue(Coin outpointValue) {
+        if (outpointValue == null || outpointValue.isNegative()){
+            throw new InvalidOutpointValueException(String.format("Invalid outpoint value: %s. Negative and null values are not allowed.",
+                outpointValue));
+        }
     }
 }
--- a/rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
+++ b/rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
@@ -19,8 +19,8 @@ public final class UtxoUtils {
      * Decode a {@code byte[]} of encoded outpoint values.
      *
      * @param encodedOutpointValues
-     * @return {@code List<Coin>} the list of outpoint values decoded in the same ordered they were
-     * before encoding. Or an {@code Collections.EMPTY_LIST} when {@code encodedOutpointValues} is
+     * @return {@code List<Coin>} the list of outpoint values decoded preserving
+     * the order of the entries. Or an {@code Collections.EMPTY_LIST} when {@code encodedOutpointValues} is
      * {@code null} or {@code empty byte[]}.
      */
     public static List<Coin> decodeOutpointValues(byte[] encodedOutpointValues) {
@@ -45,7 +45,7 @@ public final class UtxoUtils {
 
             offset += valueAsVarInt.getSizeInBytes();
             Coin outpointValue = Coin.valueOf(valueAsVarInt.value);
-            assertValidOutpointValue(outpointValue);
+            validateOutpointValue(outpointValue);
 
             outpointValues.add(outpointValue);
         }
@@ -68,7 +68,7 @@ public final class UtxoUtils {
 
         ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
         for (Coin outpointValue : outpointValues) {
-            assertValidOutpointValue(outpointValue);
+            validateOutpointValue(outpointValue);
             VarInt varIntOutpointValue = new VarInt(outpointValue.getValue());
             try {
                 outputStream.write(varIntOutpointValue.encode());
@@ -84,7 +84,7 @@ public final class UtxoUtils {
         return outputStream.toByteArray();
     }
 
-    private static void assertValidOutpointValue(Coin outpointValue) {
+    private static void validateOutpointValue(Coin outpointValue) {
         if (outpointValue == null || outpointValue.isNegative()) {
             throw new InvalidOutpointValueException(String.format(
                 "Invalid outpoint value: %s. Negative and null values are not allowed.",
--- a/rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
+++ b/rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
@@ -1,0 +1,40 @@
+package co.rsk.peg.bitcoin;
+
+import co.rsk.bitcoinj.core.Coin;
+import co.rsk.bitcoinj.core.VarInt;
+import com.google.common.primitives.Bytes;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class UtxoUtils {
+    private UtxoUtils() {}
+
+    public static List<Coin> decodeOutpointValues(byte[] encodedOutpointValues){
+        if (encodedOutpointValues == null || encodedOutpointValues.length == 0) {
+            return Collections.EMPTY_LIST;
+        }
+
+        int offset = 0;
+        List<Coin> outpointValues = new ArrayList<>();
+        while (encodedOutpointValues.length > offset){
+            VarInt entryEncoded = new VarInt(encodedOutpointValues, offset);
+            offset += entryEncoded.getSizeInBytes();
+            outpointValues.add(Coin.valueOf(entryEncoded.value));
+        }
+        return outpointValues;
+    }
+
+    public static byte[] encodeOutpointValues(List<Coin> outpointValues) {
+        if (outpointValues == null || outpointValues.isEmpty()) {
+            return new byte[]{};
+        }
+
+        List<byte[]> encodedOutpointValues = new ArrayList<>();
+        for (Coin outpointValue : outpointValues) {
+            VarInt varIntOutpointValue = new VarInt(outpointValue.getValue());
+            encodedOutpointValues.add(varIntOutpointValue.encode());
+        }
+        return Bytes.concat(encodedOutpointValues.toArray(new byte[][]{{}}));
+    }
+}
--- a/rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
+++ b/rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
@@ -2,13 +2,15 @@ package co.rsk.peg.bitcoin;
 
 import co.rsk.bitcoinj.core.Coin;
 import co.rsk.bitcoinj.core.VarInt;
-import com.google.common.primitives.Bytes;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import org.spongycastle.util.encoders.Hex;
 
-public class UtxoUtils {
+public final class UtxoUtils {
 
     private UtxoUtils() {
     }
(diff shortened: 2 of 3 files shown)


rskj-core/src/main/java/co/rsk/peg/bitcoin/InvalidOutpointValueException.java
```diff
@@ -0,0 +1,12 @@
+package co.rsk.peg.bitcoin;
+
+public class InvalidOutpointValueException extends RuntimeException{
```
apancorb, line 3: nit: add the extra space before the bracket
```suggestion
public class InvalidOutpointValueException extends RuntimeException {
```

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,78 @@
...
+    private UtxoUtils() {}
+
+    /**
+     * Decode a {@code byte[]} of encoded outpoint values.
+     *
+     * @param encodedOutpointValues
+     * @return {@code List<Coin>} the list of outpoint values decoded in the same ordered they were before encoding.
+     * Or an {@code Collections.EMPTY_LIST} when {@code encodedOutpointValues} is {@code null} or {@code empty byte[]}.
+     */
+    public static List<Coin> decodeOutpointValues(byte[] encodedOutpointValues){
```
apancorb, line 21: Something interesting could be to also have either a public or private `decodeOutpointValue` method that gets called from this method. This can lead to this method being shorter. And if it is public maybe it can be used in other places in the future where only one outpoint value is needed. wdyt?

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,78 @@
...
+        if (encodedOutpointValues == null || encodedOutpointValues.length == 0) {
+            return Collections.emptyList();
```
apancorb, line 23: I really like how the method returns an empty list. Makes it very clean and easy to use :)

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+        int offset = 0;
+        List<Coin> outpointValues = new ArrayList<>();
+
+        while (encodedOutpointValues.length > offset){
```
apancorb, line 28: There is a bunch of places where you are not adding a space before the bracket. If possible, make sure we are formatting this right.

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,78 @@
...
+            return new byte[]{};
+        }
+
+        List<byte[]> encodedOutpointValues = new ArrayList<>();
+        for (Coin outpointValue : outpointValues) {
+            assertValidOutpointValue(outpointValue);
+            VarInt varIntOutpointValue = new VarInt(outpointValue.getValue());
+            encodedOutpointValues.add(varIntOutpointValue.encode());
+        }
+        return Bytes.concat(encodedOutpointValues.toArray(new byte[][]{{}}));
```
apancorb, line 69: `new byte[][]{{}}` creates an array of byte arrays containing a single empty byte array. Is this empty array intended?

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    private static void assertValidOutpointValue(Coin outpointValue) {
+        if (outpointValue == null || outpointValue.isNegative()){
```
apancorb, line 73: I was wondering, should this also check that the outpoint value is not greater than the `MAX_BTC` value found in the tests?

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,78 @@
...
+            throw new InvalidOutpointValueException(String.format("Invalid outpoint value: %s. Negative and null values are not allowed.",
```
apancorb, line 74: This line seems to be too long, we can wrap it into various lines. Something like this: 

```suggestion
      throw new InvalidOutpointValueException(
          String.format("Invalid outpoint value: %s. Negative and null values are not allowed.",
               outpointValue));
```

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
nathanieliov: Nice idea. Let's discuss it further to see its benefits.

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
nathanieliov: I will change it to `new byte[][]{}`. 

But by the way what is happening here is that `Bytes.concat` is waiting for an array of arrays, so then it can concat all of them into a single array.

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
nathanieliov: Both are very interesting questions, and I agree that we can add those checks into the utils methods if we want to be the most realistic possible. It is not supposed to receive a greater value than `MAX_BTC`, nor a zero outpoint value.

 I would just like to know what the rest of the team thinks about this before adding it.

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
nathanieliov: It will get fixed with the reformat over the file.

rskj-core/src/main/java/co/rsk/peg/bitcoin/InvalidOutpointValueException.java
nathanieliov: It's not nit, it is necessary. I reformatted the whole file, so it will get fixed.

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
nathanieliov: Thanks! If there's no further comment about this feel free to mark this discussion as resolved whenever you ready :).

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,83 @@
...
+        List<byte[]> encodedOutpointValues = new ArrayList<>();
+        for (Coin outpointValue : outpointValues) {
+            assertValidOutpointValue(outpointValue);
+            VarInt varIntOutpointValue = new VarInt(outpointValue.getValue());
+            encodedOutpointValues.add(varIntOutpointValue.encode());
+        }
+        return Bytes.concat(encodedOutpointValues.toArray(new byte[][]{}));
+    }
+
+    private static void assertValidOutpointValue(Coin outpointValue) {
```
wilmerrootstock, line 76: What do you think about changing this method name `assertValidOutpointValue` with `isValidOutpointValue`?
Assertions are not recommended in production code, and this name might generate confusion about it.

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
```diff
@@ -0,0 +1,231 @@
...
+        assertArrayEquals(expectedDecodedValues.toArray(), decodeOutpointValues.toArray());
+    }
+
+    @Test
+    void decodeOutpointValues_nullEncodedValues_shouldReturnEmptyList() {
+        // act
+        List<Coin> decodeOutpointValues = UtxoUtils.decodeOutpointValues(null);
+
+        // assert
+        List<Coin> expectedDecodedValues = Collections.EMPTY_LIST;
```
wilmerrootstock, line 180: WDYT about putting `expectedResult` before `actualResult` and not in the assertion section?

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,83 @@
...
+
+import co.rsk.bitcoinj.core.Coin;
+import co.rsk.bitcoinj.core.VarInt;
+import com.google.common.primitives.Bytes;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.spongycastle.util.encoders.Hex;
+
+public class UtxoUtils {
```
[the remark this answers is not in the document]
nathanieliov: Good catch

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public static List<Coin> decodeOutpointValues(byte[] encodedOutpointValues) {
+        if (encodedOutpointValues == null || encodedOutpointValues.length == 0) {
+            return Collections.emptyList();
+        }
+        int offset = 0;
+        List<Coin> outpointValues = new ArrayList<>();
+
+        while (encodedOutpointValues.length > offset) {
+            VarInt valueAsVarInt;
```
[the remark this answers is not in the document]
nathanieliov: `VarInt` is a special class to encode and decode integers efficiently. Check [this](https://wiki.bitcoinsv.io/index.php/VarInt)

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
wilmerrootstock: After a technical discussion, we get an agreement as a team: 

"Put the `expectedResult` right at the beginning of the assert section".

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,83 @@
...
+            try {
+                valueAsVarInt = new VarInt(encodedOutpointValues, offset);
+            } catch (Exception ex) {
+                throw new InvalidOutpointValueException(
+                    String.format("Invalid value with invalid VarInt format: %s",
+                        Hex.toHexString(encodedOutpointValues).toUpperCase()
```
marcos-iov, line 38: Why upper case?

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,83 @@
...
+            return new byte[]{};
+        }
+
+        List<byte[]> encodedOutpointValues = new ArrayList<>();
+        for (Coin outpointValue : outpointValues) {
+            assertValidOutpointValue(outpointValue);
+            VarInt varIntOutpointValue = new VarInt(outpointValue.getValue());
+            encodedOutpointValues.add(varIntOutpointValue.encode());
+        }
+        return Bytes.concat(encodedOutpointValues.toArray(new byte[][]{}));
```
marcos-iov, line 73: I'd try to stay away from this dependency. I know it's already in the project, but keep adding usages will not help remove it someday.

You can achieve this with no so much work 
```
private byte[] concatenate(byte[] left, byte[] right) {
        byte[] leftRight = Arrays.copyOf(left, left.length + right.length);
        arraycopy(right, 0, leftRight, left.length, right.length);
        return leftRight;
    }
```

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
```diff
@@ -0,0 +1,231 @@
...
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.spongycastle.util.encoders.Hex;
+
+class UtxoUtilsTest {
+
+    private final static Coin MAX_BTC = Coin.valueOf(21_000_000, 0);
```
marcos-iov, line 22: You can also get this from BridgeConstants.getMaxRbtc()

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+    private static Stream<Arguments> validOutpointValues() {
```
marcos-iov, line 24: This is extremely hard to read and follow. Think you can find a way to simplify?

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
nathanieliov: Because [VarInt documentation](https://wiki.bitcoinsv.io/index.php/VarInt) is expressed using upper case.

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
nathanieliov: Extracting each case into separate method.

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
nathanieliov: I refactored it. I dit it using `ByteArrayOutputStream`.

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
apancorb: I will mark it this as resolved, and leave it for you to decide ;)

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
```diff
@@ -0,0 +1,94 @@
...
+public final class UtxoUtils {
+
+    private UtxoUtils() {
+    }
+
+    /**
+     * Decode a {@code byte[]} of encoded outpoint values.
+     *
+     * @param encodedOutpointValues
+     * @return {@code List<Coin>} the list of outpoint values decoded in the same ordered they were
```
julia-zack, line 22: `... in the same order` (just a typo)

julia-zack, line 22: `the list of outpoint values decoded preserving the order of the entry` sounds better to me. More aligned with what you put below

julia-zack: i think saying "same order they were before encoded" is confusing, you are just decoding not encoding and decoding

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
```diff
@@ -0,0 +1,213 @@
...
+    void encodeOutpointValues(byte[] expectedEncodedOutpointValues, List<Coin> outpointValues) {
+        // act
+        byte[] encodeOutpointValues = UtxoUtils.encodeOutpointValues(outpointValues);
+
+        // assert
+        assertArrayEquals(expectedEncodedOutpointValues, encodeOutpointValues);
+    }
+
+    @Test
+    void decodeOutpointValues_nullOutpointValues_shouldReturnEmptyArray() {
```
julia-zack, line 177: this test does exactly the same as [this](https://github.com/rsksmart/rskj/pull/2328/files#diff-bc219d6855f7bc8d87e68b50b242d584b2b4b7268cd3b2ed97e230c1b22178c3R157) one.
I think in this one you were supposed to test `encodeOutpointValues`

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
```diff
@@ -0,0 +1,213 @@
...
+        // act
+        List<Coin> decodeOutpointValues = UtxoUtils.decodeOutpointValues(null);
+
+        // assert
+        List<Coin> expectedDecodedValues = Collections.EMPTY_LIST;
+        assertArrayEquals(expectedDecodedValues.toArray(), decodeOutpointValues.toArray());
+    }
+
+    @ParameterizedTest
+    @MethodSource("validOutpointValues")
```
julia-zack, line 167: it would be great to have the arguments closer to the test instead of at the beginning of the class

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
julia-zack: agree with marcos

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
```diff
@@ -0,0 +1,213 @@
...
+                187,
+                187,
+                187,
+                // 13_337 = FE9145DC00 in VarInt format
+                13_337,
+                // 14_435_729 = FEDC4591 in VarInt format
+                14_435_729
+            ))
+        );
+
```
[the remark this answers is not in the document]
nathanieliov: Looks good. I updated it based on your suggestion.

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
nathanieliov: Updated.

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
nathanieliov: Fixed

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
nathanieliov: Solved

nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
```diff
@@ -0,0 +1,176 @@
...
+        final byte[] bigOutpointValueArray = Hex.decode(
+            Stream.iterate("FCFCBBBBBBFD1934FE9145DC00", UnaryOperator.identity()).limit(1000)
+                .collect(Collectors.joining()));
+        List<Coin> bigListOfOutpointValues = Stream.generate(
+                () -> coinListOf(252, 252, 187, 187, 187, 13_337, 14_435_729)).limit(1000)
+            .flatMap(Collection::stream).collect(Collectors.toList());
+        arguments.add(Arguments.of(
+            bigOutpointValueArray,
+            bigListOfOutpointValues
+        ));
```
julia-zack, line 64:
```
infiniteStream = Stream.iterate("FCFCBBBBBBFD1934FE9145DC00", UnaryOperator.identity());
limitedInfiniteStream = infiniteStream.limit(1000);
final byte[] bigOutpointValueArray = Hex.decode(limitedInfiniteStream.collect(Collectors.joining()));
```

please use meaningful name variables when possible

julia-zack: sorry to be nit, but theres a bunch of numbers and letters all over the file. It visually helps to have them placed in variables

nathanieliov: I don't follow. The reason why this has a variable it's to give context. `bigOutpointValueArray` expressed that we are dealing with a big array with many values. 

`infiniteStream = Stream.iterate("FCFCBBBBBBFD1934FE9145DC00", UnaryOperator.identity());` This is not a `iniftiveStream` and to me separate a lambda in this way just a more confusion.

julia-zack: actually you are right. I was confused with all the numbers and letters.

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
jeremy-then: I agree.

jeremy-then (approved):
LGTM.

wilmerrootstock (approved):
LGTM!

apancorb (approved):

rskj-core/src/main/java/co/rsk/peg/bitcoin/UtxoUtils.java
julia-zack: agree!

marcos-iov (approved):

PR #2332, opened 2024-05-16
Create federation change functions enum inside /federation package

julia-zack:
Create an enum to hold FEDERATION_CHANGE_FUNCTIONS from BridgeSupport.

_Disclaimer:_ this pr just creates the enum.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/federation/FederationChangeFunctions.java
+++ b/rskj-core/src/main/java/co/rsk/peg/federation/FederationChangeFunctions.java
@@ -1,0 +1,16 @@
+package co.rsk.peg.federation;
+
+public enum FederationChangeFunctions {
+    CREATE("create"),
+    ADD("add"),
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/peg/federation/FederationChangeFunctions.java
```diff
@@ -0,0 +1,16 @@
+package co.rsk.peg.federation;
+
+public enum FederationChangeFunctions {
```
marcos-iov, line 3: Let's use singular names for enums :)

julia-zack: you right, i'll change it! Thanks

julia-zack: done

marcos-iov (approved):

PR #2333, opened 2024-05-16
Create FederationSupport interface

julia-zack:
Create FederationSupport interface in /federation package with all federation-related methods

_Disclaimer:_ added `getPendingFederation` method to replace that the pending federation is being get directly from the provider.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/federation/FederationSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/federation/FederationSupport.java
@@ -18,7 +18,7 @@ public interface FederationSupport {
     Optional<Script> getActiveFederationRedeemScript();
     Address getActiveFederationAddress();
     int getActiveFederationSize();
-    Integer getActiveFederationThreshold();
+    int getActiveFederationThreshold();
     Instant getActiveFederationCreationTime();
     long getActiveFederationCreationBlockNumber();
     byte[] getActiveFederatorBtcPublicKey(int index);
@@ -28,8 +28,8 @@ public interface FederationSupport {
     @Nullable
     Federation getRetiringFederation();
     Address getRetiringFederationAddress();
-    Integer getRetiringFederationSize();
-    Integer getRetiringFederationThreshold();
+    int getRetiringFederationSize();
+    int getRetiringFederationThreshold();
     Instant getRetiringFederationCreationTime();
     long getRetiringFederationCreationBlockNumber();
     byte[] getRetiringFederatorPublicKey(int index);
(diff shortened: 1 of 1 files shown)


rskj-core/src/main/java/co/rsk/peg/federation/FederationSupport.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.List;
+import java.util.Optional;
+
+public interface FederationSupport {
+
+    Federation getActiveFederation();
+    Optional<Script> getActiveFederationRedeemScript();
+    Address getActiveFederationAddress();
+    int getActiveFederationSize();
+    Integer getActiveFederationThreshold();
```
apancorb, line 21: I notice some of the return types are primitives while others are wrappers. Do you think it would be possible to use the primitive type always? That way we avoid doing null checks and such

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupport.java
```diff
@@ -0,0 +1,48 @@
...
+    Instant getActiveFederationCreationTime();
+    long getActiveFederationCreationBlockNumber();
+    byte[] getActiveFederatorBtcPublicKey(int index);
+    byte[] getActiveFederatorPublicKeyOfType(int index, FederationMember.KeyType keyType);
+    List<UTXO> getActiveFederationBtcUTXOs();
+
+    @Nullable
+    Federation getRetiringFederation();
```
marcos-iov, line 29: For another task, but what do you think about changing these functions to return an Optional instead of null?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupport.java
julia-zack: very good point actually. @marcos-iov you want me to create a task to take care of this? Or should I just do it here?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupport.java
julia-zack: I like the idea. I'll add it to the backlog

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupport.java
marcos-iov: You can probably simply do it here. You'll probably need to do some adjustments when this gets integrated but that will be on different task

julia-zack: Perfect, thanks !!!!

julia-zack: done

marcos-iov (approved):

PR #2340, opened 2024-05-17
Replace fed related methods impl in bridge support

julia-zack:
- Replace federation-related methods implementation in BridgeSupport by calling the ones from federationSupport instance.
- Remove `/peg/FederationSupport` class.
- Fix broken uses of removed FederationSupport class.


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1864,110 +1855,87 @@ public Federation getRetiringFederation() {
...
-        return getActiveFederation().getCreationBlockNumber();
+    public long getActiveFederationCreationBlockNumber() {
+        return federationSupport.getActiveFederationCreationBlockNumber();
     }
 
     /**
      * Returns the retiring federation bitcoin address.
      * @return the retiring federation bitcoin address, null if no retiring federation exists
      */
     public Address getRetiringFederationAddress() {
```
marcos-iov, line 1921: We already talked about making this return an Optional, right?

marcos-iov (approved):

PR #2344, opened 2024-05-20
Emit pegout_tranaction_created event when a peg-out is created

nathanieliov:
Add use of BridgeEventLogger.logPegoutTransactionCreated. Once RSKIP428 is activated it emits the pegout_transaction_created event in each of the peg-out flows.


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1053,28 +1053,37 @@ private void migrateFunds(
...
                 amountMigrated
             );
-            eventLogger.logReleaseBtcRequested(rskTxHash.getBytes(), btcTx, amountMigrated);
+            eventLogger.logReleaseBtcRequested(rskTxHash.getBytes(), migrationTransaction, amountMigrated);
         } else {
-            pegoutsWaitingForConfirmations.add(btcTx, rskExecutionBlock.getNumber());
+            pegoutsWaitingForConfirmations.add(migrationTransaction, rskExecutionBlock.getNumber());
+        }
+
+        if (activations.isActive(RSKIP428)) {
```
julia-zack, line 1069: `if (shouldLogPegoutCreation())`

descriptive names please :))

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1187,67 +1196,80 @@ private void processPegoutsInBatch(
...
-                while (pegoutEntries.size() > 1 && result.getResponseCode() == ReleaseTransactionBuilder.Response.EXCEED_MAX_TRANSACTION_SIZE) {
-                    logger.info("[processPegoutsInBatch] Max size exceeded, going to divide {} requests in half", pegoutEntries.size());
-                    int firstHalfSize = pegoutEntries.size() / 2;
-                    pegoutEntries = pegoutEntries.subList(0, firstHalfSize);
-                    result = txBuilder.buildBatchedPegouts(pegoutEntries);
-                }
+        if (pegoutEntries.isEmpty()) {
+            logger.warn("[processPegoutsInBatch] No pegout requests to process");
+        } else {
+            logger.info("[processPegoutsInBatch] going to create a batched pegout transaction for {} requests, total amount {}", pegoutEntries.size(), totalPegoutValue);
```
julia-zack, line 1217: mmmm, i don't like this order. 
I would do this instead:

```
if (pegoutEntries.isEmpty()) {
            logger.warn("[processPegoutsInBatch] No pegout requests to process");
            // update next Pegout height even if there were no request in queue
            long nextPegoutHeight = currentBlockNumber + bridgeConstants.getNumberOfBlocksBetweenPegouts();
            provider.setNextPegoutHeight(nextPegoutHeight);
            logger.info("[processPegoutsInBatch] Next Pegout Height updated from {} to {}", currentBlockNumber, nextPegoutHeight);
            return;
}
```

and then proceed with the behavior when we have entries, with no more elses and ifs. wdyt?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportPegoutTransactionCreatedEventTest.java
julia-zack, line 1: no need to add _test__ before each test name

rskj-core/src/test/java/co/rsk/peg/BridgeSupportPegoutTransactionCreatedEventTest.java
```diff
@@ -0,0 +1,323 @@
...
+    private static Stream<Arguments> pegoutTransactionCreatedEventArgsProvider() {
+        return Stream.of(
+            Arguments.of(ActivationConfigsForTest.arrowhead600().forBlock(0)),
+            Arguments.of(ActivationConfigsForTest.lovell700().forBlock(0))
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("pegoutTransactionCreatedEventArgsProvider")
+    void test_pegoutTransactionCreatedEvent_when_pegout_batch_is_created(ActivationConfig.ForBlock activations) throws IOException {
```
julia-zack, line 82: `pegoutTransactionCreatedEvent_whenPegoutBatchCreation_ shouldDoSomething`
something like that, wdyt?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportPegoutTransactionCreatedEventTest.java
```diff
@@ -0,0 +1,323 @@
...
+        Coin totalTransactionAmount = pegoutRequests.stream().map(Entry::getAmount).reduce(Coin.ZERO, Coin::add);
+
+        verify(eventLogger, times(1)).logBatchPegoutCreated(btcTxHash, pegoutRequestRskTxHashes);
+        verify(eventLogger, times(1)).logReleaseBtcRequested(pegoutCreationRskTxHash.getBytes(), pegoutBatchTransaction, totalTransactionAmount);
+
+        if (activations.isActive(ConsensusRule.RSKIP428)){
+            verify(eventLogger, times(1)).logPegoutTransactionCreated(btcTxHash, outpointValues);
+        } else {
+            verify(eventLogger, never()).logPegoutTransactionCreated(any(), any());
+        }
```
julia-zack, line 136: i'm not a fan of doing this inside the test.


`assertPegoutTransactionCreatedEmmittedWhenSupposedTo(activations, btcTxHash, outpointValues);`

```
private void assertPegoutTransactionCreatedEmmittedWhenSupposedTo(activations, btcTxHash, outpointValues) {
     if (!activations.isActive(ConsensusRule.RSKIP428)) {
          verify(eventLogger, never()).logPegoutTransactionCreated(any(), any());
          return;
     }
     
     verify(eventLogger, times(1)).logPegoutTransactionCreated(btcTxHash, outpointValues);
}
```

wdyt?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
nathanieliov: I don't consider this needed here. Even though, if we agree to do this for every RSKIP, I think we should do it in a better way, but not creating this kind of method in all the places where the RSKIP is being checked. Sometimes RSKIP checks are done in more than one class, so it will imply having this kind of method you proposed in more than one place by doing as you proposed.

julia-zack: you are right, we should find a better way. I'll resolve this conversation.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
nathanieliov: Good catch. This can be way better. I got confused with the use of different variables for checking if there are no pegout requests.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportPegoutTransactionCreatedEventTest.java
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/BridgeSupportPegoutTransactionCreatedEventTest.java
nathanieliov: I don't think this `if` is a big deal compared to the benefits of the refactoring you are proposing. It's just one line each block and it is pretty clear what's being done.

In my opinion, moving it to a method does not add many benefits in terms of improving legibility and simplifying the test.

rskj-core/src/test/java/co/rsk/peg/PowpegMigrationTest.java
```diff
@@ -473,9 +475,19 @@ private void testChangePowpeg(
...
         bridgeSupport.save();
-        bridgeStorageProvider.getPegoutsWaitingForConfirmations().getEntries().stream()
-            .map(PegoutsWaitingForConfirmations.Entry::getBtcTransaction)
-            .forEach(pegoutTx -> verifyPegoutTxSigHashIndex(activations, bridgeStorageProvider, pegoutTx));
+        List<BtcTransaction> pegoutsTxs = bridgeStorageProvider.getPegoutsWaitingForConfirmations()
+            .getEntries().stream()
+            .map(Entry::getBtcTransaction).collect(Collectors.toList());
+
+        pegoutsTxs.stream().peek(
+            pegoutTx -> verifyPegoutTxSigHashIndex(activations, bridgeStorageProvider, pegoutTx));
```
apancorb, line 483:
```suggestion
        pegoutsTxs.stream().forEach(
            pegoutTx -> verifyPegoutTxSigHashIndex(activations, bridgeStorageProvider, pegoutTx));
```

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1187,67 +1196,80 @@ private void processPegoutsInBatch(
...
-            Coin totalPegoutValue = pegoutEntries
-                .stream()
-                .map(ReleaseRequestQueue.Entry::getAmount)
-                .reduce(Coin.ZERO, Coin::add);
-
-            if (wallet.getBalance().isLessThan(totalPegoutValue)) {
-                logger.warn("[processPegoutsInBatch] wallet balance {} is less than the totalPegoutValue {}", wallet.getBalance(), totalPegoutValue);
-                return;
-            }
+        if (isCurrentBlockBelowNextPegoutCreationBlockNumber(currentBlockNumber, nextPegoutCreationBlockNumber)) {
```
apancorb, line 1199: Not sure, but I find it clearer just using the less than operator here.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1187,67 +1196,80 @@ private void processPegoutsInBatch(
...
-                    pegoutEntries = pegoutEntries.subList(0, firstHalfSize);
-                    result = txBuilder.buildBatchedPegouts(pegoutEntries);
-                }
+        if (pegoutEntries.isEmpty()) {
+            logger.warn("[processPegoutsInBatch] No pegout requests to process");
+        } else {
+            logger.info("[processPegoutsInBatch] going to create a batched pegout transaction for {} requests, total amount {}", pegoutEntries.size(), totalPegoutValue);
+            ReleaseTransactionBuilder.BuildResult result = txBuilder.buildBatchedPegouts(pegoutEntries);
+
+            while (pegoutEntries.size() > 1 && result.getResponseCode() == ReleaseTransactionBuilder.Response.EXCEED_MAX_TRANSACTION_SIZE) {
```
apancorb, line 1220: nit: wrap in two lines

rskj-core/src/test/java/co/rsk/peg/PowpegMigrationTest.java
```diff
@@ -473,9 +475,19 @@ private void testChangePowpeg(
...
+
+        if (activations.isActive(ConsensusRule.RSKIP428)) {
+            pegoutsTxs.stream()
+                .peek(pegoutTx -> verifyPegoutTransactionCreatedEvent(bridgeEventLogger, pegoutTx));
```
apancorb, line 487: Same here

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: i think i agree

rskj-core/src/test/java/co/rsk/peg/PowpegMigrationTest.java
nathanieliov: Solved

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
nathanieliov: Done

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
nathanieliov: Undo.

rskj-core/src/test/java/co/rsk/peg/PowpegMigrationTest.java
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -36,11 +38,13 @@ public static List<BtcECKey> getBtcEcKeysFromSeeds(String[] seeds, boolean sorte
...
     public static Address createP2PKHAddress(NetworkParameters networkParameters, String seed) {
-        BtcECKey key = BtcECKey.fromPrivate(HashUtil.keccak256(seed.getBytes(StandardCharsets.UTF_8)));
+        BtcECKey key = BtcECKey.fromPrivate(
+            HashUtil.keccak256(seed.getBytes(StandardCharsets.UTF_8)));
         return key.toAddress(networkParameters);
     }
 
-    public static Address createP2SHMultisigAddress(NetworkParameters networkParameters, List<BtcECKey> keys) {
+    public static Address createP2SHMultisigAddress(NetworkParameters networkParameters,
+        List<BtcECKey> keys) {
```
julia-zack, line 47: i won't split this

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -57,7 +61,8 @@ public static Sha256Hash createHash(int nHash) {
         return Sha256Hash.wrap(bytes);
     }
 
-    public static List<BtcECKey.ECDSASignature> extractSignaturesFromTxInput(TransactionInput txInput) {
+    public static List<BtcECKey.ECDSASignature> extractSignaturesFromTxInput(
+        TransactionInput txInput) {
```
julia-zack, line 65: same here, i found it harder to read

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
julia-zack, line 47: i won't split this

julia-zack (approved):

apancorb (approved):
LGTM

rskj-core/src/test/java/co/rsk/peg/BridgeSupportPegoutTransactionCreatedEventTest.java
```diff
@@ -0,0 +1,325 @@
...
+        when(provider.getNewFederationBtcUTXOs())
+            .thenReturn(fedUTXOs);
+
+        List<Entry> pegoutRequests = PegTestUtils.createReleaseRequestQueueEntries(3);
+        when(provider.getReleaseRequestQueue())
+            .thenReturn(new ReleaseRequestQueue(pegoutRequests));
+
+        PegoutsWaitingForConfirmations pegoutsWaitingForConfirmations = provider.getPegoutsWaitingForConfirmations();
+
+        BridgeSupport bridgeSupport = new BridgeSupportBuilder()
```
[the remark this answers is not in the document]
apancorb: For the second option I would have the name as `from`, `of`, etc (since `getInstance` might allude to the same instance being returned)

rskj-core/src/test/java/co/rsk/peg/federation/FederationTestUtils.java
```diff
@@ -46,6 +48,36 @@ public final class FederationTestUtils {
     private FederationTestUtils() {
     }
 
+    public static final List<BtcECKey> TEST_GENESIS_FED_SIGNERS = BitcoinTestUtils.getBtcEcKeysFromSeeds(
+        new String[]{"fa01", "fa02", "fa03", "fa04", "fa05", "fa06", "fa07", "fa08", "fa09"}, true
+    );
+
+    public static List<BtcECKey> TEST_GENESIS_ERP_SIGNERS = BitcoinTestUtils.getBtcEcKeysFromSeeds(
```
marcos-iov, line 55: Do these need to be public?

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
```diff
@@ -168,4 +182,75 @@ private static Stream<Arguments> invalidEncodedOutpointValues() {
...
+        fundingTransaction.addInput(
+            BitcoinTestUtils.createHash(1),
+            FIRST_OUTPUT_INDEX,
+            new Script(new byte[]{})
+        );
+        fundingTransaction.addOutput(amountToSend, TEST_ERP_FEDERATION.getAddress());
+
+        BtcTransaction pegout = new BtcTransaction(btcMainnetParams);
+        pegout.addInput(fundingTransaction.getOutput(FIRST_OUTPUT_INDEX));
+        pegout.addOutput(amountToSend, testAddress);
```
marcos-iov, line 200: Do you really need all this setup? Looks like all you really need is a Bitcoin transaction with some inputs

rskj-core/src/test/java/co/rsk/peg/federation/FederationTestUtils.java
nathanieliov: I put it as public just in case they are need it somewhere else to test signing a transaction

rskj-core/src/test/java/co/rsk/peg/bitcoin/UtxoUtilsTest.java
nathanieliov: The inputs must have the outpoints because the outpoint value is obtained from them. So, at least the funding transactions are needed. I can remove the outputs but we will be testing the method against an unrealistic transaction. I prefer it as it is now.

rskj-core/src/test/java/co/rsk/peg/federation/FederationTestUtils.java
```diff
@@ -46,6 +48,36 @@ public final class FederationTestUtils {
     private FederationTestUtils() {
     }
 
+    public static final List<BtcECKey> TEST_GENESIS_FED_SIGNERS = BitcoinTestUtils.getBtcEcKeysFromSeeds(
```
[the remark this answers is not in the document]
nathanieliov: Done

rskj-core/src/test/java/co/rsk/peg/federation/FederationTestUtils.java
marcos-iov: If it's not needed then I wouldn't make it public. Public keys can be obtained from the federation object

nathanieliov: Done

marcos-iov (approved):

PR #2349, opened 2024-05-22
Fix some tests that broke after FederationSupport refactor

julia-zack:


rskj-core/src/test/java/co/rsk/test/builders/BridgeSupportBuilder.java
```diff
@@ -96,7 +102,8 @@ public BridgeSupportBuilder withSignatureCache(SignatureCache signatureCache) {
 
     public BridgeSupport build() {
         StorageAccessor bridgeStorageAccessor = new BridgeStorageAccessorImpl(repository);
-        FeePerKbStorageProvider feePerKbStorageProvider = new FeePerKbStorageProviderImpl(bridgeStorageAccessor);
+        FeePerKbSupport feePerKbSupport = newFeePerKbSupportInstance(bridgeStorageAccessor, bridgeConstants.getFeePerKbConstants());
```
marcos-iov, line 105: This might not work. If bridgeConstants has not been set then it will return `null`. Jeremy is working on a task to fix this, so I'll merge this change in the meantime and then we can update it

marcos-iov (approved):

PR #2353, opened 2024-05-23
Fix BridgeStorageProviderFederationTests

julia-zack:
Rename it to FederationStorageProviderTests and move it into /federation package.


jeremy-then (approved):
LGTM.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportProcessFundsMigrationTest.java
```diff
@@ -167,7 +168,7 @@ void test_processFundsMigration(
             Coin.COIN,
             oldFederation.getAddress()
         ));
-        when(provider.getOldFederationBtcUTXOs()).thenReturn(sufficientUTXOsForMigration);
+        when(federationStorageProvider.getOldFederationBtcUTXOs()).thenReturn(sufficientUTXOsForMigration);
```
julia-zack, line 171: probably i'd have to add `feePerKbSupport` with that `federationStorageProvider` to the built `bridgeSupport`. So i'll do it when that feature is added.
cc: @jeremy-then

marcos-iov (approved):

PR #2359, opened 2024-05-24
Remove feePerKb related tests from BridgeSupportTests

marcos-iov:
- Remove tests from BridgeSupportTest class that are now part of FeePerKbSupportTest
- Remove some exceptions from method signatures that are never thrown

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
@@ -24,10 +24,7 @@ import co.rsk.bitcoinj.script.ScriptBuilder;
 import co.rsk.bitcoinj.store.BlockStoreException;
 import co.rsk.bitcoinj.wallet.Wallet;
 import co.rsk.blockchain.utils.BlockGenerator;
-import co.rsk.peg.constants.BridgeConstants;
-import co.rsk.peg.constants.BridgeMainNetConstants;
-import co.rsk.peg.constants.BridgeRegTestConstants;
-import co.rsk.peg.constants.BridgeTestNetConstants;
+import co.rsk.peg.constants.*;
 import co.rsk.core.RskAddress;
 import co.rsk.crypto.Keccak256;
 import co.rsk.peg.feeperkb.FeePerKbStorageProvider;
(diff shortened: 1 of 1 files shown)


rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -88,19 +81,13 @@
 
 import static org.hamcrest.Matchers.lessThanOrEqualTo;
 import static org.hamcrest.core.Is.is;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
apancorb, line 84: This can lead to a larger JAR file because it may include unused assertions. However this should not have a huge impact because the actual code should be small

apancorb (approved):
LGTM

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
wilmerrootstock: Agree with @apancorb. Aside from that, the compilation might take longer as well. Anyway, we can discuss how to address this topic for the next PRs

wilmerrootstock (changes requested):
Good job @marcos-iov! Left some minors comments.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
jeremy-then: As discussed, using wild card to import does not affect the runtime performance at all because the compiler will take care of it and import exactly what is being used.

Besides that, I prefer to import every class individually, makes it easier to read where it is coming from in an environment like this.

jeremy-then (approved):
Left a couple of nit comments. In general, looks good to me.

PR #2375, opened 2024-05-27
Fix RemascFederationProviderTest by creating FederationSupport object

marcos-iov:


rskj-core/src/test/java/co/rsk/remasc/RemascFederationProviderTest.java
```diff
@@ -30,81 +36,84 @@ class RemascFederationProviderTest {
...
-                bridgeMainNetConstants,
-                activations
+        return new RemascFederationProvider(
+            activations,
+            federationSupport
         );
+    }
 
-        FederationSupport federationSupport = new FederationSupport(bridgeMainNetConstants, bridgeStorageProvider, blockchain.getBestBlock(), activations);
+    private static FederationSupport getFederationSupport(Repository repository, Blockchain blockchain, ForBlock activations) {
```
julia-zack, line 107: no big deal, but since we are setting the network as mainnet inside the method, it could be named `getMainnetFederationSupport`. Just an idea

PR #2378, opened 2024-05-27
Fix BridgeSupportRegisterBtcTransactionTest class

marcos-iov:


rskj-core/src/test/java/co/rsk/peg/BridgeSupportRegisterBtcTransactionTest.java
```diff
@@ -91,8 +94,8 @@ class BridgeSupportRegisterBtcTransactionTest {
     private BtcLockSenderProvider btcLockSenderProvider;
     private PeginInstructionsProvider peginInstructionsProvider;
 
-    private List<UTXO> retiringFederationUtxos = new ArrayList<>();
-    private List<UTXO> activeFederationUtxos = new ArrayList<>();
+    private final List<UTXO> retiringFederationUtxos = new ArrayList<>();
+    private final List<UTXO> activeFederationUtxos = new ArrayList<>();
```
nathanieliov, line 98: I think these two fields should be instantiated in the setup method since every test using it possibly adds elements to the instance list.

nathanieliov (approved):
LGTM, just a suggestion

rskj-core/src/test/java/co/rsk/peg/BridgeSupportRegisterBtcTransactionTest.java
marcos-iov: Just trying to fix the tests to compile at this stage :) Don't want to get into too much refactoring or it will be a never ending story

julia-zack (approved):

PR #2409, opened 2024-05-28
Fee per kb tests improvements

marcos-iov:
- Removed some tests that were redundant after the new test cases added
- Create a function to return an AbiCallElection object with a pre-existing vote


wilmerrootstock (approved):
Great job, @marcos-iov!

PR #2430, opened 2024-05-31
Fix BridgeSupportTestIntegration

julia-zack:


jeremy-then (approved):
LGTM.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestIntegration.java
```diff
@@ -484,32 +539,34 @@ void callUpdateCollectionsThrowsCouldNotAdjustDownwards() throws IOException {
...
+            activationsBeforeForks
         );
 
+        federationStorageProvider = createFederationStorageProvider(track);
+        FederationSupport federationSupport = federationSupportBuilder
+            .withFederationConstants(federationConstants)
+            .withFederationStorageProvider(federationStorageProvider)
+            .withActivations(activationsBeforeForks)
+            .withRskExecutionBlock(rskCurrentBlock)
+            .build();
```
apancorb, line 566: Since you are using a bunch of times this builder. Maybe have a private method that returns this?

julia-zack: yes, that's a good idea.
Since we are using bridge support and federation support builders in many test files i thought about creating a "builder helper" class to avoid having the same private method in all of them.
Lets discuss it

nathanieliov (approved):
LGTM

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestIntegration.java
```diff
@@ -4034,121 +4239,217 @@ private BridgeSupport getBridgeSupportWithMocksForWhitelistTests(LockWhitelist m
...
+        return bridgeSupportBuilder
+            .withBridgeConstants(bridgeConstants)
+            .withProvider(provider)
+            .withRepository(track)
+            .withBtcBlockStoreFactory(blockStoreFactory)
+            .withFederationSupport(federationSupport)
+            .build();
+    }
+
+    private BridgeSupport getBridgeSupport(
```
marcos-iov, line 4332: Can we avoid adding all these new methods and use BridgeSupportBuilder instead? That was the original purpose

julia-zack: done

marcos-iov (approved):

PR #2433, opened 2024-06-01
Integration Tests For FeePerKb

wilmerrootstock:
## Description
Making use of the newly created InMemoryStorage class, create tests simulating different real-life scenarios, avoiding using mocks as much as possible.

Integration tests are written to observe the behavior simulating real-life scenarios and stressing the InMemoryStorage executing them one after another.

## Motivation and Context
Get Coverage with more realistic scenarios, even more than the Unit Tests.

## How Has This Been Tested?
Through Integration tests.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
```diff
@@ -0,0 +1,478 @@
...
+        Coin differentFeePerKbVote = Coin.valueOf(60_000L);
+        Coin maxFeePerKbVote = feePerKbConstants.getMaxFeePerKb();
+        Coin excessiveFeePerKbVote = maxFeePerKbVote.add(Coin.SATOSHI);
+        Coin genesisFeePerKb = feePerKbConstants.getGenesisFeePerKb();
+
+        // Get fee per kb before any voting, should return the genesis fee per kb
+        assertFeePerKbValue(genesisFeePerKb);
+
+        /*
+         *  1st voting: authorizer 1 and 2 vote the same value
```
apancorb, line 45: I am wandering, can each of these sections be placed in its own test? Or is there a need for all of them to be chained in this manner?

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
```diff
@@ -0,0 +1,478 @@
...
+    }
+
+    private Transaction getTransactionFromCaller(RskAddress authorizer) {
+        Transaction txFromAuthorizedCaller = mock(Transaction.class);
+        when(txFromAuthorizedCaller.getSender(signatureCache)).thenReturn(authorizer);
+
+        return txFromAuthorizedCaller;
+    }
+
+    private List<RskAddress> getAuthorizedRskAddresses(){
```
apancorb, line 467: nit: add space before open brace

apancorb (commented):
Looks very nice and readable, left two comments

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
wilmerrootstock: Hey, Antonio!
Although we're working with UT's tools, this approach creates integration tests, that is, test the whole Fee Per KB module, for this reason, it is important to make different test cases stressing the InMemoryStorage executing one test after another, and observe the behavior interacting like a real-life scenario without isolate them.

marcos-iov: We should this with the rest of the team @apancorb. The plan here was to create a test that simulates a real interaction with the fee per kb feature. That implies keeping the state and not reseting it between tests.

One option is to do it this way, in a single big test executing one action after the other. The other option is to do it in separate tests, sharing the same environment between tests. This last option implies that those tests will not be independent.

What do you think? Any other ideas?

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
```diff
@@ -0,0 +1,475 @@
...
+            "a02db0ed94a5894bc6f9079bb9a2d93ada1917f3",
+            "180a7edda4e640ea5a3e495e17a1efad260c39e9",
+            "8418edc8fea47183116b4c8cd6a12e51a7e169c1"
+        ).map(RskAddress::new).collect(Collectors.toList());
+    }
+
+    private RskAddress getUnauthorizedRskAddress() {
+        return new RskAddress("e2a5070b4e2cb77fe22dff05d9dcdc4d3eaa6ead");
+    }
+}
```
jeremy-then, line 475: What happens if, for example, the current fee per kb is 100_000L and the authorizers decide to vote for the same 100_000L? Does it get rejected? Does it throw an error?

wilmerrootstock: Hey, Jere!
Brilliant contribution! In this case, 100_000L is set again as "the new fee per KB value".
I just added this case to the tests:
`16th voting: authorized callers vote the same current fee per KB value`

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
apancorb: I see, for integration tests sharing state between tests is fine. I would recommend having each Xth voting section in its own small test and sharing state at the class level.

Check [this article](https://www.baeldung.com/junit-testinstance-annotation#:~:text=Sharing%20state%20is%20usually%20an,tests%20that%20intentionally%20share%20state.) out that will describe how to set per-class lifecycle and how to set the ordering of tests if it matters. This might help you achieve what you want to create. 

_As a side note, integration tests are usually placed under the `intTest` directory next to `main` and `test`. Gradle natively supports this convention. Maybe in the future this is something we can adopt._

nathanieliov: I agree with @apancorb. This long test should be split using one of the options available for this case using `junit`.

nathanieliov (dismissed):
Very well done, Wilmer! Just a single comment, same as @apancorb: I would rather split the test into separate tests sharing the same context instance.

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
wilmerrootstock: ### Refactoring Test Integration for Fee Per KB
- Adding some annotations such as @testinstance, and @BeforeAll to help share state in the test.
- Adding annotations such as @TestMethodOrder, and @order to create an order of test execution.
- Adding Caller enum to get the code more expressive.
- After adding the Caller enum, adapt the voteFeePerKb method to receive as an In Parameter RskAddress.
- Removing voteFeePerKbByUnauthorizedCaller, getAuthorizedRskAddresses, and getUnauthorizedRskAddress methods since they are no longer needed after the adaptation of the Caller enum in the code.

_P.S.: it's just missing to move this integration test to the IntegrationTest directory._

rskj-core/src/test/java/co/rsk/peg/utils/Caller.java
```diff
@@ -0,0 +1,24 @@
+package co.rsk.peg.utils;
+
+import co.rsk.core.RskAddress;
+
+/**
+ * Created to makes easier the way to get RskAddress of 3 authorized and 1 unauthorized Callers. It
+ * contains hex encoded string associated to each caller and return the RskAddress of each one.
+ */
+public enum Caller {
```
marcos-iov, line 9: I wouldn't create this, at least not in a separate file. If it makes sense to have this enum then simply it put it inside the test class, it's the only place it's used

wilmerrootstock: But, we could use it in the UTs as well. We can think about having an isolated library project in the future for tests and create a `.jar` to be used as a kind of test `library/framework`, that could be used in `cross-projects (RSKj/Powpeg-Node)` and `cross-tests (UTs and ITs)`. In the meantime,  we might put all these objects together in a `utils` package. Creating it inside the class breaks the Single Responsibility Principle.

nathanieliov: I like @wilmerrootstock  idea very much about moving this out of the test for the benefit of reusability.

marcos-iov: Ok, in that case I'd move this to feeperkb package and rename it FeePerKbVoteCaller. What do you think?

I'm not a fan of generic `util` packages or classes

wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
```diff
@@ -0,0 +1,475 @@
...
+        feePerKbSupport = new FeePerKbSupportImpl(feePerKbConstants, feePerKbStorageProvider);
+        signatureCache = mock(SignatureCache.class);
+    }
+
+    @Test
+    void feePerKbIntegrationTest() {
+        Coin differentFeePerKbVote = Coin.valueOf(60_000L);
+        Coin maxFeePerKbVote = feePerKbConstants.getMaxFeePerKb();
+        Coin excessiveFeePerKbVote = maxFeePerKbVote.add(Coin.SATOSHI);
+        Coin genesisFeePerKb = feePerKbConstants.getGenesisFeePerKb();
```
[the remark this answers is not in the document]
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
```diff
@@ -0,0 +1,567 @@
...
+        assertFeePerKbValue(secondFeePerKbVote);
+    }
+
+    /**
+     * Authorizers 1 and 2 vote different values. Then authorizer 3 votes the same value as
+     * authorizer 1
+     */
+    @Test
+    @Order(3)
+    void thirdVoting() {
```
julia-zack, line 98: i would make the test names more descriptive instead of having the comments above them.
wdyt?

wilmerrootstock: Totally agree, @julia-zack!

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
```diff
@@ -0,0 +1,567 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.TestInstance.Lifecycle;
+import org.junit.jupiter.api.TestMethodOrder;
+
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+@TestInstance(Lifecycle.PER_CLASS)
```
apancorb, line 24: Not sure, but I don't think the `@TestInstance(Lifecycle.PER_CLASS)` annotation is not needed here since it is implied by `@TestMethodOrder(MethodOrderer.OrderAnnotation.class)`

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbVoteCaller.java
```diff
@@ -0,0 +1,24 @@
+package co.rsk.peg.feeperkb;
+
+import co.rsk.core.RskAddress;
+
+/**
+ * Created to makes easier the way to get RskAddress of 3 authorized and 1 unauthorized Callers. It
+ * contains hex encoded string associated to each caller and return the RskAddress of each one.
+ */
+public enum FeePerKbVoteCaller {
```
apancorb, line 9: Are we expecting this to only be used by `FeePerKbIntegrationTest`? If that is the case, we can just declare them as constants in `FeePerKbIntegrationTest`

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
wilmerrootstock: Hey, @apancorb!
@TestMethodOrder(MethodOrderer.OrderAnnotation.class) doesn't have built-in @TestInstance(Lifecycle.PER_CLASS) in it. If we don't set `@TestInstance(Lifecycle.PER_CLASS)` it goes work `Lifecycle.PER_METHOD` by default. That means, if we are going to share a state of a variable in all methods, for instance, `currentFeePerKb`, it must be static. Apart from that, the method annotated with `@BeforeAll` must be static and all into it, static as well. That being said, I believe our best option is to use `@TestInstance(Lifecycle.PER_CLASS)`.

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbVoteCaller.java
wilmerrootstock: Hey, @apancorb! Take a look at this thread, this topic was discussed in it:

https://github.com/rsksmart/rskj/pull/2433#discussion_r1638970808

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
apancorb: Thanks!

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
```diff
@@ -0,0 +1,567 @@
...
+    private Coin currentFeePerKb;
+    private Coin differentFeePerKbVote;
+
+    @BeforeAll
+    void setUp() {
+        StorageAccessor inMemoryStorage = new InMemoryStorage();
+        FeePerKbStorageProvider feePerKbStorageProvider = new FeePerKbStorageProviderImpl(inMemoryStorage);
+        feePerKbSupport = new FeePerKbSupportImpl(feePerKbConstants, feePerKbStorageProvider);
+        signatureCache = mock(SignatureCache.class);
+        differentFeePerKbVote = Coin.valueOf(60_000L);
```
nathanieliov, line 38: This field is only set to a value here; it seems to be a constant to me. It could be moved to fed-level declaration. Wdyt?

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
wilmerrootstock: Done! @jeremy-then @julia-zack @marcos-iov

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIntegrationTest.java
wilmerrootstock: Done! Yes, it is a constant. Good catch! differentFeePerKbVote was changed from variable to constant.

marcos-iov (approved):

apancorb (dismissed):

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
```diff
@@ -0,0 +1,497 @@
...
+        FeePerKbStorageProvider feePerKbStorageProvider = new FeePerKbStorageProviderImpl(inMemoryStorage);
+        feePerKbSupport = new FeePerKbSupportImpl(feePerKbConstants, feePerKbStorageProvider);
+        signatureCache = mock(SignatureCache.class);
+    }
+
+    @Test
+    @Order(0)
+    void getGenesisFeePerKb_whenNotExistsPreviousVoting_shouldReturnGenesisFeePerKbValue() {
+        Coin genesisFeePerKb = feePerKbConstants.getGenesisFeePerKb();
+        currentFeePerKb = genesisFeePerKb;
```
julia-zack, line 44: shouldn't we access the `currentFeePerKb` value through `getFeePerKb()` method?

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
```diff
@@ -0,0 +1,497 @@
...
+        voteFeePerKb(seventhFeePerKbVote, FeePerKbVoteCaller.THIRD_AUTHORIZED.getRskAddress());
+        currentFeePerKb = seventhFeePerKbVote;
+
+        // Get fee per kb, should return the third vote value
+        assertFeePerKbValue(seventhFeePerKbVote);
+    }
+
+    @Test
+    @Order(8)
+    void voteFeePerKbChange_whenAuthorizerOneAndThreeVoteAfterAuthorizerOneVoteAuthorizerTwoDifferentValueAfterThatUnauthorizedCallerVotesSameThanAuthorizerTwoAndLasVoteIsEmittedByAuthorizerThree_shouldSetValueVotedByOneAndThree() {
```
julia-zack, line 194: kind of nit, but I find this test name really hard to follow

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
julia-zack: or at least set it after the assertion

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
```diff
@@ -0,0 +1,497 @@
...
+        voteFeePerKb(differentFeePerKbVote, FeePerKbVoteCaller.UNAUTHORIZED.getRskAddress());
+        //It still has the previous voting round value
+        assertFeePerKbValue(currentFeePerKb);
+
+        // Second vote is by authorizer 1
+        voteFeePerKb(ninthFeePerKbVote, FeePerKbVoteCaller.FIRST_AUTHORIZED.getRskAddress());
+        //It still has the previous voting round value
+        assertFeePerKbValue(currentFeePerKb);
+
+        // Third vote is by authorizer 2. it votes similar then the unauthorized caller
```
julia-zack, line 235: nit: `Third vote is by authorizer 2, who votes the same as the unauthorized caller`

julia-zack: (i mean, the vote is not _similar_ but the same)

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
```diff
@@ -0,0 +1,497 @@
...
+        voteFeePerKb(fifteenthFeePerKbVote, FeePerKbVoteCaller.SECOND_AUTHORIZED.getRskAddress());
+        currentFeePerKb = fifteenthFeePerKbVote;
+
+        //It still has the previous voting round value
+        assertFeePerKbValue(fifteenthFeePerKbVote);
+    }
+
+    @Test
+    @Order(16)
+    void voteFeePerKbChange_whenAuthorizerOneAndThreeVoteSameCurrentFeePerKbValueAndInBetweenAuthorizerTwoVotesDifferentValue_shouldSetValueVotedByOneAndThree() {
```
julia-zack, line 460: isn't this test [duplicated](https://github.com/rsksmart/rskj/pull/2433/files#diff-72f83e7f5b9745bce799c9de61da92d8ffb07cd4646ef1da1758d11431208939R84)?

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbVoteCaller.java
```diff
@@ -0,0 +1,24 @@
...
+
+    private final String RskAddress;
+
+    FeePerKbVoteCaller(String rskAddress) {
+        RskAddress = rskAddress;
+    }
+
+    public RskAddress getRskAddress() {
+        return new RskAddress(RskAddress);
+    }
```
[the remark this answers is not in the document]
wilmerrootstock: Done! Good catch Juli!

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
wilmerrootstock: We use this value in the `assertFeePerKbValue` method. It stands for the expected `currentFeePerKb`, and we compare it with `getFeePerKb()` to validate we are getting the correct current value.

About the position of this line of code, it is set in all the methods after the current Fee per Kb is set. Here corresponds after call `genesisFeePerKb()`, which initiates the current fee per KB value.

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
julia-zack: pretty sure you are not using the `currentFeePerKb` value in the assertion...
You are passing the `genesisFeePerKb` [as an input](https://github.com/rsksmart/rskj/pull/2433/files/f4e06f289f974c733e66c139fcabee30e240c30f#diff-72f83e7f5b9745bce799c9de61da92d8ffb07cd4646ef1da1758d11431208939R45) and then comparing it to the one [being obtained from the `getFeePerKb`](https://github.com/rsksmart/rskj/blob/f4e06f289f974c733e66c139fcabee30e240c30f/rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java#L483) method.

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
wilmerrootstock: Please let me know if the following method name explains better the sense of the test:

`voteFeePerKbChange_whenMajorityOfAuthorizersVoteTheSameValueIsAlreadySetAsCurrentFeePerKbValue_shouldSetValueVotedByMajority`

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
julia-zack: after asserting that the value obtained from the method is the genesis one, it makes sense to set
`currentFeePerKb = genesisFeePerKb`.

wilmerrootstock: Yes, that's correct! In each method, you can see the `assertFeePerKbValue` method is used a couple of times or more with different inputs such as `currentFeePerKb` or the value voted by the majority of authorizers. 

**Context:** 
When you use it the first time, we want to ensure the current fee per kb doesn't change its value because it doesn't have to change in the first moment (for this case it doesn't apply since the first voting doesn't happen yet. In this first case, it is used to guarantee it has the value of the previous voting round, for this reason, we used `currentFeePerKb` as an input parameter for this case). On the other hand, in the case when the `assertFeePerKbValue` method is used at the end of methods, we want to ensure the current fee per kb has changed correctly and it has the value voted by the majority of the authorizers. 
In this case, we assume `genesisFeePerKb` is a "kind of value voted", and we use the last approach explained.

wilmerrootstock: About changing the position of the line indicated, I wouldn't put at the end of the test that line. I'd choose to use assertions to keep using the `AAA` pattern in the tests.

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
wilmerrootstock: Done! It was renamed in this way:

`voteFeePerKbChange_whenVotingSameCurrentFeePerKbValue_shouldSetValueVoted`

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
wilmerrootstock: Please let me know if this test name is more readable and easier to get:

`voteFeePerKbChange_whenAuthorizerOneAndThreeVoteSameValueThenInBetweenAuthorizerTwoAndUnauthorizedCallerVoteSame_shouldSetValueVotedByOneAndThree`

rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java
julia-zack: sounds good! But i think you should remove [this](https://github.com/rsksmart/rskj/blob/f4e06f289f974c733e66c139fcabee30e240c30f/rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbIT.java#L468) part then

lucasvuotto:
pipeline: run

marcos-iov (approved):

nathanieliov (approved):

marcos-iov:
pipeline:run

PR #2439, opened 2024-06-03
Fix getPendingFederationHash bridge's method to avoid npe 

julia-zack:
In the federation support refactor, the bridge support's method `getPendingFederationHash` [was modified](https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java#L1996) to return a `Keccak256 hash` [instead of](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java#L2348) `bytes`, and therefore, the same bridge's method [was modified](https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/Bridge.java#L978) to get the bytes from the hash [instead of](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/Bridge.java#L977) getting the bytes directly.

This pr refactors the bridge method to avoid a npe when pending federation hash returns null.
Also, the method's name have been renamed to be more accurate

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/Bridge.java
+++ b/rskj-core/src/main/java/co/rsk/peg/Bridge.java
@@ -972,7 +972,7 @@ public class Bridge extends PrecompiledContracts.PrecompiledContract {
         );
     }
 
-    public byte[] getPendingFederationHashBytes(Object[] args) {
+    public byte[] getPendingFederationHashSerialized(Object[] args) {
         logger.trace("getPendingFederationHash");
 
         Keccak256 hash = bridgeSupport.getPendingFederationHash();
--- a/rskj-core/src/main/java/co/rsk/peg/Bridge.java
+++ b/rskj-core/src/main/java/co/rsk/peg/Bridge.java
@@ -972,17 +972,17 @@ public class Bridge extends PrecompiledContracts.PrecompiledContract {
         );
     }
 
-    public byte[] getPendingFederationHash(Object[] args) {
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -972,17 +972,17 @@ public Integer rollbackFederation(Object[] args) throws BridgeIllegalArgumentExc
         );
     }
 
-    public byte[] getPendingFederationHash(Object[] args) {
+    public byte[] getPendingFederationHashBytes(Object[] args) {
```
nathanieliov, line 975:
```suggestion
    public byte[] getPendingFederationHashSerialized(Object[] args) {
```

marcos-iov: Good one, we shouldn't use data types or structures as part of the names

julia-zack: on it! thanks guys

julia-zack: done

nathanieliov (approved):

marcos-iov (approved):

PR #2440, opened 2024-06-03
Fix PowpegMigrationTest

julia-zack:


rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java
```diff
@@ -123,18 +124,6 @@ private void testChangePowpeg(
...
-            .withRepository(repository)
-            .withEventLogger(bridgeEventLogger)
-            .withExecutionBlock(initialBlock)
-            .withActivations(activations)
-            .withBridgeConstants(bridgeConstants)
-            .withBtcBlockStoreFactory(btcBlockStoreFactory)
-            .withPeginInstructionsProvider(new PeginInstructionsProvider())
-            .withFeePerKbSupport(feePerKbSupport)
-            .build();
-
```
julia-zack, line 137: this was moved further down to create it after federationSupport is defined

rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java
```diff
@@ -198,8 +187,8 @@ private void testChangePowpeg(
         federationStorageProvider.save(btcParams, activations);
 
         // Proceed with the powpeg change
-        FederationSupportImpl federationSupport = new FederationSupportImpl(bridgeConstants.getFederationConstants(), federationStorageProvider, initialBlock, activations);
-        federationSupport.commitFederation(false, pendingFederation.getHash(), bridgeEventLogger);
+        FederationSupportImpl federationSupportImpl = new FederationSupportImpl(bridgeConstants.getFederationConstants(), federationStorageProvider, initialBlock, activations);
+        federationSupportImpl.commitFederation(false, pendingFederation.getHash(), bridgeEventLogger);
```
julia-zack, line 191: here we need to create a federation support implementation to can call the commitFederation method (that does not exist in the interface)

julia-zack (commented):
It is needed to create a new federation support instance right before creating a new bridge support instance, to have all the variables updated

wilmerrootstock (approved):
LGTM!

rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java
nathanieliov: Why this method is not included as part of the interface? Isn't the `commitFederation` an operation done by a `federationSupport` type?

julia-zack: the `commitFederation` method is an [internal operation](https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java#L488) done after voting the commit function to be performed by the bridge

julia-zack: is the only one that is `protected` (instead of `private`) so we can test it with less complexity

jeremy-then (approved):
LGTM.

nathanieliov (approved):

apancorb (approved):

marcos-iov (approved):

PR #2443, opened 2024-06-04
Fix BridgeStorageProviderTest

julia-zack:


rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -318,7 +323,7 @@ void getNewFederation_initialVersion() {
             RskAddress contractAddress = invocation.getArgument(0);
             DataWord address = invocation.getArgument(1);
 
-            assertArrayEquals(Hex.decode("aabbccdd"), contractAddress.getBytes());
+            assertArrayEquals(bridgeAddressBytes, contractAddress.getBytes());
```
julia-zack, line 326: before, the bridge storage provider was being created [with a mock "aabbccdd" address](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java#L321). 
But now we are using the federation storage provider, that is always created with the bridge address. That's why now the assertion should check if the contractAddress is the real bridge address.

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -323,8 +323,8 @@ private void saveNewFederation(ActivationConfig.ForBlock activations) {
             return;
         }
 
-        bridgeStorageAccessor.safeSaveToRepository(NEW_FEDERATION_KEY.getKey(), newFederation, BridgeSerializationUtils::serializeFederation);
         saveFederationFormatVersion(NEW_FEDERATION_FORMAT_VERSION.getKey(), newFederation.getFormatVersion());
+        bridgeStorageAccessor.safeSaveToRepository(NEW_FEDERATION_KEY.getKey(), newFederation, BridgeSerializationUtils::serializeFederation);
```
julia-zack, line 327: with the refactor, these keys were being stored upside down [as they were](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java#L358), causing some tests to fail.

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -98,9 +98,12 @@ class BridgeStorageProviderTest {
     private final ActivationConfig.ForBlock activationsAllForks = ActivationConfigsForTest.all().forBlock(0);
     private final BridgeTestNetConstants bridgeTestnetInstance = BridgeTestNetConstants.getInstance();
     private final FederationConstants federationTestnetConstants = bridgeTestnetInstance.getFederationConstants();
-    private final NetworkParameters testnetBtcParams = bridgeTestnetInstance.getBtcParams();
+    private final NetworkParameters testnetBtcParams = NetworkParameters.fromID(NetworkParameters.ID_TESTNET);
 
     private final RskAddress bridgeAddress = PrecompiledContracts.BRIDGE_ADDR;
+    private final byte[] bridgeAddressBytes = new byte[]{(byte) 0x00, (byte) 0x00, (byte) 0x00, (byte) 0x00,
+        (byte) 0x00, (byte) 0x00, (byte) 0x00, (byte) 0x00, (byte) 0x00, (byte) 0x00, (byte) 0x00, (byte) 0x00,
+        (byte) 0x00, (byte) 0x00, (byte) 0x00, (byte) 0x00, (byte) 0x01, (byte) 0x00, (byte) 0x00, (byte) 0x06};
```
jeremy-then, line 106: Hard to read. You can try:

```Java
byte[] bridgeAddressBytes = new byte[20];
        bridgeAddressBytes2[16] = 1;
        bridgeAddressBytes2[19] = 6;
```

Or:

```Java
byte[] bridgeAddressBytes = new byte[]{ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 6 };
```

marcos-iov: Or `bridgeAddress.getBytes()`?

jeremy-then: Or:

```Java
Hex.decode("0000000000000000000000000000000001000006")
```

jeremy-then: Or:

```Java
PrecompiledContracts.BRIDGE_ADDR.getBytes();
```

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -1362,18 +1365,18 @@ void saveLockWhitelist() {
...
-                        return null;
-                    })
-                    .when(repositoryMock).addStorageBytes(any(RskAddress.class), eq(LOCK_ONE_OFF_WHITELIST_KEY.getKey()), any(byte[].class));
+                .doAnswer((InvocationOnMock invocation) -> {
+                    storageBytesCalls.add(0);
+                    RskAddress contractAddress = invocation.getArgument(0);
+                    DataWord address = invocation.getArgument(1);
+                    byte[] data = invocation.getArgument(2);
+                    // Make sure the bytes are set to the correct address in the repo and that what's saved is what was serialized
+                    assertArrayEquals(Hex.decode("aabbccdd"), contractAddress.getBytes());
```
jeremy-then, line 1374: I thought we were going to change all these instances of `"aabbccdd"`, `"ccdd"`, etc.

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -1436,11 +1439,11 @@ void saveLockWhiteListAfterGetWithData() {
...
-                    .doAnswer((InvocationOnMock invocation) -> {
-                        storageCalled.set(Boolean.TRUE);
-                        return null;
-                    })
-                    .when(repositoryMock).addStorageBytes(any(RskAddress.class), eq(LOCK_ONE_OFF_WHITELIST_KEY.getKey()), eq(new byte[]{(byte) 0xee}));
+                .doAnswer((InvocationOnMock invocation) -> {
+                    storageCalled.set(Boolean.TRUE);
+                    return null;
+                })
+                .when(repositoryMock).addStorageBytes(any(RskAddress.class), eq(LOCK_ONE_OFF_WHITELIST_KEY.getKey()), eq(new byte[]{(byte) 0xee}));
```
jeremy-then, line 1446: nit. If we can pass the actual values, it would be better and less error pronne, like `eq(actualRskAddress)`, if possible.

jeremy-then (commented):
Left a couple of comments.

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -357,14 +357,15 @@ private void savePendingFederation(ActivationConfig.ForBlock activations) {
...
-        bridgeStorageAccessor.saveToRepository(PENDING_FEDERATION_KEY.getKey(), serializedPendingFederation);
 
         if (!activations.isActive(RSKIP123)) {
+            bridgeStorageAccessor.saveToRepository(PENDING_FEDERATION_KEY.getKey(), serializedPendingFederation);
             return;
         }
 
         // we only need to save the standard part of the fed since the emergency part is constant
         saveFederationFormatVersion(PENDING_FEDERATION_FORMAT_VERSION.getKey(), STANDARD_MULTISIG_FEDERATION.getFormatVersion());
+        bridgeStorageAccessor.saveToRepository(PENDING_FEDERATION_KEY.getKey(), serializedPendingFederation);
```
marcos-iov, line 368: This turned out a bit confusing, same call to saveToRepository repeated twice. The old way was clearer

```
if (activations.isActive(RSKIP123)) {
  saveFederationFormatVersion(PENDING_FEDERATION_FORMAT_VERSION.getKey(), STANDARD_MULTISIG_FEDERATION.getFormatVersion());
}

bridgeStorageAccessor.saveToRepository(PENDING_FEDERATION_KEY.getKey(), serializedPendingFederation);
```

By the way, `saveToRepository` should actually be `safeSaveToRepository` probably. There's no `saveToRepository` defined in StorageAcessor

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
marcos-iov: Why compare bytes? You could simply compare RskAddress objects

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
julia-zack: not all of them, the ones that are still using the bridge storage provider are defined with that mocked address. The change was made for the ones that now use federation storage provider

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
julia-zack: all good ones guys haha, thanks

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
julia-zack: i agree, but thats out of scope

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julia-zack: mmmm, i had in mind the discussion we had a couple of days ago about returning earlier when possible.


about the `saveToRepository` thing, you right, i messed up things trying to solve the error that appeared

marcos-iov: Yeah, but in this case I don't think it helps. Having the call repeated in 2 different flows is error prone. If you ever need to change it, you risk only changing it in one place

rskj-core/src/main/java/co/rsk/peg/storage/BridgeStorageAccessorImpl.java
```diff
@@ -50,7 +50,7 @@ private <T> byte[] getSerializedData(T object, RepositorySerializer<T> serialize
     }
 
     @Override
-    public void saveToRepository(DataWord addressKey, byte[] serializedData) {
+    public void safeSaveToRepository(DataWord addressKey, byte[] serializedData) {
```
marcos-iov, line 53: Any reason this one is not typed in the interface?

rskj-core/src/main/java/co/rsk/peg/storage/BridgeStorageAccessorImpl.java
```diff
@@ -50,7 +50,7 @@ private <T> byte[] getSerializedData(T object, RepositorySerializer<T> serialize
...
         repository.addStorageBytes(CONTRACT_ADDRESS, addressKey, serializedData);
     }
-}
+}
```
marcos-iov, line 56: Blank line :)

rskj-core/src/main/java/co/rsk/peg/storage/BridgeStorageAccessorImpl.java
julia-zack: it is https://github.com/rsksmart/rskj/blob/fix/test/bridge_storage_provider_test/rskj-core/src/main/java/co/rsk/peg/storage/StorageAccessor.java#L11

rskj-core/src/main/java/co/rsk/peg/storage/BridgeStorageAccessorImpl.java
julia-zack: done, thanks

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julia-zack: ok, it makes sense
Actually we should have something like 

```
bridgeStorageAccessor.safeSaveToRepository(PENDING_FEDERATION_KEY.getKey(), 
serializedPendingFederation);

if (!activations.isActive(RSKIP123)) {
       return;
}

// we only need to save the standard part of the fed since the emergency part is constant
saveFederationFormatVersion(PENDING_FEDERATION_FORMAT_VERSION.getKey(), STANDARD_MULTISIG_FEDERATION.getFormatVersion());
```
but that order (saving first the serialized federation and then the format version) was causing a test error

rskj-core/src/main/java/co/rsk/peg/storage/BridgeStorageAccessorImpl.java
julia-zack: To use the first method, you have to pass a serialization method as an input.
The idea is that, eventually, each object knows how to serialize itself (this is already happening in the [PendingFederation](https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/federation/PendingFederation.java#L172) class).

When that happens for all of the objects, we would no longer need to use the first method [but this one](https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java#L360), since we will be able to have the serialized data.

marcos-iov (approved):

PR #2445, opened 2024-06-05
Whitelist Constants

wilmerrootstock:
## Description
- Following the Bridge refactors tasks, it is time to refactor and remove whitelisting logic from the BridgeConstants class.

## Motivation and Context
This is related to the Bridge Refactor to decouple some objects that are tied closely to the Bridge.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/constants/BridgeTestNetConstants.java
+++ b/rskj-core/src/main/java/co/rsk/peg/constants/BridgeTestNetConstants.java
@@ -22,6 +22,7 @@ import co.rsk.bitcoinj.core.Coin;
 import co.rsk.bitcoinj.core.NetworkParameters;
 import co.rsk.peg.federation.constants.FederationTestNetConstants;
 import co.rsk.peg.vote.AddressBasedAuthorizer;
+import co.rsk.peg.whitelist.constants.WhitelistTestNetConstants;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
@@ -36,6 +37,8 @@ public class BridgeTestNetConstants extends BridgeConstants {
         btcParamsString = NetworkParameters.ID_TESTNET;
         feePerKbConstants = FeePerKbTestNetConstants.getInstance();
         federationConstants = FederationTestNetConstants.getInstance();
+        whitelistConstants = WhitelistTestNetConstants.getInstance();
+
 
         btc2RskMinimumAcceptableConfirmations = 10;
         btc2RskMinimumAcceptableConfirmationsOnRsk = 10;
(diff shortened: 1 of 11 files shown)


rskj-core/src/main/java/co/rsk/peg/whitelist/constants/WhitelistConstants.java
```diff
@@ -0,0 +1,12 @@
+package co.rsk.peg.whitelist.constants;
+
+import co.rsk.peg.vote.AddressBasedAuthorizer;
+
+public class WhitelistConstants {
+
+    protected AddressBasedAuthorizer lockWhitelistChangeAuthorizer;
```
apancorb, line 7: What about making this field private, and providing a constructor where each sub-class call it using `super(AddressBasedAuthorizer)`

rskj-core/src/main/java/co/rsk/peg/whitelist/constants/WhitelistConstants.java
```diff
@@ -0,0 +1,12 @@
+package co.rsk.peg.whitelist.constants;
+
+import co.rsk.peg.vote.AddressBasedAuthorizer;
+
+public class WhitelistConstants {
```
apancorb, line 5: I see that the intention here is to have this class be extended by a mainnet, testnet, etc. Would it make sense to have this class as an abstract class?

rskj-core/src/main/java/co/rsk/peg/whitelist/constants/WhitelistConstants.java
marcos-iov: I'm curious about this @apancorb. What advantage do you see in doing it that way?

marcos-iov (changes requested):
Looks like we are missing tests for WhitelistConstants @wilmerrootstock

rskj-core/src/main/java/co/rsk/peg/constants/BridgeTestNetConstants.java
```diff
@@ -36,6 +37,8 @@ public class BridgeTestNetConstants extends BridgeConstants {
         btcParamsString = NetworkParameters.ID_TESTNET;
         feePerKbConstants = FeePerKbTestNetConstants.getInstance();
         federationConstants = FederationTestNetConstants.getInstance();
+        whitelistConstants = WhitelistTestNetConstants.getInstance();
+
```
julia-zack, line 41: nit: remove extra blank line

rskj-core/src/main/java/co/rsk/peg/whitelist/constants/WhitelistConstants.java
apancorb: Since `WhitelistConstants` is declaring this instance variable it should also have the responsibility of initializing it. IMO, the advantage is that it provides better encapsulation. You can still get the reference by the getter present in the class, if needed by a sub-class.

nathanieliov (approved):
LGTM

rskj-core/src/main/java/co/rsk/peg/whitelist/constants/WhitelistConstants.java
marcos-iov: As discussed, let's keep it as it is for the time being. For consistency with the rest of the codebase. We can figure out a better way to access all the Constants later.

rskj-core/src/main/java/co/rsk/peg/whitelist/constants/WhitelistConstants.java
wilmerrootstock: @apancorb, Amazing comment! Indeed, it shouldn't be an abstract class. It should be an Interface. Some weeks ago, I was discussing this topic with @marcos-iov  since in these cases we should apply a behavioral design pattern, Strategy! It would be interesting to discuss this at a technical meeting. WDYT, @marcos-iov ?

wilmerrootstock:
> Looks like we are missing tests for WhitelistConstants @wilmerrootstock

I believe I misunderstood. I was thinking we are going to solve the issues related to UTs in a different story to create smaller PRs. Well, let me work on it.

rskj-core/src/main/java/co/rsk/peg/whitelist/constants/WhitelistConstants.java
wilmerrootstock: I know that we are going to keep it as it is, however, I want to share my perspective on this case for future refactors.
IMO, if we apply a Strategy design pattern, we can create a private variable on the implementation, and give access only to the getter method through the interface. Definitely, as @apancorb said, it provides a better encapsulation.

rskj-core/src/main/java/co/rsk/peg/constants/BridgeTestNetConstants.java
wilmerrootstock: Done!

marcos-iov (approved):

PR #2447, opened 2024-06-05
Whitelist Storage Provider

wilmerrootstock:
## Description
Create an interface WhitelistStorageProvider under co.rsk.peg.whitelist package. Include the signature of the 2 methods Whitelist-related from BridgeStorageProvider.

## Motivation and Context
This is related to the Bridge Refactor to decouple some objects that are tied closely to the Bridge.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistStorageProvider.java
+++ b/rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistStorageProvider.java
@@ -1,8 +1,11 @@
 package co.rsk.peg.whitelist;
 
+/**
+ * Interface for storage access for whitelisting.
+ */
 public interface WhitelistStorageProvider {
 
-    void saveLockWhitelist();
+    void save();
 
     LockWhitelist getLockWhitelist();
 }
--- a/rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistStorageProvider.java
+++ b/rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistStorageProvider.java
@@ -1,0 +1,8 @@
+package co.rsk.peg.whitelist;
+
+public interface WhitelistStorageProvider {
+
+    void saveLockWhitelist();
+
+    LockWhitelist getLockWhitelist();
+}


rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistStorageProvider.java
```diff
@@ -0,0 +1,8 @@
+package co.rsk.peg.whitelist;
+
+public interface WhitelistStorageProvider {
+
+    void saveLockWhitelist();
```
apancorb, line 5: If possible, providing javadoc would be great :)

apancorb (approved):
LGTM

rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistStorageProvider.java
```diff
@@ -0,0 +1,8 @@
...
+
+    LockWhitelist getLockWhitelist();
+}
```
marcos-iov, line 8: We are also gonna need a new method `void save();`

julia-zack: actually, the `saveLockWhitelist()` method shouldn't be part of the interface.

All the [specific `save...()` methods are private](https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java#L299), and just part of the implementation.
The only one that is also part of the interface is the [`save()`](https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProvider.java#L40) one, since is the one that would be [called from the supports](https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java#L775)

nathanieliov: @julia-zack really good point.  Specific save methods should be private and only be called by the `save` method.

marcos-iov: Good catch

wilmerrootstock: Hey, guys! @julia-zack @nathanieliov @marcos-iov 
I was looking into BridgeStorageProvider, and I noticed that `void save()` method involves other processes apart from whitelisting. 
Should we really expose this `save()` method in `WhitelistStorageProvider`?

image

julia-zack: I don't think I follow the question.
The `saveLockWhiteList()` method should be removed from the `BridgeStorageProvider`, since we now have specific providers. In this situation, the `WhiteListStorageProvider`.

As I said, the `WhiteListStorageProvider` interface should just have the `save()` method. The `WhiteListStorageProviderImpl` will override that method, and call the private saving methods.

See the links i provided above, showing how this was solved in the FederationSupport refactor.

rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistStorageProvider.java
wilmerrootstock: Done!

rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistStorageProvider.java
wilmerrootstock: Done!

julia-zack (approved):

marcos-iov (approved):

PR #2449, opened 2024-06-05
Fix storage access thrown exceptions

julia-zack:
Fix the errors between `StorageAccessor` and its implementations that were causing the project not to compile.
Create new `SerializationException` to be thrown when needed.
Remove _IOException_ thrown from methods related to repository access. Remove no more needed `StorageAccessException`.


rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
```diff
@@ -132,11 +133,15 @@ public static List<UTXO> deserializeUTXOList(byte[] data) throws IOException {
 
         int nutxos = rlpList.size();
 
-        for (int k = 0; k < nutxos; k++) {
-            byte[] utxoBytes = rlpList.get(k).getRLPData();
-            InputStream istream = new ByteArrayInputStream(utxoBytes);
-            UTXO utxo = new UTXO(istream);
-            list.add(utxo);
+        try {
```
wilmerrootstock, line 136: Just in case: this way of handling this exception breaks the loop and does not keep on processing the following if something is wrong with one element. is it ok?

wilmerrootstock (commented):
God job! Just left one question (it's not a request)

rskj-core/src/main/java/co/rsk/peg/storage/BridgeStorageAccessorImpl.java
```diff
@@ -18,29 +16,21 @@ public BridgeStorageAccessorImpl(Repository repository) {
 
     @Override
     public <T> T safeGetFromRepository(DataWord keyAddress, RepositoryDeserializer<T> deserializer) {
```
nathanieliov, line 18: Since we are no longer handling potential exceptions, I see no sense in continuing to use the prefix "safe" in all these refactored methods. Wdyt?

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
```diff
@@ -105,23 +102,27 @@ public static SortedMap<Keccak256, BtcTransaction> deserializeMap(byte[] data, N
...
+            for (UTXO utxo : list) {
+                try (ByteArrayOutputStream ostream = new ByteArrayOutputStream()) {
+                    utxo.serializeToStream(ostream);
+                    bytes[n++] = RLP.encodeElement(ostream.toByteArray());
+                }
             }
+        } catch (IOException ioe) {
+            throw new SerializationException("Unable to serialize UTXO list ", ioe);
         }
```
nathanieliov, line 120: What is the reason behind using double `try` here? Moving the catch to the inner try seems to be enough. No need for the wrapper try.

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
```diff
@@ -105,23 +102,27 @@ public static SortedMap<Keccak256, BtcTransaction> deserializeMap(byte[] data, N
...
-                bytes[n++] = RLP.encodeElement(ostream.toByteArray());
+        try {
+            for (UTXO utxo : list) {
+                try (ByteArrayOutputStream ostream = new ByteArrayOutputStream()) {
+                    utxo.serializeToStream(ostream);
+                    bytes[n++] = RLP.encodeElement(ostream.toByteArray());
+                }
             }
+        } catch (IOException ioe) {
+            throw new SerializationException("Unable to serialize UTXO list ", ioe);
```
nathanieliov, line 119: The utxo list as string can be part of the exception message as well. Wdyt?

rskj-core/src/main/java/co/rsk/peg/SerializationException.java
```diff
@@ -0,0 +1,16 @@
...
+
+
+public class SerializationException extends RuntimeException {
+    public SerializationException() {
+        super();
+    }
+
+    public SerializationException(String message) {
+        super(message);
+    }
```
nathanieliov, line 11: Do we really need these constructors? I think for logging consistency we should always use the one below.

nathanieliov (commented):
Well done. Left some comments.

rskj-core/src/main/java/co/rsk/peg/storage/BridgeStorageAccessorImpl.java
julia-zack: very good point actually

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
julia-zack: good idea! Thanks

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
julia-zack: good point. Actually we should discuss what wilmer [pointed out](https://github.com/rsksmart/rskj/pull/2449#discussion_r1628045542). 
Do we want the method to serialize the utxos it can? Or should we stop the process if there's any we can't serialize?
I think the second approach is the correct one. If we want to de/serialize a list of utxos, we shouldn't be satisfied with the de/serialization of only _some_ of them

julia-zack: you right, thanks natha

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
julia-zack: wasn't it like that before as well?

wilmerrootstock: You're right! Only I wanted to double-check. Thanks, @julia-zack !

wilmerrootstock (approved):
LGTM!

rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java
julia-zack: thanks to you :)

rskj-core/src/main/java/co/rsk/peg/SerializationException.java
julia-zack: not sure, wdyt @marcos-iov ?

marcos-iov: True, if not needed then no reason to add it

rskj-core/src/main/java/co/rsk/peg/storage/BridgeStorageAccessorImpl.java
marcos-iov: Yeap, just call getFromRepository directly

rskj-core/src/main/java/co/rsk/peg/SerializationException.java
julia-zack: done

rskj-core/src/main/java/co/rsk/peg/storage/BridgeStorageAccessorImpl.java
julia-zack: done

jeremy-then (approved):
LGTM.

nathanieliov (approved):
LGTM

marcos-iov (approved):

PR #2451, opened 2024-06-06
Whitelist Support Interface

wilmerrootstock:
## Description
Create an interface called WhitelistSupport, and put all the method signatures Whitelist-related on it. Take these methods from BridgeSupport. The plan is to remove all of them from BridgeSupport into their own WhitelistSupport class.

## Motivation and Context

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupport.java
@@ -3,6 +3,9 @@ package co.rsk.peg.whitelist;
 import java.math.BigInteger;
 import org.ethereum.core.Transaction;
 
+/**
+ * Interface for the whitelist support. This interface is used to add, remove and get whitelist entries.
+ */
 public interface WhitelistSupport {
 
     Integer getLockWhitelistSize();
--- a/rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupport.java
@@ -1,0 +1,21 @@
+package co.rsk.peg.whitelist;
+
+import java.math.BigInteger;
+import org.ethereum.core.Transaction;
+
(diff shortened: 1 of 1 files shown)


rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupport.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.math.BigInteger;
+import org.ethereum.core.Transaction;
+
+/**
+ * Interface for the whitelist support. This interface is used to add, remove and get whitelist entries.
+ */
+public interface WhitelistSupport {
+
+    Integer getLockWhitelistSize();
```
apancorb, line 11: Can the methods that return a boxed primitive type be changed to just a primitive instead? This way we avoid doing null checks.

rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupport.java
```diff
@@ -0,0 +1,24 @@
...
+
+    LockWhitelistEntry getLockWhitelistEntryByIndex(int index);
+
+    LockWhitelistEntry getLockWhitelistEntryByAddress(String addressBase58);
```
apancorb, line 15: Will these get methods return a `null` reference in the case of not being able to retrieve the `LockWhitelistEntry`? What about returning an `Optional<LockWhitelistEntry>` and if the get fails return an `Optional.empty()`?

marcos-iov: That's a good point, I'll create a separate task to do that change

rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupport.java
marcos-iov: Agree

marcos-iov (approved):

PR #2482, opened 2024-06-11
Add getActiveFederatorBtcPublicKey test

julia-zack:


rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -158,6 +158,26 @@ void getActiveFederationCreationBlockNumber_returnsGenesisFederationCreationBloc
...
+        @Tag("getActiveFederatorBtcPublicKey")
+        void getActiveFederatorBtcPublicKey_withNegativeIndex_throwsIndexOutOfBoundsException() {
+            assertThrows(IndexOutOfBoundsException.class, () -> federationSupport.getActiveFederatorBtcPublicKey(-1));
+        }
+
+        @Test
+        @Tag("getActiveFederatorBtcPublicKey")
+        void getActiveFederatorBtcPublicKey_withIndexGreaterThanActiveFederationSize_throwsIndexOutOfBoundsException() {
+            int activeFederationSize = genesisFederation.getSize();
+            assertThrows(IndexOutOfBoundsException.class, () -> federationSupport.getActiveFederatorBtcPublicKey(activeFederationSize));
```
julia-zack, line 179: not sure if these two tests should be here, since the check we are doing is not related to have two null federations.

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -598,6 +631,38 @@ private Stream<Arguments> expectedCreationBlockNumberArgs() {
...
+            BtcECKey federatorFromOldFederationBtcPublicKey = oldFederation.getBtcPublicKeys().get(0);
+            BtcECKey federatorFromNewFederationBtcPublicKey = newFederation.getBtcPublicKeys().get(0);
+
+            return Stream.of(
+                Arguments.of(blockNumberFederationActivationHop - 1, hopActivations, federatorFromOldFederationBtcPublicKey.getPubKey()),
+                Arguments.of(blockNumberFederationActivationHop, hopActivations, federatorFromNewFederationBtcPublicKey.getPubKey()),
+                Arguments.of(blockNumberFederationActivationHop, fingerrootActivations, federatorFromOldFederationBtcPublicKey.getPubKey()),
+                Arguments.of(blockNumberFederationActivationFingerroot - 1, fingerrootActivations, federatorFromOldFederationBtcPublicKey.getPubKey()),
+                Arguments.of(blockNumberFederationActivationFingerroot, fingerrootActivations, federatorFromNewFederationBtcPublicKey.getPubKey()),
+                Arguments.of(blockNumberFederationActivationFingerroot, hopActivations, federatorFromNewFederationBtcPublicKey.getPubKey())
```
jeremy-then, line 663: I would rather pass the `BtcECKey` instead of the `byte[]`, since a `byte[]` can be anything, but a `BtcECKey` can only be a `BtcECKey`.
Then, inside of the test, you do `btcEcKey.getPubKey()`.

julia-zack: yep, good catch. Thanks jere!

PR #2330, opened 2024-05-15
Create federation storage provider implementation

julia-zack:
Create FederationStorageProviderImpl to implement the interface.

PR #2490, opened 2024-06-13
Whitelist support creation

marcos-iov:
- Use WhitelistResponseCode in Bridge class
- Add network parameters to WhitelistConstants. Remove network parameters from WhitelistSupport, get the value from WhitelistConstants
- Update BridgeSupportFactory to build WhitelistSupport
- Include WhitelistConstants tests

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/Bridge.java
+++ b/rskj-core/src/main/java/co/rsk/peg/Bridge.java
@@ -36,6 +36,7 @@ import co.rsk.peg.flyover.FlyoverTxResponseCodes;
 import co.rsk.peg.utils.BtcTransactionFormatUtils;
 import co.rsk.peg.whitelist.LockWhitelistEntry;
 import co.rsk.peg.whitelist.OneOffWhiteListEntry;
+import co.rsk.peg.whitelist.WhitelistResponseCode;
 import co.rsk.rpc.modules.trace.ProgramSubtrace;
 import com.google.common.annotations.VisibleForTesting;
 import org.ethereum.config.Constants;
@@ -205,10 +206,6 @@ public class Bridge extends PrecompiledContracts.PrecompiledContract {
     public static final CallTransaction.Function GET_BTC_BLOCKCHAIN_PARENT_BLOCK_HEADER_BY_HASH = BridgeMethods.GET_BTC_BLOCKCHAIN_PARENT_BLOCK_HEADER_BY_HASH.getFunction();
 
     public static final CallTransaction.Function GET_ACTIVE_POWPEG_REDEEM_SCRIPT = BridgeMethods.GET_ACTIVE_POWPEG_REDEEM_SCRIPT.getFunction();
-
-    public static final int LOCK_WHITELIST_UNLIMITED_MODE_CODE = 0;
-    public static final int LOCK_WHITELIST_ENTRY_NOT_FOUND_CODE = -1;
-    public static final int LOCK_WHITELIST_INVALID_ADDRESS_FORMAT_ERROR_CODE = -2;
 
     // Log topics used by Bridge Contract pre RSKIP146
     public static final DataWord RELEASE_BTC_TOPIC = DataWord.fromString("release_btc_topic");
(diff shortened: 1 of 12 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeSupportFactory.java
```diff
@@ -93,25 +104,42 @@ public BridgeSupport newInstance(Repository repository, Block executionBlock,
...
+
         return new FeePerKbSupportImpl(feePerKbConstants, feePerKbStorageProvider);
     }
+
+    private WhitelistSupport getWhitelistSupportInstance(StorageAccessor bridgeStorageAccessor, ActivationConfig.ForBlock activations) {
+        WhitelistConstants whitelistConstants = bridgeConstants.getWhitelistConstants();
+        WhitelistStorageProvider whitelistStorageProvider = new WhitelistStorageProviderImpl(
+            whitelistConstants.getBtcParams(),
+            activations,
+            bridgeStorageAccessor
```
julia-zack, line 136: i just noticed something when seeing this.
When doing the FederationSupport refactor, we had [this](https://github.com/rsksmart/rskj/pull/2330#discussion_r1603951907) discussion about the activations and network parameters being part of the class.

This situation is different because the _activations_ are not part of the `WhitelistSupportImpl`, but the _networkParameters_ are. So probably they should be removed from the `WhitelistStorageProvider` class.

About the activations, I think we should take advantage of the fact that the `BridgeSupport` already has an _activations_ field, and create the supports with it.
That way, all the storage providers would only need the bridge storage accessor to be created, standardizing it.
I find this approach cleaner.

wdyt guys? @marcos-iov @nathanieliov @jeremy-then @apancorb

rskj-core/src/main/java/co/rsk/peg/whitelist/constants/WhitelistConstants.java
```diff
@@ -1,11 +1,16 @@
 package co.rsk.peg.whitelist.constants;
 
+import co.rsk.bitcoinj.core.NetworkParameters;
 import co.rsk.peg.vote.AddressBasedAuthorizer;
 
 public class WhitelistConstants {
-
+    protected NetworkParameters btcParams;
```
wilmerrootstock, line 7: Is it necessary to include `btcParams` as part of the Whitelist objects?  I noticed this attribute is already in BridgeConstants, and putting it in other places could affect the data modeling. If we see that data model to get a good normalization we have to avoid spreading the same information/attribute in different objects.

rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -1055,20 +1052,20 @@ public long getLockWhitelistEntryByAddress(Object[] args) {
         try {
             addressBase58 = (String) args[0];
         } catch (Exception e) {
-            logger.warn("Exception in getLockWhitelistEntryByAddress", e);
-            return LOCK_WHITELIST_INVALID_ADDRESS_FORMAT_ERROR_CODE;
+            logger.warn("[getLockWhitelistEntryByAddress] Exception in getLockWhitelistEntryByAddress", e);
```
wilmerrootstock, line 1055: Awesome! Looks much better with the pattern: logger.type("[method]"...). 
WDYT if we add a variable with some information and remove "Exception in getLockWhitelistEntryByAddress" (it seems redundant with the name of the method):

`logger.warn("[removeLockWhitelistAddress] {}", e.getMessage(), e);`

rskj-core/src/main/java/co/rsk/peg/whitelist/constants/WhitelistConstants.java
marcos-iov: Makes life easier :) I can get the network params from the constants instead of passing the NetowrkParam object around.

I agree that it adds some duplication, but the value is also tied to the class name (MainnetConstants => MainnetParams). I'm open for discussing it, there's room for improvement in many thing s we've done so far definitely

rskj-core/src/main/java/co/rsk/peg/Bridge.java
marcos-iov: Done, thanks!

rskj-core/src/main/java/co/rsk/peg/BridgeSupportFactory.java
marcos-iov: Thanks for the idea! Just implemented it, let me know what you think

rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupportImpl.java
```diff
@@ -50,9 +58,9 @@ public LockWhitelistEntry getLockWhitelistEntryByIndex(int index) {
     @Override
     public LockWhitelistEntry getLockWhitelistEntryByAddress(String addressBase58) {
         try {
-            Address address = Address.fromBase58(networkParameters, addressBase58);
+            Address address = Address.fromBase58(whitelistConstants.getBtcParams(), addressBase58);
```
julia-zack, line 61: why this change?

rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupportImpl.java
```diff
@@ -62,7 +70,7 @@ public LockWhitelistEntry getLockWhitelistEntryByAddress(String addressBase58) {
     @Override
     public int addOneOffLockWhitelistAddress(Transaction tx, String addressBase58, BigInteger maxTransferValue) {
         try {
-            Address address = Address.fromBase58(networkParameters, addressBase58);
+            Address address = Address.fromBase58(whitelistConstants.getBtcParams(), addressBase58);
```
julia-zack, line 73: same here

rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupportImpl.java
```diff
@@ -76,7 +84,7 @@ public int addOneOffLockWhitelistAddress(Transaction tx, String addressBase58, B
     @Override
     public int addUnlimitedLockWhitelistAddress(Transaction tx, String addressBase58) {
         try {
-            Address address = Address.fromBase58(networkParameters, addressBase58);
+            Address address = Address.fromBase58(whitelistConstants.getBtcParams(), addressBase58);
```
julia-zack, line 87: and here

wilmerrootstock (approved):

PR #2463, opened 2024-06-10
Renames FederationStorageProvider derived and test classes. Moves get…

jeremy-then:
Renames FederationStorageProvider derived and test classes.

Moves getNewFederationBtcUTXOs tests from BridgeStorageProviderTest to FedreationStorageProviderImplTest.

Created new `getNewFederationBtcUTXOs_calledTwice_returnCachedUtxos` to increase coverage by and test a caching logic that was not being tested.

PR #2498, opened 2024-06-17
Whitelist Storage Provider Test

wilmerrootstock:
## Description
Create a new class WhitelistStorageProviderImplTest, look for test cases that use Whitelist related methods in BridgeStorageProviderTest, and move those tests to the new class.

## Motivation and Context
Get Coverage.

## How Has This Been Tested?
Unit Tests

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
+++ b/rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
@@ -29,6 +29,18 @@ public class BitcoinTestUtils {
         return keys;
     }
 
+    /**
+     * Returns a BTC address derived from the seed.
+     *
+     * @param seed              the seed to derive the address from
+     * @param networkParameters the network parameters
+     * @return the BTC address
+     */
+    public static Address getBtcAddress(String seed, NetworkParameters networkParameters) {
+        BtcECKey btcECKey = BtcECKey.fromPrivate(HashUtil.keccak256(seed.getBytes(StandardCharsets.UTF_8)));
+        return btcECKey.toAddress(networkParameters);
+    }
+
     public static Address createP2PKHAddress(NetworkParameters networkParameters, String seed) {
         BtcECKey key = BtcECKey.fromPrivate(
             HashUtil.keccak256(seed.getBytes(StandardCharsets.UTF_8)));
--- a/rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
+++ b/rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
@@ -29,18 +29,6 @@ public class BitcoinTestUtils {
         return keys;
     }
 
-    /**
-     * Returns a BTC address derived from the seed.
-     *
-     * @param seed              the seed to derive the address from
-     * @param networkParameters the network parameters
-     * @return the BTC address
-     */
-    public static Address getBtcAddress(String seed, NetworkParameters networkParameters) {
-        BtcECKey btcECKey = BtcECKey.fromPrivate(HashUtil.keccak256(seed.getBytes(StandardCharsets.UTF_8)));
-        return btcECKey.toAddress(networkParameters);
-    }
-
     public static Address createP2PKHAddress(NetworkParameters networkParameters, String seed) {
         BtcECKey key = BtcECKey.fromPrivate(
             HashUtil.keccak256(seed.getBytes(StandardCharsets.UTF_8)));
--- a/rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
@@ -1,256 +1,124 @@
 package co.rsk.peg.whitelist;
 
 import static co.rsk.peg.whitelist.WhitelistStorageIndexKey.LOCK_ONE_OFF;
-import static co.rsk.peg.whitelist.WhitelistStorageIndexKey.LOCK_UNLIMITED;
-import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.mockStatic;
 import static org.mockito.Mockito.when;
 
 import co.rsk.bitcoinj.core.Address;
+import co.rsk.bitcoinj.core.BtcECKey;
 import co.rsk.bitcoinj.core.Coin;
 import co.rsk.bitcoinj.core.NetworkParameters;
-import co.rsk.core.RskAddress;
 import co.rsk.peg.BridgeSerializationUtils;
 import co.rsk.peg.InMemoryStorage;
 import co.rsk.peg.storage.StorageAccessor;
 import co.rsk.peg.whitelist.constants.WhitelistConstants;
 import co.rsk.peg.whitelist.constants.WhitelistMainNetConstants;
-import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicReference;
 import org.apache.commons.lang3.tuple.Pair;
-import org.bouncycastle.util.encoders.Hex;
-import org.ethereum.TestUtils;
 import org.ethereum.config.blockchain.upgrades.ActivationConfig;
-import org.ethereum.core.Repository;
-import org.ethereum.vm.DataWord;
-import org.junit.jupiter.api.Assertions;
+import org.ethereum.config.blockchain.upgrades.ConsensusRule;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.mockito.MockedStatic;
-import org.mockito.Mockito;
-import org.mockito.invocation.InvocationOnMock;
 
 class WhitelistStorageProviderImplTest {
-
     private final WhitelistConstants whitelistConstants = WhitelistMainNetConstants.getInstance();
     private final NetworkParameters networkParameters = whitelistConstants.getBtcParams();
     private WhitelistStorageProvider whitelistStorageProvider;
+    private StorageAccessor inMemoryStorage;
     private ActivationConfig.ForBlock activationConfig;
 
     @BeforeEach
     void setUp() {
-        StorageAccessor inMemoryStorage = new InMemoryStorage();
+        inMemoryStorage = new InMemoryStorage();
         whitelistStorageProvider = new WhitelistStorageProviderImpl(inMemoryStorage);
         activationConfig = mock(ActivationConfig.ForBlock.class);
     }
 
     @Test
-    void getLockWhitelist_nonNullBytes() {
-        List<Integer> calls = new ArrayList<>();
-        LockWhitelist whitelistMock = new LockWhitelist(new HashMap<>());
-        LockWhitelistEntry oneOffEntry = new OneOffWhiteListEntry(getBtcAddress("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"), Coin.COIN);
-        LockWhitelistEntry unlimitedEntry = new UnlimitedWhiteListEntry(getBtcAddress("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"));
-        whitelistMock.put(oneOffEntry.address(), oneOffEntry);
-        whitelistMock.put(unlimitedEntry.address(), unlimitedEntry);
-        Repository repositoryMock = mock(Repository.class);
+    void save_whenLockWhitelistIsNull_shouldReturnZeroEntries() {
+        whitelistStorageProvider.save(activationConfig);
 
-        when(repositoryMock.getStorageBytes(any(RskAddress.class), eq(LOCK_ONE_OFF.getKey())))
-            .then((InvocationOnMock invocation) -> {
-                calls.add(0);
-                RskAddress contractAddress = invocation.getArgument(0);
-                DataWord address = invocation.getArgument(1);
-                // Make sure the bytes are got from the correct address in the repo
-                assertArrayEquals(new byte[]{(byte) 0xaa, (byte) 0xbb, (byte) 0xcc, (byte) 0xdd}, contractAddress.getBytes());
-                assertEquals(LOCK_ONE_OFF.getKey(), address);
-                return new byte[]{(byte) 0xaa};
-            });
-        when(repositoryMock.getStorageBytes(any(RskAddress.class), eq(LOCK_UNLIMITED.getKey())))
-            .then((InvocationOnMock invocation) -> {
-                calls.add(0);
-                RskAddress contractAddress = invocation.getArgument(0);
-                DataWord address = invocation.getArgument(1);
-                // Make sure the bytes are got from the correct address in the repo
-                assertArrayEquals(new byte[]{(byte) 0xaa, (byte) 0xbb, (byte) 0xcc, (byte) 0xdd}, contractAddress.getBytes());
-                assertEquals(LOCK_UNLIMITED.getKey(), address);
-                return new byte[]{(byte) 0xbb};
-            });
-        try (MockedStatic<BridgeSerializationUtils> bridgeSerializationUtilsMocked = mockStatic(BridgeSerializationUtils.class)) {
-            bridgeSerializationUtilsMocked
-                .when(() -> BridgeSerializationUtils.deserializeOneOffLockWhitelistAndDisableBlockHeight(any(byte[].class), any(NetworkParameters.class)))
-                .then((InvocationOnMock invocation) -> {
-                    calls.add(0);
-                    byte[] data = invocation.getArgument(0);
-                    NetworkParameters parameters = invocation.getArgument(1);
-                    assertEquals(NetworkParameters.fromID(NetworkParameters.ID_TESTNET), parameters);
-                    // Make sure we're deserializing what just came from the repo with the correct AddressBasedAuthorizer
-                    assertArrayEquals(new byte[]{(byte) 0xaa}, data);
-                    HashMap<Address, LockWhitelistEntry> map = new HashMap<>();
-                    map.put(oneOffEntry.address(), oneOffEntry);
-                    return Pair.of(map, 0);
-                });
-            bridgeSerializationUtilsMocked.when(() -> BridgeSerializationUtils.deserializeUnlimitedLockWhitelistEntries(any(byte[].class), any(NetworkParameters.class)))
-                .then((InvocationOnMock invocation) -> {
-                    calls.add(0);
-                    byte[] unlimitedData = invocation.getArgument(0);
-                    NetworkParameters parameters = invocation.getArgument(1);
-                    assertEquals(NetworkParameters.fromID(NetworkParameters.ID_TESTNET), parameters);
-                    // Make sure we're deserializing what just came from the repo with the correct AddressBasedAuthorizer
-                    assertArrayEquals(new byte[]{(byte) 0xbb}, unlimitedData);
-                    HashMap<Address, LockWhitelistEntry> map = new HashMap<>();
-                    map.put(unlimitedEntry.address(), unlimitedEntry);
-                    return map;
-                });
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
 
-            assertEquals(whitelistMock.getAll(), whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters).getAll());
-            assertEquals(4, calls.size()); // 1 for each call to deserializeFederationOnlyBtcKeys & getStorageBytes (we call getStorageBytes twice)
-        }
+        assertEquals(0, actualLockWhitelist.getAll().size());
     }
 
     @Test
-    void getLockWhitelist_nullBytes() {
-        List<Integer> calls = new ArrayList<>();
-        Repository repositoryMock = mock(Repository.class);
+    void save_whenLockWhiteListIsNotNull_ShouldReturnSavedValue() {
+        saveOneOffWhiteListEntry();
 
-        when(repositoryMock.getStorageBytes(any(RskAddress.class), any(DataWord.class)))
-            .then((InvocationOnMock invocation) -> {
-                calls.add(0);
-                RskAddress contractAddress = invocation.getArgument(0);
-                DataWord address = invocation.getArgument(1);
-                // Make sure the bytes are got from the correct address in the repo
-                assertArrayEquals(new byte[]{(byte) 0xaa, (byte) 0xbb, (byte) 0xcc, (byte) 0xdd}, contractAddress.getBytes());
-                assertEquals(LOCK_ONE_OFF.getKey(), address);
-                return new byte[]{(byte) 0xee};
-            });
-        try (MockedStatic<BridgeSerializationUtils> bridgeSerializationUtilsMocked = mockStatic(BridgeSerializationUtils.class)) {
-            bridgeSerializationUtilsMocked.when(() -> BridgeSerializationUtils.deserializeOneOffLockWhitelistAndDisableBlockHeight(any(byte[].class), any(NetworkParameters.class)))
-                .then((InvocationOnMock invocation) -> {
-                    calls.add(0);
-                    return null;
-                });
-            bridgeSerializationUtilsMocked.when(() -> BridgeSerializationUtils.deserializeUnlimitedLockWhitelistEntries(any(byte[].class), any(NetworkParameters.class)))
-                .then((InvocationOnMock invocation) -> {
-                    calls.add(0); // THIS ONE WON'T BE CALLED BECAUSE ONEOFF IS EMPTY
-                    Assertions.fail(
-                        "As we don't have data for one-off, we shouldn't have called deserialize unlimited");
-                    return null;
-                });
+        whitelistStorageProvider.save(activationConfig);
 
-            LockWhitelist result = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
-            Assertions.assertNotNull(result);
-            assertEquals(0, result.getSize().intValue());
-            assertEquals(2, calls.size()); // 1 for each call to deserializeFederationOnlyBtcKeys & getStorageBytes
-        }
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+        assertEquals(1, actualLockWhitelist.getAll().size());
     }
 
     @Test
-    void saveLockWhitelist() {
-        LockWhitelist whitelistMock = mock(LockWhitelist.class);
-        List<Integer> storageBytesCalls = new ArrayList<>();
-        List<Integer> serializeCalls = new ArrayList<>();
-        Repository repositoryMock = mock(Repository.class);
+    void save_whenIsActiveRSKIP87_ShouldSavedUnlimitedLockWhitelist() {
+        saveOneOffWhiteListEntry();
+        saveUnlimitedWhiteListEntry();
+        when(activationConfig.isActive(ConsensusRule.RSKIP87)).thenReturn(true);
 
-        try (MockedStatic<BridgeSerializationUtils> bridgeSerializationUtilsMocked = mockStatic(BridgeSerializationUtils.class)) {
-            // Mock the One-Off serialization
-            bridgeSerializationUtilsMocked.when(() -> BridgeSerializationUtils.serializeOneOffLockWhitelist(any(Pair.class)))
-                .then((InvocationOnMock invocation) -> {
-                    Pair<List<OneOffWhiteListEntry>, Integer> data = invocation.getArgument(0);
-                    assertEquals(whitelistMock.getAll(OneOffWhiteListEntry.class), data.getLeft());
-                    Assertions.assertSame(whitelistMock.getDisableBlockHeight(), data.getRight());
-                    serializeCalls.add(0);
-                    return Hex.decode("ccdd");
-                });
+        whitelistStorageProvider.save(activationConfig);
 
-            Mockito.doAnswer((InvocationOnMock invocation) -> {
-                    storageBytesCalls.add(0);
-                    RskAddress contractAddress = invocation.getArgument(0);
-                    DataWord address = invocation.getArgument(1);
-                    byte[] data = invocation.getArgument(2);
-                    // Make sure the bytes are set to the correct address in the repo and that what's saved is what was serialized
-                    assertArrayEquals(Hex.decode("aabbccdd"), contractAddress.getBytes());
-                    assertEquals(LOCK_ONE_OFF.getKey(), address);
-                    assertArrayEquals(Hex.decode("ccdd"), data);
-                    return null;
-                })
-                .when(repositoryMock)
-                .addStorageBytes(any(RskAddress.class), eq(LOCK_ONE_OFF.getKey()), any(byte[].class));
-
-            // Mock the Unlimited serialization
-            bridgeSerializationUtilsMocked.when(() -> BridgeSerializationUtils.serializeUnlimitedLockWhitelist(any(List.class)))
-                .then((InvocationOnMock invocation) -> {
-                    List<UnlimitedWhiteListEntry> unlimitedWhiteListEntries = invocation.getArgument(0);
-                    assertEquals(whitelistMock.getAll(UnlimitedWhiteListEntry.class), unlimitedWhiteListEntries);
-                    serializeCalls.add(0);
-                    return Hex.decode("bbcc");
-                });
-
-            Mockito.doAnswer((InvocationOnMock invocation) -> {
-                    storageBytesCalls.add(0);
-                    RskAddress contractAddress = invocation.getArgument(0);
-                    DataWord address = invocation.getArgument(1);
-                    byte[] data = invocation.getArgument(2);
-                    // Make sure the bytes are set to the correct address in the repo and that what's saved is what was serialized
-                    assertArrayEquals(Hex.decode("aabbccdd"), contractAddress.getBytes());
-                    assertEquals(LOCK_UNLIMITED.getKey(), address);
-                    assertArrayEquals(Hex.decode("bbcc"), data);
-                    return null;
-                })
-                .when(repositoryMock)
-                .addStorageBytes(any(RskAddress.class), eq(LOCK_UNLIMITED.getKey()), any(byte[].class));
-
-            whitelistStorageProvider.save(activationConfig);
-            // Shouldn't have tried to save nor serialize anything
-            assertEquals(0, storageBytesCalls.size());
-            assertEquals(0, serializeCalls.size());
-            TestUtils.setInternalState(whitelistStorageProvider, "lockWhitelist", whitelistMock);
-            whitelistStorageProvider.save(activationConfig);
-            assertEquals(2, storageBytesCalls.size());
-            assertEquals(2, serializeCalls.size());
-        }
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+        assertEquals(2, actualLockWhitelist.getAll().size());
     }
 
     @Test
-    void saveLockWhiteListAfterGetWithData() {
-        AtomicReference<Boolean> storageCalled = new AtomicReference<>();
-        storageCalled.set(Boolean.FALSE);
-        Repository repositoryMock = mock(Repository.class);
-        OneOffWhiteListEntry oneOffEntry = new OneOffWhiteListEntry(getBtcAddress("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"), Coin.COIN);
+    void getLockWhitelist_whenLockWhitelistIsNull_shouldReturnZeroEntries() {
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
 
-        when(repositoryMock.getStorageBytes(any(RskAddress.class), eq(LOCK_ONE_OFF.getKey())))
-            .then((InvocationOnMock invocation) -> new byte[]{(byte) 0xaa});
-
-        try (MockedStatic<BridgeSerializationUtils> bridgeSerializationUtilsMocked = mockStatic(BridgeSerializationUtils.class)) {
-            bridgeSerializationUtilsMocked.when(() -> BridgeSerializationUtils.deserializeOneOffLockWhitelistAndDisableBlockHeight(any(byte[].class), any(NetworkParameters.class)))
-                .then((InvocationOnMock invocation) -> {
-                    HashMap<Address, LockWhitelistEntry> map = new HashMap<>();
-                    map.put(oneOffEntry.address(), oneOffEntry);
-                    return Pair.of(map, 0);
-                });
-
-            bridgeSerializationUtilsMocked.when(() -> BridgeSerializationUtils.serializeOneOffLockWhitelist(any(Pair.class)))
-                .thenReturn(new byte[]{(byte) 0xee});
-
-            Mockito.doAnswer((InvocationOnMock invocation) -> {
-                    storageCalled.set(Boolean.TRUE);
-                    return null;
-                })
-                .when(repositoryMock)
-                .addStorageBytes(any(RskAddress.class), eq(LOCK_ONE_OFF.getKey()), eq(new byte[]{(byte) 0xee}));
-
-            Assertions.assertTrue(whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters).getSize() > 0);
-
-            whitelistStorageProvider.save(activationConfig);
-
-            Assertions.assertTrue(storageCalled.get());
-        }
+        assertEquals(0, actualLockWhitelist.getAll().size());
     }
 
-    private Address getBtcAddress(String address) {
-        return new Address(networkParameters, Hex.decode(address));
+    @Test
+    void getLockWhitelist_whenLockWhitelistIsNotNull_shouldReturnOneEntry() {
+        saveOneOffWhiteListEntry();
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        assertEquals(1, actualLockWhitelist.getAll().size());
+    }
+
+    @Test
+    void getWhitelist_whenIsActiveRSKIP87_shouldReturnTwoEntries() {
+        saveOneOffWhiteListEntry();
+        saveUnlimitedWhiteListEntry();
+        when(activationConfig.isActive(ConsensusRule.RSKIP87)).thenReturn(true);
+
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        assertEquals(2, actualLockWhitelist.getAll().size());
+    }
+
+    private void saveOneOffWhiteListEntry() {
+        Coin maxTransferValue = networkParameters.getMaxMoney();
+        OneOffWhiteListEntry oneOffWhiteListEntry = new OneOffWhiteListEntry(getBtcAddress(), maxTransferValue);
+        List<OneOffWhiteListEntry> oneOffWhiteListEntries = Collections.singletonList(oneOffWhiteListEntry);
+        Pair<List<OneOffWhiteListEntry>, Integer> pairValue = Pair.of(oneOffWhiteListEntries, 100);
+
+        inMemoryStorage.safeSaveToRepository(
+            LOCK_ONE_OFF.getKey(),
+            pairValue,
+            BridgeSerializationUtils::serializeOneOffLockWhitelist
+        );
+    }
+
+    private void saveUnlimitedWhiteListEntry() {
+        UnlimitedWhiteListEntry unlimitedWhiteListEntry = new UnlimitedWhiteListEntry(getBtcAddress());
+        List<UnlimitedWhiteListEntry> unlimitedWhiteListEntries = Collections.singletonList(unlimitedWhiteListEntry);
+
+        inMemoryStorage.safeSaveToRepository(
+            WhitelistStorageIndexKey.LOCK_UNLIMITED.getKey(),
+            unlimitedWhiteListEntries,
+            BridgeSerializationUtils::serializeUnlimitedLockWhitelist
+        );
+    }
+
+    private Address getBtcAddress() {
+        BtcECKey btcECKey = new BtcECKey();
+        return btcECKey.toAddress(whitelistConstants.getBtcParams());
     }
 }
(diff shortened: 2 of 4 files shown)


rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,114 @@
...
+class WhitelistStorageProviderImplTest {
+    private final WhitelistConstants whitelistConstants = WhitelistMainNetConstants.getInstance();
+    private final NetworkParameters networkParameters = whitelistConstants.getBtcParams();
+    private WhitelistStorageProvider whitelistStorageProvider;
+    private ActivationConfig.ForBlock activationConfig;
+
+    @BeforeEach
+    void setUp() {
+        StorageAccessor inMemoryStorage = new InMemoryStorage();
+        whitelistStorageProvider = new WhitelistStorageProviderImpl(inMemoryStorage);
```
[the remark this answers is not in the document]
wilmerrootstock: I need to have this variable apart since I added another method that uses the `inMemoryStorage` variable.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,114 @@
...
+        LockWhitelistEntry unlimitedWhiteListEntry = addUnlimitedWhiteListEntry();
+        lockWhitelist.put(getBtcAddress(), oneOffWhiteListEntry);
+        lockWhitelist.put(getBtcAddress(), unlimitedWhiteListEntry);
+
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        assertEquals(2, actualLockWhitelist.getAll().size());
+    }
+
+    private LockWhitelistEntry addOneOffWhiteListEntry() {
```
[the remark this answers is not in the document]
wilmerrootstock: I'm following the same pattern as WhitelistSupportImpl which uses terms such as: `add`, and `remove`. These terms are commonly used as synonyms of `create`, and `delete`.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,114 @@
...
+        Coin maxTransferValue = networkParameters.getMaxMoney();
+        return new OneOffWhiteListEntry(getBtcAddress(), maxTransferValue);
+    }
+
+    private LockWhitelistEntry addUnlimitedWhiteListEntry() {
+        return new UnlimitedWhiteListEntry(getBtcAddress());
+    }
+
+    private Address getBtcAddress() {
+        BtcECKey btcECKey = new BtcECKey();
```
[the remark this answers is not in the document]
marcos-iov: We have a method for this in BitcoinTestUtils class, let's use that

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,114 @@
...
+        assertEquals(0, actualLockWhitelist.getAll().size());
+    }
+
+    @Test
+    void save_whenLockWhiteListIsNotNull_ShouldReturnSavedValue() {
+        LockWhitelist lockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+        LockWhitelistEntry oneOffWhiteListEntry =  addOneOffWhiteListEntry();
+        lockWhitelist.put(getBtcAddress(), oneOffWhiteListEntry);
+
+        whitelistStorageProvider.save(activationConfig);
```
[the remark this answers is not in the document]
marcos-iov: Perhaps we could check the InMemoryStorage, make sure the value was stored under the expected key. What do you think?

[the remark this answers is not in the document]
nathanieliov: Yes, that's a good option.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
nathanieliov: I see you updated it. Now it is fine but the previous version was using it just as a local variable.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,114 @@
...
+        LockWhitelist lockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+        LockWhitelistEntry oneOffWhiteListEntry =  addOneOffWhiteListEntry();
+        LockWhitelistEntry unlimitedWhiteListEntry = addUnlimitedWhiteListEntry();
+        lockWhitelist.put(getBtcAddress(), oneOffWhiteListEntry);
+        lockWhitelist.put(getBtcAddress(), unlimitedWhiteListEntry);
+
+        whitelistStorageProvider.save(activationConfig);
+
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+        assertEquals(2, actualLockWhitelist.getAll().size());
```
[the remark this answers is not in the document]
wilmerrootstock: do you mean looking into `inMemoryStorage` and ensuring it is empty?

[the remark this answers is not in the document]
nathanieliov: It could be either directly checking the `inMemoryStorage` or something like `assertEquals(0, actualLockWhitelist.getAll().size())`. Wdyt?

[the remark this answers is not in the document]
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
wilmerrootstock: Done! Created the `BitcoinTestUtils.getBtcAddress` method to return a BTC address derived from a seed. I didn't use `getBtcEcKeysFromSeeds` method since it converts from an `array` of `seeds` to `List<BtcECKey>`. 
Additionally, I created 2 instance variables that get the value from this new method, to be used in different UTs.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,114 @@
...
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        assertEquals(0, actualLockWhitelist.getAll().size());
+    }
+
+    @Test
+    void save_whenLockWhiteListIsNotNull_ShouldReturnSavedValue() {
+        LockWhitelist lockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+        LockWhitelistEntry oneOffWhiteListEntry =  addOneOffWhiteListEntry();
+        lockWhitelist.put(getBtcAddress(), oneOffWhiteListEntry);
```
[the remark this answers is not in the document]
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
wilmerrootstock: Done! Added:
- New assertions to verify were saved with the correct address value InMemoryStorage.
- New assertions to verify were whitelisted the correct addresses.

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
```diff
@@ -29,6 +29,18 @@ public static List<BtcECKey> getBtcEcKeysFromSeeds(String[] seeds, boolean sorte
...
     }
 
+    /**
+     * Returns a BTC address derived from the seed.
+     *
+     * @param seed              the seed to derive the address from
+     * @param networkParameters the network parameters
+     * @return the BTC address
+     */
+    public static Address getBtcAddress(String seed, NetworkParameters networkParameters) {
```
marcos-iov, line 39: This already exists in this class @wilmerrootstock. See `createP2PKHAddress`

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+        //Making sure the saved value is correct
+        Map<Address, OneOffWhiteListEntry> oneOffWhiteListEntryMap = getAddressFromOneOffWhiteListEntryInMemoryStorage();
+        assertTrue(oneOffWhiteListEntryMap.containsKey(firstBtcAddress));
+
+        Map<Address, LockWhitelistEntry> lockWhitelistEntryMap = getAddressFromUnlimitedWhiteListEntryInMemoryStorage();
+        assertTrue(lockWhitelistEntryMap.containsKey(secondBtcAddress));
+    }
+
+    @Test
+    void getLockWhitelist_whenLockWhitelistIsNull_shouldReturnZeroEntries() {
```
marcos-iov, line 103: The name isn't completely right. It's not about lockWhitelist being null or not.

The variable `lockWhitelist` only serves as cache, to not fetch the data from storage every time.

In this case it's returning 0 entries simply because there are no entries in storage

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
marcos-iov: I think those are separate things @wilmerrootstock.

WhitelistSupportImpl uses `add` and `remove` because it's adding and removing from storage.

Here you are creating a `LockWhitelistEntry` object, not adding anything

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+        inMemoryStorage.safeSaveToRepository(
+            LOCK_ONE_OFF.getKey(),
+            pairValue,
+            BridgeSerializationUtils::serializeOneOffLockWhitelist
+        );
+    }
+
+    private LockWhitelistEntry addOneOffWhiteListEntry() {
+        Coin maxTransferValue = networkParameters.getMaxMoney();
```
marcos-iov, line 152: Why use this value?

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        assertEquals(0, actualLockWhitelist.getAll().size());
+    }
+
+    @Test
+    void getLockWhitelist_whenLockWhitelistIsNotNull_shouldReturnOneEntry() {
+        LockWhitelist lockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+        LockWhitelistEntry oneOffWhiteListEntry =  addOneOffWhiteListEntry();
+        lockWhitelist.put(firstBtcAddress, oneOffWhiteListEntry);
```
marcos-iov, line 113: I don't think this is properly set up. You are getting the LockWhitelist object from the storage provider and modifying the object reference.

See how something similar was implemented here https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java#L662

Basically store the value directly in the storage and then assert getting it

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        assertEquals(1, actualLockWhitelist.getAll().size());
+        //Making sure the correct value was whitelisted
+        assertTrue(actualLockWhitelist.isWhitelisted(firstBtcAddress));
+    }
+
+    @Test
+    void getWhitelist_whenIsActiveRSKIP87_shouldReturnTwoEntries() {
```
marcos-iov, line 123: I don't think this is right. RSKIP87 active doesn't mean it should return 2 elements. See the implementation of the method

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
wilmerrootstock: What do you guys think about this method name: `createOneOffWhiteListEntry`? 
Take into account we have 2 methods that return a `LockWhitelistEntry`. One related to `OneOffWhiteListEntry` and another one related to `UnlimitedWhiteListEntry`

`createOneOffWhiteListEntry`
`UnlimitedWhiteListEntry`

Do you agree, @nathanieliov @marcos-iov ?

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinTestUtils.java
wilmerrootstock: Oooh, Thanks for sharing!

wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
nathanieliov: Better now.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+        //make sure the lockWhitelist is empty
+        assertEquals(0, lockWhitelist.getAll().size());
+
+        LockWhitelistEntry oneOffWhiteListEntry =  addOneOffWhiteListEntry();
+        lockWhitelist.put(firstBtcAddress, oneOffWhiteListEntry);
+
+        whitelistStorageProvider.save(activationConfig);
+
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
```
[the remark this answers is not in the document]
nathanieliov: I was thinking about that. I think we should do it.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: @marcos-iov What do you think about this name: `getLockWhitelist_whenNoEntriesInStorage_shouldReturnZeroEntries()`?

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,176 @@
...
+        secondBtcAddress = BitcoinTestUtils.getBtcAddress("secondBtcAddress", networkParameters);
+    }
+
+    @Test
+    void save_whenLockWhitelistIsNull_shouldReturnZeroEntries() {
+        whitelistStorageProvider.save(activationConfig);
+
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        assertEquals(0, actualLockWhitelist.getAll().size());
```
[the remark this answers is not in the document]
wilmerrootstock: `whitelistStorageProvider.getLockWhitelist` looks for elements in storage so it is making sure InMemoryStorage didn't change. Please let me know if anything is needed.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: I was taking a look at the method, and first, it is looking for the `LOCK_ONE_OFF` key, and after that, in `if (activations.isActive(RSKIP87))`, it is looking for the `LOCK_UNLIMITED` key. If both exist in storage, they are added to the map that later is returned. This scenario might happen. 
WDYT? Would you like an additional test or modify this?

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
wilmerrootstock: Done!
Liked it this suggestion!
Applied to the `save()` tests.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
marcos-iov: Makes sense

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
marcos-iov: I don't think the value used makes much difference. So I wouldn't add a dependency just to get some random value. Use a hardcoded one, Coin.COIN or whatever

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
marcos-iov: `getLockWhitelist` might get values from storage or might return the cached values. That's my concern. I'd check directly in the storage

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
wilmerrootstock: As `lockWhitelist` is `null`, it goes to query the storage directly with the getLockWhitelist method.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+        assertEquals(2, actualLockWhitelist.getAll().size());
+    }
+
+    private LockWhitelistEntry addOneOffWhiteListEntry() {
+        Coin maxTransferValue = networkParameters.getMaxMoney();
+        return new OneOffWhiteListEntry(getBtcAddress(), maxTransferValue);
+    }
+
+    private LockWhitelistEntry addUnlimitedWhiteListEntry() {
```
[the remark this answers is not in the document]
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,185 @@
...
+    void save_whenLockWhitelistIsNull_shouldReturnZeroEntries() {
+        whitelistStorageProvider.save(activationConfig);
+
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        assertEquals(0, actualLockWhitelist.getAll().size());
+    }
+
+    @Test
+    void save_whenLockWhiteListIsNotNull_ShouldReturnSavedValue() {
```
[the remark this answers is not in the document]
marcos-iov: We could also add a setup for the activations object to make it explicit

```
when(activation.isActive(RSKIP87).thenReturn(false)`

File: rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java [hunk 16]
```diff
@@ -0,0 +1,185 @@
...
+
+        assertEquals(2, actualLockWhitelist.getAll().size());
+        //Making sure the correct value was whitelisted
+        assertTrue(actualLockWhitelist.isWhitelisted(firstBtcAddress));
+        assertTrue(actualLockWhitelist.isWhitelisted(secondBtcAddress));
+    }
+
+    private void saveInMemoryStorageOneOffWhiteListEntry() {
+        Coin maxTransferValue = Coin.COIN;
+        OneOffWhiteListEntry oneOffWhiteListEntry = new OneOffWhiteListEntry(firstBtcAddress, maxTransferValue);
```
marcos-iov, line 138: Use `createOneOffWhiteListEntry` method?

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+    @Test
+    void getLockWhitelist_whenNoEntriesInStorage_shouldReturnZeroEntries() {
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        assertEquals(0, actualLockWhitelist.getAll().size());
+    }
+
+    @Test
+    void getLockWhitelist_whenLockWhitelistIsNotNull_shouldReturnOneEntry() {
```
marcos-iov, line 112: There's a difference between the test name and the actual test scenario.

The scenario tested here is that there is a value in storage and it returns it. lockWhitelist is still null in memory.

To test what happens when lockWhitelist is not null, you should call the storageProvider twice. The second time lockWhitelist will not be null. See how Jere did it https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java#L662 

Call storageProvider once, get the values. Insert something directly in the storage. Call the storageProvider again, should use the cache so the value inserted directly in storage will not show. Recreate the storageProvider instance, should return both values now

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
marcos-iov: The scenario you describe is correct but I think the test name is confusing. It implies that after RSKIP87 is active 2 elements will be returned. That is not really the case, it could still return 1, 2,3, n elements.

`getWhitelist_whenIsActiveRSKIP87_shouldReturnUnlimitedEntries`

Something like that makes more sense to me

wilmerrootstock: Done! the method was renamed.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,185 @@
...
+        return oneOffWhitelistAndDisableBlockHeightData.getLeft();
+    }
+
+    private Map<Address, LockWhitelistEntry> getAddressFromUnlimitedWhiteListEntryInMemoryStorage() {
+        Map<Address, LockWhitelistEntry> whitelistedAddresses = new HashMap<>();
+        whitelistedAddresses.putAll(inMemoryStorage.safeGetFromRepository(
+            LOCK_UNLIMITED.getKey(),
+            data -> BridgeSerializationUtils.deserializeUnlimitedLockWhitelistEntries(data, networkParameters)
+        ));
+        return whitelistedAddresses;
```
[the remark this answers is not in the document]
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    private LockWhitelistEntry createOneOffWhiteListEntry() {
+        Coin maxTransferValue = Coin.COIN;
+        return new OneOffWhiteListEntry(firstBtcAddress, maxTransferValue);
+    }
+
+    private LockWhitelistEntry createUnlimitedWhiteListEntry() {
+        return new UnlimitedWhiteListEntry(secondBtcAddress);
+    }
```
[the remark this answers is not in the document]
wilmerrootstock: Done! Good catch!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
wilmerrootstock: Done!:

- Changed Method Name To Indicate RSKIP87 Is Not Active.
- `when(activation.isActive(RSKIP87).thenReturn(false)` added at the beginning of the method.
- Make sure there are no values in storage related to UnlimitedWhiteListEntry.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done! Test Case improved. Please let me know if the following accomplishes what we want:

- Changed the name to get closer to what it does.
- This test case is based on these steps:
1. The first time it gets values, `LockWhitelist` is `null` and queries the storage.
2. It should return zero entries as no value is saved in storage. As of here, `LockWhitelist` is `not null`.
3. Save a value directly in storage.
4. The second time it gets values, `LockWhitelist` is `not null` so it doesn't query the storage and get the value from the cache.
5. Return zero entries as it doesn't query in storage.
6. Recreate `whitelistStorageProvider` to ensure it is querying the storage.
7. Return one entry that was saved directly in storage.
8. Making sure the correct value was whitelisted.

_P.S.: I didn't save another entry directly in storage related to `OneOffWhiteListEntry` since it replaces the previous one because they have the same `WhitelistStorageIndexKey`, `LOCK_ONE_OFF`._

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,216 @@
...
+    private OneOffWhiteListEntry createOneOffWhiteListEntry(Address btcAddress) {
+        Coin maxTransferValue = Coin.COIN;
+        return new OneOffWhiteListEntry(btcAddress, maxTransferValue);
+    }
+
+    private UnlimitedWhiteListEntry createUnlimitedWhiteListEntry(Address btcAddress) {
+        return new UnlimitedWhiteListEntry(btcAddress);
+    }
+
+    private Map<Address, OneOffWhiteListEntry> getAddressFromOneOffWhiteListEntryInMemoryStorage() {
```
julia-zack, line 200: no need to add the `WhiteList` part

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,216 @@
...
+        Pair<HashMap<Address, OneOffWhiteListEntry>, Integer> oneOffWhitelistAndDisableBlockHeightData = inMemoryStorage.safeGetFromRepository(
+            LOCK_ONE_OFF.getKey(),
+            data -> BridgeSerializationUtils.deserializeOneOffLockWhitelistAndDisableBlockHeight(data, networkParameters)
+        );
+        return Objects.isNull(oneOffWhitelistAndDisableBlockHeightData) ? new HashMap<>() : oneOffWhitelistAndDisableBlockHeightData.getLeft();
+    }
+
+    private Map<Address, UnlimitedWhiteListEntry> getAddressFromUnlimitedWhiteListEntryInMemoryStorage() {
```
julia-zack, line 208: same here

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    private OneOffWhiteListEntry createOneOffWhiteListEntry(Address btcAddress) {
+        Coin maxTransferValue = Coin.COIN;
+        return new OneOffWhiteListEntry(btcAddress, maxTransferValue);
+    }
+
+    private UnlimitedWhiteListEntry createUnlimitedWhiteListEntry(Address btcAddress) {
+        return new UnlimitedWhiteListEntry(btcAddress);
+    }
```
julia-zack, line 198: are these methods necessary? they aren't really doing anything

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
julia-zack, line 1: please add a blank space after the _//_ when starting comments :)

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,216 @@
...
+        secondBtcAddress = BitcoinTestUtils.createP2PKHAddress(networkParameters, "secondBtcAddress");
+    }
+
+    @Test
+    void save_whenLockWhitelistIsNull_shouldReturnZeroEntries() {
+        whitelistStorageProvider.save(activationConfig);
+
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        //make sure the lockWhitelist and storage is empty
```
julia-zack, line 55: what does this mean?

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,216 @@
...
+        assertEquals(0, actualLockWhitelist.getAll().size());
+
+        //Making sure there is no value saved in Storage
+        Map<Address, OneOffWhiteListEntry> oneOffWhiteListEntryMap = getAddressFromOneOffWhiteListEntryInMemoryStorage();
+        assertEquals(0, oneOffWhiteListEntryMap.size());
+
+        //Making sure there is no value saved in Storage
+        Map<Address, UnlimitedWhiteListEntry> unlimitedWhiteListEntryMap = getAddressFromUnlimitedWhiteListEntryInMemoryStorage();
+        assertEquals(0, unlimitedWhiteListEntryMap.size());
+    }
```
julia-zack, line 65: since we are checking the storage for both keys, lets specify it:
`// should be no value saved in one off entry`
`// should be no value saved in unlimited entry`

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,216 @@
...
+        Map<Address, OneOffWhiteListEntry> oneOffWhiteListEntryMap = getAddressFromOneOffWhiteListEntryInMemoryStorage();
+        assertTrue(oneOffWhiteListEntryMap.containsKey(firstBtcAddress));
+
+        //Making sure there is no value saved in storage related to UnlimitedWhiteListEntry
+        Map<Address, UnlimitedWhiteListEntry> unlimitedWhiteListEntryMap = getAddressFromUnlimitedWhiteListEntryInMemoryStorage();
+        assertEquals(0, unlimitedWhiteListEntryMap.size());
+    }
+
+    @Test
+    void save_whenIsActiveRSKIP87_ShouldSavedUnlimitedLockWhitelist() {
```
julia-zack, line 94: `_shouldSave`

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,216 @@
...
+    void getLockWhitelist_whenNoEntriesInStorage_shouldReturnZeroEntries() {
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        assertEquals(0, actualLockWhitelist.getAll().size());
+    }
+
+    @Test
+    void getLockWhitelist_whenSavedValueDirectlyInStorageAndLockWhitelistIsNotNull_shouldReturnZeroEntries() {
+        //The first time LockWhitelist is null and it queries the storage
+        LockWhitelist firstTimeQuery = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
```
julia-zack, line 130: i think `firstTimeQuery` is not a super-descriptive name, wdyt about something like `lockWhitelistFromEmptyStorage` ?

julia-zack: or better, something like

```
LockWhitelist lockWhitelist;
// The first time LockWhitelist is null and it queries the storage
lockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
```

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+        //It should return zero entries as there is no value saved in storage. As of here LockWhitelist is not null
+        assertEquals(0, firstTimeQuery.getAll().size());
+
+        //Saving a value directly in storage
+        saveInMemoryStorageOneOffWhiteListEntry(firstBtcAddress);
+        //The second time LockWhitelist is not null so that it doesn't query the storage and get the value from cache
+        LockWhitelist secondTimeQuery = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
```
julia-zack, line 138: and here:
```
// The second time LockWhitelist is not null so that it doesn't query the storage and get the value from cache
lockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
```

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+        //Return zero entries as it doesn't query in storage
+        assertEquals(0, secondTimeQuery.getAll().size());
+        assertFalse(secondTimeQuery.isWhitelisted(firstBtcAddress));
+
+        //Recreating whitelistStorageProvider to make sure it is querying the storage
+        whitelistStorageProvider = new WhitelistStorageProviderImpl(inMemoryStorage);
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
```
julia-zack, line 146: and here:
`lockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);`

julia-zack (commented):
great improvements from previous tests!
Left some comments

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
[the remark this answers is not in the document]
nathanieliov: Well done

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
nathanieliov: Agree with @julia-zack

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+    @Test
+    void getLockWhitelist_whenNoEntriesInStorage_shouldReturnZeroEntries() {
+        LockWhitelist actualLockWhitelist = whitelistStorageProvider.getLockWhitelist(activationConfig, networkParameters);
+
+        assertEquals(0, actualLockWhitelist.getAll().size());
+    }
+
+    @Test
+    void getLockWhitelist_whenSavedValueDirectlyInStorageAndLockWhitelistIsNotNull_shouldReturnZeroEntries() {
```
nathanieliov, line 128: This is not possible. The implementation does not save directly into the storage and then checks against an out-of-date source(like the one in the cache). I think we don't need this test case since it is not allowed and differs from the implementation.

nathanieliov: I thought about this better, and maybe it would be good to have this test. The right thing to do is add a comment clarifying that this asserts that the value returned is the one in the cache.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: The sense is to keep the complete name of the object we are getting, `OneOffWhiteListEntry`. Agree regarding removing `InMemory` from the name. I propose you this name: `getAddressFromOneOffWhiteListEntryStorage`. Please let me know your insights!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: They improve the readability and maintainability of the code.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: It doesn't have any entry in LockWhitelist/"cache", and in storage either.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Hey, @nathanieliov!
About the comment in the last assertion, it retrieves the entry from storage, we recreate the `provider` to be able to do it.

About this test case, this is a suggestion requested by @marcos-iov. You can go deeper on this thread to get more details:

https://github.com/rsksmart/rskj/pull/2498#discussion_r1648883687

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Do you mean to create the `lockWhitelist` as a kind of temporary variable? Could it make sense to call it `tempLockWhitelist`? WDYT?

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
julia-zack: we are in the whitelist context, there's no need to have the _whitelist_ part.
And i think saying "storage entry" es cleaner and easier to read than "entry in storage"

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
julia-zack: so, `// make sure the lockWhitelist cache and storage entries are empty`

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
julia-zack: i dont see any benefits of calling it `tempLockWhitelist`, aren't most of variables temporary? jaja
i'd simply call it `lockWhitelist`.

btw, to avoid doing [this](https://github.com/rsksmart/rskj/pull/2498/files/3ee384652a2b3b342a702c7f480acc68f210fc8f#diff-4c11f95737338a52ebfaae1f3c570691c244e8eb14eedb9888743489488fd06cR144) you can do [what jere did](https://github.com/rsksmart/rskj/pull/2463/files#diff-f42b277f39730e846e15f899b88e1e7e718280d3ea3d4c4d5b035a5d66385923R634) (getting the value directly from repository)

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: The benefit is related to maintainability. It helps to understand at a glance what the variable's function is, and that it'll be updated several times.

wilmerrootstock: For the second suggestion, we need to recreate the provider, indeed, it is part of the challenge in this test. You can go deeper into it on this thread:
https://github.com/rsksmart/rskj/pull/2498#discussion_r1649755505

julia-zack: but you are doing it so its query the storage, right? 
If you query the storage directly you avoid the recreation.

wilmerrootstock: Done! Created the `tempLockWhitelist` variable. Nevertheless, we can discuss the variable name further to get an agreement about this topic.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: Done!

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
wilmerrootstock: But, we want the recreation to get the data from storage using `whitelistStorageProvider.getLockWhitelist`.

rskj-core/src/test/java/co/rsk/peg/whitelist/WhitelistStorageProviderImplTest.java
nathanieliov: Thanks for pointing me to that comment. It clarifies everything.

nathanieliov (approved):

julia-zack (approved):
lgtm! very well done

marcos-iov (approved):

PR #2501, opened 2024-06-18
Add getRetiringFederationCreationBlockNumber tests

julia-zack:


rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -880,6 +887,13 @@ void getRetiringFederationCreationTime_returnsNull() {
             Instant retiringFederationCreationTime = federationSupport.getRetiringFederationCreationTime();
             assertThat(retiringFederationCreationTime, is(nullValue()));
         }
+
+        @Test
+        @Tag("getRetiringFederationCreationBlockNumber")
+        void getRetiringFederationCreationBlockNumber_returnsLongMinusOne() {
```
julia-zack, line 893: not so sure about the _returnsLongMinusOne_ part, but I couldn't think of a better name

wilmerrootstock:
> _ No description provided. _

Hey, @julia-zack !
Could you please provide a PR description?

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
wilmerrootstock: or `shouldReturnNegativeValue`

julia-zack: i like the `minusOne` part since it is the exactly expected value

marcos-iov: Aren't we expecting `RETIRING_FEDERATION_NON_EXISTENT` response code? Why not compare against that value?

julia-zack: because that response code is a `-1`, so i'd have to cast it to be long. I wasn't sure if that was ok

apancorb: I think it will be better to compare it against `RETIRING_FEDERATION_NON_EXISTENT` and do a cast if needed. Are we using `RETIRING_FEDERATION_NON_EXISTENT` in prod code?

julia-zack: ok! yes we are

nathanieliov (approved):

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java

marcos-iov: @julia-zack looks like the code in FederationSupportImpl is returning a hardcoded -1. Maybe we should also update it to return the response code instead?

julia-zack: yep thats true. I'll cast it to return the response code. Thanks guys!

marcos-iov (approved):

PR #2506, opened 2024-06-18
Add getPendingFederationHash tests

julia-zack:


rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -1540,6 +1548,13 @@ void getPendingFederationSize_returnsPendingFederationSize() {
             int pendingFederationSize = federationSupport.getPendingFederationSize();
             assertThat(pendingFederationSize, is(pendingFederation.getSize()));
         }
+
+        @Test
+        @Tag("getPendingFederationHash")
+        void getPendingFederationHash_returnsPendingFederationHash() {
+            Keccak256 pendingFederationHash = federationSupport.getPendingFederationHash();
+            assertThat(pendingFederationHash, is(pendingFederation.getHash()));
```
apancorb, line 1556: Couldn't `pendingFedaration` be declared and instantiated as an instance var of the outer class? Ot it only makes sense to use it in this grouping of tests?

apancorb (approved):
LGTM

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: it only makes sense to use it in this group of tests, yes. 
But thanks for bringing it up, i've already thought about reviewing which global variables should be actually declared

nathanieliov (approved):

marcos-iov (approved):

PR #2507, opened 2024-06-19
Add getPendingFederatorBtcPublicKey tests

julia-zack:
- Add getPendingFederatorBtcPublicKey tests
- Rename `getPendingFederatorPublicKey` to `getPendingFederatorBtcPublicKey` in `BridgeSupport` and `FederationSupport` to be more accurate

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
@@ -1519,6 +1519,13 @@ class FederationSupportImplTest {
             Keccak256 pendingFederationHash = federationSupport.getPendingFederationHash();
             assertThat(pendingFederationHash, is(nullValue()));
         }
+
+        @Test
+        @Tag("getPendingFederatorBtcPublicKey")
+        void getPendingFederatorBtcPublicKey_returnsNull() {
+            byte[] pendingFederatorBtcPublicKey = federationSupport.getPendingFederatorBtcPublicKey(0);
+            assertThat(pendingFederatorBtcPublicKey, is(nullValue()));
+        }
     }
 
     @Nested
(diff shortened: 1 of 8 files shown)


rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -1555,6 +1562,26 @@ void getPendingFederationHash_returnsPendingFederationHash() {
...
+        @Tag("getPendingFederatorBtcPublicKey")
+        void getPendingFederatorBtcPublicKey_returnsFederatorFromPendingFederationBtcPublicKey() {
+            byte[] pendingFederatorBtcPublicKey = federationSupport.getPendingFederatorBtcPublicKey(0);
+            assertThat(pendingFederatorBtcPublicKey, is(pendingFederation.getBtcPublicKeys().get(0).getPubKey()));
+        }
+
+        @Test
+        @Tag("getPendingFederatorBtcPublicKey")
+        void getPendingFederatorBtcPublicKey_withNegativeIndex_throwsIndexOutOfBoundsException() {
+            assertThrows(IndexOutOfBoundsException.class, () -> federationSupport.getPendingFederatorBtcPublicKey(-1));
```
apancorb, line 1576: I am wandering, are we expecting this to happen in prod? Does it make sense to include this test? If that is the case we should probably be catching this exception inside `getPendingFederatorBtcPublicKey` and throwing a more domain custom one. Just an idea not sure.

apancorb (approved):
LGTM

nathanieliov (approved):
LGTM

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
apancorb: I see, that exception should already be being thrown by the `get` method to the list. Not sure why that logic is there.

marcos-iov: Good catch @apancorb, it's the very same exact exception.

We could simply remove that piece of code, or catch the exception and throw a custom one. Thoughts? @julia-zack @apancorb

julia-zack: ahhh you right guys. 
i don't see any benefits from catching the exception, its just an index out of bounds exception.
I think I would simply remove the piece of code

apancorb: If this is not expected to happen in prod then I would just remove that logic. Otherwise, catch and throw a custom one.

marcos-iov (approved):

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
marcos-iov: Apparently this method can only be called before RSKIP123, it was then replaced for `getPendingFederatorPublicKeyOfType`.

So I simply removed the check

PR #2511, opened 2024-06-20
Add getActiveFederationCreationBlockHeight tests

julia-zack:


rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -775,13 +776,151 @@ private Stream<Arguments> expectedFederatorPublicKeyOfTypeArgs() {
         }
     }
 
+    @Nested
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    @Tag("active federation creation block height tests")
+    class ActiveFederationCreationBlockHeightTests {
```
julia-zack, line 782: since these tests don't need new and old federation context, a new separated class was created

marcos-iov (approved):

PR #2518, opened 2024-06-24
Add getActiveFederationBtcUTXOs tests

julia-zack:
-Add `getActiveFederationBtcUTXOs` tests
-Rename other tests to be more accurate

_Disclaimer:_
The values had to be saved directly in the repository because there are no `setFederationUTXOs` methods in the implementation

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
@@ -17,11 +17,12 @@
  */
 package co.rsk.peg.federation;
 
+import static co.rsk.peg.storage.FederationStorageIndexKey.NEW_FEDERATION_BTC_UTXOS_KEY;
+import static co.rsk.peg.storage.FederationStorageIndexKey.OLD_FEDERATION_BTC_UTXOS_KEY;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.*;
-import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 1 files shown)


rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -199,6 +201,24 @@ void getActiveFederatorPublicKeyOfType_returnsFederatorPublicKeysFromGenesisFede
             byte[] activeFederatorMstPublicKey = federationSupport.getActiveFederatorPublicKeyOfType(0, KeyType.MST);
             assertThat(activeFederatorMstPublicKey, is(federatorFromGenesisFederationMstPublicKey.getPubKey(true)));
         }
+
+        @Test
+        @Tag("getActiveFederationBtcUTXOs")
+        void getActiveFederationUTXOs_returnsGenesisFederationUTXOs() {
+            storageProvider = mock(FederationStorageProviderImpl.class);
```
marcos-iov, line 208: Any reason to mock instead of using the inMemoeryStorage as in the other test?

julia-zack: yes! i added a disclaimer in the pr description:

Disclaimer:
The values being obtained from the provider had to be mocked because there are no setFederationUTXOs methods in the implementation

marcos-iov: Sorry, didn't see that 😬

What about setting the value directly in the storage under the proper key? Like this https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java#L669-L673

julia-zack: jaja no worries :))
i thought about doing that, but i wasn't convinced. But yes i think is more realistic doing it that way, ill change it. Thanks!

marcos-iov (approved):

PR #2532, opened 2024-07-01
Adds new getPendingFederation tests to FederationStorageProviderImplT…

jeremy-then:
Adds new getPendingFederation tests to FederationStorageProviderImplTest and removes the ones from BridgeStorageProviderTest


rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -144,13 +147,110 @@ void testGetOldFederation(
...
+        if(federationFormat != INVALID_FEDERATION_FORMAT && federationFormat != EMPTY_FEDERATION_FORMAT) {
+            when(activations.isActive(ConsensusRule.RSKIP123)).thenReturn(true);
+        }
+
+        StorageAccessor storageAccessor = new InMemoryStorage();
+        byte[] federationFormatSerialized = getFederationFormatSerialized(federationFormat);
+        storageAccessor.saveToRepository(PENDING_FEDERATION_FORMAT_VERSION.getKey(), federationFormatSerialized);
+
+        byte[] serializedFederation = expectedFederation.serialize(activations);
+        System.out.println("serializedFederation: " + ByteUtil.toHexString(serializedFederation));
```
marcos-iov, line 188: Should this be here?

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -144,13 +147,110 @@ void testGetOldFederation(
...
+
+        // Arrange
+
+        ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
+
+        when(activations.isActive(ConsensusRule.RSKIP123)).thenReturn(false);
+
+        if(federationFormat != INVALID_FEDERATION_FORMAT && federationFormat != EMPTY_FEDERATION_FORMAT) {
+            when(activations.isActive(ConsensusRule.RSKIP123)).thenReturn(true);
+        }
```
marcos-iov, line 181: Not sure I follow this, what's the idea?

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -144,13 +147,110 @@ void testGetOldFederation(
...
+        // Assert
+
+        // Call the method again and assert the same cached federation is returned
+        assertEquals(obtainedFederation, federationStorageProvider.getPendingFederation());
+
+        if(federationFormat == INVALID_FEDERATION_FORMAT || federationFormat == EMPTY_FEDERATION_FORMAT) {
+            assertArrayEquals(expectedFederation.serialize(activations), obtainedFederation.serialize(activations));
+        } else {
+            assertEquals(expectedFederation, obtainedFederation);
+        }
```
marcos-iov, line 208: This is confusing, should it be 2 separate test methods?

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -673,4 +773,9 @@ private byte[] getSerializedFederation(Federation federation, int federationForm
 
     }
 
+
+    private static PendingFederation buildMockPendingFederation() {
```
marcos-iov, line 777: Julia created a `PendingFederationBuilder`

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
apancorb: I agree

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -144,13 +147,110 @@ void testGetOldFederation(
 
     @Test
     void getOldFederation_previouslySetToNull_returnsNull() {
-        StorageAccessor storageAccessor = new InMemoryStorage();
-        FederationStorageProvider federationStorageProvider = new FederationStorageProviderImpl(storageAccessor);
+        FederationStorageProvider federationStorageProvider = new FederationStorageProviderImpl(null);
```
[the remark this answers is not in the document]
jeremy-then: The `null` part is to show in the test that it will not be used. It is not expected to be used. I think for test it is reasonable to have it this way. What do you think? This is mainly because we are avoiding mocks. With a mock, I would simply assert that certain mocked function is not called, etc. But with an actual StorageAccessor, there's no way to ensure it was not accessed, unless I don't set it, and if it were to be used when we are not expecting it, then the test would fail.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
julia-zack: agree too

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
julia-zack: +1

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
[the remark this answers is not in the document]
marcos-iov: I don't like this approach. It's basically setting a null value to then show that if a NPE is not thrown then the value was not used. Not very nice in my opinion.

You can always check the storage before/after to make sure it didn't change if that's what we need to assert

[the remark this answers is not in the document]
nathanieliov: Well, I kind of agree with Marcos; I don't like the NPE idea. What about using a mock and verifying that the mock method is a `never` call?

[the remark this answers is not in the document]
jeremy-then: Updated.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: Updated.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: Updated. Separeated the tests.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: Updated.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: Updated.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -144,13 +147,110 @@ void testGetOldFederation(
...
         federationStorageProvider.setOldFederation(null);
         Federation oldFederation = federationStorageProvider.getOldFederation(federationConstants, activations);
         assertNull(oldFederation);
     }
 
+    private static Stream<Arguments> providePendingFederationAndFormatArguments() {
+        return Stream.of(
+            Arguments.of(P2SH_ERP_FEDERATION_FORMAT_VERSION, buildMockPendingFederation()),
```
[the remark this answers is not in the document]
jeremy-then: Updated.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -144,13 +147,110 @@ void testGetOldFederation(
...
+        PendingFederation expectedFederation
+    ) {
+
+        // Arrange
+
+        ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
+
+        when(activations.isActive(ConsensusRule.RSKIP123)).thenReturn(false);
+
+        if(federationFormat != INVALID_FEDERATION_FORMAT && federationFormat != EMPTY_FEDERATION_FORMAT) {
```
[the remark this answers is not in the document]
jeremy-then: Updated.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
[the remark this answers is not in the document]
julia-zack: i doesn't appear updated for me, still `null`

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -144,13 +146,160 @@ void testGetOldFederation(
...
+        PendingFederation expectedFederation
+    ) {
+
+        // Arrange
+
+        ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
+
+        when(activations.isActive(ConsensusRule.RSKIP123)).thenReturn(false);
+
+        when(activations.isActive(ConsensusRule.RSKIP123)).thenReturn(true);
```
julia-zack, line 176: :thinking_face:

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -144,13 +146,160 @@ void testGetOldFederation(
...
         Federation oldFederation = federationStorageProvider.getOldFederation(federationConstants, activations);
         assertNull(oldFederation);
     }
 
+    private static Stream<Arguments> providePendingFederationAndFormatArguments() {
+        return Stream.of(
+            Arguments.of(P2SH_ERP_FEDERATION_FORMAT_VERSION, new PendingFederationBuilder().build()),
+            Arguments.of(NON_STANDARD_ERP_FEDERATION_FORMAT_VERSION, new PendingFederationBuilder().build()),
+            Arguments.of(STANDARD_MULTISIG_FEDERATION_FORMAT_VERSION, new PendingFederationBuilder().build())
+        );
```
julia-zack, line 160:
```
pendingFed = new PendingFederationBuilder().build();
return Stream.of(
            Arguments.of(P2SH_ERP_FEDERATION_FORMAT_VERSION, pendingFed),
            Arguments.of(NON_STANDARD_ERP_FEDERATION_FORMAT_VERSION, pendingFed),
            Arguments.of(STANDARD_MULTISIG_FEDERATION_FORMAT_VERSION, pendingFed)
        )
```

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -144,13 +146,160 @@ void testGetOldFederation(
...
+        // Directly saving a null federation in storage to then assert that the method returns the cached federation
+        storageAccessor.saveToRepository(PENDING_FEDERATION_KEY.getKey(), null);
+
+        // Assert
+
+        // Call the method again and assert the same cached federation is returned
+        assertEquals(obtainedFederation, federationStorageProvider.getPendingFederation());
+
+        assertEquals(expectedFederation, obtainedFederation);
+
```
julia-zack, line 199: i would create a separated test for this, wdyt? so every test does only one thing

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -144,13 +146,160 @@ void testGetOldFederation(
...
+        // Arrange
+
+        ActivationConfig.ForBlock activations = mock(ActivationConfig.ForBlock.class);
+
+        when(activations.isActive(ConsensusRule.RSKIP123)).thenReturn(false);
+
+        StorageAccessor storageAccessor = new InMemoryStorage();
+        byte[] federationFormatSerialized = getFederationFormatSerialized(STANDARD_MULTISIG_FEDERATION_FORMAT_VERSION);
+        storageAccessor.saveToRepository(PENDING_FEDERATION_FORMAT_VERSION.getKey(), federationFormatSerialized);
```
julia-zack, line 288: rskip123 isn't supposed to be active if you're saving the format version? 
https://github.com/rsksmart/rskj/blob/add-getPendingFederation-set-of-tests/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java#L361-#L364

julia-zack: btw, i don't really follow why you activate or deactivate that rskip in each case

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: I thought about that, but I wanted a different fed object for each tests. Probably no big deal though.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: We would be duplicating exactly the same thing. This part is dependent on the above part, meaning, I would need to copy and paste the above part exactly as is anyways.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: Good catch, but the `getPendingFederation` doesn't check the rskip to know if it should use the format version or not, it simply tries to get it from storage. If some other method directly saves that format version in storage, then the `getPendingFederation` method will use it, with rskip active or not.

Is this behavior something we should discuss with the team much deeper? This looks like something that can be 'worked around' to make the `getPendingFederation` use the version key even when the rskip is not active.

What do you guys think? @nathanieliov , @marcos-iov , @julia-zack .

jeremy-then: Should we have a check for the rskip here? https://github.com/rsksmart/rskj/blob/c4402b35b0e536bcdc5936064e51196c51ad3544/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java#L191

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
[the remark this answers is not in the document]
jeremy-then: Updated.

marcos-iov (approved):

PR #2536, opened 2024-07-03
Adds getFederationElection tests in FederationStorageProviderImplTest…

jeremy-then:
Adds getFederationElection tests in FederationStorageProviderImplTest and removes the ones from BridgeStorageProviderTest


rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -686,6 +696,65 @@ void getOldFederationBtcUTXOs_calledTwice_returnCachedUtxos() {
...
+
+    @Test
+    void getFederationElection_whenElectionIsNotInStorage_shouldReturnDefault() {
+
+        // Arrange
+
+        AddressBasedAuthorizer authorizer = getTestingAddressBasedAuthorizer();
+
+        ABICallElection expectedElection = new ABICallElection(authorizer);
+        byte[] expectedElectionEncoded = BridgeSerializationUtils.serializeElection(expectedElection);
```
nathanieliov, line 742: Let's move this variable to the `assert` section. I think it's better to have the `expected` variables close to the assertion where is being used(when possible, of course). I think this case it's suitable for being there.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -686,6 +696,65 @@ void getOldFederationBtcUTXOs_calledTwice_returnCachedUtxos() {
...
 
+    @Test
+    void getFederationElection_whenElectionIsInStorage_shouldReturnNewElection() {
+
+        // Arrange
+
+        AddressBasedAuthorizer authorizer = getTestingAddressBasedAuthorizer();
+
+        ABICallElection expectedElection = getSampleElection("function1", authorizer);
+        byte[] expectedElectionEncoded = BridgeSerializationUtils.serializeElection(expectedElection);
```
nathanieliov, line 707: Same here. I think this variable should be move to the line above where the assertion is using it.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -686,6 +696,65 @@ void getOldFederationBtcUTXOs_calledTwice_returnCachedUtxos() {
...
+
+        // Setting a different election in storage to assert that calling the method again returns the cached election
+
+        ABICallElection secondElectionSample = getSampleElection("function2", authorizer);
+        storageAccessor.saveToRepository(FEDERATION_ELECTION_KEY.getKey(), BridgeSerializationUtils.serializeElection(secondElectionSample));
+
+        ABICallElection cachedElection = federationStorageProvider.getFederationElection(authorizer);
+
+        assertArrayEquals(serializeElection(actualElection), serializeElection(cachedElection));
+        assertFalse(Arrays.equals(expectedElectionEncoded, serializeElection(secondElectionSample)));
```
julia-zack, line 730: this should probably go in another test, wdyt?

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -686,6 +696,65 @@ void getOldFederationBtcUTXOs_calledTwice_returnCachedUtxos() {
...
+
+        FederationStorageProvider federationStorageProvider = new FederationStorageProviderImpl(storageAccessor);
+
+        // Act
+
+        ABICallElection actualElection = federationStorageProvider.getFederationElection(authorizer);
+
+        // Assert
+
+        assertArrayEquals(expectedElectionEncoded, serializeElection(actualElection));
```
julia-zack, line 720: shouldn't we test that `expectedElection` matches `actualElection`?

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: Not possible here. It is being used in the next line.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: It would look exactly like this one. For this part of the test, the above part needs to happen first. Meaning, it is dependent on it.

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: It does. It is asserting it in this line, asserting the serialized versions. The `ABICallElection` class does not implement an `equals` method, so, not good for comparing equality as is. I had to convert them to their serialized form (byte[]) and compare those arrays for equality.

nathanieliov (approved):
LGTM

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
julia-zack: i would put it in separated tests anyways, so the test tests what its name describes

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -734,6 +744,64 @@ void getOldFederationBtcUTXOs_calledTwice_returnCachedUtxos() {
 
     }
 
+    @Test
+    void getFederationElection_whenElectionIsInStorage_shouldReturnNewElection() {
+
+        // Arrange
+
+        AddressBasedAuthorizer authorizer = getTestingAddressBasedAuthorizer();
```
marcos-iov, line 752: What do you think of implementing this for fed change callers? https://github.com/rsksmart/rskj/blob/master/rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbVoteCaller.java

It will probably be useful in voteFedChange tests

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: Updated.

julia-zack (approved):

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
jeremy-then: Updated.

PR #2572, opened 2024-07-11
Add get federation election tests2

jeremy-then:
Add get federation election tests


rskj-core/src/test/java/co/rsk/peg/vote/AddressBasedAuthorizerVoteCaller.java
```diff
@@ -0,0 +1,22 @@
+package co.rsk.peg.vote;
+
+import co.rsk.core.RskAddress;
+
+public enum AddressBasedAuthorizerVoteCaller {
```
marcos-iov, line 5: This isn't really what I meant, who are these callers?

In the example I pointed out https://github.com/rsksmart/rskj/blob/master/rskj-core/src/test/java/co/rsk/peg/feeperkb/FeePerKbVoteCaller.java, those are the addresses authorized to vote for feePerKb changes. Taken from https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/feeperkb/constants/FeePerKbMainNetConstants.java#L17

My proposal is to create an enum with federation change authorized callers, taken from https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/federation/constants/FederationMainNetConstants.java#L41

This enum should go in federation package, an probably be called `FederationChangeCaller` to be consistent

jeremy-then: Thanks. That was much clear.
Updated.

nathanieliov (approved):

marcos-iov (approved):

PR #2589, opened 2024-07-16
LockingCapConstants Refactor

wilmerrootstock:
## Description
Following with the Bridge refactors, it is now time to refactor and remove the locking cap business logic from the Bridge.

Create its package `lockingcap` with a Support, StorageProvider, and Constants classes. Since it will have its classes, should be removed from the variable names, `lockingCap`, to avoid being redundant.

## Motivation and Context

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapRegTestConstants.java
+++ b/rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapRegTestConstants.java
@@ -1,0 +1,33 @@
+package co.rsk.peg.lockingcap.constants;
+
+import co.rsk.bitcoinj.core.Coin;
+import co.rsk.peg.vote.AddressBasedAuthorizer;
+import java.util.Collections;
+import java.util.List;
+import org.bouncycastle.util.encoders.Hex;
+import org.ethereum.crypto.ECKey;
+
+public class LockingCapRegTestConstants extends LockingCapConstants {
+
+    private static final LockingCapRegTestConstants instance = new LockingCapRegTestConstants();
+
+    private LockingCapRegTestConstants() {
+        ECKey authorizerPublicKey = ECKey.fromPublicOnly(Hex.decode(
+            "04450bbaab83ec48b3cb8fbb077c950ee079733041c039a8c4f1539e5181ca1a27589eeaf0fbf430e49d2909f14c767bf6909ad6845831f683416ee12b832e36ed"
+        ));
+
+        List<ECKey> increaseAuthorizedKeys = Collections.singletonList(authorizerPublicKey);
+
+        increaseAuthorizer = new AddressBasedAuthorizer(
+            increaseAuthorizedKeys,
+            AddressBasedAuthorizer.MinimumRequiredCalculation.ONE
+        );
+
+        initialValue = Coin.COIN.multiply(1_000L); // 1000 BTC
+        incrementsMultiplier = 2;
+    }
+
+    public static LockingCapRegTestConstants getInstance() {
+        return instance;
+    }
+}
--- a/rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapTestNetConstants.java
+++ b/rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapTestNetConstants.java
@@ -2,9 +2,10 @@ package co.rsk.peg.lockingcap.constants;
 
 import co.rsk.bitcoinj.core.Coin;
 import co.rsk.peg.vote.AddressBasedAuthorizer;
-import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.bouncycastle.util.encoders.Hex;
 import org.ethereum.crypto.ECKey;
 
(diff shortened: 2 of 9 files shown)


rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapTestNetConstants.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.stream.Collectors;
+import org.bouncycastle.util.encoders.Hex;
+import org.ethereum.crypto.ECKey;
+
+public class LockingCapTestNetConstants extends LockingCapConstants {
+
+    private static final LockingCapTestNetConstants instance = new LockingCapTestNetConstants();
+
+    private LockingCapTestNetConstants() {
+        List<ECKey> increaseAuthorizedKeys = Arrays.stream(new String[]{
```
apancorb, line 16: nit: for this scenarios, we can also use `Arrays.asList(...)`

rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapRegTestConstants.java
```diff
@@ -0,0 +1,33 @@
...
+        ));
+
+        List<ECKey> increaseAuthorizedKeys = Collections.singletonList(authorizerPublicKey);
+
+        increaseAuthorizer = new AddressBasedAuthorizer(
+            increaseAuthorizedKeys,
+            AddressBasedAuthorizer.MinimumRequiredCalculation.ONE
+        );
+
+        initialValue = Coin.COIN.multiply(1_000L); // 1000 BTC
```
apancorb, line 26: why multiplying here?

apancorb (approved):
LGTM, left minor comments

rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapTestNetConstants.java
wilmerrootstock: Thanks for your feedback, @apancorb! For this opportunity, we can't apply `Arrays.asList(...)` to replace `new String[]` since the `Arrays.stream` method receives an `array` and not a `List<>`.

rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapRegTestConstants.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.List;
+import org.bouncycastle.util.encoders.Hex;
+import org.ethereum.crypto.ECKey;
+
+public class LockingCapRegTestConstants extends LockingCapConstants {
+
+    private static final LockingCapRegTestConstants instance = new LockingCapRegTestConstants();
+
+    private LockingCapRegTestConstants() {
+        ECKey authorizerPublicKey = ECKey.fromPublicOnly(Hex.decode(
```
julia-zack, line 15: check that here you named it `authorizerPublicKey` and in [mainnet](https://github.com/rsksmart/rskj/pull/2589/files#diff-7eb4e14993a959654a6715a0c76922524bc297d4116994ad9c15ce6bc40e383bR16) you named it `increaseAuthorizedKeys`. Lets standardize it

rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapTestNetConstants.java
julia-zack: i think we can use `Collections.unmodifiableList(Stream.of(...` like in FederationConstants. Check [FederationMainNetConstants](https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/federation/constants/FederationMainNetConstants.java#L22)

julia-zack (commented):
left two comments but looks good! Well done

rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapRegTestConstants.java
wilmerrootstock: I'm not sure about this answer. Maybe @marcos-iov @nathanieliov @jeremy-then @julia-zack  might shed some light on it.

julia-zack: not a clue tbh

marcos-iov: Just a way to represent 1,000 BTC, `Coin.COIN` represents 1 BTC. Using `Coin.valueOf(long)` expects the value in satoshis so that would be a pretty long number with many 0s

rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapTestNetConstants.java
wilmerrootstock: Done!

rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapRegTestConstants.java
wilmerrootstock: Yes, this is because, we are dealing with only one key in `RegTest`, unlike `MainNet` and `TestNet`. We don't need to deal with it like a list of keys. For this reason, I deal with it differently. Nevertheless, after this variable, you'll find the `increaseAuthorizedKeys`.

marcos-iov: No big deal, but I would still call it `increaseAuthorizedKeys`

The variable name shouldn't change depending on the amount of elements. It's still a list, with one, zero or many elements

wilmerrootstock: @marcos-iov, The name `increaseAuthorizedKeys` remains. I didn't change that variable. I only changed how we deal with the public key. Actually, it is a suggestion by @antonio and you in Whitelist where we deal with one public key as well and not a list of keys. I used the same variable name indeed.  Please take a look at [Whitelist code](https://github.com/rsksmart/rskj/blob/5be021fbb3ba0a11f96160de57ad3437e9acac5f/rskj-core/src/main/java/co/rsk/peg/whitelist/constants/WhitelistMainNetConstants.java#L17)

rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapTestNetConstants.java
apancorb: You should be able to do so as something like this:

```java
List<ECKey> increaseAuthorizedKeys = Arrays.asList(
    "0470...",
    "045e...",
    "0455...")
   .stream()
   .map(hex -> ECKey.fromPublicOnly(Hex.decode(hex)))
```

But I agree with what Julia proposed this way we are more consistent in the codebase.

rskj-core/src/main/java/co/rsk/peg/lockingcap/constants/LockingCapRegTestConstants.java
marcos-iov: Ok, got it. Thanks

marcos-iov (approved):

PR #2595, opened 2024-07-16
LockingCapStorageProvider Implementation

wilmerrootstock:
## Description
Move the following methods from `BridgeStorageProvider` to `LockingCapStorageProviderImpl`:

- `getLockingCap`
- `setLockingCap`
- `saveLockingCap`

Create a new class `LockingCapStorageProviderImpl` under the `co.rsk.peg.lockingcap` package that implements the methods defined in the `LockingCapStorageProvider` interface

## Motivation and Context

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
@@ -41,9 +41,10 @@ public class LockingCapStorageProviderImpl implements LockingCapStorageProvider 
     @Override
     public void save(ActivationConfig.ForBlock activations) {
         if (activations.isActive(RSKIP134)) {
+            Coin currentLockingCap = getLockingCap(activations).orElse(null);
             bridgeStorageAccessor.saveToRepository(
                 LOCKING_CAP.getKey(),
-                getLockingCap(activations).get(),
+                currentLockingCap,
                 BridgeSerializationUtils::serializeCoin
             );
         }
--- a/rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
@@ -40,13 +40,15 @@ public class LockingCapStorageProviderImpl implements LockingCapStorageProvider 
 
     @Override
     public void save(ActivationConfig.ForBlock activations) {
-        if (activations.isActive(RSKIP134)) {
-            Coin currentLockingCap = getLockingCap(activations).orElse(null);
-            bridgeStorageAccessor.saveToRepository(
-                LOCKING_CAP.getKey(),
-                currentLockingCap,
-                BridgeSerializationUtils::serializeCoin
-            );
+        if (!activations.isActive(RSKIP134)) {
+            return;
         }
+
+        Coin currentLockingCap = getLockingCap(activations).orElse(null);
+        bridgeStorageAccessor.saveToRepository(
+            LOCKING_CAP.getKey(),
+            currentLockingCap,
+            BridgeSerializationUtils::serializeCoin
+        );
     }
 }
--- a/rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
@@ -12,18 +12,14 @@ import org.ethereum.config.blockchain.upgrades.ActivationConfig;
 public class LockingCapStorageProviderImpl implements LockingCapStorageProvider {
 
     private Coin lockingCap;
-    private final ActivationConfig.ForBlock activations;
     private final StorageAccessor bridgeStorageAccessor;
 
-    public LockingCapStorageProviderImpl(
-        ActivationConfig.ForBlock activations,
-        StorageAccessor bridgeStorageAccessor) {
-        this.activations = activations;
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
```diff
@@ -43,9 +39,13 @@ public void setLockingCap(Coin lockingCap) {
     }
 
     @Override
-    public void save() {
+    public void save(ActivationConfig.ForBlock activations) {
         if (activations.isActive(RSKIP134)) {
-            bridgeStorageAccessor.safeSaveToRepository(LOCKING_CAP.getKey(), getLockingCap().get(), BridgeSerializationUtils::serializeCoin);
+            bridgeStorageAccessor.safeSaveToRepository(
+                LOCKING_CAP.getKey(),
+                getLockingCap(activations).get(),
```
marcos-iov, line 46: This is the kind of thing that shold come up in test, but careful here.

You changed `getLockingCap` to return Optional.empty instead of null. So now what happens when you call `save` with empty instead of null?

wilmerrootstock: Done! To avoid an `NoSuchElementException` trying to get the value from `Optional.get()` when is `Optional.empty()`, I used `orElse(null)` method to keep the same behavior.

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
```diff
@@ -0,0 +1,52 @@
...
+            if (lockingCap == null) {
+                lockingCap = initializeLockingCap();
+            }
+            return Optional.of(lockingCap);
+        }
+        return Optional.empty();
+    }
+
+    private synchronized Coin initializeLockingCap() {
+        return bridgeStorageAccessor.getFromRepository(LOCKING_CAP.getKey(), BridgeSerializationUtils::deserializeCoin);
```
julia-zack, line 33: why moving this here?

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
julia-zack: i think its better to keep things as they are on this task and create a new one to approach this kind of things. That way is easier to see the impact of the change

wilmerrootstock: @julia-zack, the solution related to `Optional` has already been implemented.

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
wilmerrootstock: Because It's a lazy initialization.

marcos-iov, line 33:
```suggestion
    private synchronized void initializeLockingCap() {
lockingCap = bridgeStorageAccessor.getFromRepository(LOCKING_CAP.getKey(), BridgeSerializationUtils::deserializeCoin);
```
If you want to instantiate a class variable, then simply assign the value. No need to return I guess

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    public void save(ActivationConfig.ForBlock activations) {
+        if (activations.isActive(RSKIP134)) {
+            Coin currentLockingCap = getLockingCap(activations).orElse(null);
+            bridgeStorageAccessor.saveToRepository(
+                LOCKING_CAP.getKey(),
+                currentLockingCap,
+                BridgeSerializationUtils::serializeCoin
+            );
+        }
```
marcos-iov, line 50:
```suggestion
        if (!activations.isActive(RSKIP134)) {
            return;
        }
            Coin currentLockingCap = getLockingCap(activations).orElse(null);
            bridgeStorageAccessor.saveToRepository(
                LOCKING_CAP.getKey(),
                currentLockingCap,
                BridgeSerializationUtils::serializeCoin
            );
        }
```
Just an idea, to return early

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
marcos-iov: Better now, thanks

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
apancorb: I agree with this approach

wilmerrootstock: Done!

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapStorageProviderImpl.java
wilmerrootstock: Done! Thanks for your contribution @apancorb!

marcos-iov (approved):

PR #2602, opened 2024-07-17
LockingCapSupport Implementation

wilmerrootstock:
## Description
After the `LockingCapSupport` interface has been created, it is time to implement it. Create a new class, `LockingCapSupportImpl`, under the `co.rsk.peg.lockingcap` package that implements the methods defined in the `LockingCapSupport` interface. Please take the implementation from the `BridgeSupport` class, and move to this new class.

## Motivation and Context

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapSupportImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapSupportImpl.java
@@ -32,7 +32,7 @@ public class LockingCapSupportImpl implements LockingCapSupport {
     }
 
     @Override
-    public Coin getLockingCap() {
+    public Optional<Coin> getLockingCap() {
         Optional<Coin> lockingCap = storageProvider.getLockingCap(activations);
 
         // Before returning the locking cap, check if it was already set
@@ -41,7 +41,7 @@ public class LockingCapSupportImpl implements LockingCapSupport {
             logger.debug("[getLockingCap] {}", "Setting initial Locking Cap value");
             storageProvider.setLockingCap(constants.getInitialValue());
         }
-        return storageProvider.getLockingCap(activations).get();
+        return storageProvider.getLockingCap(activations);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapSupportImpl.java
```diff
@@ -41,7 +41,7 @@ public Optional<Coin> getLockingCap() {
             logger.debug("[getLockingCap] {}", "Setting initial Locking Cap value");
             storageProvider.setLockingCap(constants.getInitialValue());
         }
-        return storageProvider.getLockingCap(activations);
+        return storageProvider.getLockingCap(activations).get();
```
marcos-iov, line 44: What happens when the storage returns empty?

wilmerrootstock: Done! We return the `Optional`. This business logic is handled on LockingCapSupportImpl.

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapSupportImpl.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        storageProvider.setLockingCap(newLockingCap);
+        logger.info("[increaseLockingCap] {} {}", "Increased locking cap: ", newLockingCap.value);
+
+        return true;
+    }
+
+    @Override
+    public void save(){
```
apancorb, line 76: nit
```suggestion
    public void save() {
```

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapSupportImpl.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        // New Locking Cap must be bigger than current Locking Cap
+        Optional<Coin> currentLockingCap = getLockingCap();
+        if (currentLockingCap.isPresent() && newLockingCap.compareTo(currentLockingCap.get()) < 0) {
+            logger.warn("[increaseLockingCap] {} {}", "Attempted value doesn't increase Locking Cap. Value attempted: ", newLockingCap.value);
+            return false;
+        }
+
+        Coin maxLockingCap = currentLockingCap.get().multiply(constants.getIncrementsMultiplier());
```
apancorb, line 63: is there any chance the `get()` will fail?

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapSupportImpl.java
wilmerrootstock: Done!

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapSupportImpl.java
wilmerrootstock: To use `Optional` unveiled different places where we can get `NPE`. In this case, I added a new `IF` statement to ensure we have a current LockingCap value to avoid this NPE. If we don't have a current Locking Cap value that means RSKIP134 is not Active, so we can't set a Locking Cap value.

C.C.: @marcos-iov @apancorb

marcos-iov (approved):

PR #2604, opened 2024-07-18
Adapt BridgeSupport Methods for LockingCap Refactor

wilmerrootstock:
## Description
After the `LockingCapSupport` interface has been created, we will use it on the `BridgeSupport`. Adapt `LockingCap` methods in the `BridgeSupport`. Use the new interface created, `LockingCapSupport`, under `co.rsk.peg.lockingcap` package to create `LockingCap` behavior.

## Motivation and Context

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -187,7 +187,7 @@ public class BridgeSupport {
         feePerKbSupport.save();
         whitelistSupport.save();
         federationSupport.save();
-        // TODO: save lockingCapSupport.save();
+        lockingCapSupport.save(activations);
     }
 
     /**
@@ -2113,14 +2113,7 @@ public class BridgeSupport {
     }
 
     public Coin getLockingCap() {
-        // Before returning the locking cap, check if it was already set
-        if (activations.isActive(ConsensusRule.RSKIP134) && this.provider.getLockingCap() == null) {
-            // Set the initial locking cap value
-            logger.debug("Setting initial locking cap value");
-            this.provider.setLockingCap(bridgeConstants.getInitialLockingCap());
-        }
-
-        return this.provider.getLockingCap();
+        return lockingCapSupport.getLockingCap().get();
     }
 
     /**
@@ -2132,29 +2125,8 @@ public class BridgeSupport {
         return federationSupport.getActiveFederationRedeemScript();
     }
 
-    public boolean increaseLockingCap(Transaction tx, Coin newCap) {
-        // Only pre-configured addresses can modify locking cap
-        AddressBasedAuthorizer authorizer = bridgeConstants.getIncreaseLockingCapAuthorizer();
-        if (!authorizer.isAuthorized(tx, signatureCache)) {
-            logger.warn("not authorized address tried to increase locking cap. Address: {}", tx.getSender(signatureCache));
-            return false;
-        }
-        // new locking cap must be bigger than current locking cap
-        Coin currentLockingCap = this.getLockingCap();
-        if (newCap.compareTo(currentLockingCap) < 0) {
-            logger.warn("attempted value doesn't increase locking cap. Attempted: {}", newCap.value);
-            return false;
-        }
-        Coin maxLockingCap = currentLockingCap.multiply(bridgeConstants.getLockingCapIncrementsMultiplier());
-        if (newCap.compareTo(maxLockingCap) > 0) {
-            logger.warn("attempted value increases locking cap above its limit. Attempted: {}", newCap.value);
-            return false;
-        }
-
-        logger.info("increased locking cap: {}", newCap.value);
-        this.provider.setLockingCap(newCap);
-
-        return true;
+    public boolean increaseLockingCap(Transaction tx, Coin newLockingCap) {
+        return lockingCapSupport.increaseLockingCap(tx, newLockingCap);
     }
 
     public void registerBtcCoinbaseTransaction(byte[] btcTxSerialized, Sha256Hash blockHash, byte[] pmtSerialized, Sha256Hash witnessMerkleRoot, byte[] witnessReservedValue) throws VMException {
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -2740,16 +2740,16 @@ public class BridgeSupport {
     }
 
     private boolean verifyLockDoesNotSurpassLockingCap(BtcTransaction btcTx, Coin totalAmount) {
-        if (!activations.isActive(ConsensusRule.RSKIP134)) {
+        Optional<Coin> lockingCap = lockingCapSupport.getLockingCap();
+        if (!lockingCap.isPresent()) {
             return true;
         }
 
         Coin fedCurrentFunds = getBtcLockedInFederation();
-        Optional<Coin> lockingCap = lockingCapSupport.getLockingCap();
         logger.trace("Evaluating locking cap for: TxId {}. Value to lock {}. Current funds {}. Current locking cap {}", btcTx.getHash(true), totalAmount, fedCurrentFunds, lockingCap);
         Coin fedUTXOsAfterThisLock = fedCurrentFunds.add(totalAmount);
         // If the federation funds (including this new UTXO) are smaller than or equals to the current locking cap, we are fine.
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -2108,14 +2113,7 @@ public Integer setLockWhitelistDisableBlockDelay(Transaction tx, BigInteger disa
...
     public Coin getLockingCap() {
-        // Before returning the locking cap, check if it was already set
-        if (activations.isActive(ConsensusRule.RSKIP134) && this.provider.getLockingCap() == null) {
-            // Set the initial locking cap value
-            logger.debug("Setting initial locking cap value");
-            this.provider.setLockingCap(bridgeConstants.getInitialLockingCap());
-        }
-
-        return this.provider.getLockingCap();
+        return lockingCapSupport.getLockingCap().get();
```
apancorb, line 2116: This might throw a `NoSuchElementException` if there is no value present. Maybe adding something like `.orElse` in order to ensure a value is always being returned. wdyt?

wilmerrootstock: This might be redundant since in the `lockingCapSupport.getLockingCap()` method we ensure this returns a valid value. If this method hasn't any value, it sets the locking cap value from `LockingCapConstants.getInitialValue()`. We can do it again in an `orElse(LockingCapConstants.getInitialValue())` method, but I think this is redundant. WDYT?

[Take a look at the code of this PR.](https://github.com/rsksmart/rskj/blob/0b5663dac1a2354405099e4d03ec7e1b1a70612e/rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapSupportImpl.java#L39)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -183,6 +187,7 @@ public void save() throws IOException {
         feePerKbSupport.save();
         whitelistSupport.save();
         federationSupport.save();
+        lockingCapSupport.save(activations);
```
julia-zack, line 190: as discussed, this `save` method shouldn't receive `activations` as a parameter, because the field would be part of locking cap support implementation.

See how this was implemented in the other supports, like in whitelist support impl for example:
https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupportImpl.java#L29,
https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/main/java/co/rsk/peg/whitelist/WhitelistSupportImpl.java#L172

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapSupport.java
```diff
@@ -2,11 +2,14 @@
...
+import org.ethereum.config.blockchain.upgrades.ActivationConfig;
 import org.ethereum.core.Transaction;
 
 public interface LockingCapSupport {
 
     Optional<Coin> getLockingCap();
 
     boolean increaseLockingCap(Transaction tx, Coin newCap);
+
+    void save(ActivationConfig.ForBlock activations);
```
julia-zack, line 14: remove the `activations` argument

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
julia-zack, line 1: pretty sure this change is already part of other prs

wilmerrootstock: Yes, it's a queue of PRs and we are putting what we need to move forward.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
apancorb: I see, if it guranteed to always return a value there is no need to use an Optional here. wdyt?

wilmerrootstock: Make sense. I'll remove it.

wilmerrootstock: Done!

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
wilmerrootstock: Done!

rskj-core/src/main/java/co/rsk/peg/lockingcap/LockingCapSupport.java
```diff
@@ -1,12 +1,13 @@
 package co.rsk.peg.lockingcap;
 
 import co.rsk.bitcoinj.core.Coin;
-import java.util.Optional;
 import org.ethereum.core.Transaction;
 
 public interface LockingCapSupport {
 
-    Optional<Coin> getLockingCap();
+    Coin getLockingCap();
```
marcos-iov, line 8: Didn't we talk about this? It returns empty pre RSKIP 134, doesn't it?

wilmerrootstock: You're right! Then, I'll apply this method for this Optional, `orElse(Optional.empty())`

wilmerrootstock: Done!

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
wilmerrootstock: @apancorb We got back the `Optional` since there's a case where we need to handle a `null` value. This is the workaround: `orElse(null)`, keeping the same behavior that existed previously.

apancorb (approved):
LGTM

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -2768,11 +2745,11 @@ private boolean verifyLockDoesNotSurpassLockingCap(BtcTransaction btcTx, Coin to
         }
 
         Coin fedCurrentFunds = getBtcLockedInFederation();
-        Coin lockingCap = this.getLockingCap();
+        Optional<Coin> lockingCap = lockingCapSupport.getLockingCap();
         logger.trace("Evaluating locking cap for: TxId {}. Value to lock {}. Current funds {}. Current locking cap {}", btcTx.getHash(true), totalAmount, fedCurrentFunds, lockingCap);
         Coin fedUTXOsAfterThisLock = fedCurrentFunds.add(totalAmount);
         // If the federation funds (including this new UTXO) are smaller than or equals to the current locking cap, we are fine.
-        if (fedUTXOsAfterThisLock.compareTo(lockingCap) <= 0) {
+        if (lockingCap.isPresent() && fedUTXOsAfterThisLock.compareTo(lockingCap.get()) <= 0) {
```
marcos-iov, line 2752: This is a bit confusing to me, it implies that lockingCap value could not be present and in that case it will return `false` indicating that the locking cap exceeded.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -2768,11 +2745,11 @@ private boolean verifyLockDoesNotSurpassLockingCap(BtcTransaction btcTx, Coin to
         }
 
         Coin fedCurrentFunds = getBtcLockedInFederation();
-        Coin lockingCap = this.getLockingCap();
+        Optional<Coin> lockingCap = lockingCapSupport.getLockingCap();
```
marcos-iov, line 2748: The only case when this returns empry is when RSKIP134 is not active, is that so?

Because if that's the case we could get rid change the part above 
```if (!activations.isActive(ConsensusRule.RSKIP134)) {
            return true;
        }```

for 

```
if (!lockingCap.isPresent()) return true
```
Would make it cleaner and clearer I believe

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
wilmerrootstock: Done!

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
wilmerrootstock: Done! Removed the `IF` statement related to `activations`, and added a new validation related to the `LockingCap` value to improve Readiness and Maintainability.

marcos-iov (approved):

PR #2609, opened 2024-07-18
Adds create call twice by same authorizer test

jeremy-then:
Add create already voted test

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
@@ -2283,12 +2283,12 @@ class FederationSupportImplTest {
         }
 
         @Test
-        void voteFederationChange_whenCreateCallTwiceBySameAuthorizer_returnsGenericResponseCode() {
+        void voteFederationChange_whenCreateIsCalledTwiceBySameAuthorizer_returnsGenericErrorResponseCode() {
 
             // Arrange
 
-            Transaction tx = getTransactionFromCaller(FederationChangeCaller.FIRST_AUTHORIZED.getRskAddress());
-            ABICallSpec abiCallSpec = new ABICallSpec("create", new byte[][]{});
+            Transaction tx = TransactionUtils.getTransactionFromCaller(signatureCache, FederationChangeCaller.FIRST_AUTHORIZED.getRskAddress());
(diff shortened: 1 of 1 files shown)


rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2270,6 +2270,28 @@ void voteFederationChange_create_returnsSuccessfulResponseCode() {
 
         }
 
+        @Test
+        void voteFederationChange_whenCreateCallTwiceBySameAuthorizer_returnsGenericResponseCode() {
```
julia-zack, line 2274: __returnsGenericErrorResponseCode_

julia-zack: __whenCreateIsCalledTwiceBySameAuthorizer_

marcos-iov (approved):
Looks like we are missing a test where we call create with 2 different authorizers, then assert that a new pending federation was created.

Unless that test is somewhere else, please a create a task to do it

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java

julia-zack (approved):

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2282,6 +2282,28 @@ void voteFederationChange_create_returnsSuccessfulResponseCode() {
...
+
+            // Arrange
+
+            Transaction tx = TransactionUtils.getTransactionFromCaller(signatureCache, FederationChangeCaller.FIRST_AUTHORIZED.getRskAddress());
+            ABICallSpec abiCallSpec = new ABICallSpec(FederationChangeFunction.CREATE.getKey(), new byte[][]{});
+
+            // Act
+
+            // First create call
+            federationSupport.voteFederationChange(tx, abiCallSpec, signatureCache, bridgeEventLogger);
```
marcos-iov, line 2296: We should assert the first call was successful

jeremy-then: I was following previous suggestions, about avoiding to test what has been already tested. The test exactly right before this up to this point is exactly the same and tested this. But it won't hurt to assert it.

marcos-iov (approved):

PR #2617, opened 2024-07-18
Add validation period duration to federation constants

julia-zack:


rskj-core/src/main/java/co/rsk/peg/federation/constants/FederationConstants.java
```diff
@@ -18,6 +18,7 @@ public class FederationConstants {
 
     protected AddressBasedAuthorizer federationChangeAuthorizer;
     protected String oldFederationAddress;
+    protected long validationPeriodDuration;
```
marcos-iov, line 21:
```suggestion
    protected long validationPeriodAge;
```
Not sure if it improves the naming, but it's consisten with the rest. What do you think?

marcos-iov (approved):

PR #2618, opened 2024-07-18
Add new storage entries

julia-zack:
- Add `proposedFederation` and `svpFundTxHashUnsigned` storage entries in `FederationStorageProviderImpl`


rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -29,6 +30,7 @@ public class FederationStorageProviderImpl implements FederationStorageProvider
     private Federation oldFederation;
     private boolean shouldSaveOldFederation = false;
 
+    private Federation proposedFederation;
```
marcos-iov, line 33: Should we add these 2 fields at this time? They are not used

rskj-core/src/main/java/co/rsk/peg/storage/FederationStorageIndexKey.java
```diff
@@ -12,18 +12,21 @@ public enum FederationStorageIndexKey {
     NEW_FEDERATION_KEY("newFederation"),
     OLD_FEDERATION_KEY("oldFederation"),
     PENDING_FEDERATION_KEY("pendingFederation"),
+    PROPOSED_FEDERATION_KEY("proposedFederation"),
```
marcos-iov, line 15: We should probably get rid of the _KEY suffix. Getting rid of it in all the keys might have a big impact so we could do it in a separate task. But let's at least not add it to new keys

julia-zack: yes, i've signed up a note for doing exactly that. But i'll not add this `_KEY`, you right.

julia-zack: Done. And i created the task to take care of the rest

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julia-zack: Done

marcos-iov (approved):

PR #2619, opened 2024-07-18
Add methods to save proposedFederation entries

julia-zack:
- Create `setProposedFederation` public method in `FederationStorageProviderImpl` to set the proposed federation

- Create `saveProposedFederation` private method in `FederationStorageProviderImpl` to save the serialized proposed federation in the repository

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
@@ -373,8 +373,7 @@ public class FederationStorageProviderImpl implements FederationStorageProvider 
     }
 
     private void saveProposedFederation(Federation proposedFederation) {
-        // we only need to save the standard part of the fed since the emergency part is constant
-        saveFederationFormatVersion(PROPOSED_FEDERATION_FORMAT_VERSION.getKey(), STANDARD_MULTISIG_FEDERATION.getFormatVersion());
+        saveFederationFormatVersion(PROPOSED_FEDERATION_FORMAT_VERSION.getKey(), proposedFederation.getFormatVersion());
         bridgeStorageAccessor.saveToRepository(PROPOSED_FEDERATION.getKey(), proposedFederation, BridgeSerializationUtils::serializeFederation);
     }
 
--- a/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
@@ -29,9 +29,11 @@ public class FederationStorageProviderImpl implements FederationStorageProvider 
     private Federation oldFederation;
     private boolean shouldSaveOldFederation = false;
 
-    private Federation proposedFederation;
     private PendingFederation pendingFederation;
     private boolean shouldSavePendingFederation = false;
+
+    private Federation proposedFederation;
+    private boolean proposedFederationIsSet = false;
 
     private ABICallElection federationElection;
 
@@ -209,6 +211,7 @@ public class FederationStorageProviderImpl implements FederationStorageProvider 
     @Override
     public void setProposedFederation(Federation proposedFederation) {
         this.proposedFederation = proposedFederation;
+        proposedFederationIsSet = true;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -367,6 +372,11 @@ private void savePendingFederation(ActivationConfig.ForBlock activations) {
         bridgeStorageAccessor.saveToRepository(PENDING_FEDERATION_KEY.getKey(), serializedPendingFederation);
     }
 
+    private void saveProposedFederation(Federation proposedFederation) {
+        saveFederationFormatVersion(PROPOSED_FEDERATION_FORMAT_VERSION.getKey(), proposedFederation.getFormatVersion());
```
julia-zack, line 376: since the proposed federation is something added after rskip123, there's no need to check that when saving the format version

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -367,6 +375,15 @@ private void savePendingFederation(ActivationConfig.ForBlock activations) {
         bridgeStorageAccessor.saveToRepository(PENDING_FEDERATION_KEY.getKey(), serializedPendingFederation);
     }
 
+    private void saveProposedFederation(Federation proposedFederation) {
+        if (!isProposedFederationSet) {
```
marcos-iov, line 379: Should we also check for RSKIP activation? To make sure nothing is stored before the RSKIP activates

julia-zack: i thought about that but i don't think that's needed, because before rskip `isProposedFederationSet` should be always false

julia-zack: Done

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
jeremy-then: What if the `setProposedFederation(Federation)` method is called with `null`? This line would fail.

Should we reset the `isProposedFederationSet` flag if the `setProposedFederation(Federation)` method is called with `null`?

julia-zack: good catch jere. Thanks. This requires some analysis.

julia-zack: solved.

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -367,6 +375,28 @@ private void savePendingFederation(ActivationConfig.ForBlock activations) {
...
+    private void saveProposedFederation(ActivationConfig.ForBlock activations) {
+        if (!activations.isActive(RSKIP419)) {
+            return;
+        }
+
+        if (!isProposedFederationSet) {
+            return;
+        }
+
+        Integer formatVersion = getProposedFederationFormatVersion();
```
marcos-iov, line 387:
```suggestion
        Integer formatVersion = Optional.ofNullable(proposedFederation.getFormatVersion()).orElse(null);
```

jeremy-then (approved):
LGTM.

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -367,6 +375,20 @@ private void savePendingFederation(ActivationConfig.ForBlock activations) {
         bridgeStorageAccessor.saveToRepository(PENDING_FEDERATION_KEY.getKey(), serializedPendingFederation);
     }
 
+    private void saveProposedFederation(ActivationConfig.ForBlock activations) {
+        if (!activations.isActive(RSKIP419)) {
+            return;
+        }
+
+        if (!isProposedFederationSet) {
```
marcos-iov, line 383:
```suggestion
        if (!activations.isActive(RSKIP419) || !isProposedFederationSet) {
            return;
        }
```

marcos-iov: Or move it a private method if you prefer

julia-zack: Done

marcos-iov (approved):

PR #2621, opened 2024-07-19
Add same fed twice with same authorizer test

jeremy-then:
Add same fed twice with same authorizer test

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
@@ -2565,18 +2565,22 @@ class FederationSupportImplTest {
 
             // Arrange
 
-            BtcECKey expectedBtcECKey = new BtcECKey();
+            List<BtcECKey> fedKeys = BitcoinTestUtils.getBtcEcKeysFromSeeds(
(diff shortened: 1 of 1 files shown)


rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2549,6 +2549,44 @@ void voteFederationChange_rollback_returnsSuccessResponseCode() {
 
         }
 
+        @Test
+        void voteFederationChange_addSameFederatorPublicKeyBySameAuthorizerTwice_returnsGenericResponseCodeOnSecondCall() {
```
julia-zack, line 2553: why is it failing? because of the same key or because the same authorizer voted? Is not clear by the test name.
I'd simply put `_addFederatorPublicKeyBySameAuthorizerTwice`, sounds better imho

julia-zack: and just `_returnsGenericErrorResponseCode`, i think is implied that it would be on the second call :)

nathanieliov (approved):

marcos-iov (approved):

PR #2623, opened 2024-07-19
Add 100 members test

jeremy-then:
Add 100 members test

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
@@ -2673,8 +2673,14 @@ class FederationSupportImplTest {
             // Voting add new fed with m of n authorizers
 
             for(int i = 0; i < EXPECTED_COUNT_OF_MEMBERS; i++) {
-                BtcECKey expectedBtcECKey = new BtcECKey();
-                ABICallSpec addFederationAbiCallSpec = new ABICallSpec(FederationChangeFunction.ADD.getKey(), new byte[][]{expectedBtcECKey.getPubKey()});
+                BtcECKey federatorBtcECKey = BtcECKey.fromPrivate(BigInteger.valueOf(i + 100));
+                ECKey federatorRskKey = ECKey.fromPrivate(BigInteger.valueOf(i + 101));
+                ECKey federatorMstKey = ECKey.fromPrivate(BigInteger.valueOf(i + 102));
+                ABICallSpec addFederationAbiCallSpec = new ABICallSpec(FederationChangeFunction.ADD_MULTI.getKey(), new byte[][]{
+                    federatorBtcECKey.getPubKey(),
+                    federatorRskKey.getPubKey(),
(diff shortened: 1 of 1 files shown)


rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack, line 1: what is the purpose of this test?

jeremy-then: The purpose is to show that the Bridge is open to add any number of federators to pending federation. Although we know that it will not be commited due to the script size. But in a future, a larger federation will be able to be commited.

julia-zack: but a test showing that creating a federation with over 10 members is possible but that committing it throws an error. what's the point of creating one with 100 members?

jeremy-then: This is more like stress testing for the Bridge. We can discuss it with the rest of the team.

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2637,6 +2637,41 @@ void voteFederationChange_addFederatorMultiKeyWithDifferentMstKey_returnsSuccess
...
+            // Voting with m of n authorizers to create the pending federation
+            federationSupport.voteFederationChange(firstAuthorizedTx, createFederationAbiCallSpec, signatureCache, bridgeEventLogger);
+            federationSupport.voteFederationChange(secondAuthorizedTx, createFederationAbiCallSpec, signatureCache, bridgeEventLogger);
+
+            // Act
+
+            // Voting add new fed with m of n authorizers
+
+            for(int i = 0; i < EXPECTED_COUNT_OF_MEMBERS; i++) {
+                BtcECKey expectedBtcECKey = new BtcECKey();
```
marcos-iov, line 2661: Let's make it deterministic, you can use BitcoinTestUtils class and pass `i` value as the seed

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2637,6 +2637,41 @@ void voteFederationChange_addFederatorMultiKeyWithDifferentMstKey_returnsSuccess
...
+                ABICallSpec addFederationAbiCallSpec = new ABICallSpec(FederationChangeFunction.ADD.getKey(), new byte[][]{expectedBtcECKey.getPubKey()});
```
marcos-iov, line 2662: This is not right, this method is disabled post RSKIP 123

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java

marcos-iov (approved):

PR #2626, opened 2024-07-21
Commit federation with over 15 members throws exception test

jeremy-then:


rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2766,6 +2765,62 @@ void voteFederationChange_commit10MembersFederation_returnsSuccessResponseCodeAn
...
+                federationSupport.voteFederationChange(tx2, addFederationAbiCallSpec, signatureCache, bridgeEventLogger);
+            }
+
+            Keccak256 pendingFederationHash = federationSupport.getPendingFederationHash();
+
+            ABICallSpec commitFederationAbiCallSpec = new ABICallSpec(FederationChangeFunction.COMMIT.getKey(), new byte[][]{pendingFederationHash.getBytes()});
+
+            // Act
+
+            // Voting commit new fed with m of n authorizers
```
julia-zack, line 2813: wdyt about adding _last needed vote throws an exception_, something like that?

jeremy-then: The `Exception exception = assertThrows...` is already clearly doing that.

marcos-iov (approved):

PR #2628, opened 2024-07-22
Adding SNAP capability

rmoreliovlabs:
## Description
I have implemented the analysis documentation that explores various options for enabling SNAP feature discovery among connected peers. You can view the detailed analysis here: [SNAP Feature Discovery Analysis](https://www.notion.so/iovlabs/DRAFT-Analyse-possible-options-for-SNAP-feature-discovery-among-connected-peers-148f805e06fc4b778cb5bf8a21059a7c)

## Motivation and Context
Establishing a decentralized approach for RSKj nodes to seamlessly discover and establish connections with SNAP-sync-provider peers during their initial connection phase. This enhancement will focus on integrating a new capability into the HelloMessage and HandshakeHandler classes, thereby facilitating efficient peer discovery and connectivity within the network

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
+++ b/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
@@ -411,6 +411,7 @@ public class RskSystemProperties extends SystemProperties {
     public boolean isServerSnapshotSyncEnabled() { return configFromFiles.getBoolean("sync.snapshot.server.enabled");}
     public boolean isClientSnapshotSyncEnabled() { return configFromFiles.getBoolean("sync.snapshot.client.enabled");}
 
+    @Override
     public List<String> peerCapabilities() {
         List<String> capabilities = super.peerCapabilities();
 
--- a/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
+++ b/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
@@ -411,7 +411,6 @@ public class RskSystemProperties extends SystemProperties {
     public boolean isServerSnapshotSyncEnabled() { return configFromFiles.getBoolean("sync.snapshot.server.enabled");}
     public boolean isClientSnapshotSyncEnabled() { return configFromFiles.getBoolean("sync.snapshot.client.enabled");}
 
-    @Override
     public List<String> peerCapabilities() {
         List<String> capabilities = super.peerCapabilities();
 
--- a/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
+++ b/rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
@@ -31,6 +31,7 @@ import org.ethereum.core.Account;
 import org.ethereum.crypto.ECKey;
 import org.ethereum.crypto.HashUtil;
 import org.ethereum.listener.GasPriceCalculator;
+import org.ethereum.net.client.Capability;
 
 import javax.annotation.Nullable;
 import java.nio.charset.StandardCharsets;
@@ -409,6 +410,17 @@ public class RskSystemProperties extends SystemProperties {
 
     public boolean isServerSnapshotSyncEnabled() { return configFromFiles.getBoolean("sync.snapshot.server.enabled");}
     public boolean isClientSnapshotSyncEnabled() { return configFromFiles.getBoolean("sync.snapshot.client.enabled");}
+
+    @Override
+    public List<String> peerCapabilities() {
+        List<String> capabilities = super.peerCapabilities();
+
+        if (isServerSnapshotSyncEnabled()) {
+            capabilities.add(Capability.SNAP);
+        }
+
+        return capabilities;
+    }
 
     public int getSnapshotChunkTimeout() {
         return configFromFiles.getInt("sync.snapshot.client.chunkRequestTimeout");
--- a/rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
+++ b/rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
@@ -156,7 +156,7 @@ public class PeersInformation implements SnapshotPeersInformation {
     public Optional<Peer> getBestSnapPeer() {
         return getBestPeer(
                 getBestCandidatesStream()
-                .filter(this::isSnapPeerCandidate)
+                        .filter(entry -> isSnapPeerCandidate(entry) && entry.getKey().isSnapCapable())
         );
     }
 
@@ -199,7 +199,7 @@ public class PeersInformation implements SnapshotPeersInformation {
     @Override
     public List<Peer> getBestSnapPeerCandidates() {
         return getBestCandidatesStream()
-                .filter(this::isSnapPeerCandidate)
+                .filter(entry -> isSnapPeerCandidate(entry) && entry.getKey().isSnapCapable())
                 .map(Map.Entry::getKey)
                 .collect(Collectors.toList());
     }
--- a/rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
+++ b/rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
@@ -156,7 +156,7 @@ public class PeersInformation implements SnapshotPeersInformation {
     public Optional<Peer> getBestSnapPeer() {
         return getBestPeer(
                 getBestCandidatesStream()
-                        .filter(entry -> isSnapPeerCandidate(entry) || entry.getKey().isSnapCapable())
+                        .filter(this::isSnapPeerCandidateOrCapable)
         );
     }
 
@@ -199,7 +199,7 @@ public class PeersInformation implements SnapshotPeersInformation {
     @Override
     public List<Peer> getBestSnapPeerCandidates() {
         return getBestCandidatesStream()
-                .filter(entry -> isSnapPeerCandidate(entry) || entry.getKey().isSnapCapable())
+                .filter(this::isSnapPeerCandidateOrCapable)
                 .map(Map.Entry::getKey)
                 .collect(Collectors.toList());
     }
@@ -288,5 +288,9 @@ public class PeersInformation implements SnapshotPeersInformation {
 
     private boolean isSnapPeerCandidate(Map.Entry<Peer, SyncPeerStatus> entry) {
         return syncConfiguration.getNodeIdToSnapshotTrustedPeerMap().containsKey(entry.getKey().getPeerNodeID().toString());
+    }
+
+    private boolean isSnapPeerCandidateOrCapable(Map.Entry<Peer, SyncPeerStatus> entry) {
+        return isSnapPeerCandidate(entry) || entry.getKey().isSnapCapable();
     }
 }
--- a/rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
+++ b/rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
@@ -199,7 +199,7 @@ public class PeersInformation implements SnapshotPeersInformation {
     @Override
     public List<Peer> getBestSnapPeerCandidates() {
         return getBestCandidatesStream()
-                .filter(this::isSnapPeerCandidateOrCapable)
+                .filter(entry -> entry.getKey().isSnapCapable())
                 .map(Map.Entry::getKey)
                 .collect(Collectors.toList());
     }
--- a/rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
+++ b/rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
@@ -156,7 +156,7 @@ public class PeersInformation implements SnapshotPeersInformation {
     public Optional<Peer> getBestSnapPeer() {
         return getBestPeer(
                 getBestCandidatesStream()
-                        .filter(entry -> isSnapPeerCandidate(entry) && entry.getKey().isSnapCapable())
+                        .filter(entry -> isSnapPeerCandidate(entry) || entry.getKey().isSnapCapable())
         );
     }
 
@@ -199,7 +199,7 @@ public class PeersInformation implements SnapshotPeersInformation {
     @Override
     public List<Peer> getBestSnapPeerCandidates() {
         return getBestCandidatesStream()
-                .filter(entry -> isSnapPeerCandidate(entry) && entry.getKey().isSnapCapable())
+                .filter(entry -> isSnapPeerCandidate(entry) || entry.getKey().isSnapCapable())
                 .map(Map.Entry::getKey)
                 .collect(Collectors.toList());
     }
--- a/rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
+++ b/rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
@@ -43,15 +43,19 @@ public class ConfigCapabilitiesImpl implements ConfigCapabilities{
 
     public ConfigCapabilitiesImpl(RskSystemProperties config) {
         if (config.syncVersion() != null) {
-            EthVersion eth =fromCode(config.syncVersion());
+            EthVersion eth = fromCode(config.syncVersion());
             if (eth != null) {
                 allCaps.add(new Capability(RSK, eth.getCode()));
-                if (config.isServerSnapshotSyncEnabled()) allCaps.add(new Capability(SNAP, eth.getCode()));
+                if (config.isServerSnapshotSyncEnabled()){
+                    allCaps.add(new Capability(SNAP, eth.getCode()));
+                }
             }
         } else {
             for (EthVersion v : EthVersion.supported()) {
                 allCaps.add(new Capability(RSK, v.getCode()));
-                if (config.isServerSnapshotSyncEnabled()) allCaps.add(new Capability(SNAP, v.getCode()));
+                if (config.isServerSnapshotSyncEnabled()){
+                    allCaps.add(new Capability(SNAP, v.getCode()));
+                }
             }
         }
         this.config = config;
--- a/rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
+++ b/rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
@@ -39,27 +39,27 @@ public class ConfigCapabilitiesImpl implements ConfigCapabilities{
 
     private final RskSystemProperties config;
 
-    private SortedSet<Capability> allCaps = new TreeSet<>();
+    private SortedSet<Capability> allCapabilities = new TreeSet<>();
 
     public ConfigCapabilitiesImpl(RskSystemProperties config) {
         if (config.syncVersion() != null) {
             EthVersion eth = fromCode(config.syncVersion());
             if (eth != null) {
-                allCaps.add(new Capability(RSK, eth.getCode()));
-                if (config.isServerSnapshotSyncEnabled()){
-                    allCaps.add(new Capability(SNAP, eth.getCode()));
-                }
+                allCapabilities.add(new Capability(RSK, eth.getCode()));
             }
         } else {
             for (EthVersion v : EthVersion.supported()) {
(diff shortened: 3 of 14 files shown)


rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
```diff
@@ -29,26 +29,29 @@
...
-    private final SystemProperties config;
+    private final RskSystemProperties config;
 
     private SortedSet<Capability> allCaps = new TreeSet<>();
 
-    public ConfigCapabilitiesImpl(SystemProperties config) {
+    public ConfigCapabilitiesImpl(RskSystemProperties config) {
         if (config.syncVersion() != null) {
-            EthVersion eth = fromCode(config.syncVersion());
+            EthVersion eth =fromCode(config.syncVersion());
```
asoto-iov, line 46: minor thing but it is better with the space 😓

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
```diff
@@ -29,26 +29,29 @@
...
             if (eth != null) {
                 allCaps.add(new Capability(RSK, eth.getCode()));
+                if (config.isServerSnapshotSyncEnabled()) allCaps.add(new Capability(SNAP, eth.getCode()));
```
asoto-iov, line 49: Please use if with '{ }'. Why are we using `eth.getCode()` for snap aswell?  Maybe it is worth to have Snap protocol version aswell, just in case we modify something in the future.

rskj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
```diff
@@ -377,14 +377,14 @@ private void processHelloMessage(ChannelHandlerContext ctx, HelloMessage helloMe
         List<Capability> capInCommon = configCapabilities.getSupportedCapabilities(helloMessage);
         channel.initMessageCodes(capInCommon);
         for (Capability capability : capInCommon) {
-            // It seems that the only supported capability is RSK, and everything else is ignored.
-            if (Capability.RSK.equals(capability.getName())) {
+            // Accept RSK or SNAP capabilities.
+            if (Capability.RSK.equals(capability.getName()) || Capability.SNAP.equals(capability.getName())) {
```
asoto-iov, line 381: It is correct to have an OR condition here? I mean, does it has sense to do snap sync if the node does not support RSK?

asoto-iov (changes requested):
As a summary I think that before approve we should discuss about:
-  having a snap sync versioning. 
- Review if it has sense to finish the handshake if the RSK protocol is not supported.

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
nagarev, line 46:
```suggestion
            EthVersion eth = fromCode(config.syncVersion());
```

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
nagarev, line 49:
```suggestion
                if (config.isServerSnapshotSyncEnabled()) {
                    allCaps.add(new Capability(SNAP, eth.getCode()));
                }
```

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
```diff
@@ -29,26 +29,29 @@
...
             }
         } else {
             for (EthVersion v : EthVersion.supported()) {
                 allCaps.add(new Capability(RSK, v.getCode()));
+                if (config.isServerSnapshotSyncEnabled()) allCaps.add(new Capability(SNAP, v.getCode()));
```
nagarev, line 54:
```suggestion
                if (config.isServerSnapshotSyncEnabled()){
                    allCaps.add(new Capability(SNAP, v.getCode()));
                }
```

rskj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
rmoreliovlabs: As discussed in the document [here](https://www.notion.so/iovlabs/Analyse-possible-options-for-SNAP-feature-discovery-among-connected-peers-148f805e06fc4b778cb5bf8a21059a7c), both capabilities are independent from one another. A peer supporting the SNAP protocol does not necessarily need to support the RSK protocol. Our current design doesn't support nested capabilities

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
rmoreliovlabs: solved

rmoreliovlabs: solved

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
rmoreliovlabs: solved

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
rmoreliovlabs: solved

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
```diff
@@ -29,26 +29,33 @@
...
                 allCaps.add(new Capability(RSK, eth.getCode()));
+                if (config.isServerSnapshotSyncEnabled()){
+                    allCaps.add(new Capability(SNAP, eth.getCode()));
+                }
             }
         } else {
             for (EthVersion v : EthVersion.supported()) {
                 allCaps.add(new Capability(RSK, v.getCode()));
+                if (config.isServerSnapshotSyncEnabled()){
+                    allCaps.add(new Capability(SNAP, v.getCode()));
```
Vovchyk, line 57: I guess it should be version `1` for `snap`, as `snap` and `rsk` protocol versions are kinda independent

Also I'd suggest to extract adding `snap` capability out of this `if-else`, eg:
```
        if (config.isServerSnapshotSyncEnabled() && allCaps.stream().anyMatch(Capability::isRSK)) {
            allCaps.add(new Capability(SNAP, (byte) 1));
        }
```

Vovchyk (changes requested):
now we would also need to filer snap-capable peers in [here](https://github.com/rsksmart/rskj/blob/8c8d72e47314f4df67ca7ba10e6bc2ef7e91424b/rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java#L200). And for that we'd need to extend `Peer` interface with new method - `boolean isSnapCapable()`, wdyt?

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
rmoreliovlabs: addressed, thanks for the suggestion! I've implemented it so that the SNAP capability now uses version` (byte) 1`

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
rmoreliovlabs: solved, SNAP capability now uses version `(byte) 1`

asoto-iov:
it is still missing what Vlad mentioned here: https://github.com/rsksmart/rskj/pull/2628#pullrequestreview-2212688058

rskj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
```diff
@@ -377,13 +377,15 @@ private void processHelloMessage(ChannelHandlerContext ctx, HelloMessage helloMe
         List<Capability> capInCommon = configCapabilities.getSupportedCapabilities(helloMessage);
         channel.initMessageCodes(capInCommon);
         for (Capability capability : capInCommon) {
-            // It seems that the only supported capability is RSK, and everything else is ignored.
+            // Accept RSK capability.
```
asoto-iov, line 380: I would say, RSK is the condition needed to be able to finish the Handshake successfully.

rskj-core/src/main/java/org/ethereum/net/server/Channel.java
```diff
@@ -263,6 +280,12 @@ public void imported(boolean best) {
         stats.imported(best);
     }
 
+    @Override
+    public boolean isSnapCapable() {
+        return capabilities.stream()
```
asoto-iov, line 285: Capabilities shouldn't vary over time, so I won't do this filtering every time we need to check this condition. I would filter only once and store the value of the result instead of the capabilities list.

rskj-core/src/test/java/org/ethereum/net/rlpx/HandshakeHandlerTest.java
```diff
@@ -92,12 +100,19 @@ void shouldDisconnectIfNoCapabilityIsPresent() throws Exception {
     }
 
     @Test
-    void shouldDisconnectIfRskCapabilityIsMissing() throws Exception {
+    void shouldDisconnectIfRskOrSnapCapabilityAreMissing() throws Exception {
```
asoto-iov, line 103: I think the title of this test shouldn't change, as we want to test if the RSK client is not connecting without RSK capability. 

Now as we have another real capability you can use the SNAP capability aswell to verify it.

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
```diff
@@ -29,39 +29,46 @@
...
             }
         } else {
             for (EthVersion v : EthVersion.supported()) {
-                allCaps.add(new Capability(RSK, v.getCode()));
+                allCapabilities.add(new Capability(RSK, v.getCode()));
             }
         }
+
+        if (config.isServerSnapshotSyncEnabled() && allCapabilities.stream().anyMatch(Capability::isRSK)) {
+            allCapabilities.add(new Capability(SNAP, (byte) 1));
```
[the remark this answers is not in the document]
Vovchyk: yeah, why not.. as an improvement we could extract `1` value to consts.

version has type - `byte`, so it cannot be in the form of `0.X.X`. `1` is a good first version, which would be incremented in future if needed with protocol changes

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
```diff
@@ -29,39 +29,46 @@
...
+                allCapabilities.add(new Capability(RSK, eth.getCode()));
             }
         } else {
             for (EthVersion v : EthVersion.supported()) {
-                allCaps.add(new Capability(RSK, v.getCode()));
+                allCapabilities.add(new Capability(RSK, v.getCode()));
             }
         }
+
+        if (config.isServerSnapshotSyncEnabled() && allCapabilities.stream().anyMatch(Capability::isRSK)) {
```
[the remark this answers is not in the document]
rmoreliovlabs: Just to clarify, the snapshot capability is only added if two conditions are met:

- **Snapshot Sync is Enabled on the Server**: We check if the server-side snapshot sync is enabled.
- **RSK Capability is Already Present**: We confirm that the RSK capability is already in the list.

The idea is that the snapshot capability should only be added if it’s relevant and if RSK capabilities are already present. This prevents adding the snapshot capability in cases where it wouldn't make sense

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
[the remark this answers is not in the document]
rmoreliovlabs: Addressed. Added SNAP_VERSION const in the Capability class

rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
```diff
@@ -199,7 +199,7 @@ public List<Peer> getBestPeerCandidates() {
     @Override
     public List<Peer> getBestSnapPeerCandidates() {
         return getBestCandidatesStream()
-                .filter(this::isSnapPeerCandidate)
+                .filter(entry -> isSnapPeerCandidate(entry) && entry.getKey().isSnapCapable())
```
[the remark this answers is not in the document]
rmoreliovlabs: Got it! I’ve updated the methods to use an OR condition so that both discovered snap capable and config-listed peers are considered. Thanks for the feedback!

rskj-core/src/main/java/org/ethereum/net/rlpx/HandshakeHandler.java
rmoreliovlabs: solved

rskj-core/src/main/java/org/ethereum/net/server/Channel.java
rmoreliovlabs: Thanks for the suggestion! I've implemented it by adding a `isSnapCapable` variable to store the result of the check. This avoids filtering the capabilities list every time and should improve efficiency

rskj-core/src/test/java/org/ethereum/net/rlpx/HandshakeHandlerTest.java
rmoreliovlabs: solved

rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
fmacleal, line 1: suggestion:

Instead of have the same logic repeated, you could create a function and use static reference on both places.

```
private boolean isSnapPeerCandidateOrCapable(Map.Entry<Peer, SyncPeerStatus> entry) {
    return isSnapPeerCandidate(entry) || entry.getKey().isSnapCapable();
}
```

Then you could use this filter like this:

```
public Optional<Peer> getBestSnapPeer() {
    return getBestPeer(
        getBestCandidatesStream()
            .filter(this::isSnapPeerCandidateOrCapable)
    );
}
```

and

```
@Override
public List<Peer> getBestSnapPeerCandidates() {
    return getBestCandidatesStream()
        .filter(this::isSnapPeerCandidateOrCapable)
        .map(Map.Entry::getKey)
}
```

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
```diff
@@ -29,39 +29,47 @@
...
+            allCapabilities.add(new Capability(SNAP, (byte) SNAP_VERSION));
```
fmacleal, line 58: suggestion: 

Change the constant `SNAP_VERSION` to byte. It's just used here, it's 1, I don't see advantage create it as int to be casted after.

rskj-core/src/test/java/org/ethereum/net/HelloMessageTest.java
fmacleal, line 1: suggestion:

Instead of change the current Hello test, why not add another one with SNAP capability and other without it. I would also suggest have a test with `SNAP` and without `RSK` capability so we can validate that it won't be SNAP capable on that situation.

rskj-core/src/test/java/org/ethereum/net/rlpx/HandshakeHandlerTest.java
```diff
@@ -98,6 +107,13 @@ void shouldDisconnectIfRskCapabilityIsMissing() throws Exception {
         assertFalse(ch.isOpen());
     }
 
+    @Test
+    void shouldConnectWithSnapCapability() throws Exception {
+        simulateHandshakeStartedByPeer(Arrays.asList(new Capability(SNAP, (byte) SNAP_VERSION), new Capability(RSK, EthVersion.UPPER)));
+        // this will only happen when an exception is raised
+        assertTrue(ch.isOpen());
+    }
+
```
fmacleal, line 116: suggestion:

Here seems a good place to add a test with variations of the capabilities. Like, one with RSK without SNAP and check if the channel will be closed if a handshake with SNAP is attempted. Another without RSK and only with SNAP, where the channel for SNAP should also be closed.

fmacleal (commented):
Good job. I have a few comments, some improvements suggestion and also other related with tests. Let me know what you think.

rskj-core/src/main/java/co/rsk/config/RskSystemProperties.java
```diff
@@ -410,6 +411,17 @@ public int getLongSyncLimit() {
     public boolean isServerSnapshotSyncEnabled() { return configFromFiles.getBoolean("sync.snapshot.server.enabled");}
     public boolean isClientSnapshotSyncEnabled() { return configFromFiles.getBoolean("sync.snapshot.client.enabled");}
 
+    @Override
```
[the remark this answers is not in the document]
rmoreliovlabs: addressed, removed it bc it wasn't necessary for the implementation

rskj-core/src/test/java/org/ethereum/net/HelloMessageTest.java
rmoreliovlabs: Added `test5` (following the naming conventions for tests in this class) to include the SNAP capability

rskj-core/src/test/java/org/ethereum/net/rlpx/HandshakeHandlerTest.java
rmoreliovlabs: Addressed by adding a new test, `shouldDisconnectWithSnapCapabilityIfRskCapabilityIsMissing()`, to ensure that the Handshake doesn't complete if the SNAP capability is present but the RSK capability is missing

rskj-core/src/main/java/org/ethereum/net/client/ConfigCapabilitiesImpl.java
rmoreliovlabs: solved!

fmacleal (approved):
LGTM

asoto-iov (approved):

rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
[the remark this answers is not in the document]
asoto-iov: I think we shouldn't check the list of snap candidates here, what it is important is the capability. 
I mean if a node is in the list but it is not snapCapable, then we are going to start the communication but it won't attend the snap status request.

The list should be used to add these nodes to the node list if the snap sync is enabled.

asoto-iov (changes requested):
Please review the new comment. I think it is needed to review the getBestSnapPeer candidate filtering to avoid unnecessary communication with non snap capable peers.

rskj-core/src/main/java/co/rsk/net/sync/PeersInformation.java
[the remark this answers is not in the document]
rmoreliovlabs: I see your point about prioritizing snap capability over the candidate list, but that makes me wonder—what’s the purpose of having the snap candidate list in the config then? If we’re already filtering based on capability, should we just focus on all snap-capable peers, whether or not they’re in the list?

It feels like the list might not be adding much value if we’re not using it to narrow down the best snap peers. Should we be considering a different approach to how we use the list?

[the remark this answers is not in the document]
asoto-iov: This was already discussed before, the reason is to have a list of nodes pre-configured for snap sync if the feature is enabled, as same is happening with the peer discovery list. 
Apart from that I don't see any positive thing to send snap messages to a client, knowing forehand that it is not snap capable. It will only delay the sync and spread unnecessary packages in the network.

asoto-iov (approved):

PR #2629, opened 2024-07-22
Create new commit federation to be performed after RSKIP419

julia-zack:
-Create new commitFederation that adds some logic needed to have the validation process, and removes some other.
Existing commitFederation should remain the same to keep consensus before RSKIP419.
-Modify createFederation to also check for proposedFederation.
-Create new `PROPOSED_FEDERATION_ALREADY_EXISTS` response code.
-Refactor some things to make the code cleaner.
-Add tests for new behavior

_Disclaimer_
There's a test in `PowpegMigrationTest` class that is failing now. This makes sense, since is testing the behavior of old commitFederation method but with lovell activated (and therefore, with RSKIP419 activated).
It has been disabled, and another task was created to fix this test.


rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -208,12 +208,47 @@ public void setPendingFederation(PendingFederation federation) {
...
+        pendingFederation = null;
+    }
+
+    @Override
+    public Optional<Federation> getProposedFederation(FederationConstants federationConstants, ActivationConfig.ForBlock activations) {
+        if (!activations.isActive(RSKIP419)) {
+            return Optional.empty();
+        }
+
+        if (proposedFederation != null || isProposedFederationSet) {
```
jeremy-then, line 222: Similar to [this](https://github.com/rsksmart/rskj/pull/2619/files#r1687419115), with this suggested change, I think this would change to use `&&` instead of `||`.

But let me know why we are setting `isProposedFederationSet` to `true` even when the `proposedFederation` is `null`, so I can understand better.

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -208,12 +208,47 @@ public void setPendingFederation(PendingFederation federation) {
...
+            return Optional.ofNullable(proposedFederation);
+        }
+
+        Optional<Integer> storageVersion = getStorageVersion(PROPOSED_FEDERATION_FORMAT_VERSION.getKey());
+        proposedFederation = bridgeStorageAccessor.getFromRepository(
+            PROPOSED_FEDERATION.getKey(),
+            data -> {
+                if (data == null) {
```
jeremy-then, line 230: `if(isNull(data))`? From `import static java.util.Objects.isNull`. No big deal though.

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -667,6 +681,87 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLo
         return FederationChangeResponseCode.SUCCESSFUL.getCode();
     }
 
+    /**
+     * Commits the currently pending federation to be the proposed one.
+     * That is, the proposed federation is set to be the currently pending federation,
+     * and the pending federation is wiped out.
+     * @param dryRun whether to just do a dry run
+     * @param hash the pending federation's hash. This is checked the execution block's pending federation hash for equality.
```
jeremy-then, line 689: `This is checked the execution block's pending federation hash for equality.`. I don't get it is easy to get this. Could you reword it?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -667,6 +681,87 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLo
...
+    private Integer commitFederationPostRSKIP419(boolean dryRun, Keccak256 hash, BridgeEventLogger eventLogger) {
+        PendingFederation currentPendingFederation = provider.getPendingFederation();
+
+        if (currentPendingFederation == null) {
+            logger.warn("[commitFederation] Pending federation does not exist.");
+            return FederationChangeResponseCode.PENDING_FEDERATION_NON_EXISTENT.getCode();
+        }
+
+        if (!currentPendingFederation.isComplete()) {
+            logger.warn("[commitFederation] Pending federation has {} members, so it does not meet the minimum required.", currentPendingFederation.getMembers().size());
```
jeremy-then, line 702: Could the minimum required be shown in the log as well?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -667,6 +681,87 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLo
...
+            return FederationChangeResponseCode.PENDING_FEDERATION_MISMATCHED_HASH.getCode();
+        }
+
+        if (dryRun) {
+            logger.info("[commitFederation] DryRun execution successful.");
+            return FederationChangeResponseCode.SUCCESSFUL.getCode();
+        }
+
+        // Network parameters for the proposed federation are taken from the bridge constants.
+        // Creation time is the block's timestamp.
```
jeremy-then, line 717: nit, but I don't think this comment is adding any value. No big deal though. We can discuss it with the rest of the team.

rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java
```diff
@@ -225,7 +225,6 @@ private void testChangePowpeg(
         for (UTXO utxo : existingUtxos) {
             assertTrue(utxosToMigrate.stream().anyMatch(storedUtxo -> storedUtxo.equals(utxo)));
         }
-        assertTrue(federationStorageProvider.getNewFederationBtcUTXOs(btcParams, activations).isEmpty());
 
         // Trying to create a new powpeg again should fail
         // -2 corresponds to a new powpeg was elected and the Bridge is waiting for this new powpeg to activate
```
jeremy-then, line 230: I guess tests will come in different prs, like we are doing with other classes. Right?

julia-zack: yes, i created a task for fixing this test and another one for adding tests

rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java
julia-zack, line 1: this test now fails. There's a task created to address that.

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -208,12 +208,47 @@ public void setPendingFederation(PendingFederation federation) {
...
+    }
+
+    @Override
+    public Optional<Federation> getProposedFederation(FederationConstants federationConstants, ActivationConfig.ForBlock activations) {
+        if (!activations.isActive(RSKIP419)) {
+            return Optional.empty();
+        }
+
+        if (proposedFederation != null || isProposedFederationSet) {
+            return Optional.ofNullable(proposedFederation);
```
marcos-iov, line 223: If `proposedFederation` is null it will not enter here. So why use `ofNullable`?

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -208,12 +208,47 @@ public void setPendingFederation(PendingFederation federation) {
...
+        }
+
+        Optional<Integer> storageVersion = getStorageVersion(PROPOSED_FEDERATION_FORMAT_VERSION.getKey());
```
marcos-iov, line 226: And if `storageVersion` is null? Consider this when writing the tests

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -208,12 +208,47 @@ public void setPendingFederation(PendingFederation federation) {
...
+        proposedFederation = bridgeStorageAccessor.getFromRepository(
+            PROPOSED_FEDERATION.getKey(),
+            data -> {
+                if (data == null) {
+                    return null;
+                }
+                // storage version should be always present for proposed federation
```
marcos-iov, line 233: Not really, right? If proposed federation is null the storage version is also null right?

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -208,12 +208,47 @@ public void setPendingFederation(PendingFederation federation) {
...
+                return BridgeSerializationUtils.deserializeFederationAccordingToVersion(data, storageVersion.get(), federationConstants, activations);
+            }
+        );
+
+        return Optional.ofNullable(proposedFederation);
```
marcos-iov, line 238: Can it be null? Isn't that covered in `if (data == null) return null`?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -510,12 +511,17 @@ private ABICallVoteResult executeVoteFederationChangeFunction(boolean dryRun, AB
...
     private Integer createPendingFederation(boolean dryRun) {
         PendingFederation currentPendingFederation = provider.getPendingFederation();
-
         if (currentPendingFederation != null) {
             logger.warn("[createPendingFederation] A pending federation already exists.");
             return FederationChangeResponseCode.PENDING_FEDERATION_ALREADY_EXISTS.getCode();
         }
 
+        Optional<Federation> currentProposedFederation = provider.getProposedFederation(constants, activations);
+        if (currentProposedFederation.isPresent()) {
```
marcos-iov, line 520: Not a big deal, buy maybe put this in a method? `boolean proposedFederationExists`. Maybe it's useful to reuse in other places

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -587,6 +593,14 @@ private Integer addFederatorPublicKeyMultikey(boolean dryRun, BtcECKey btcKey, E
         return FederationChangeResponseCode.SUCCESSFUL.getCode();
     }
 
+    protected int commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLogger eventLogger) {
+        if (!activations.isActive(ConsensusRule.RSKIP419)) {
+            return commitFederationPreRSKIP419(dryRun, hash, eventLogger);
+        }
+
+        return commitFederationPostRSKIP419(dryRun, hash, eventLogger);
```
marcos-iov, line 601: I prefer to call them `legacyCommitFederation` and `commitFederation`

Makes it clearer to me at least

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julia-zack: when we clear the proposed federation, we set it to null. And `isProposedFederationSet` to true. So it will enter here and return a null value

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julia-zack: yes, but shouldn't that be covered returning null when data is null?

julia-zack: i mean, if we reach that line of code is because a non null proposed federation is saved. Im rewording the comment though to make it clearer.

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: for me its clearer to have the rskip, but i'll change it. No problem

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julia-zack: Done

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: Done

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: Done

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -208,12 +208,47 @@ public void setPendingFederation(PendingFederation federation) {
...
+            return Optional.ofNullable(proposedFederation);
+        }
+
+        proposedFederation = bridgeStorageAccessor.getFromRepository(
+            PROPOSED_FEDERATION.getKey(),
+            data -> {
+                if (data == null) {
+                    return null;
+                }
+                // storage version should be always present for not null proposed federation
```
jeremy-then, line 232: nit: `non-null`.

jeremy-then, line 232: nit:
```suggestion
                // storage version should be always present for a non-null proposed federation
```

julia-zack: good catch jere, thanks. Solved

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julia-zack: i think its better to create another task and refactor all of these `data == null`

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: i tried to but its not that straight forward. We could create a new task to do it

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -758,10 +841,6 @@ public void updateFederationCreationBlockHeights() {
         provider.clearNextFederationCreationBlockHeight();
     }
 
-    private boolean thereIsNoNextFederationCreation(Optional<Long> nextFederationCreationBlockHeight) {
-        return !nextFederationCreationBlockHeight.isPresent();
-    }
-
```
julia-zack, line 764: this change was made to get rid of a sonar complain

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
jeremy-then: 👍🏽

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -667,16 +682,84 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLo
...
+    /**
+     * Commits the currently pending federation to be the proposed one.
+     * That is, the proposed federation is set to be the currently pending federation,
+     * and the pending federation is wiped out.
+     * @param dryRun whether to just do a dry run
+     * @param hash the pending federation's hash. This is checked to match the execution block's pending federation hash.
+     * @return 1 upon success, -1 if there was no pending federation, -2 if the pending federation was incomplete,
+     * -3 if the given hash doesn't match the current pending federation's hash.
+     */
+    private int commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLogger eventLogger) {
```
apancorb, line 694: Why not return the enum directly here?

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -208,12 +208,47 @@ public void setPendingFederation(PendingFederation federation) {
...
+            return Optional.ofNullable(proposedFederation);
+        }
+
+        proposedFederation = bridgeStorageAccessor.getFromRepository(
+            PROPOSED_FEDERATION.getKey(),
+            data -> {
+                if (data == null) {
+                    return null;
+                }
+                // storage version should be always present for non-null proposed federation
```
apancorb, line 232: what is the purpose of the storage version?

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julia-zack: done

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julia-zack: to know how to deserialize the obtained fed. The redeem script format depends on the version, see [this](https://github.com/rsksmart/rskj/blob/split_commit_federation/rskj-core/src/main/java/co/rsk/peg/BridgeSerializationUtils.java#L311) method

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: nice! done

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -208,12 +208,47 @@ public void setPendingFederation(PendingFederation federation) {
...
+                Optional<Integer> storageVersion = getStorageVersion(PROPOSED_FEDERATION_FORMAT_VERSION.getKey());
```
marcos-iov, line 233: Should we check that the StorageVersion is present?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -531,7 +535,7 @@ private Integer createPendingFederation(boolean dryRun) {
             return FederationChangeResponseCode.SUCCESSFUL.getCode();
         }
 
-        currentPendingFederation = new PendingFederation(Collections.emptyList());
+        PendingFederation currentPendingFederation = new PendingFederation(Collections.emptyList());
```
marcos-iov, line 538: currentPendingFederation? 🤔 I thought we checked above that there is not an existing pending federation

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -630,20 +653,15 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLo
         oldFederationUTXOs.clear();
         oldFederationUTXOs.addAll(utxosToMove);
 
-        // Network parameters for the new federation are taken from the bridge constants.
-        // Creation time is the block's timestamp.
         Instant creationTime = Instant.ofEpochMilli(rskExecutionBlock.getTimestamp());
+        long rskExecutionBlockNumber = rskExecutionBlock.getNumber();
```
marcos-iov, line 657: Maybe call it federationCreationBlockNumber? A bit more explanatory I think

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -630,20 +653,15 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLo
...
+
         provider.setOldFederation(activeFederation);
         provider.setNewFederation(
-            currentPendingFederation.buildFederation(
-                creationTime,
-                rskExecutionBlock.getNumber(),
-                constants,
-                activations
-            )
+            currentPendingFederation.buildFederation(creationTime, rskExecutionBlockNumber, constants, activations)
```
marcos-iov, line 662: Not a big deal, but I'd put this in its own variable

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -664,7 +680,77 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLo
...
+            // Preserve federation change info
+            provider.setNextFederationCreationBlockHeight(rskExecutionBlockNumber);
+            Script oldFederationP2SHScript = getFederationP2SHScript(activeFederation);
+            provider.setLastRetiredFederationP2SHScript(oldFederationP2SHScript);
+        }
+
+        return FederationChangeResponseCode.SUCCESSFUL;
+    }
+
+    private Script getFederationP2SHScript(Federation federation) {
```
marcos-iov, line 744: Careful with the method name, this can be confusing. This is not a method to get the p2sh script for a given federation. It is for a very specific use case.

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -664,7 +680,77 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLo
...
+        if (!activations.isActive(RSKIP377)) {
+            return federation.getP2SHScript();
+        }
+
+        if (!(federation instanceof ErpFederation)) {
+            return federation.getP2SHScript();
+        }
```
marcos-iov, line 751: Why not join them?
```suggestion
        if (!activations.isActive(RSKIP377) || !federation instanceof ErpFederation) {
            return federation.getP2SHScript();
        }
```

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -587,6 +601,14 @@ private Integer addFederatorPublicKeyMultikey(boolean dryRun, BtcECKey btcKey, E
         return FederationChangeResponseCode.SUCCESSFUL.getCode();
     }
 
+    protected FederationChangeResponseCode commitFederationAccordingToActivations(boolean dryRun, Keccak256 hash, BridgeEventLogger eventLogger) {
```
marcos-iov, line 604: I'd consider renaming the `hash` param. Sounds too generic

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -664,7 +680,77 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLo
...
+
+        // Clear pending federation and votes on election
+        provider.clearPendingFederation();
+        provider.getFederationElection(constants.getFederationChangeAuthorizer()).clear();
+
+        Federation activeFederation = getActiveFederation();
+        eventLogger.logCommitFederation(rskExecutionBlock, activeFederation, proposedFederation);
+        logger.debug("[commitProposedFederation] Proposed Federation committed: {}", proposedFederation.getAddress());
+
+        if (activations.isActive(RSKIP186)) {
```
marcos-iov, line 734: Shouldn't we assume this is active at this point?

rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java
```diff
@@ -225,7 +227,6 @@ private void testChangePowpeg(
         for (UTXO utxo : existingUtxos) {
             assertTrue(utxosToMigrate.stream().anyMatch(storedUtxo -> storedUtxo.equals(utxo)));
         }
-        assertTrue(federationStorageProvider.getNewFederationBtcUTXOs(btcParams, activations).isEmpty());
```
marcos-iov, line 228: Why remove this?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -664,7 +680,77 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLo
...
+     * Commits the currently pending federation to be the proposed one.
+     * That is, the proposed federation is set to be a new federation generated from the currently pending federation,
+     * and the pending federation is wiped out.
+     * @param dryRun whether to just do a dry run
+     * @param hash the pending federation's hash. This is checked to match the execution block's pending federation hash.
+     * @return SUCCESSFUL upon success, PENDING_FEDERATION_NON_EXISTENT if there was no pending federation,
+     * INSUFFICIENT_MEMBERS if the pending federation was incomplete,
+     * PENDING_FEDERATION_MISMATCHED_HASH if the given hash doesn't match the current pending federation's hash.
+     */
+    private FederationChangeResponseCode commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLogger eventLogger) {
```
marcos-iov, line 696: Seems to be a lot of code repeated between both `commitFederation` methods. Let's create a task to analyze re-using this code

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julia-zack: after rskip186 storage version is always saved too. So if the saved version is null, is because the saved fed is null. And that's covered when checking for null data

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: wdym? the check [here](https://github.com/rsksmart/rskj/pull/2629/files#diff-4a4b5a028a8ff057d136592eae64d90f7d7bf630169a6b6e3f778505756f83e8R513) is that the pending federation already exists 🤔

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -664,7 +680,77 @@ protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLo
         logger.debug("[commitFederation] New Federation committed: {}", newFederation.getAddress());
         eventLogger.logCommitFederation(rskExecutionBlock, oldFederation, newFederation);
```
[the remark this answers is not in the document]
julia-zack: sure

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: sure

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: good one, thanks

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: you are right. Thanks.

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: it just seems cleaner to me to understand the cases, but no big deal. I can change it

rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java
julia-zack: wow, good catch. This was not intended.

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: done

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: done

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: done

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: done

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: already addressed in this pr

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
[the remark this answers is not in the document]
julia-zack: task created

rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java
julia-zack: solved

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -588,83 +602,144 @@ private Integer addFederatorPublicKeyMultikey(boolean dryRun, BtcECKey btcKey, E
...
+
+    /**
+     * UTXOs are moved from active to retiring federation so that the transfer of funds can begin.
+     * Then, the retiring federation (old federation) is set to be the currently active federation,
+     * the active federation (new federation) is replaced with a new federation generated from the pending federation,
+     * and the pending federation is wiped out.
+     * The federation change info is preserved, and the commitment with the voted federation is logged.
+     */
+    private FederationChangeResponseCode legacyCommitFederation(PendingFederation currentPendingFederation, BridgeEventLogger eventLogger) {
+        moveUTXOsFromNewToOldFederation();
```
nathanieliov, line 653: Should this happen before setting the active fed as the old fed, and the same for the new fed?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -588,83 +602,144 @@ private Integer addFederatorPublicKeyMultikey(boolean dryRun, BtcECKey btcKey, E
...
-        eventLogger.logCommitFederation(rskExecutionBlock, oldFederation, newFederation);
+        // when the federation also has erp keys,
+        // the members p2sh script is the default p2sh script
+        return ((ErpFederation) federation).getDefaultP2SHScript();
+    }
 
-        return FederationChangeResponseCode.SUCCESSFUL.getCode();
+    private void logCommitmentWithVotedFederation(BridgeEventLogger eventLogger, Federation federationToBeRetired, Federation votedFederation) {
+        eventLogger.logCommitFederation(rskExecutionBlock, federationToBeRetired, votedFederation);
+        logger.debug("[logCommitmentWithVotedFederation] Voted federation committed: {}", votedFederation.getAddress());
```
nathanieliov, line 742: Why debug log level?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: thats how it was before, see [previous code](https://github.com/rsksmart/rskj/blob/feature/powpeg_validation_protocol/rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java#L625-#L631)

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: same, thats just how it was before. See [previous code](https://github.com/rsksmart/rskj/blob/feature/powpeg_validation_protocol/rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java#L664)

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2194,6 +2195,186 @@ void save_callsStorageProviderSave() {
...
+        @Test
+        void commitFederationAccordingToActivations_preRSKIP186_shouldCallLegacyCommitFederationAndNotPreserveFederationChangeInfo() {
+            // arrange
+            activations = ActivationConfigsForTest.papyrus200().forBlock(0L);
+            federationSupport = new FederationSupportImpl(federationMainnetConstants, federationStorageProvider, rskExecutionBlock, activations);
+            List<UTXO> utxosToMove = new ArrayList<>(federationStorageProvider.getNewFederationBtcUTXOs(federationMainnetConstants.getBtcParams(), activations));
+
+            // act
+            FederationChangeResponseCode commitFederationCall = federationSupport.commitFederationAccordingToActivations(false, pendingFederation.getHash(), bridgeEventLogger);
+
```
nathanieliov, line 2237:
```suggestion
```

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2194,6 +2195,186 @@ void save_callsStorageProviderSave() {
...
+            assertPendingFederationVotingWasCleaned();
+
+            assertFederationChangeInfoWasNotSet();
+
+            Federation oldFederation = federationStorageProvider.getOldFederation(federationMainnetConstants, activations);
+            Federation newFederation = federationStorageProvider.getNewFederation(federationMainnetConstants, activations);
+            assertLogCommitFederation(oldFederation, newFederation);
+
+            assertCallWasSuccessful(commitFederationCall);
```
nathanieliov, line 2253: I like this! Using methods that abstract each kind of assertion for the test. Well done! The test looks very clear.

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2194,6 +2195,186 @@ void save_callsStorageProviderSave() {
...
+            Optional<Federation> proposedFederation =
+                federationStorageProvider.getProposedFederation(federationMainnetConstants, activations);
+            assertTrue(proposedFederation.isPresent());
+            assertEquals(federationBuiltFromPendingFederation, proposedFederation.get());
+
+            assertPendingFederationVotingWasCleaned();
+
+            assertFederationChangeInfoWasSet();
+
+            assertLogCommitFederation(activeFederation, proposedFederation.get());
```
nathanieliov, line 2309: I believe we should assert that this test is really calling the new `commitFederation` method. As it is now, pre and post RSKIP419 both use the same assertions. I think we missing an assertion that allows us to differentiate which one was called. Wdyt?

rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java
```diff
@@ -62,6 +63,7 @@
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.*;
 
+@Disabled
```
nathanieliov, line 66: Is this now an official integration test? This explains why you decided to disable it, right?

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: thanks! it was the idea <3

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: yes, i had thought about that but is not an assertion that can be added in this same test (because in that case the federationSupport should be mocked, and we don't want that).

I also think that is kind of covered since the legacyCommitFederation does not have the proposed federation part. So if the proposedFederation is set, we know that the method called was the new one.

That being said, i think that adding a new test that proves what you say with mocks will add some value.
wdyt?

rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java
julia-zack: not exactly. A new task was created to address what to do with this test since it requires deeper analysis

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: i added the test cases :)

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: done

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -588,83 +602,148 @@ private Integer addFederatorPublicKeyMultikey(boolean dryRun, BtcECKey btcKey, E
...
-     * @return 1 upon success, -1 if there was no pending federation, -2 if the pending federation was incomplete,
-     * -3 if the given hash doesn't match the current pending federation's hash.
+     * @param pendingFederationHash the pending federation's hash. This is checked to match the execution block's pending federation hash.
+     * @return PENDING_FEDERATION_NON_EXISTENT if there was no pending federation,
+     * INSUFFICIENT_MEMBERS if the pending federation was incomplete,
+     * PENDING_FEDERATION_MISMATCHED_HASH if the given hash doesn't match the current pending federation's hash.
+     * SUCCESSFUL upon success.
      */
-    protected Integer commitFederation(boolean dryRun, Keccak256 hash, BridgeEventLogger eventLogger) {
+    protected FederationChangeResponseCode commitFederation(boolean dryRun, Keccak256 pendingFederationHash, BridgeEventLogger eventLogger) {
```
marcos-iov, line 614: Is this protected just to be accessed by the tests or is there another reason?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -588,83 +602,148 @@ private Integer addFederatorPublicKeyMultikey(boolean dryRun, BtcECKey btcKey, E
...
+        // first check that we can commit the pending federation
         PendingFederation currentPendingFederation = provider.getPendingFederation();
 
         if (currentPendingFederation == null) {
-            logger.warn("[commitFederation] Pending federation does not exist.");
-            return FederationChangeResponseCode.PENDING_FEDERATION_NON_EXISTENT.getCode();
+            logger.warn("[commitFederationAccordingToActivations] Pending federation does not exist.");
```
marcos-iov, line 619: Guess this change doesn't apply anymore

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -588,83 +602,148 @@ private Integer addFederatorPublicKeyMultikey(boolean dryRun, BtcECKey btcKey, E
...
+            return FederationChangeResponseCode.SUCCESSFUL;
         }
 
-        // Move UTXOs from the new federation into the old federation
-        // and clear the new federation's UTXOs
+        // proceed with the commitment
+        return commitPendingFederationAccordingToActivations(currentPendingFederation, eventLogger);
+    }
+
+    protected FederationChangeResponseCode commitPendingFederationAccordingToActivations(PendingFederation currentPendingFederation, BridgeEventLogger eventLogger) {
```
marcos-iov, line 642: protected?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -588,83 +602,148 @@ private Integer addFederatorPublicKeyMultikey(boolean dryRun, BtcECKey btcKey, E
...
+    }
+
+    /**
+     * UTXOs are moved from active to retiring federation so that the transfer of funds can begin.
+     * Then, the retiring federation (old federation) is set to be the currently active federation,
+     * the active federation (new federation) is replaced with a new federation generated from the pending federation,
+     * and the pending federation is wiped out.
+     * The federation change info is preserved, and the commitment with the voted federation is logged.
+     */
+    protected FederationChangeResponseCode legacyCommitPendingFederation(PendingFederation currentPendingFederation, BridgeEventLogger eventLogger) {
```
marcos-iov, line 656: protected?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -588,83 +602,148 @@ private Integer addFederatorPublicKeyMultikey(boolean dryRun, BtcECKey btcKey, E
...
-        Federation newFederation = provider.getNewFederation(constants, activations);
+    protected Script getFederationMembersP2SHScript(Federation federation) {
+        // when the federation is a standard multisig,
+        // the members p2sh script is the p2sh script
+        if (!activations.isActive(RSKIP377)) {
+            return federation.getP2SHScript();
+        }
+        if (!(federation instanceof ErpFederation)) {
+            return federation.getP2SHScript();
+        }
```
marcos-iov, line 737:
```suggestion
        if (!activations.isActive(RSKIP377) || !(federation instanceof ErpFederation)) {
            return federation.getP2SHScript();
        }
```
Why not just join?

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2194,6 +2195,218 @@ void save_callsStorageProviderSave() {
...
+
+        @BeforeEach
+        void setUp() {
+            StorageAccessor bridgeStorageAccessor = new InMemoryStorage();
+            federationStorageProvider = new FederationStorageProviderImpl(bridgeStorageAccessor);
+            
+            pendingFederation = new PendingFederation(FederationTestUtils.getFederationMembers(5));
+            federationStorageProvider.setPendingFederation(pendingFederation);
+
+            bridgeEventLogger = mock(BridgeEventLogger.class);
```
marcos-iov, line 2221: Maybe we can use an actual objet here instead of the mock?

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2194,6 +2195,218 @@ void save_callsStorageProviderSave() {
...
+
+            rskExecutionBlock = mock(Block.class);
```
marcos-iov, line 2223: Here as well. There is a BlockHeaderBuilder class that can help you create a Block object with the required properties

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2194,6 +2195,218 @@ void save_callsStorageProviderSave() {
...
+            when(rskExecutionBlock.getNumber()).thenReturn(rskExecutionBlockNumber);
+            when(rskExecutionBlock.getTimestamp()).thenReturn(rskExecutionBlockTimestamp);
+        }
+
+        @Test
+        void commitPendingFederationAccordingToActivations_preRSKIP419_shouldCallPreRSKIP419Method() {
```
marcos-iov, line 2229: Maybe call it preLovell instead of preRskipX?

We should standardize that

marcos-iov: Also, not sure this test is a good idea.

We should be testing behaviour, not the internals of a class. Instead of spying federationSupport object I think we should simply call commitFederation pre RSKIP and assert whatever is different from when it's called post RSKIP

marcos-iov: Similar situation with the test names. Rather than `shouldCallXMethod` I think it should be `shouldDoX`

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: yes, to be accessed by the tests

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: you right, thanks

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: it just feels easier for me to understand the flow. But i can change it

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: sure, thanks

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: sure

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: done

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: done

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: done

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: i removed those two tests.

Regarding the test name, preLovell or preRskip, I found putting the rskip better. I think is more accurate, since he behavior changes because of the rskip activation

nathanieliov (approved):

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -3113,6 +3116,224 @@ private ABICallSpec getAddMultiKeysVote(FederationMember federationMember) {
...
+
+            // assertions
+
+            assertUTXOsWereMoved(utxosToMove);
+
+            assertNewAndOldFederationsWereSet();
+
+            assertPendingFederationVotingWasCleaned();
+
+            assertFederationChangeInfoWasNotSet();
```
marcos-iov, line 3176: Super nice setup, very clean. Well done 👏👏

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -3113,6 +3116,224 @@ private ABICallSpec getAddMultiKeysVote(FederationMember federationMember) {
...
+
+            Federation oldFederation = federationStorageProvider.getOldFederation(federationMainnetConstants, activations);
+            Federation newFederation = federationStorageProvider.getNewFederation(federationMainnetConstants, activations);
+            assertLogCommitFederation(oldFederation, newFederation);
+
+            assertCallWasSuccessful(commitFederationCall);
+        }
+
+        @Test
+        void commitFederation_postRSKIP419_shouldCallCommitPendingFederation() {
```
marcos-iov, line 3215: Maybe rename this to be more specific? shouldSetProposedFederation?

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -3113,6 +3116,224 @@ private ABICallSpec getAddMultiKeysVote(FederationMember federationMember) {
...
+
+
+            // assertions
+
+            // assert proposed federation was set correctly
+            Optional<Federation> proposedFederation =
+                federationStorageProvider.getProposedFederation(federationMainnetConstants, activations);
+            assertTrue(proposedFederation.isPresent());
+            assertEquals(federationBuiltFromPendingFederation, proposedFederation.get());
+
```
marcos-iov, line 3235: Assert new and old federations didn't change?

marcos-iov (commented):
Nice work, like how the tests were set up

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -3113,6 +3116,224 @@ private ABICallSpec getAddMultiKeysVote(FederationMember federationMember) {
...
+            federationSupport = new FederationSupportImpl(federationMainnetConstants, federationStorageProvider, rskExecutionBlock, activations);
+            List<UTXO> utxosToMove = new ArrayList<>(federationStorageProvider.getNewFederationBtcUTXOs(federationMainnetConstants.getBtcParams(), activations));
+
+            // act
+            FederationChangeResponseCode commitFederationCall = federationSupport.commitFederation(false, pendingFederation.getHash(), bridgeEventLogger);
+
+
+            // assertions
+
+            assertUTXOsWereMoved(utxosToMove);
```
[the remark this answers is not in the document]
julia-zack: thanks! you right. I'll change it :)

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: thanks!! :)

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: good one, thanks :)

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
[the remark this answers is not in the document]

rskj-core/src/test/java/co/rsk/peg/federation/FederationTestUtils.java
```diff
@@ -255,4 +257,19 @@ private static Script createInputScriptSig(
             .data(fedRedeemScript.getProgram())
             .build();
     }
+
+    public static Script getFederationMembersP2SHScript(ActivationConfig.ForBlock activations, Federation federation) {
```
marcos-iov, line 261: Should we have this here? Not sure the value it adds as a util. It is only used for a specific case, when storing the last retired fed script

rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java
marcos-iov: Fails to run or to compile? If it's the first we could simply disable

marcos-iov (approved):

marcos-iov (approved):

PR #2633, opened 2024-07-23
Showcase the Bridge behaviour under different scenarios

apancorb:
## Description

Adds unit tests to analyze a new pegout minimum. This PR is just focused on analisis purposes. 

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/MinimumPegValueTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/MinimumPegValueTest.java
@@ -18,7 +18,11 @@
 package co.rsk.peg;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 import co.rsk.bitcoinj.core.Address;
 import co.rsk.bitcoinj.core.BtcECKey;
@@ -28,17 +32,40 @@ import co.rsk.bitcoinj.core.NetworkParameters;
 import co.rsk.bitcoinj.core.Sha256Hash;
 import co.rsk.bitcoinj.core.UTXO;
 import co.rsk.bitcoinj.wallet.Wallet;
+import co.rsk.db.MutableTrieCache;
+import co.rsk.db.MutableTrieImpl;
 import co.rsk.peg.constants.BridgeConstants;
 import co.rsk.peg.constants.BridgeMainNetConstants;
+import co.rsk.peg.federation.*;
+import co.rsk.peg.feeperkb.FeePerKbSupport;
+import co.rsk.peg.feeperkb.FeePerKbSupportImpl;
+import co.rsk.peg.storage.BridgeStorageAccessorImpl;
+import co.rsk.peg.storage.StorageAccessor;
+import co.rsk.peg.utils.BridgeEventLogger;
+import co.rsk.peg.utils.BridgeEventLoggerImpl;
+import co.rsk.test.builders.BridgeSupportBuilder;
+import co.rsk.test.builders.FederationSupportBuilder;
+import co.rsk.trie.Trie;
+import java.io.IOException;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Stream;
-import co.rsk.peg.federation.*;
+import org.bouncycastle.util.encoders.Hex;
+import org.ethereum.config.Constants;
 import org.ethereum.config.blockchain.upgrades.ActivationConfig;
 import org.ethereum.config.blockchain.upgrades.ActivationConfigsForTest;
+import org.ethereum.core.BlockTxSignatureCache;
+import org.ethereum.core.ReceivedTxSignatureCache;
+import org.ethereum.core.Repository;
+import org.ethereum.core.SignatureCache;
+import org.ethereum.core.Transaction;
+import org.ethereum.crypto.ECKey;
+import org.ethereum.db.MutableRepository;
+import org.ethereum.vm.LogInfo;
+import org.ethereum.vm.PrecompiledContracts;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
@@ -48,6 +75,12 @@ class MinimumPegValueTest {
 
   private static final Coin MINIMUM_PEGIN_VALUE = Coin.valueOf(100_000L);
   private static final Coin MINIMUM_PEGOUT_VALUE = Coin.valueOf(80_000L);
+
+  private static final BigInteger NONCE = new BigInteger("0");
+  private static final BigInteger GAS_PRICE = new BigInteger("100");
+  private static final BigInteger GAS_LIMIT = new BigInteger("1000");
+  private static final String DATA = "80af2871";
+  private static final ECKey SENDER = new ECKey();
 
   private ActivationConfig.ForBlock activations;
   private NetworkParameters networkParameters;
@@ -163,13 +196,42 @@ class MinimumPegValueTest {
     assertEquals(ReleaseTransactionBuilder.Response.SUCCESS, result.getResponseCode());
   }
 
+  @ParameterizedTest()
+  @MethodSource("providePegMinimumParameters")
+  void whenReleaseBtcCalledWithMinimumPegoutValue_shouldLogReleaseBtcRequestReceived(Coin feePerKb) throws IOException {
+    BridgeConstants bridgeConstants = mock(BridgeConstants.class);
+    when(bridgeConstants.getBtcParams()).thenReturn(bridgeMainNetConstants.getBtcParams());
+    when(bridgeConstants.getMinimumPegoutValuePercentageToReceiveAfterFee())
+        .thenReturn(bridgeMainNetConstants.getMinimumPegoutValuePercentageToReceiveAfterFee());
+    when(bridgeConstants.getMinimumPegoutTxValue()).thenReturn(MINIMUM_PEGOUT_VALUE);
+
+    List<LogInfo> logInfo = new ArrayList<>();
+    SignatureCache signatureCache = new BlockTxSignatureCache(new ReceivedTxSignatureCache());
+    BridgeEventLoggerImpl eventLogger = spy(new BridgeEventLoggerImpl(
+        bridgeConstants, activations, logInfo, signatureCache));
+
+    FeePerKbSupport feePerKbSupport = mock(FeePerKbSupportImpl.class);
+    when(feePerKbSupport.getFeePerKb()).thenReturn(feePerKb);
+
+    BridgeSupport bridgeSupport = initBridgeSupport(
+        bridgeConstants, eventLogger, activations, signatureCache, feePerKbSupport);
+
+    bridgeSupport.releaseBtc(buildReleaseRskTx(MINIMUM_PEGOUT_VALUE));
+
+    verify(eventLogger).logReleaseBtcRequestReceived(any(), any(), any());
+  }
+
   private static Stream<Arguments> providePegMinimumParameters() {
     return Stream.of(
         // current feePerKb value
         Arguments.of(Coin.valueOf(24_000L)),
         // maximum feePerKb value that allows all cases to succeed
-        Arguments.of(Coin.valueOf(81_423L)));
+        Arguments.of(Coin.valueOf(24_750L)));
   }
+
+  /**********************************
+   * ------- UTILS ------- *
+   *********************************/
 
   private Address getAddress(int pk) {
     return BtcECKey.fromPrivate(BigInteger.valueOf(pk))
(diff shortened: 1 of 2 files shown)


rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
```diff
@@ -904,6 +909,112 @@ void test_VerifyTXFeeIsSpentEquallyForBatchedPegouts_three_pegouts() {
...
+        // pegin value, but since a change output is also an UTXO that the current wallet implementation can use
+        // then we will find and use that value.
+        Coin minSpendableUTXOValue = Coin.SATOSHI;
+        tx.clearOutputs();
+        tx.addOutput(minSpendableUTXOValue, fed.getP2SHScript());
+        while (tx.getOutput(0).isDust()) {
+            minSpendableUTXOValue = minSpendableUTXOValue.add(Coin.SATOSHI);
+            tx.clearOutputs();
+            tx.addOutput(minSpendableUTXOValue, fed.getP2SHScript());
+        }
```
apancorb, line 975: Interesting enough this value turns out to be `2700 SATS`. Which is exactly the lowest value UTXO I was able to find in the Bridge address. Meaning given the current wallet implementation, this is the lowest value that the Bridge will not consider it as dust and will proceed to spend it if so desired by the selector UTXO algorithm. 

Here is the tx: https://mempool.space/tx/ab94ce7abfe178025b593ce0f0c7468a984087ae7977800b62cf47b4284d00dd

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
```diff
@@ -904,6 +909,112 @@ void test_VerifyTXFeeIsSpentEquallyForBatchedPegouts_three_pegouts() {
...
+        // build batch peg-out transaction
+        ReleaseTransactionBuilder.BuildResult result = releaseTransactionBuilder.buildBatchedPegouts(entries);
+
+        // the proposed feePerKb and minPegoutValue are compatible
+        assertEquals(ReleaseTransactionBuilder.Response.SUCCESS, result.getResponseCode());
+    }
+
+    private static Stream<Arguments> providePegoutMinimumParameters() {
+        return Stream.of(
+            Arguments.of(Coin.valueOf(24_000L), Coin.valueOf(28_500L))
```
apancorb, line 1014: I just gave this initial pair, but any more permutations would be welcomed.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
```diff
@@ -904,6 +909,112 @@ void test_VerifyTXFeeIsSpentEquallyForBatchedPegouts_three_pegouts() {
...
+        ReleaseTransactionBuilder releaseTransactionBuilder = new ReleaseTransactionBuilder(
+            networkParameters,
+            fedWallet,
+            fed.getAddress(),
+            feePerKb,
+            activations
+        );
+
+        // build batch peg-out transaction
+        ReleaseTransactionBuilder.BuildResult result = releaseTransactionBuilder.buildBatchedPegouts(entries);
```
apancorb, line 1006: With the current feePerKb of `24,000 SATS` and my initial proposed minPegoutValue of `28,500 SATS`, this batch peg-out transaction will not be built with a `Wallet.CouldNotAdjustDownwards` exception. Meaning the user output could not be adjusted downwards to pay the tx fees. This makes sense since this scenario I believe is the worst case scenario, since we have N P2SH inputs with the min possible value to be non-dust and only one user output that is going to take all the load. So I think if we can find two values that pass this scenario, then we should be good for any other scenario. Wdyt @marcos-iov ?

apancorb: Nevermind, this test scenario will never pass no matter what min peg-out value you choose. Maybe we can just split the min peg-out value by using 10 inputs. Like if min peg-out value is 20 then have 10 inputs of 2 each. Wdyt?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
```diff
@@ -904,6 +909,159 @@ void test_VerifyTXFeeIsSpentEquallyForBatchedPegouts_three_pegouts() {
         Assertions.assertEquals(inputsValue.minus(totalPegoutAmount), changeOutput.getValue());
     }
 
+    @ParameterizedTest()
+    @MethodSource("providePegoutMinimumParameters")
+    void buildBatchedPegouts_whenSpendableUXTO_shouldBuildTransactionSuccessfully(Coin feePerKb, Coin minPegoutValue) {
+        // build federation with 9 members
+        Integer[] memberPKs = new Integer[]{ 100, 200, 300, 400, 500, 600, 700, 800, 900 };
+        Federation fed = FederationTestUtils.getFederation(memberPKs);
```
marcos-iov, line 917: This is creating a standard multisig federation. We should test this with an ERP fed, that is the one used at the moment. Larger redeem script since it includes the emergency multisig. You can use FederationFactory class to build it

marcos-iov: It should have 9 members and 4 in the emergency multisig

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
```diff
@@ -904,6 +909,159 @@ void test_VerifyTXFeeIsSpentEquallyForBatchedPegouts_three_pegouts() {
...
+        // the federation wallet, without flyover support
+        Wallet fedWallet = BridgeUtils.getFederationSpendWallet(
+            Context.getOrCreate(networkParameters),
+            fed,
+            utxos,
+            false,
+            mock(BridgeStorageProvider.class)
+        );
+        // build release transaction builder for current fed
+        when(activations.isActive(ConsensusRule.RSKIP201)).thenReturn(true);
```
marcos-iov, line 933: Just in case, let's use ActivationConfigForTest class. And get the `all` object, that includes all RSKIPs active

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
marcos-iov: So this is the lowest change output the Bridge will create, right? If a UTXO is used that generates a smaller change, it will remove sats from the user output until it completes 2,700 sats as the change. So we could say that the max it will drain from a user output is 2,700 sats in the case that the change was 0.

And this does not depend on the feePerKb value set in the Bridge, is that right?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
marcos-iov: Not sure I followed the logic here. There is only one pegout request for the min pegout value, right? And how many UTXOs available and for what value?

marcos-iov (commented):
Should we also consider the size of the transaction after signing?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
julia-zack: you can check [P2shErpFederationBuilder](https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/test/java/co/rsk/peg/federation/P2shErpFederationBuilder.java) as a guidance

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
```diff
@@ -904,6 +909,159 @@ void test_VerifyTXFeeIsSpentEquallyForBatchedPegouts_three_pegouts() {
         Assertions.assertEquals(inputsValue.minus(totalPegoutAmount), changeOutput.getValue());
     }
 
+    @ParameterizedTest()
+    @MethodSource("providePegoutMinimumParameters")
+    void buildBatchedPegouts_whenSpendableUXTO_shouldBuildTransactionSuccessfully(Coin feePerKb, Coin minPegoutValue) {
```
julia-zack, line 914: typo in UTXO

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
apancorb: > So this is the lowest change output the Bridge will create, right? If a UTXO is used that generates a smaller change, it will remove sats from the user output until it completes 2,700 sats as the change. So we could say that the max it will drain from a user output is 2,700 sats in the case that the change was 0.

That is right.

> And this does not depend on the feePerKb value set in the Bridge, is that right?

That is right, reason being because it always uses a [hardcoded feePerKb value of 15,000 SATS](https://github.com/rsksmart/bitcoinj-thin/blob/master/src/main/java/co/rsk/bitcoinj/core/TransactionOutput.java#L246)

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
apancorb: Yes only one pegout request. And there will be as many UTXOs available to cover the request. Each UTXO will have a value of 2,700 SATS. The problem with this test is that is impossible to pass since it is too much fees that has to be taken from the user. I have written a similar one but that takes the pegout request value and divides it by 10 equal UTXOs.

rskj-core/src/test/java/co/rsk/peg/MinimumPegValueTest.java
```diff
@@ -0,0 +1,189 @@
...
+    // build batch peg-out transaction
+    ReleaseTransactionBuilder.BuildResult result = releaseTransactionBuilder.buildBatchedPegouts(entries);
+
+    // the proposed feePerKb and peg values are compatible
+    assertEquals(ReleaseTransactionBuilder.Response.SUCCESS, result.getResponseCode());
+  }
+
+  private static Stream<Arguments> providePegMinimumParameters() {
+    return Stream.of(
+        Arguments.of(Coin.valueOf(24_000L), Coin.valueOf(100_000L), Coin.valueOf(80_000L)));
```
apancorb, line 174: Is there a way to get the current feePerKb programatically?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
apancorb: I have updated to use the defaults of [P2shErpFederationBuilder](https://github.com/rsksmart/rskj/blob/federation-support-refactor-integration/rskj-core/src/test/java/co/rsk/peg/federation/P2shErpFederationBuilder.java), let me know if this looks good

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
apancorb: How would it look in code? If I understand correctly, I should still mock `isActive(ConsensusRule.RSKIP201)` right?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderTest.java
apancorb: Done!

rskj-core/src/test/java/co/rsk/peg/MinimumPegValueTest.java
```diff
@@ -0,0 +1,189 @@
...
+    // build federation
+    Federation fed = new P2shErpFederationBuilder().withNetworkParameters(networkParameters).build();
+    // list of peg-out requests with the given minimum peg-out value
+    List<ReleaseRequestQueue.Entry> entries = new ArrayList<>();
+    for (int i = 0; i < 10; i++) {
+      entries.add(createTestEntry(1000 + i, minPegoutValue));
+    }
+    // list of utxos that contains one utxo with the minimum peg-in value
+    List<UTXO> utxos = new ArrayList<>();
+    for (int i = 0; i < 10; i++) {
```
apancorb, line 145: It is fine to add 10 here instead of 9 since the selector algorithm will only choose the needed amount of inputs to cover the total outputs. Let me know if that is fine. @marcos-iov

rskj-core/src/test/java/co/rsk/peg/MinimumPegValueTest.java
marcos-iov: Current as in mainnet? You'd have to do a call to the Bridge using a mainnet node. Probably out of scope for a test.

In the code, you can use FeePerKbSupport class to get it

marcos-iov (commented):
Looks good overall. I think we just need to test for different fee per kb values

rskj-core/src/test/java/co/rsk/peg/MinimumPegValueTest.java
apancorb: I see , thanks

rskj-core/src/test/java/co/rsk/peg/MinimumPegValueTest.java
```diff
@@ -0,0 +1,189 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+class MinimumPegValueTest {
+  private ActivationConfig.ForBlock activations;
+  private NetworkParameters networkParameters;
+  private BridgeConstants bridgeMainNetConstants;
+
+  @BeforeEach
+  void setup() {
+    activations = mock(ActivationConfig.ForBlock.class);
```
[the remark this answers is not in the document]
apancorb: Updated, let me know if it look better now

rskj-core/src/test/java/co/rsk/peg/MinimumPegValueTest.java
```diff
@@ -0,0 +1,189 @@
...
+    // build batch peg-out transaction
+    ReleaseTransactionBuilder.BuildResult result = releaseTransactionBuilder.buildBatchedPegouts(entries);
+
+    // the proposed feePerKb and peg values are compatible
+    assertEquals(ReleaseTransactionBuilder.Response.SUCCESS, result.getResponseCode());
+  }
+
+  private static Stream<Arguments> providePegMinimumParameters() {
+    return Stream.of(
+        Arguments.of(Coin.valueOf(24_000L), Coin.valueOf(100_000L), Coin.valueOf(80_000L)));
```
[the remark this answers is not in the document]
apancorb: Sounds good, I have updated using this approach

rskj-core/src/test/java/co/rsk/peg/MinimumPegValueTest.java
```diff
@@ -0,0 +1,189 @@
...
+    bridgeMainNetConstants = BridgeMainNetConstants.getInstance();
+    networkParameters = bridgeMainNetConstants.getBtcParams();
+  }
+
+  @ParameterizedTest()
+  @MethodSource("providePegMinimumParameters")
+  void checkProposedPegValues_whenOneInputofMinPeginValueAndOneOutputOfMinPegoutValue_shouldBuildTransactionSuccessfully(
+      Coin feePerKb, Coin minPeginValue, Coin minPegoutValue) {
+    // build federation
+    Federation fed = new P2shErpFederationBuilder().withNetworkParameters(networkParameters).build();
```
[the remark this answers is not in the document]
apancorb: Updated

rskj-core/src/test/java/co/rsk/peg/MinimumPegValueTest.java
```diff
@@ -0,0 +1,189 @@
...
+  @BeforeEach
+  void setup() {
+    activations = mock(ActivationConfig.ForBlock.class);
+    bridgeMainNetConstants = BridgeMainNetConstants.getInstance();
+    networkParameters = bridgeMainNetConstants.getBtcParams();
+  }
+
+  @ParameterizedTest()
+  @MethodSource("providePegMinimumParameters")
+  void checkProposedPegValues_whenOneInputofMinPeginValueAndOneOutputOfMinPegoutValue_shouldBuildTransactionSuccessfully(
```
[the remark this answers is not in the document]
apancorb: Sounds good, I have renamed the tests names

PR #2653, opened 2024-07-29
Enable snap syncing with RSK cli args

rmoreliovlabs:
# **Add Support for Sync Mode and Snap Nodes RSK CLI Options**

## Description
This PR introduces two new command-line options to the RSKj node to enhance its synchronization capabilities:

- `--sync-mode=<full|snap>`: Allows setting the synchronization mode to either full or snap, with full being the default.
- `--snap-nodes=<enode URLs>`: Enables specifying multiple SNAP-capable nodes using encoded enode URLs.

**Key Changes:**

- Added support for `--sync-mode` to configure the sync mode.
- Added support for `--snap-nodes` to configure SNAP boot nodes.
- Implemented validation to ensure only valid values are accepted for these options, with appropriate error messages for invalid inputs.

## Motivation and Context
The goal is to improve the user experience when configuring SNAP protocol-related settings. Previously, enabling SNAP synchronization required **only** manual configuration changes. By introducing these CLI options, users can now easily configure their node's sync mode and specify SNAP boot nodes directly via command-line arguments.

- The `--sync-mode` argument ensures that only full or snap are valid inputs. Any other values will result in an error. **Example**: `--sync-mode=snap`
- The `--snap-nodes` argument parses and validates each enode URL to ensure it follows the format `enode://PUBKEY@HOST:PORT`. Invalid URLs will trigger an error message. **Example**: `--snap-nodes=[enode://PUBKEY1@HOST1:PORT1,enode://PUBKEY2@HOST2:PORT2]`

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/RskContext.java
+++ b/rskj-core/src/main/java/co/rsk/RskContext.java
@@ -1010,7 +1010,7 @@ public class RskContext implements NodeContext, NodeBootstrapper {
 
         if (getRskSystemProperties().isSyncEnabled()) {
             internalServices.add(getSyncPool());
-            if (getSyncConfiguration().isServerSnapSyncEnabled()) {
+            if (getSyncConfiguration().isServerSnapSyncEnabled() || getRskSystemProperties().getSyncMode().equalsIgnoreCase("snap")) {
                 internalServices.add(getSnapshotProcessor());
             }
         }
--- a/rskj-core/src/main/java/co/rsk/RskContext.java
+++ b/rskj-core/src/main/java/co/rsk/RskContext.java
@@ -1010,7 +1010,7 @@ public class RskContext implements NodeContext, NodeBootstrapper {
 
         if (getRskSystemProperties().isSyncEnabled()) {
             internalServices.add(getSyncPool());
-            if (getSyncConfiguration().isServerSnapSyncEnabled() || getRskSystemProperties().getSyncMode().equalsIgnoreCase("snap")) {
+            if (getSyncConfiguration().isServerSnapSyncEnabled()) {
                 internalServices.add(getSnapshotProcessor());
             }
         }
--- a/rskj-core/src/main/java/co/rsk/cli/RskCli.java
+++ b/rskj-core/src/main/java/co/rsk/cli/RskCli.java
@@ -26,6 +26,7 @@ import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+
 @CommandLine.Command(name = "rskj", mixinStandardHelpOptions = true, versionProvider = VersionProviderUtil.class,
         description = "RSKJ blockchain node implementation in Java")
 public class RskCli implements Runnable {
@@ -75,6 +76,12 @@ public class RskCli implements Runnable {
 
     @CommandLine.Option(names = {"-X"}, description = "Read arguments in command line")
     private List<String> xArguments;
+
+    @CommandLine.Option(names = {"--sync-mode"}, description = "Set Synchronization mode. Valid options are <full | snap>")
+    private String syncMode;
+
+    @CommandLine.Option(names = {"--snap-nodes"}, description = "Set snapboot nodes")
+    private List<String> snapBootNodes;
 
     private boolean help;
     private boolean version;
@@ -151,6 +158,14 @@ public class RskCli implements Runnable {
             } else if (networkFlags.networkMainnet) {
                 activatedFlags.add(NodeCliFlags.NETWORK_MAINNET);
             }
+        }
+
+        if (syncMode != null) {
+            activatedOptions.put(NodeCliOptions.SYNC_MODE, syncMode);
+        }
+
+        if (snapBootNodes != null) {
+            activatedOptions.put(NodeCliOptions.SNAP_NODES, String.join(",", snapBootNodes));
         }
 
         cliArgs = CliArgs.of(activatedOptions, activatedFlags, paramValueMap);
--- a/rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
+++ b/rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
@@ -58,7 +58,6 @@ public enum NodeCliOptions implements OptionalizableCliArg {
             if(configValue.equalsIgnoreCase("full") || configValue.equalsIgnoreCase("snap")){
                 return config.withValue(RskSystemProperties.PROPERTY_SYNC_MODE, ConfigValueFactory.fromAnyRef(configValue));
             }else {
-                System.err.println("Invalid sync mode provided: " + configValue);
                 throw new IllegalArgumentException("Invalid sync mode: " + configValue + ". The valid options are <full> or <snap>.");
             }
         }
--- a/rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
+++ b/rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
@@ -19,8 +19,19 @@ package co.rsk.config;
 
 import co.rsk.cli.OptionalizableCliArg;
 import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+import com.typesafe.config.ConfigObject;
+import com.typesafe.config.ConfigValue;
 import com.typesafe.config.ConfigValueFactory;
 import org.ethereum.config.SystemProperties;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * Options that the node can receive via command line arguments.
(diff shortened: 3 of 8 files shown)


rskj-core/src/main/java/co/rsk/cli/RskCli.java
```diff
@@ -76,6 +77,12 @@ static class NetworkFlags {
     @CommandLine.Option(names = {"-X"}, description = "Read arguments in command line")
     private List<String> xArguments;
 
+    @CommandLine.Option(names = {"--sync-mode"}, description = "Set Synchronization mode. Valid options are <full | snap>")
+    private String syncMode;
```
Vovchyk, line 81: I'd suggest to make it strongly typed, eg. with help enum:
```
enum SyncMode { full, snap }
```

rskj-core/src/main/resources/reference.conf
```diff
@@ -296,6 +296,8 @@ sync {
     # block chain synchronization can be: [true/false]
     enabled = true
 
+    mode = full
```
Vovchyk, line 299: not sure we need a new config. We could simply make use of existing one -  `sync.snapshot.client.enabled`

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
```diff
@@ -41,6 +52,62 @@ public Config withConfig(Config config, String configValue) {
...
+            }else {
+                throw new IllegalArgumentException("Invalid sync mode: " + configValue + ". The valid options are <full> or <snap>.");
+            }
+        }
+    },
+    SNAP_NODES("snap-nodes", true) {
+        @Override
+        public Config withConfig(Config config, String configValue) {
+            try {
+                List<ConfigObject> snapConfigObjects = Arrays.stream(configValue.substring(1, configValue.length() - 1).split(","))
```
Vovchyk, line 69: could you pls explain why do we need to do `substring()` here?

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
```diff
@@ -41,6 +52,62 @@ public Config withConfig(Config config, String configValue) {
...
+                }
+                return config;
+            } catch (Exception e) {
+                e.printStackTrace();
+                return config;
+            }
+        }
+        private ConfigObject createConfigObjectFromNode(String snapNode) {
+            try {
+                URI snapURI = new URI(snapNode);
```
Vovchyk, line 87: in cli we expect value to be in this format: `enode://PUBKEY@HOST:PORT`. So we could simply make a ConfigObject in the form of: `{ "url": "<value>" }` without a need to parse it, wdyt?

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -1010,7 +1010,7 @@ public synchronized List<InternalService> buildInternalServices() {
 
         if (getRskSystemProperties().isSyncEnabled()) {
             internalServices.add(getSyncPool());
-            if (getSyncConfiguration().isServerSnapSyncEnabled()) {
+            if (getSyncConfiguration().isServerSnapSyncEnabled() || getRskSystemProperties().getSyncMode().equalsIgnoreCase("snap")) {
```
Vovchyk, line 1013: I guess we should only add this service only if the server part is enable (as it is now), regardless of client is enabled or not

Vovchyk (changes requested):

rskj-core/src/main/resources/reference.conf
fmacleal: Agreed.
But if we still proceed having a new config, we should at least have a comment explaining the use.

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
```diff
@@ -41,6 +52,62 @@ public Config withConfig(Config config, String configValue) {
...
+                String host = snapURI.getHost();
+                int port = snapURI.getPort();
+
+                Map<String, ConfigValue> map = new HashMap<>();
+                map.put("port", ConfigValueFactory.fromAnyRef(port));
+                map.put("ip", ConfigValueFactory.fromAnyRef(host));
+                map.put("nodeId", ConfigValueFactory.fromAnyRef(id));
+
+                return ConfigFactory.parseMap(map).root();
+            } catch (URISyntaxException e) {
```
fmacleal, line 100: suggestion:

Add unit tests for this scenario where we have invalid URL format, so we can have coverage for these situations.

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
```diff
@@ -41,6 +52,62 @@ public Config withConfig(Config config, String configValue) {
...
+                throw new IllegalArgumentException("Invalid URL format: " + snapNode, e);
```
fmacleal, line 101:
```suggestion
                throw new IllegalArgumentException("Invalid URL format for snap node: " + snapNode, e);
```

Just to let it clear straight away from where come from this errorMessage.

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
```diff
@@ -41,6 +52,62 @@ public Config withConfig(Config config, String configValue) {
...
+            }
+        }
+
+        private void validateSnapNodeURI(URI uri) {
+            if (!"enode".equals(uri.getScheme()) || uri.getUserInfo() == null || uri.getHost() == null || uri.getPort() == -1) {
```
fmacleal, line 106: suggestion:

If we are still keeping this validation, we could add an unit test to cover this piece of code.

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
```diff
@@ -41,6 +52,62 @@ public Config withConfig(Config config, String configValue) {
...
+                throw new IllegalArgumentException("expecting URL in the format enode://PUBKEY@HOST:PORT");
```
fmacleal, line 107:
```suggestion
                throw new IllegalArgumentException("expecting URL for snap node in the format enode://PUBKEY@HOST:PORT");
```

rskj-core/src/main/java/co/rsk/RskContext.java
fmacleal, line 1013: suggestion:

Add this string to a constant.

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
rmoreliovlabs: Got it, that makes sense. I’ll simplify the method to directly wrap the URI string in a ConfigObject without parsing it into components

Just to make sure, should we still include some validation to check if the URL is correctly formatted before accepting it? It might help catch any errors early and make sure everything is in the expected format. Let me know what you think

rskj-core/src/main/java/co/rsk/cli/RskCli.java
rmoreliovlabs: solved, thanks for the suggestion!

rskj-core/src/main/resources/reference.conf
rmoreliovlabs: solved, removed the new config

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
rmoreliovlabs: solved, removed the substring. The flag should be formatted as `--snap-nodes=<url1>`` --snap-nodes=<url2>`, with each URL specified separately

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
rmoreliovlabs: Added the test [here](https://github.com/rsksmart/rskj/pull/2653/files#diff-f6561bc832eef1ae925e5bd576c6e88f42c7e6919f60d8f805fbb11b0ca09162R165)

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
rmoreliovlabs: addressed, removed the substring. The flag should be formatted as --snap-nodes=<url1>`` --snap-nodes=<url2>, with each URL specified separately

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
rmoreliovlabs: removed the validation following @Vovchyk 's suggestion

rskj-core/src/main/java/co/rsk/RskContext.java
rmoreliovlabs: created enum for the SyncMode

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
rmoreliovlabs: solved, implemented you suggestion [here](https://github.com/rsksmart/rskj/pull/2653/files#diff-2936d53bde166899a1f013108ef53a23fa931a04c7f06dae1df7f4c309a7ce12R92)

rskj-core/src/main/java/co/rsk/RskContext.java
rmoreliovlabs: Addressed the issue by reverting the change, the service is now only added if the snap server is enabled

rskj-core/src/main/java/co/rsk/cli/RskCli.java
Vovchyk, line 81: I guess you could use `SyncMode` instead of `String`, so you won't need to parse it manually

rskj-core/src/main/java/co/rsk/config/SyncMode.java
```diff
@@ -0,0 +1,6 @@
+package co.rsk.config;
```
Vovchyk, line 1: cc is missing

rskj-core/src/main/java/co/rsk/config/SyncMode.java
```diff
@@ -0,0 +1,6 @@
...
+
+public enum SyncMode {
+    FULL,
```
Vovchyk, line 4: will it still support lowercase values?

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
```diff
@@ -41,6 +50,53 @@ public Config withConfig(Config config, String configValue) {
...
+                e.printStackTrace();
+                throw new RuntimeException("expecting URL in the format enode://PUBKEY@HOST:PORT", e);
+            }
+        }
+        private ConfigObject createConfigObjectFromSnapNode(String snapNode) {
+            try {
+                Map<String, ConfigValue> properties = new HashMap<>();
+                properties.put("url", ConfigValueFactory.fromAnyRef(snapNode));
+
+                return ConfigFactory.parseMap(properties).root();
```
Vovchyk, line 94: `ConfigValueFactory.fromMap( Collections.singletonMap("url", ConfigValueFactory.fromAnyRef(snapNode)) )` ?

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/cli/RskCli.java
rmoreliovlabs: Because picocli is case-sensitive and setting up a custom converter seems like overkill for this case, it’s simpler to stick with syncMode as a String.

This way, we handle case insensitivity by converting the string to uppercase before processing it. This approach will handle any case variations like full, FULL, or Snap correctly:

```
SYNC_MODE("sync-mode", true) {
    @Override
    public Config withConfig(Config config, String configValue) {
        SyncMode mode;
        try {
            mode = SyncMode.valueOf(configValue.toUpperCase());
        } catch (IllegalArgumentException e) {
            throw new IllegalArgumentException("Invalid sync mode: " + configValue + ". Valid options are <full> or <snap>.");
        }

        if (mode == SyncMode.SNAP) {
            return config.withValue(RskSystemProperties.PROPERTY_SNAP_CLIENT_ENABLED, ConfigValueFactory.fromAnyRef(true));
        }
        return config;
    }
}
```
Using the SyncMode enum directly would require a custom converter to handle case insensitivity, which seems like extra work for our needs.

Let me know if this approach works for you or if you have any other thoughts!

rskj-core/src/main/java/co/rsk/config/SyncMode.java
rmoreliovlabs: Given that the `NodeCliOptions` approach converts the string to uppercase before mapping it to the enum value, it will indeed handle lowercase values as well

rskj-core/src/main/java/co/rsk/config/SyncMode.java
rmoreliovlabs: solved

rskj-core/src/main/java/co/rsk/config/NodeCliOptions.java
rmoreliovlabs: solved

asoto-iov (approved):

PR #2656, opened 2024-07-30
Adds add invalid btc public key test

jeremy-then:
Adds add invalid btc public key test


rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2910,6 +2909,26 @@ void voteFederationChange_addMultiFederatorPublicKeyWithBadMstKey_returnsGeneric
 
         }
 
+        @Test
+        void voteFederationChange_callMultiKeyWithOnlyOneKey_ThrowsArrayIndexOutOfBoundsException() {
```
jeremy-then, line 2913: The name of the test changed, because `FederationChangeFunction.ADD.getKey()` was replaced with `FederationChangeFunction.ADD_MULTI.getKey()`, the logic changed, now it doesn't match with the branch name.

marcos-iov (approved):

PR #2659, opened 2024-07-31
Add getProposedFederation method

julia-zack:
- Create `getProposedFederation` method to get the proposed federation saved in the repository.
- Create tests for different scenarios

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
@@ -236,8 +236,10 @@ public class FederationStorageProviderImpl implements FederationStorageProvider 
                     return null;
                 }
 
-                // storage version is always present for non-null proposed federation
                 Optional<Integer> storageVersion = getStorageVersion(PROPOSED_FEDERATION_FORMAT_VERSION.getKey());
+                if (!storageVersion.isPresent()) {
+                    throw new IllegalStateException("Storage version should be present for non-null proposed federation");
+                }
                 return BridgeSerializationUtils.deserializeFederationAccordingToVersion(data, storageVersion.get(), federationConstants, activations);
             }
         );
--- a/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
@@ -4,12 +4,15 @@ import co.rsk.bitcoinj.core.NetworkParameters;
 import co.rsk.bitcoinj.core.UTXO;
 import co.rsk.bitcoinj.script.Script;
 import co.rsk.peg.BridgeSerializationUtils;
+import co.rsk.peg.bitcoin.ErpRedeemScriptBuilderUtils;
 import co.rsk.peg.federation.constants.FederationConstants;
 import co.rsk.peg.storage.StorageAccessor;
 import co.rsk.peg.vote.ABICallElection;
 import co.rsk.peg.vote.AddressBasedAuthorizer;
 import org.ethereum.config.blockchain.upgrades.ActivationConfig;
 import org.ethereum.vm.DataWord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 import java.util.HashMap;
@@ -21,6 +24,7 @@ import static co.rsk.peg.federation.FederationFormatVersion.*;
 import static org.ethereum.config.blockchain.upgrades.ConsensusRule.*;
 
 public class FederationStorageProviderImpl implements FederationStorageProvider {
+    private static final Logger logger = LoggerFactory.getLogger(FederationStorageProviderImpl.class);
     private final StorageAccessor bridgeStorageAccessor;
     private List<UTXO> newFederationBtcUTXOs;
     private List<UTXO> oldFederationBtcUTXOs;
@@ -238,7 +242,9 @@ public class FederationStorageProviderImpl implements FederationStorageProvider 
 
                 Optional<Integer> storageVersion = getStorageVersion(PROPOSED_FEDERATION_FORMAT_VERSION.getKey());
                 if (!storageVersion.isPresent()) {
-                    throw new IllegalStateException("Storage version should be present for non-null proposed federation");
+                    String message = "Storage version should be present for non-null proposed federation";
+                    logger.warn("[getProposedFederation] {}", message);
+                    throw new IllegalStateException(message);
                 }
                 return BridgeSerializationUtils.deserializeFederationAccordingToVersion(data, storageVersion.get(), federationConstants, activations);
             }
--- a/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
@@ -215,6 +215,31 @@ public class FederationStorageProviderImpl implements FederationStorageProvider 
     }
 
     @Override
+    public Optional<Federation> getProposedFederation(FederationConstants federationConstants, ActivationConfig.ForBlock activations) {
+        if (!activations.isActive(RSKIP419)) {
+            return Optional.empty();
+        }
+
+        if (proposedFederation != null || isProposedFederationSet) {
+            return Optional.ofNullable(proposedFederation);
+        }
+
+        proposedFederation = bridgeStorageAccessor.getFromRepository(
+            PROPOSED_FEDERATION.getKey(),
+            data -> {
+                if (data == null) {
+                    return null;
+                }
+                // storage version should be always present for non-null proposed federation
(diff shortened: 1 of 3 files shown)


rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -214,6 +214,37 @@ public void setProposedFederation(Federation proposedFederation) {
...
+        }
+
+        if (proposedFederation != null) {
+            return Optional.of(proposedFederation);
+        }
+
+        // reaching this point means the proposed federation was set to null
+        if (isProposedFederationSet) {
+            return Optional.empty();
+        }
```
marcos-iov, line 230: Although correct,  this got me confused and forced me to think about it for a while. I think we need to make it simpler and more straightforward

```suggestion
        if (proposedFederation != null || isProposedFederationSet) {
            return Optional.ofNullable(proposedFederation);
        }
```
This way it's consistent with getOldFederation for example.

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -214,6 +214,37 @@ public void setProposedFederation(Federation proposedFederation) {
...
+        proposedFederation = bridgeStorageAccessor.getFromRepository(
+            PROPOSED_FEDERATION.getKey(),
+            data -> {
+                if (data == null) {
+                    return null;
+                }
+
+                // storage version is always present for non-null proposed federation
+                Optional<Integer> storageVersion = getStorageVersion(PROPOSED_FEDERATION_FORMAT_VERSION.getKey());
+                return BridgeSerializationUtils.deserializeFederationAccordingToVersion(data, storageVersion.get(), federationConstants, activations);
```
marcos-iov, line 241: Isn't Sonar complaining that you are calling `get` on an Optional without checking if the value exists?

I think we should check that and throw an exception if for some reason there is no storage version present

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -999,6 +999,69 @@ private Integer getProposedFederationFormatVersionFromRepository() {
                 .map(BridgeSerializationUtils::deserializeInteger)
                 .orElse(null);
         }
+
+        @Test
+        void getProposedFederation_whenThereIsNoProposedFederationSavedNorSet_shouldReturnEmpty() {
+            assertFalse(federationStorageProvider.getProposedFederation(federationConstants, allActivations).isPresent());
```
marcos-iov, line 1005: Nit, but I prefer to store the result of the call to the storage provider in a variable, then apply the assertion on that variable. Cleaner and easier to read

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -999,6 +999,69 @@ private Integer getProposedFederationFormatVersionFromRepository() {
...
+        @Test
+        void getProposedFederation_whenProposedFederationIsCached_shouldReturnCachedFederation() {
+            // first we have to save the proposed fed so the repo is not empty
+            bridgeStorageAccessor.saveToRepository(PROPOSED_FEDERATION_FORMAT_VERSION.getKey(), proposedFederation.getFormatVersion(), BridgeSerializationUtils::serializeInteger);
+            bridgeStorageAccessor.saveToRepository(PROPOSED_FEDERATION.getKey(), proposedFederation, BridgeSerializationUtils::serializeFederation);
+            // this should set the proposed fed in proposedFederation field
+            federationStorageProvider.getProposedFederation(federationConstants, allActivations);
+
+            // saving in the repo another fed to make sure the cached value is the one being returned
+            Federation anotherFederation = new StandardMultiSigFederationBuilder().build();
```
marcos-iov, line 1059: Let's use the P2shErp fed, more realistic

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -999,6 +999,69 @@ private Integer getProposedFederationFormatVersionFromRepository() {
...
+            bridgeStorageAccessor.saveToRepository(PROPOSED_FEDERATION.getKey(), proposedFederation, BridgeSerializationUtils::serializeFederation);
+
+            bridgeStorageAccessor.saveToRepository(PROPOSED_FEDERATION_FORMAT_VERSION.getKey(), null, BridgeSerializationUtils::serializeInteger);
+            bridgeStorageAccessor.saveToRepository(PROPOSED_FEDERATION.getKey(), null, BridgeSerializationUtils::serializeFederation);
+
+            assertFalse(federationStorageProvider.getProposedFederation(federationConstants, allActivations).isPresent());
+        }
+
+        @Test
+        void getProposedFederation_whenProposedFederationIsCached_shouldReturnCachedFederation() {
```
marcos-iov, line 1051: Nice, maybe another case when there is a fed in storage. Call `set`, not `save`, and call `get`. Should return the one setted and not the one in storage

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
marcos-iov: Remember to check sonar and intellij warnings

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
julia-zack: good one, thanks

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julia-zack: yeah at first i wrote it that way, but i thought it was better to actually know when it returns empty and when it returns an actual value. But no worries, i can change it

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -236,8 +236,10 @@ public Optional<Federation> getProposedFederation(FederationConstants federation
                     return null;
                 }
 
-                // storage version is always present for non-null proposed federation
                 Optional<Integer> storageVersion = getStorageVersion(PROPOSED_FEDERATION_FORMAT_VERSION.getKey());
+                if (!storageVersion.isPresent()) {
```
marcos-iov, line 240: Maybe log this also?

rskj-core/src/test/java/co/rsk/peg/federation/FederationStorageProviderImplTests.java
```diff
@@ -1056,11 +1088,21 @@ void getProposedFederation_whenProposedFederationIsCached_shouldReturnCachedFede
...
+            Federation anotherFederation = createAnotherFederation();
             bridgeStorageAccessor.saveToRepository(PROPOSED_FEDERATION_FORMAT_VERSION.getKey(), anotherFederation.getFormatVersion(), BridgeSerializationUtils::serializeInteger);
             bridgeStorageAccessor.saveToRepository(PROPOSED_FEDERATION.getKey(), anotherFederation, BridgeSerializationUtils::serializeFederation);
 
-            assertEquals(Optional.of(proposedFederation), federationStorageProvider.getProposedFederation(federationConstants, allActivations));
+            Optional<Federation> actualProposedFederation = federationStorageProvider.getProposedFederation(federationConstants, allActivations);
+            assertEquals(Optional.of(proposedFederation), actualProposedFederation);
+        }
+
+        private Federation createAnotherFederation() {
```
marcos-iov, line 1099: Pretty sure we have something like this in FederationTestUtils. Let's use that to avoid ending up with methods like this all over the code base

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -236,8 +236,10 @@ public Optional<Federation> getProposedFederation(FederationConstants federation
...
+                    throw new IllegalStateException("Storage version should be present for non-null proposed federation");
```
marcos-iov, line 241: A test for this case?

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -4,12 +4,15 @@
 import co.rsk.bitcoinj.core.UTXO;
 import co.rsk.bitcoinj.script.Script;
 import co.rsk.peg.BridgeSerializationUtils;
+import co.rsk.peg.bitcoin.ErpRedeemScriptBuilderUtils;
```
marcos-iov, line 7: unused?

julia-zack: solved

marcos-iov (approved):

PR #2677, opened 2024-08-08
Add fund transaction creation

julia-zack:


rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -748,6 +749,74 @@ private void logCommitmentWithVotedFederation(BridgeEventLogger eventLogger, Fed
...
+        int neededSignaturesSize = 72 * federation.getNumberOfSignaturesRequired();
+        return redeemScriptSize + neededSignaturesSize;
+    }
+
+    private SendRequest createSendRequest(BtcTransaction transaction, Coin feePerKb) {
+        SendRequest sendRequest = SendRequest.forTx(transaction);
+        sendRequest.changeAddress = getActiveFederationAddress();
+        sendRequest.feePerKb = feePerKb;
+        sendRequest.missingSigsMode = Wallet.MissingSigsMode.USE_OP_ZERO;
+        sendRequest.recipientsPayFees = false;
```
julia-zack, line 815: we dont want the proposed fed to pay for the fees since the amount to send was calculated quite precise

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -748,6 +749,74 @@ private void logCommitmentWithVotedFederation(BridgeEventLogger eventLogger, Fed
...
+    }
+
+    private void addOutputToProposedFederation(BtcTransaction svpFundTransaction, Federation proposedFederation, Coin feePerKb) {
+        long transactionSizeToBeSentFromProposedFederation = calculateTransactionSizeToBeSentFromFederation(proposedFederation);
+        Coin neededValueToSendFromProposedFederation = feePerKb.multiply(transactionSizeToBeSentFromProposedFederation);
+
+        svpFundTransaction.addOutput(neededValueToSendFromProposedFederation, proposedFederation.getAddress());
+    }
+
+    private long calculateTransactionSizeToBeSentFromFederation(Federation federation) {
```
julia-zack, line 782: this calculation is a guide for now, it will be re-calculated in another task.

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
marcos-iov: Interesting, this way no matter how high the current network fees are we ensure the proposed will receive enough funds to send back, right?

Let's take a note to double check this anyway

rskj-core/src/main/java/co/rsk/peg/bitcoin/BitcoinUtils.java
```diff
@@ -61,4 +62,9 @@ public static Optional<Script> extractRedeemScriptFromInput(TransactionInput txI
             return Optional.empty();
         }
     }
+
+    public static Script createBaseInputScriptThatSpendsFromTheFederation(Federation federation) {
```
marcos-iov, line 66: Is this still needed?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -747,6 +748,38 @@ private void logCommitmentWithVotedFederation(BridgeEventLogger eventLogger, Fed
         logger.debug("[logCommitmentWithVotedFederation] Voted federation committed: {}", votedFederation.getAddress());
     }
 
+    @Override
+    public BtcTransaction createAndSetSvpFundTransactionWithoutSignatures(Coin spendableValueFromProposedFederation, Coin feePerKb, Wallet activeFederationWallet) throws InsufficientMoneyException {
```
marcos-iov, line 752: The fact that the method name has an `and` as part of it's name makes me think it's doing more than one thing and probably Uncle Bob does not approve. Should we split this into 2 separate methods?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupport.java
```diff
@@ -53,5 +53,7 @@ public interface FederationSupport {
 
     void updateFederationCreationBlockHeights();
 
+    BtcTransaction createAndSetSvpFundTransactionWithoutSignatures(Coin spendableValueFromProposedFederation, Coin feePerKb, Wallet activeFederationWallet) throws InsufficientMoneyException;
```
marcos-iov, line 56: Maybe the `spendableValueFromProposedFederation` should come from BridgeConstants?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -962,6 +962,22 @@ public void updateCollections(Transaction rskTx) throws IOException {
         updateFederationCreationBlockHeights();
     }
 
+    protected void createAndProcessSvpFundTransactionWithoutSignatures() throws InsufficientMoneyException, IOException {
```
marcos-iov, line 965: Where is this used?

Same comment regarding the `and` in the method name, although in this case I think it'd be wiser to renamed rather than split.

Let's discuss this a bit, I guess it depends a bit on who the caller is going to be

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2191,6 +2195,50 @@ void clearRetiredFederation_whenHavingOldFederation_removesOldFederation() {
         assertThat(oldFederation, is(nullValue()));
     }
 
+    @Test
+    @Tag("create and process svp fund transaction unsigned")
+    void createAndProcessSvpFundTransactionWithoutSignatures_savesExpectedTransactionInStorage() throws InsufficientMoneyException {
```
marcos-iov, line 2200: Any unexpected corner case we'd like to test?

No UTXOs available at the moment? Fees too high? InsufficientMoneyException?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -747,6 +748,38 @@ private void logCommitmentWithVotedFederation(BridgeEventLogger eventLogger, Fed
...
+            String message = "Proposed federation should be present when creating SVP fund transaction.";
+            logger.warn(message);
+            throw new IllegalStateException(message);
+        }
+        Federation proposedFederation = proposedFederationOpt.get();
+
+        BtcTransaction svpFundTransactionUnsigned = new BtcTransaction(constants.getBtcParams());
+        svpFundTransactionUnsigned.setVersion(BTC_TX_VERSION_2);
+
+        svpFundTransactionUnsigned.addOutput(spendableValueFromProposedFederation, proposedFederation.getAddress());
```
marcos-iov, line 764: There's supposed to be an output for the flyover federation as well, right? Is that part of another task?

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
```diff
@@ -2191,6 +2195,50 @@ void clearRetiredFederation_whenHavingOldFederation_removesOldFederation() {
...
+        // assert the amount sent was correct
+        TransactionOutput outputToProposedFederation = outputToProposedFederationOpt.get();
+        assertEquals(outputToProposedFederation.getValue(), spendableValueFromProposedFederation);
+
+        // assert all the other outputs are the change sent to the active federation
+        Script activeFederationScriptPubKey = activeFederation.getP2SHScript();
+        assertTrue(svpFundTransactionUnsignedOutputs.stream()
+            .filter(output -> !output.equals(outputToProposedFederation))
+            .allMatch(output -> output.getScriptPubKey().equals(activeFederationScriptPubKey)));
+    }
```
marcos-iov, line 2240: assert the transaction is added to the pegout tx index?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
marcos-iov: Should we emit release_requested event also? Where does that come into play? And `pegout_transaction_created` event

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: isn't that [this assertion](https://github.com/rsksmart/rskj/pull/2677/files#diff-bca5614b0a0ce7ba81d79322090b66cee8d9289e3430798d67767568abc8491aR637-R641)?

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
julia-zack: i think that's always a risk, because the fees could vary between the fund transaction and the spend transaction are sent. But I wanted to make sure that the value chosen to be the one received from the proposed fed will actually be the value sent to it

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupport.java
julia-zack: good one. Thanks.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: the `protected` is for testing the method.
`release_requested` event can't be emitted because the rskTxHash is null right?
`pegout_transaction_created` should be added. Thanks!

julia-zack: regarding the method name / behavior, yep maybe we can discuss it a bit :)

rskj-core/src/test/java/co/rsk/peg/federation/FederationSupportImplTest.java
julia-zack: mmm, I think that should be a test for `completeTx` method. Wdyt?

julia-zack: i mean, if there are no utxos, `completeTx` should throw the exception. Is not being thrown from these new methods.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -962,6 +962,86 @@ public void updateCollections(Transaction rskTx) throws IOException {
...
+            String message = "The output for the proposed federation should be present in the svp fund transaction.";
+            logger.warn(message);
+            throw new IllegalStateException(message);
+        }
+        TransactionOutput outputToProposedFederation = outputToProposedFederationOpt.get();
+
+        return outputToProposedFederation.getValue();
+    }
+
+    private void performPegoutActions(BtcTransaction pegoutTransaction, Keccak256 rskTxHash, Coin requestedPegoutAmount) throws IOException {
```
marcos-iov, line 1024:
```suggestion
    private void confirmReleaseRequest(BtcTransaction pegoutTransaction, Keccak256 releaseCreationTxHash, Coin requestedAmount) throws IOException {
```
Tried to give a more meaningful name to the method and params. The idea is that this can be used for migration, pegouts, and fundTx right?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -962,6 +962,86 @@ public void updateCollections(Transaction rskTx) throws IOException {
...
+        addPegoutToPegoutsWaitingForConfirmations(pegoutTransaction, rskTxHash);
+        savePegoutTxSigHash(pegoutTransaction);
+        logReleaseRequested(rskTxHash.getBytes(), pegoutTransaction, requestedPegoutAmount);
+        logPegoutTransactionCreated(pegoutTransaction);
+    }
+
+    private void addPegoutToPegoutsWaitingForConfirmations(BtcTransaction pegoutTransaction, Keccak256 rskTxHash) throws IOException {
```
marcos-iov, line 1031: Isn't this gonna clash with the already existing method that has a very similar name?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -962,6 +962,86 @@ public void updateCollections(Transaction rskTx) throws IOException {
...
+        PegoutsWaitingForConfirmations pegoutsWaitingForConfirmations = provider.getPegoutsWaitingForConfirmations();
+        pegoutsWaitingForConfirmations.add(pegoutTransaction, rskExecutionBlock.getNumber(), rskTxHash);
+    }
+
+    private void logReleaseRequested(byte[] rskTxHashSerialized, BtcTransaction pegoutTransaction, Coin requestedPegoutAmount) {
```
marcos-iov, line 1036: Why not receive the hash object here and serialize it in the method? Seems like that's a responsibility of this method

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -522,6 +553,223 @@ void save_callsFederationSupportSave() throws IOException {
...
+        private Sha256Hash svpFundTransactionHashUnsigned;
+
+        private Transaction rskTx;
+        private Keccak256 rskTxHash;
+
+        private PegoutsWaitingForConfirmations pegoutsWaitingForConfirmations;
+
+        @BeforeEach
+        void setUp() {
+            rskExecutionBlock = mock(Block.class);
```
marcos-iov, line 573: Use an actual block like we did in other tests?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -522,6 +553,223 @@ void save_callsFederationSupportSave() throws IOException {
...
+            long rskExecutionBlockNumber = 1000L;
+            when(rskExecutionBlock.getNumber()).thenReturn(rskExecutionBlockNumber);
+
+            activeFederation = FederationTestUtils.getGenesisFederation(federationConstantsMainnet);
```
marcos-iov, line 577: Not sure using the genesis federation is a good idea, it's a legacy standard federation. I'd use the current federation type

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -522,6 +553,223 @@ void save_callsFederationSupportSave() throws IOException {
...
+            federationSupport = mock(FederationSupport.class);
+            when(federationSupport.getActiveFederation()).thenReturn(activeFederation);
+            when(federationSupport.getActiveFederationAddress()).thenReturn(activeFederation.getAddress());
+
+            List<UTXO> utxos = BitcoinTestUtils.createUTXOs(10, activeFederation.getAddress());
```
marcos-iov, line 582:
```suggestion
            List<UTXO> activeFederationUTXOs = BitcoinTestUtils.createUTXOs(10, activeFederation.getAddress());
```
Let's use meaningful variable names

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -522,6 +553,223 @@ void save_callsFederationSupportSave() throws IOException {
...
+            when(federationSupport.getActiveFederationBtcUTXOs()).thenReturn(utxos);
+
+            proposedFederation = new P2shErpFederationBuilder().build();
+            when(federationSupport.getProposedFederation()).thenReturn(Optional.of(proposedFederation));
+
+            Coin feePerKb = Coin.valueOf(1000);
+            when(feePerKbSupport.getFeePerKb()).thenReturn(feePerKb);
+
+            rskTx = mock(Transaction.class);
+            rskTxHash = PegTestUtils.createHash3(1);
```
marcos-iov, line 592: Should this method be part of `RskTestUtils` actually?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -522,6 +553,223 @@ void save_callsFederationSupportSave() throws IOException {
...
+                .withProvider(bridgeStorageProvider)
+                .withActivations(activationsAfterForks)
+                .withFederationSupport(federationSupport)
+                .withFeePerKbSupport(feePerKbSupport)
+                .withExecutionBlock(rskExecutionBlock)
+                .build();
+        }
+
+        @Test
+        void createSvpFundTransactionWithoutSignatures_createsExpectedSvpFundTransaction() throws InsufficientMoneyException {
```
marcos-iov, line 609: Any corner cases to consider besides the happy path?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: yea i though about addressing that in the task in charge of doing that refactor. wdyt?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -962,6 +962,86 @@ public void updateCollections(Transaction rskTx) throws IOException {
...
+        Optional<Federation> proposedFederationOpt = federationSupport.getProposedFederation();
+        if (!proposedFederationOpt.isPresent()) {
+            String message = "Proposed federation should be present when processing SVP fund transaction.";
+            logger.warn(message);
+            throw new IllegalStateException(message);
+        }
+        Federation proposedFederation = proposedFederationOpt.get();
+
+        BtcTransaction svpFundTransactionUnsigned = createSvpFundTransactionWithoutSignatures(proposedFederation);
+        provider.setSvpFundTxHashUnsigned(svpFundTransactionUnsigned.getHash());
```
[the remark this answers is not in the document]
julia-zack: probably the same, but lets analyze that better in a new task

PR #2686, opened 2024-08-15
Add svp fund tx creation

julia-zack:
-Add methods for creating and processing the svp fund transaction (without signatures)
-Add tests

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -36,7 +36,6 @@ import co.rsk.peg.constants.BridgeConstants;
 import co.rsk.core.RskAddress;
 import co.rsk.crypto.Keccak256;
 import co.rsk.panic.PanicProcessor;
-import co.rsk.peg.bitcoin.*;
 import co.rsk.peg.btcLockSender.BtcLockSender.TxSenderAddressType;
 import co.rsk.peg.btcLockSender.BtcLockSenderProvider;
 import co.rsk.peg.federation.*;
@@ -974,7 +973,7 @@ public class BridgeSupport {
         settleReleaseRequest(pegoutsWaitingForConfirmations, svpFundTransactionUnsigned, rskTx.getHash(), spendableValueFromProposedFederation);
     }
 
-    private BtcTransaction createSvpFundTransaction(Federation proposedFederation, Coin spendableValueFromProposedFederation) throws IOException, InsufficientMoneyException {
+    private BtcTransaction createSvpFundTransaction(Federation proposedFederation, Coin spendableValueFromProposedFederation) throws InsufficientMoneyException {
         Wallet activeFederationWallet = getActiveFederationWallet(true);
 
         BtcTransaction svpFundTransaction = new BtcTransaction(bridgeConstants.getBtcParams());
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -31,6 +31,7 @@ import co.rsk.bitcoinj.store.BlockStoreException;
 import co.rsk.bitcoinj.wallet.SendRequest;
 import co.rsk.bitcoinj.wallet.Wallet;
 import co.rsk.core.types.bytes.Bytes;
+import co.rsk.peg.bitcoin.*;
 import co.rsk.peg.constants.BridgeConstants;
 import co.rsk.core.RskAddress;
 import co.rsk.crypto.Keccak256;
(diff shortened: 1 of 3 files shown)


rskj-core/src/main/java/co/rsk/peg/constants/BridgeConstants.java
```diff
@@ -106,6 +107,8 @@ public Coin getMinimumPeginTxValue(ActivationConfig.ForBlock activations) {
 
     public Coin getSpendableValueFromProposedFederation() { return minimumPegoutTxValue.multiply(2); }
 
+    public Keccak256 getProposedFederationFlyoverPrefix() { return new Keccak256("0000000000000000000000000000000000000000000000000000000000000001"); }
```
nathanieliov, line 110: Could you tell me what the reason is for this? Why do we need this prefix to send the funds to the proposed federation? Isn't enough the address of the P2shScript?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -962,6 +962,76 @@ public void updateCollections(Transaction rskTx) throws IOException {
...
+            logger.warn(message);
+            throw new IllegalStateException(message);
+        }
+        Federation proposedFederation = proposedFederationOpt.get();
+
+        Coin spendableValueFromProposedFederation = bridgeConstants.getSpendableValueFromProposedFederation();
+        BtcTransaction svpFundTransactionUnsigned = createSvpFundTransaction(proposedFederation, spendableValueFromProposedFederation);
+
+        provider.setSvpFundTxHashUnsigned(svpFundTransactionUnsigned.getHash());
+        settleReleaseRequest(svpFundTransactionUnsigned, rskTx.getHash(), spendableValueFromProposedFederation);
```
nathanieliov, line 978: I believe we want to move away from the "release" name reference. It would be best if you used "pegout" instead.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -962,6 +962,76 @@ public void updateCollections(Transaction rskTx) throws IOException {
...
+    }
+
+    private BtcTransaction createSvpFundTransaction(Federation proposedFederation, Coin spendableValueFromProposedFederation) throws IOException, InsufficientMoneyException {
```
nathanieliov, line 981:
```suggestion
    private BtcTransaction createSvpFundBtcTransaction(Federation proposedFederation, Coin spendableValueFromProposedFederation) throws IOException, InsufficientMoneyException {
```

Wdyt?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: that name was a [suggestion](https://github.com/rsksmart/rskj/pull/2677#discussion_r1717490742) from marcos. Wdyt @marcos-iov ?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: well, the method returns a `BtcTransaction`. And the fund transaction is a btc transaction. I think adding the `Btc` part is redundant

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
marcos-iov: I think release is a good name to cover the case for peg-out, migration, or fund tx.

We are releasing funds

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -962,6 +962,76 @@ public void updateCollections(Transaction rskTx) throws IOException {
...
+    }
+
+    private void settleReleaseRequest(BtcTransaction pegoutTransaction, Keccak256 releaseCreationTxHash, Coin requestedAmount) throws IOException {
+        addPegoutToPegoutsWaitingForConfirmations(pegoutTransaction, releaseCreationTxHash);
+        savePegoutTxSigHash(pegoutTransaction);
+        logReleaseRequested(releaseCreationTxHash, pegoutTransaction, requestedAmount);
+        logPegoutTransactionCreated(pegoutTransaction);
+    }
+
+    private void addPegoutToPegoutsWaitingForConfirmations(BtcTransaction pegoutTransaction, Keccak256 releaseCreationTxHash) throws IOException {
```
marcos-iov, line 1017: There is another method in this same class with a very similar name and very similar behaviour. Shouldn't we try to consolidate this?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -522,6 +551,232 @@ void save_callsFederationSupportSave() throws IOException {
...
+            // act & assert
+            IllegalStateException exception = assertThrows(IllegalStateException.class,
+                () -> bridgeSupport.processSvpFundTransactionUnsigned(rskTx));
+
+            String expectedMessage = "Proposed federation should be present when processing SVP fund transaction.";
+            assertEquals(expectedMessage, exception.getMessage());
+        }
+
+        @Test
+        void processSvpFundTransactionWithoutSignatures_createsExpectedTransactionAndSavesTheHashInStorageEntryAndPerformsPegoutActions() throws Exception {
```
marcos-iov, line 625: No other unhappy path? Not enough UTXOs in the Bridge to create the fund tx?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: theres a new task created to do it and to reuse new settle release request method whenever creating a pegout

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
julia-zack: not enough utxos should throw an `InsufficientMoneyException`, but from `completeTx` method.
I mean, that test should be part of Wallet testing class.

The only exception being thrown from this method implementation is when not having a proposed fed

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -522,6 +551,235 @@ void save_callsFederationSupportSave() throws IOException {
...
+        }
+
+        private void assertSvpReleaseWasSettled() throws IOException {
+            PegoutsWaitingForConfirmations pegoutsWaitingForConfirmations = bridgeStorageProvider.getPegoutsWaitingForConfirmations();
+            assertPegoutWasAddedToPegoutsWaitingForConfirmations(pegoutsWaitingForConfirmations, svpFundTransactionHashUnsigned, rskTxHash);
+            getSvpFundTransactionFromPegoutsMap(pegoutsWaitingForConfirmations);
+
+            assertPegoutTxSigHashWasSaved(svpFundTransactionUnsigned);
+            assertLogReleaseRequested(rskTxHash, svpFundTransactionHashUnsigned, bridgeMainNetConstants.getSpendableValueFromProposedFederation());
+            assertLogPegoutTransactionCreated(svpFundTransactionUnsigned);
```
julia-zack, line 650: ideally we should use an `assertReleaseRequestWasSettled` that call all these methods, like in prod code. But in this case we have to "recover" the transaction from the pegouts map, first asserting it is there using its hash.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
marcos-iov: Disagree, I think that if we are testing behavior it's a good test case. Regardless of where the error happens

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -522,6 +551,234 @@ void save_callsFederationSupportSave() {
...
+            assertSvpFundTxHashUnsignedWasSavedInStorage();
+            assertSvpReleaseWasSettled();
+            assertSvpFundTransactionHasExpectedInputsAndOutputs();
+        }
+
+        private void assertSvpFundTxHashUnsignedWasSavedInStorage() {
+            Optional<Sha256Hash> svpFundTransactionHashUnsignedOpt = bridgeStorageProvider.getSvpFundTxHashUnsigned();
+            assertTrue(svpFundTransactionHashUnsignedOpt.isPresent());
+
+            svpFundTransactionHashUnsigned = svpFundTransactionHashUnsignedOpt.get();
```
marcos-iov, line 638: Anything missing after this?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -522,6 +551,234 @@ void save_callsFederationSupportSave() {
...
+            PegoutsWaitingForConfirmations pegoutsWaitingForConfirmations = bridgeStorageProvider.getPegoutsWaitingForConfirmations();
+            assertPegoutWasAddedToPegoutsWaitingForConfirmations(pegoutsWaitingForConfirmations, svpFundTransactionHashUnsigned, rskTxHash);
+            getSvpFundTransactionFromPegoutsMap(pegoutsWaitingForConfirmations);
+
+            assertPegoutTxSigHashWasSaved(svpFundTransactionUnsigned);
+            assertLogReleaseRequested(rskTxHash, svpFundTransactionHashUnsigned, bridgeMainNetConstants.getSpendableValueFromProposedFederation());
+            assertLogPegoutTransactionCreated(svpFundTransactionUnsigned);
+        }
+
+        private void getSvpFundTransactionFromPegoutsMap(PegoutsWaitingForConfirmations pegoutsWaitingForConfirmations) {
```
marcos-iov, line 651: This is a bit strange. The method is called `get` but it's void

julia-zack: good one, thanks

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
julia-zack: Done

marcos-iov (approved):
Super nice! Well done!

PR #2688, opened 2024-08-19
Adding Checkstyle linter

rmoreliovlabs:
# Add Checkstyle with Google Code Conventions and Spotless for Google Java Code Formatting

depends on powpeg pr: https://github.com/rsksmart/powpeg-node/pull/324/files

## Description
This PR introduces **Checkstyle 8.45**, configured with Google Code Conventions, and adds **Spotless** to our project for code formatting using the official Google Java Format. **Checkstyle** will analyze the codebase to enforce coding standards, while **Spotless** ensures consistent code formatting. This addition aims to improve code quality and maintain consistency for new contributions.

Due to security concerns with the previous unofficial Google Java Formatter plugin by [sherter](https://github.com/sherter/google-java-format-gradle-plugin), which was from a less-established source, **Spotless** was chosen as a more widely used and trusted alternative. **Spotless** not only integrates Google's official format but also offers better reliability and security for formatting our code.

### Highlights:
- Added **Checkstyle 8.45** configured with Google Code Conventions for code quality enforcement
- Integrated **Spotless** to use the official Google Java Format for consistent code styling
- Configured **Checkstyle** to analyze only newly added code, avoiding disruption of existing code
- **Spotless** offers enhanced reliability and security for code formatting compared to previous tools

## Motivation and Context
This update is designed to help our project follow the latest best practices and industry standards for Java-based open-source projects. By adding Checkstyle with Google Code Conventions and the Google Java Formatter, we’re making sure that all new code meets high-quality standards and maintains consistent formatting.

## Basic Implementation:

- **Checkstyle Version**: Added Checkstyle 8.45 to ensure compatibility with Java 8.
- **Configuration**: Integrated Google Code Conventions to align with industry standards.
- **Suppressions File**: Included a suppressions.xml file to manage specific rule exceptions. This file allows us to suppress certain Checkstyle rules where they might not apply or are too restrictive.

### Why Use Spotless with Official Google Java Formatter

I have chosen **Spotless** with the official Google Java Format over the unofficial `sherter` plugin due to several reasons:

- **Security**: Spotless is a widely used, trusted tool with a well-established reputation, whereas the `sherter` plugin comes from a less-established source, raising potential security concerns.
- **Reliability**: Spotless integrates directly with Google's official formatting standards, ensuring consistent and accurate code formatting.

**File Checking Mechanism:** The method for filtering files has been updated. Previously, Checkstyle and the Google Formatter relied on the `lastModified()` timestamp of files, which could lead to inconsistencies. For example, if you modify file A and then undo those changes, the `lastModified()` timestamp might still indicate that the file was recently changed, causing Checkstyle and the formatter to run on it even though the modifications were reverted.

To address this, we now use Git commands to identify modified, staged, and committed files based on a specific timestamp. This method ensures that Checkstyle and **Spotless** formatting checks are applied more accurately. For instance, with the new approach, if you revert changes to file A, Git will no longer include it in the list of modified files, so neither Checkstyle nor Spotless will check it unnecessarily. This ensures more precise and reliable code quality and formatting checks.

**To run Checkstyle** and verify the new code, use the following command:`./gradlew clean rskj-core:checkstyleMain`
**To verify files to format without formatting them**, use the following command:`./gradlew clean rskj-core:spotlessCheck`
**To format code**, use the following command:`./gradlew clean rskj-core:spotlessApply`

**Example of Suppression:**
In `suppressions.xml`, you can suppress a rule like this:

```
<suppressions>
    <suppress checks="CheckstyleRuleName" files=".*SomeFile.java"/>
</suppressions>
```
This example suppresses a specific Checkstyle rule for `SomeFile.java`. The suppressions file helps tailor the Checkstyle configuration to fit our project needs more precisely.

## Example of the Formatting Mechanism:

The file filtering mechanism now uses Git commands to determine which files have been modified, staged, or committed since a specific timestamp. This ensures that only the relevant files are processed, based on actual changes tracked by Git.

**Example**:

**Modified File** (`/rskj/rskj-core/src/main/java/co/rsk/config/GasLimitConfig.java`):

**- Before Formatting:**
```
**/**
 * Wraps configuration for Mining, which is usually derived from configuration files.
 */**
public class GasLimitConfig {
    private final int minGasLimit;
    private final long targetGasLimit;
    private final boolean isTargetGasLimitForced;

    public GasLimitConfig(int minGasLimit, long targetGasLimit, boolean isTargetGasLimitForced) {
        this.minGasLimit = minGasLimit;
        this.targetGasLimit = targetGasLimit;
        this.isTargetGasLimitForced = isTargetGasLimitForced;
    }

    public int getMininimum() {
        return minGasLimit;
    }

    public long getTarget() {
        return targetGasLimit;
    }

    public boolean isTargetForced() {
        return isTargetGasLimitForced;
    }
}

```

**- After Formatting:**
```
package co.rsk.config;

**/** Wraps configuration for Mining, which is usually derived from configuration files. */**
public class GasLimitConfig {
    private final int minGasLimit;
    private final long targetGasLimit;
    private final boolean isTargetGasLimitForced;

    public GasLimitConfig(int minGasLimit, long targetGasLimit, boolean isTargetGasLimitForced) {
        this.minGasLimit = minGasLimit;
        this.targetGasLimit = targetGasLimit;
        this.isTargetGasLimitForced = isTargetGasLimitForced;
    }

    public int getMininimum() {
        return minGasLimit;
    }

    public long getTarget() {
        return targetGasLimit;
    }

    public boolean isTargetForced() {
        return isTargetGasLimitForced;
    }
}
```
**And then, if we run Checkstyle, the console might look something like this:**
```
> Task :rskj-core:checkstyleMain
[ant:checkstyle] [WARN] /Users/reynoldmorel/Documents/projects/rskj/rskj-core/src/main/java/co/rsk/config/GasLimitConfig.java:27:5: Missing a Javadoc comment. [MissingJavadocMethod]
Checkstyle rule violations were found. See the report at: file:///Users/reynoldmorel/Documents/projects/rskj/rskj-core/build/reports/checkstyle/main.html
Checkstyle files with violations: 1
Checkstyle violations by severity: [warning:1]

```

* **Other information**:

`fed:updating_dependencies_for_spotless`


rskj-core/build.gradle
```diff
@@ -1,10 +1,127 @@
...
+    id "com.diffplug.spotless" version "6.13.0"
 }
 
 apply plugin: 'maven-publish'
 apply plugin: 'jacoco'
 
+import java.nio.file.Paths
+
+// Define static timestamp
+def staticTimestamp = '2024-09-03T02:17:00'
```
fmacleal, line 13: issue:

If this is fixed to a current date, does it mean that if we rebase or merge after this date but a file that wasn't supposed to be checked with a date after this one could be included in our commit? Because we have a lot of feature branches that lives for a long time (like the snap synch one) and we keep merging master to it to update. With this logic, will these files from master be included or not in the checking?

Maybe worth to test this on the snap-synch branch? 

A possibility could be passed the cutoff date as parameter... 🤔

rskj-core/build.gradle
```diff
@@ -1,10 +1,127 @@
...
+    allFiles = allFiles.unique().findAll { file -> file }
+
+    allFiles = allFiles.findAll { file ->
+        file.endsWith('.java')
+    }
+
+    return allFiles.collect { file -> Paths.get(file).toAbsolutePath().toFile() }
+}
+
+// Method to configure Checkstyle tasks
```
fmacleal, line 86: nitpick:

I think the name of the method it's pretty much the same what states this comment. I suggest to remove this.

```suggestion
```

rskj-core/build.gradle
```diff
@@ -1,10 +1,127 @@
...
+    configFile = file("$rootDir/config/checkstyle/checkstyle.xml")
+}
+
+spotless {
+    java {
+        googleJavaFormat()
+        indentWithTabs(2)
+        indentWithSpaces(4)
+        target filteredFiles
+
```
fmacleal, line 29: nitpick:

Unnecessary new line.

```suggestion
```

rskj-core/build.gradle
```diff
@@ -416,3 +533,8 @@ static def amendPathIfNeeded(details) {
         details.path = newPath
     }
 }
+
+tasks.named('check').configure {
+    dependsOn 'checkstyleMain'
+    dependsOn 'checkstyleTest'
```
fmacleal, line 539: question:

What does this method? It's creating a `check` task? Will this check also run the `checkstyleIntegrationTest` and  `checkstyleJmh`? Because if not, I would suggest to add them here as well. If we are adding a global check, it should check everything.

fmacleal (commented):
Great job, the spotless addition seems a very nice one. I think that these tasks will be very useful to make the code follow a pattern. 👏 

I have a few comments regarding the date, once we clear that out, I will approve.

apancorb:
Hey guys, a couple of months ago I worked on something similar. Leaving [PR](https://github.com/rsksmart/powpeg-node/pull/250) here so you can check it out. This is a great integration!

rskj-core/build.gradle
apancorb, line 13: There is also the option to use [ratchetFrom](https://github.com/rsksmart/powpeg-node/pull/250/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R176), which only formats from a given git tag

rmoreliovlabs: Thank you for the suggestion! I considered using `ratchetFrom` to format only changes from a specific git tag. However, [the last tag](https://github.com/rsksmart/rskj/tags) in the repository is from July 10, which is quite a while ago. Using ratchetFrom would include all changes made since that tag, which would affect files modified after that date—something we want to avoid in this case

rmoreliovlabs: Thanks for bringing this up! I see where you're coming from regarding the fixed timestamp. The idea was to enforce code conventions from a specific point in time, so any code merged or rebased after that date would follow the latest standards. The assumption was that branches or PRs that are older would need to rebase or merge from master to align with these conventions

However, I understand this might not work perfectly for long-lived branches that are regularly updated from master

What are your thoughts on this? One idea could be to keep the timestamp in a file and update it manually as needed, or perhaps use a GitHub Actions workflow to manage it

Let me know if you have other suggestions or ideas!

rskj-core/build.gradle
rmoreliovlabs: addressed!

rskj-core/build.gradle
rmoreliovlabs: Addressed! I’ve just added `checkstyleJmh` and `checkstyleIntegrationTest` to our `build.gradle` file. Now, when you run the `check` task, it will automatically trigger style checks for `checkstyleMain`, `checkstyleTest`, `checkstyleJmh`, and `checkstyleIntegrationTest`. This means you no longer need to manually run each check individually. Instead, running the check task will execute all these style checks in one go, making the process more efficient

rskj-core/build.gradle
rmoreliovlabs: solved!

rmoreliovlabs:
depends on powpeg pr: https://github.com/rsksmart/powpeg-node/pull/324/files

rskj-core/build.gradle
```diff
@@ -1,10 +1,156 @@
...
+checkstyle {
+    toolVersion = '8.45'
+    configFile = file("$rootDir/config/checkstyle/checkstyle.xml")
+}
+
+spotless {
+    java {
+        googleJavaFormat()
+        indentWithTabs(2)
+        indentWithSpaces(4)
```
apancorb, line 27: Should we choose only one, either indent with tabs or indent with spaces? How will it choose how to do the indentation in this case?

apancorb, line 27: Maybe we could also add `trimTrailingWhitespace()` and `endWithNewline()`. Wdyt?

apancorb (dismissed):
Looks good to me :) Added some minor comments for your consideration

rskj-core/build.gradle
```diff
@@ -1,10 +1,156 @@
...
+
+checkstyle {
+    toolVersion = '8.45'
+    configFile = file("$rootDir/config/checkstyle/checkstyle.xml")
+}
+
+spotless {
+    java {
+        googleJavaFormat()
+        indentWithTabs(2)
```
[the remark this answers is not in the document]
rmoreliovlabs: Yes, when you use `indentWithTabs(2)`, it means that each tab character is equivalent to 2 spaces. So, if you want to achieve a visual indentation that looks like 4 spaces, you would actually need to use 2 tabs

rskj-core/build.gradle
rmoreliovlabs: you're right! chose indentWithTabs(2) and removed indentWithSpaces(4). Visually, the result was the same with both

rmoreliovlabs: addressed, added both suggestions, thanks!

marcos-iov (approved):

Vovchyk (approved):

PR #2692, opened 2024-08-19
Create new settleReleaseRequest method to perform common actions when creating pegouts

julia-zack:
-Create a new settleReleaseRequest that perform all the needed actions whenever creating a pegout.
-Get rid of existing addToPegoutsWaitingForConfirmations method.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -21,6 +21,7 @@ import static co.rsk.peg.BridgeUtils.getRegularPegoutTxSize;
 import static co.rsk.peg.ReleaseTransactionBuilder.BTC_TX_VERSION_2;
 import static co.rsk.peg.bitcoin.UtxoUtils.extractOutpointValues;
 import static co.rsk.peg.pegin.RejectedPeginReason.INVALID_AMOUNT;
+import static java.util.Objects.isNull;
 import static org.ethereum.config.blockchain.upgrades.ConsensusRule.*;
 
 import co.rsk.bitcoinj.core.*;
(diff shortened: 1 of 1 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1145,24 +1123,63 @@ private void processPegoutRequests(Transaction rskTx) {
...
-            pegoutWaitingForConfirmations.add(generatedTransaction, rskExecutionBlock.getNumber(), pegoutCreationRskTxHash);
-            // For a short time period, there could be items in the pegout request queue that don't have the pegoutCreationRskTxHash
-            // (these are pegouts created right before the consensus rule activation, that weren't processed before its activation)
-            // We shouldn't generate the event for those pegouts
-            if (pegoutCreationRskTxHash != null) {
-                eventLogger.logReleaseBtcRequested(pegoutCreationRskTxHash.getBytes(), generatedTransaction, amount);
-            }
-        } else {
-            pegoutWaitingForConfirmations.add(generatedTransaction, rskExecutionBlock.getNumber());
+    private void settleReleaseRequest(PegoutsWaitingForConfirmations pegoutsWaitingForConfirmations, BtcTransaction pegoutTransaction, Keccak256 releaseCreationTxHash, Coin requestedAmount) {
```
marcos-iov, line 1126: I don't think we need to pass around the `pegoutsWaitingForConfirmations`, we can simply get it from the storageProvider right?

marcos-iov (commented):
Looking nice and clean. Well done!

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: yes, thats right, but the `getPegoutsWaitingForConfirmations` could throw an `IOException` that has to be handled all over the code. See for example in [processPegoutsInBatch](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java#L1205) and [processPegoutsIndividually](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java#L1166), that the map is an input of the method already handled in [processPegoutRequests](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java#L1120) method.

I thought it was better to do it this way to have less impact. But let me know what you think!

marcos-iov: Yeah, no big deal. Thanks

marcos-iov (approved):

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: thanks to you :)

PR #2697, opened 2024-08-23
Feature/initcode size limit using opcodes contract creation

fmacleal:
This PR introduces the validation of initcodesize during the contract creation via opcodes CREATE and CREATE2. Besides this validation, it also introduces the cost calculation according the initcode size. The bigger the initcode, the bigger the cost create this contract.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/cost/InitcodeCostCalculator.java
```diff
@@ -38,10 +38,10 @@ public static InitcodeCostCalculator getInstance() {
     }
 
     @Override
-    public long calculateCost(byte[] data, ActivationConfig.ForBlock activations) {
-        if ( activations.isActive(ConsensusRule.RSKIP438) ) {
-            return  INITCODE_WORD_COST  *  ( (long) Math.ceil( ( (double) getLength(data) ) / 32 ) );
+    public long calculateCost(int dataLength, ActivationConfig.ForBlock activations) {
+        if( activations.isActive(ConsensusRule.RSKIP438) ) {
```
nagarev, line 42:
```suggestion
        if ( activations.isActive(ConsensusRule.RSKIP438) ) {
```

rskj-core/src/test/java/org/ethereum/cost/InitcodeCostCalculatorTest.java
```diff
@@ -0,0 +1,56 @@
+package org.ethereum.cost;
```
nagarev, line 1: I think tabs spacing is not correct in this file

rskj-core/src/test/java/co/rsk/vm/VmDslTest.java
```diff
@@ -118,53 +125,189 @@ void testPush0() throws FileNotFoundException, DslProcessorException {
...
         Assertions.assertEquals(1, statusCallOld.length);
         Assertions.assertEquals(1, statusCallOld[0]);
 
         short oldGas = ByteBuffer.wrap(callTransactionReceiptOld.getGasUsed()).getShort();
-        Assertions.assertTrue(newGas < oldGas);
+        assertTrue(newGas < oldGas);
+    }
+
+    @Test
+    void testInitCodeSizeValidationSuccessWithoutInitcodeCostViaCreateOpcodeCREATE() throws FileNotFoundException, DslProcessorException {
```
nagarev, line 179: I personally try to match the dsl `.txt` files with the test name to avoid misunderstandings. I think it's clearer that way, but if it's too much work, forget about it.

rskj-core/src/test/java/co/rsk/vm/VmDslTest.java
```diff
@@ -47,28 +54,28 @@ void invokeRecursiveContractsUsing400Levels() throws FileNotFoundException, DslP
 
         Block block = world.getBlockByName("b02");
 
-        Assertions.assertNotNull(block);
+        assertNotNull(block);
```
rmoreliovlabs, line 57: I personally prefer keeping `Assertions.assertNotNull(<value>);` for consistency and clarity, as it clearly indicates the assertion library in use and helps avoid confusion, especially if multiple libraries are involved. What do you think about sticking with the explicit `Assertions` reference?

rskj-core/src/test/java/co/rsk/vm/VmDslTest.java
```diff
@@ -118,53 +125,189 @@ void testPush0() throws FileNotFoundException, DslProcessorException {
...
+    }
+
+    @Test
+    void testInitCodeSizeValidationSuccessWithoutInitcodeCostViaCreateOpcodeCREATE() throws FileNotFoundException, DslProcessorException {
+        TestSystemProperties rskip438Disabled = new TestSystemProperties(rawConfig ->
+                rawConfig.withValue("blockchain.config.hardforkActivationHeights.lovell700", ConfigValueFactory.fromAnyRef(-1))
+        );
+        DslParser parser = DslParser.fromResource("dsl/initcode_rskip438/create_opcode_test_without_initcode_cost_rskip_NOT_ACTIVE.txt");
+        World world = new World(rskip438Disabled);
+        WorldDslProcessor processor = new WorldDslProcessor(world);
```
rmoreliovlabs, line 185: nice job on this! To clean things up a bit and avoid repeating code, how about adding a helper method to handle the setup for `DslParser`, `World`, and `WorldDslProcessor`?

rskj-core/src/test/java/org/ethereum/vm/ProgramTest.java
```diff
@@ -34,7 +34,7 @@
 import static org.mockito.Mockito.*;
 
 
-public abstract class ProgramTest {
+public class ProgramTest {
```
rmoreliovlabs, line 37: why removing the abstract here?

rskj-core/src/test/java/co/rsk/vm/VmDslTest.java
fmacleal: Well, for me it makes more noise than help. Because basically all different Assertions library have the class Assertions, so I would sill need to go to definition and see which one is used no? Or is there a very different name recognisable only by the `Assertions.` ?

rskj-core/src/test/java/co/rsk/vm/VmDslTest.java
fmacleal: Ok, will do.

rskj-core/src/test/java/co/rsk/vm/VmDslTest.java
fmacleal: Ok, will do.

rskj-core/src/test/java/org/ethereum/vm/ProgramTest.java
fmacleal: I initially struggled  trying to run just this test. The way it was implemented, the `ProgramTest` could not be executed. It would allow you to execute one of the inheritors. And this abstraction created just to override the beforeEach seemed overkill to me. So I removed a test file and added one with logs and gas enabled by default and made the others override the `beforeEach` method.  It made more sense than have a test class with test code that can't be executed alone and rely others to initialise their objects. 🤷‍♂️ 

For me, it made more sense. This way if we want to add a test, we edit only the `ProgramTest` class and the others will execute this test since they inherit and override the method with the flags.

rmoreliovlabs: got it, that makes sense. Thanks for clarifying!

rskj-core/src/test/java/co/rsk/vm/VmDslTest.java
rmoreliovlabs: I see your point. My concern was mainly about clarity and avoiding confusion when multiple libraries are involved. I personally prefer using `Assertions.<method>` to ensure that clarity, but if you prefer to keep it as is, that’s totally fine!

rskj-core/src/main/java/org/ethereum/vm/program/Program.java
```diff
@@ -453,6 +481,10 @@ public void createContract2(DataWord value, DataWord memStart, DataWord memSize,
     }
 
     private void createContract( RskAddress senderAddress, byte[] nonce, DataWord value, DataWord memStart, DataWord memSize, RskAddress contractAddress, boolean isCreate2) {
+        // [1] FETCH THE CODE FROM THE MEMORY
+        byte[] programCode = memoryChunk(memStart.intValue(), memSize.intValue());
```
Vovchyk, line 485: if possible I'd try to make this check after the `isNotCovers()` one, where we were initially reading memory chunk (this reading op. can be quite heavy)

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1470,21 +1472,25 @@ protected void doCREATE(){
...
+    }
+
     protected void doCREATE2(){
         if (program.isStaticCall()) {
             throw Program.ExceptionHelper.modificationException(program);
         }
 
         if (computeGas){
-            long codeSize = stack.get(stack.size() - 3).longValueSafe();
+            DataWord codeSize = getCodeSize();
```
Vovchyk, line 1485: not sure it's beneficial to make such refactoring - I mean extracting `stack.get(stack.size() - 3)` into a method. As for me, it was more explicit which item of stack was being extracted. Also `codeSize.longValue()`, which is used in several places below does some internal calcs

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1440,17 +1441,18 @@ protected void doCREATE(){
...
             gasCost = GasCost.CREATE;
-            size = stack.get(stack.size() - 3);
-            sizeLong = Program.limitToMaxLong(size);
+            codeSize = getCodeSize();
+            sizeLong = Program.limitToMaxLong(codeSize);
             checkSizeArgument(sizeLong);
             newMemSize = memNeeded(stack.get(stack.size() - 2), sizeLong);
-            gasCost = GasCost.add(gasCost, calcMemGas(oldMemSize, newMemSize, 0));
+            gasCost = GasCost.add(gasCost, calcMemGas(oldMemSize, newMemSize, 0))
+                    + InitcodeCostCalculator.getInstance().calculateCost(codeSize.intValue(), program.getActivations());
```
Vovchyk, line 1455: do you think we could make use of the `GasCost.add()` helper for adding initcode cost?

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1470,21 +1472,25 @@ protected void doCREATE(){
...
                     GasCost.add(
                             GasCost.CREATE,
-                            calcMemGas(oldMemSize, memNeeded(stack.get(stack.size() - 2), codeSize), 0)
+                            calcMemGas(oldMemSize, memNeeded(stack.get(stack.size() - 2), codeSize.longValue()), 0)
                     ),
                     GasCost.SHA3_WORD,
-                    GasCost.add(codeSize, 31) / 32
-            );
+                    GasCost.add(codeSize.longValue(), 31) / 32
+            ) + InitcodeCostCalculator.getInstance().calculateCost( codeSize.intValue(), program.getActivations());
```
Vovchyk, line 1493: same here

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/vm/VM.java
fmacleal: Ok, will do.

rskj-core/src/main/java/org/ethereum/vm/VM.java
fmacleal: Well, in my opinion it worth to to the refactor from getCodeSize because it's more clear that you are getting the code size from the stack. Do you see a problem in re-use this logic?

The thing with the long value it's because I need the intValue in the call. So I would have to get this value again converted, if I get the object that can be used for both situations isn't better?

rskj-core/src/main/java/org/ethereum/vm/program/Program.java
Vovchyk: so there's a reason.. ok then

Vovchyk (approved):
lgtm 👍

Vovchyk (approved):

PR #2699, opened 2024-08-28
Create SVP spend tx hash unsigned storage entry

apancorb:
## Description
Just as we did with the [fund transaction unsigned hash](https://github.com/rsksmart/rskj/blob/feature/powpeg_validation_protocol/rskj-core/src/main/java/co/rsk/peg/BridgeStorageIndexKey.java#L26), we need to save the spend transaction hash to be able to identify it when trying to register it later.

## Motivation and Context

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeStorageIndexKey.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeStorageIndexKey.java
@@ -25,6 +25,7 @@ public enum BridgeStorageIndexKey {
 
     SVP_FUND_TX_HASH_UNSIGNED("svpFundTxHashUnsigned"),
     SVP_FUND_TX_HASH_SIGNED("svpFundTxHashSigned"),
+    SVP_SPEND_TX_HASH_UNSIGNED("svpSpendTxHashUnsigned"),
     ;
 
     private final String key;


rskj-core/src/main/java/co/rsk/peg/BridgeStorageIndexKey.java
```diff
@@ -25,6 +25,7 @@ public enum BridgeStorageIndexKey {
 
     SVP_FUND_TX_HASH_UNSIGNED("svpFundTxHashUnsigned"),
     SVP_FUND_TX_HASH_SIGNED("svpFundTxHashSigned"),
+    SVP_SPEND_TX_HASH_UNSIGNED("svpSpendTxHashUnsigned"),
```
apancorb, line 28: Should we put the same name as the one that appears in the [RSKIP](https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP419.md?plain=1#L222)?

nathanieliov: I prefer how you did it to be consistent with the naming convention of the rest of the entries.

nathanieliov (approved):

rskj-core/src/main/java/co/rsk/peg/BridgeStorageIndexKey.java
marcos-iov: That's a good question, if we keep this name we need to update the RSKIP. Let's check with @julia-zack when she's back

marcos-iov (approved):

PR #2701, opened 2024-08-28
Adding Rootstock Integration Tests workflow

rmoreliovlabs:
# Adding of Rootstock Integration Tests Workflow

This GitHub Actions workflow is designed to automate the execution of Rootstock Integration Tests for the RSKj repository. It replaces the previous Jenkins-based setup with a streamlined and integrated GitHub Actions workflow. The primary goal is to ensure that integration tests are run manually or automatically for opened pull requests against the `master` branch or any `*-rc` branches, as well as for direct pushes to these branches.

## Implementation Details
**Triggers**:

- **Pull Requests**: The workflow triggers on pull requests opened or reopened against any branch, specifically targeting `master` and `*-rc` branches.
- **Push Events:** The workflow triggers on pushes to `master` and `*-rc` branches.
- **Manual Trigger**: The workflow can be manually triggered using the `workflow_dispatch` event, which allows for additional customization via inputs.

### How to Trigger the Workflow
**Automatically**:
- **Pull Requests**: The workflow will automatically trigger when a pull request is opened or reopened against the `master` or `*-rc `branches.
- **Pushes**: The workflow will automatically trigger on pushes to the `master` and `*-rc `branches.

**Manually**:

**Workflow Dispatch**: You can manually trigger the workflow using the **workflow_dispatch** event. To do this, follow these steps:
- Navigate to the **Actions** tab in your GitHub repository.
- Select the **Rootstock Integration Tests** workflow from the list of workflows.
- Click on the `Run workflow` button.
- Choose the branch from which you want to run the workflow using the "**Use workflow from branch**" dropdown menu. This allows you to select the branch (_**e.g., feature-branch**_) that will be used for the `RSKJ_BRANCH` value.
- Enter the desired input parameters (if any) or leave them as their default values.

**Inputs for Manual Trigger:**

- **`rit-branch`**: The branch for Rootstock Integration Tests.
              **Default**: main
- **`powpeg-branch`**: The branch for PowPeg Node.
              **Default**: master

To manually trigger the workflow using `workflow_dispatch`, **the workflow file must be present in the master branch**. That's we can't manually test this workflow until it's merged into master. For more info, check [here](https://github.com/orgs/community/discussions/25219#discussioncomment-3246909) and [here](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow#running-a-workflow-on-github).

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/.github/workflows/rit.yml
+++ b/.github/workflows/rit.yml
@@ -6,9 +6,7 @@ on:
     branches:
       - "**"
   push:
-    branches:
-      - "master"
-      - "*-rc"
+    branches: ["master", "*-rc"]
   workflow_dispatch:
     inputs:
       rit-branch:
@@ -50,14 +48,14 @@ jobs:
 
       - name: Checkout rskj Repository
         if: github.event_name == 'pull_request'
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4.1.7
         with:
           repository: rsksmart/rskj
           ref: ${{ env.RSKJ_BRANCH }}
           fetch-depth: 1
 
       - name: Checkout Rootstock Integration Tests Repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v4.1.7
         with:
           repository: rsksmart/rootstock-integration-tests
           ref: ${{ env.RIT_BRANCH }}
@@ -68,6 +66,3 @@ jobs:
         with:
           rskj-branch: ${{ env.RSKJ_BRANCH }}
           powpeg-node-branch: ${{ env.POWPEG_BRANCH }}
-
-      - name: Notify on Status
-        run: echo "Rootstock Integration Tests pipeline triggered"
--- a/.github/workflows/rit.yml
+++ b/.github/workflows/rit.yml
@@ -3,6 +3,9 @@ name: Rootstock Integration Tests
 on:
   push:
     branches: ["master", "*-rc"]
+  pull_request:
+    types: [ opened, synchronize, reopened ]
+    branches: ["master", "*-rc"]
   workflow_dispatch:
     inputs:
       rit-branch:
@@ -32,6 +35,8 @@ jobs:
             RSKJ_BRANCH="${{ github.ref }}"
           elif [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
             RSKJ_BRANCH="${{ github.ref_name }}"
+          elif [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            RSKJ_BRANCH="${{ github.head_ref }}"
           fi
 
           echo "RSKJ_BRANCH=$RSKJ_BRANCH" >> $GITHUB_ENV
--- a/.github/workflows/rit.yml
+++ b/.github/workflows/rit.yml
@@ -28,8 +28,8 @@ jobs:
         run: |
           # Default values
           RSKJ_BRANCH="master"
-          RIT_BRANCH="${{ github.event.inputs.rit-branch || github.event.inputs['rit-branch'].default }}"
-          POWPEG_BRANCH="${{ github.event.inputs.powpeg-branch || github.event.inputs['powpeg-branch'].default }}"
+          RIT_BRANCH="${{ github.event.inputs.rit-branch || 'main' }}"
+          POWPEG_BRANCH="${{ github.event.inputs.powpeg-branch || 'master' }}"
 
           if [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" == "refs/heads/*-rc" ]]; then
             RSKJ_BRANCH="${{ github.ref }}"
--- a/.github/workflows/rit.yml
+++ b/.github/workflows/rit.yml
@@ -1,10 +1,6 @@
 name: Rootstock Integration Tests
 
 on:
-  pull_request:
-    types: [opened, reopened]
-    branches:
-      - "**"
   push:
     branches: ["master", "*-rc"]
   workflow_dispatch:
@@ -32,12 +28,8 @@ jobs:
           RIT_BRANCH="${{ github.event.inputs.rit-branch || 'main' }}"
           POWPEG_BRANCH="${{ github.event.inputs.powpeg-branch || 'master' }}"
 
-          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
-            RSKJ_BRANCH="${{ github.event.pull_request.head.ref }}"
-          elif [[ "${{ github.event_name }}" == "push" ]]; then
-            if [[ "${{ github.ref }}" == refs/heads/*-rc ]]; then
-              RSKJ_BRANCH="${{ github.ref }}"
-            fi
+          if [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" == "refs/heads/*-rc" ]]; then
+            RSKJ_BRANCH="${{ github.ref }}"
           elif [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
             RSKJ_BRANCH="${{ github.ref_name }}"
           fi
@@ -46,23 +38,9 @@ jobs:
           echo "RIT_BRANCH=$RIT_BRANCH" >> $GITHUB_ENV
           echo "POWPEG_BRANCH=$POWPEG_BRANCH" >> $GITHUB_ENV
 
-      - name: Checkout rskj Repository
-        if: github.event_name == 'pull_request'
-        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 #v4.1.7
-        with:
-          repository: rsksmart/rskj
-          ref: ${{ env.RSKJ_BRANCH }}
-          fetch-depth: 1
-
-      - name: Checkout Rootstock Integration Tests Repository
-        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 #v4.1.7
-        with:
-          repository: rsksmart/rootstock-integration-tests
-          ref: ${{ env.RIT_BRANCH }}
-          fetch-depth: 0
-
       - name: Run Rootstock Integration Tests
         uses: rsksmart/rootstock-integration-tests@497172fd38dcfaf48c77f9bb1eeb6617eef5eed6 #v1
         with:
           rskj-branch: ${{ env.RSKJ_BRANCH }}
           powpeg-node-branch: ${{ env.POWPEG_BRANCH }}
+          rit-branch: ${{ env.RIT_BRANCH }}
--- a/.github/workflows/rit.yml
+++ b/.github/workflows/rit.yml
@@ -1,0 +1,73 @@
+name: Rootstock Integration Tests
+
+on:
+  pull_request:
+    types: [opened, reopened]
+    branches:
+      - "**"
+  push:
+    branches:
+      - "master"
+      - "*-rc"
+  workflow_dispatch:
+    inputs:
+      rit-branch:
+        description: 'Branch for Rootstock Integration Tests'
+        required: false
+        default: 'main'
+      powpeg-branch:
+        description: 'Branch for PowPeg Node'
+        required: false
+        default: 'master'
+
+jobs:
+  rootstock-integration-tests:
+    name: Rootstock Integration Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Set Branch Variables
+        id: set-branch-variables
+        run: |
+          # Default values
+          RSKJ_BRANCH="master"
+          RIT_BRANCH="${{ github.event.inputs.rit-branch || 'main' }}"
+          POWPEG_BRANCH="${{ github.event.inputs.powpeg-branch || 'master' }}"
+
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            RSKJ_BRANCH="${{ github.event.pull_request.head.ref }}"
+          elif [[ "${{ github.event_name }}" == "push" ]]; then
+            if [[ "${{ github.ref }}" == refs/heads/*-rc ]]; then
+              RSKJ_BRANCH="${{ github.ref }}"
+            fi
+          elif [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            RSKJ_BRANCH="${{ github.ref_name }}"
+          fi
+
+          echo "RSKJ_BRANCH=$RSKJ_BRANCH" >> $GITHUB_ENV
+          echo "RIT_BRANCH=$RIT_BRANCH" >> $GITHUB_ENV
+          echo "POWPEG_BRANCH=$POWPEG_BRANCH" >> $GITHUB_ENV
+
+      - name: Checkout rskj Repository
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@v3
+        with:
+          repository: rsksmart/rskj
+          ref: ${{ env.RSKJ_BRANCH }}
+          fetch-depth: 1
+
+      - name: Checkout Rootstock Integration Tests Repository
+        uses: actions/checkout@v4
+        with:
+          repository: rsksmart/rootstock-integration-tests
+          ref: ${{ env.RIT_BRANCH }}
+          fetch-depth: 0
+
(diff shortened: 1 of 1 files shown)


rmoreliovlabs:
pipeline:run

rmoreliovlabs:
pipeline:run

bcodesido (changes requested):

.github/workflows/rit.yml
```diff
@@ -0,0 +1,101 @@
...
+  #     - "*-rc"
+  pull_request:
+    types: [opened, reopened, synchronize]
+    branches:
+      - "**"
+  issue_comment:
+    types:
+      - created
+      - edited
+permissions:
```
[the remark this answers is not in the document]
rmoreliovlabs: thanks for the clarification! removed it, as it is not necessary in the workflow logic

.github/workflows/rit.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Checkout rskj Repository
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@v3
+        with:
+          repository: rsksmart/rskj
+          ref: ${{ env.RSKJ_BRANCH }}
+          fetch-depth: 1
+
+      - name: Checkout Rootstock Integration Tests Repository
+        uses: actions/checkout@v4
```
bcodesido, line 60: Same as above.
Also, why we first use `checkout@v3` and then `checkout@v4`, is not possible to use the same version?

.github/workflows/rit.yml
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          repository: rsksmart/rootstock-integration-tests
+          ref: ${{ env.RIT_BRANCH }}
+          fetch-depth: 0
+
+      - name: Run Rootstock Integration Tests
+        uses: rsksmart/rootstock-integration-tests@v1
```
bcodesido, line 67: Same comment as above about pinning.
I understand that can be an issue given that the action is publishing always under `v1` tag, but also would be nice to have it pinned.

bcodesido (changes requested):

.github/workflows/rit.yml
```diff
@@ -0,0 +1,73 @@
...
+            RSKJ_BRANCH="${{ github.ref_name }}"
+          fi
+
+          echo "RSKJ_BRANCH=$RSKJ_BRANCH" >> $GITHUB_ENV
+          echo "RIT_BRANCH=$RIT_BRANCH" >> $GITHUB_ENV
+          echo "POWPEG_BRANCH=$POWPEG_BRANCH" >> $GITHUB_ENV
+
+      - name: Checkout rskj Repository
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@v3
```
diego-jeronymo, line 53: is there a reason why the actions/checkout@v3 should be used instead of actions/checkout@v4 ? if not I suggest to use the v4. thanks in advance.

diego-jeronymo (changes requested):

.github/workflows/rit.yml
```diff
@@ -0,0 +1,73 @@
...
+          fetch-depth: 0
+
+      - name: Run Rootstock Integration Tests
+        uses: rsksmart/rootstock-integration-tests@v1
+        with:
+          rskj-branch: ${{ env.RSKJ_BRANCH }}
+          powpeg-node-branch: ${{ env.POWPEG_BRANCH }}
+
+      - name: Notify on Status
+        run: echo "Rootstock Integration Tests pipeline triggered"
```
fmacleal, line 73: question:

Is it really necessary? Don't the RIT already print a lot of stuff already?

fmacleal (approved):
Good job!

I have a single comment regarding the need of a step, but doesn't prevent the approval.

.github/workflows/rit.yml
rmoreliovlabs: you're right! that's kinda redundant. Removed it

.github/workflows/rit.yml
rmoreliovlabs: Addressed, added the version v4.1.7

.github/workflows/rit.yml
```diff
@@ -0,0 +1,73 @@
+name: Rootstock Integration Tests
+
+on:
+  pull_request:
+    types: [opened, reopened]
+    branches:
+      - "**"
+  push:
+    branches:
+      - "master"
```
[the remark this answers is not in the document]
rmoreliovlabs: solved!

.github/workflows/rit.yml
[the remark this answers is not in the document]
rmoreliovlabs: good suggestion! specified the version `@v4.1.7`

.github/workflows/rit.yml
rmoreliovlabs: addressed, specified the version `@v4.1.7`

.github/workflows/rit.yml
rmoreliovlabs: addressed, specified the version `@v1.0.0`

rmoreliovlabs:
Sorry for the frequent closing and reopening of the PR! I did this to trigger the workflow as part of the testing process. I appreciate your patience!

.github/workflows/rit.yml
fmacleal: I generated a tag with the [full semantic version](https://github.com/rsksmart/rootstock-integration-tests/releases/tag/v1.0.0), if you add it it will work now.

fmacleal (dismissed):
LGTM

diego-jeronymo (approved):
LGTM

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
...
+            RSKJ_BRANCH="${{ github.ref_name }}"
+          fi
+
+          echo "RSKJ_BRANCH=$RSKJ_BRANCH" >> $GITHUB_ENV
+          echo "RIT_BRANCH=$RIT_BRANCH" >> $GITHUB_ENV
+          echo "POWPEG_BRANCH=$POWPEG_BRANCH" >> $GITHUB_ENV
+
+      - name: Checkout rskj Repository
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@v4.1.7
```
bcodesido, line 51: Please pin the hash of the version.
That should looks like:

`uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 #4.1.7`

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
...
+        with:
+          repository: rsksmart/rskj
+          ref: ${{ env.RSKJ_BRANCH }}
+          fetch-depth: 1
+
+      - name: Checkout Rootstock Integration Tests Repository
+        uses: actions/checkout@v4.1.7
```
bcodesido, line 58: Same as above.

.github/workflows/rit.yml
bcodesido: I would add the hash instead of the semantic version.

bcodesido (changes requested):

.github/workflows/rit.yml
rmoreliovlabs: thanks for clarifying! addressed

.github/workflows/rit.yml
rmoreliovlabs: addressed

.github/workflows/rit.yml
rmoreliovlabs: I’ve updated the workflow with the commit hash `497172fd38dcfaf48c77f9bb1eeb6617eef5eed6`. You can review these changes in the PowPeg repository, where the workflow has been successfully tested with the hashes. For reference, here is the related job run: [PowPeg Workflow Run](https://github.com/rsksmart/powpeg-node/actions/runs/10692919902)

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
+name: Rootstock Integration Tests
+
+on:
+  pull_request:
+    types: [opened, reopened]
+    branches:
+      - "**"
```
bcodesido, line 7: it's really necessary to run the tests against each PR created independently from the destination?

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
...
+          repository: rsksmart/rskj
+          ref: ${{ env.RSKJ_BRANCH }}
+          fetch-depth: 1
+
+      - name: Checkout Rootstock Integration Tests Repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 #v4.1.7
+        with:
+          repository: rsksmart/rootstock-integration-tests
+          ref: ${{ env.RIT_BRANCH }}
+          fetch-depth: 0
```
fmacleal, line 62: suggestion:

I don't think this part is necessary, it won't impact which RIT branch will be used to execute the tests at the end. Because the container-action clones the repo and does the checkout again, [as you can see here](https://github.com/rsksmart/rootstock-integration-tests/blob/main/container-action/entrypoint.sh#L40), on the branch passed as parameter. Which you are missing to pass it on the next step. I suggest remove this step and pass the `RIT_BRANCH` as input to the action.

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Run Rootstock Integration Tests
+        uses: rsksmart/rootstock-integration-tests@497172fd38dcfaf48c77f9bb1eeb6617eef5eed6 #v1
+        with:
+          rskj-branch: ${{ env.RSKJ_BRANCH }}
+          powpeg-node-branch: ${{ env.POWPEG_BRANCH }}
```
fmacleal, line 68: suggestion:

As commented above, you need to pass the RIT branch here. The container-action is ready to receive it, as you can see [here](https://github.com/rsksmart/rootstock-integration-tests/blob/main/action.yml#L14). 

```suggestion
          powpeg-node-branch: ${{ env.POWPEG_BRANCH }}
          rit-branch: ${{ env. RIT_BRANCH }}

```

fmacleal (changes requested):
Few minors suggested.

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
...
+          ref: ${{ env.RSKJ_BRANCH }}
+          fetch-depth: 1
+
+      - name: Checkout Rootstock Integration Tests Repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 #v4.1.7
+        with:
+          repository: rsksmart/rootstock-integration-tests
+          ref: ${{ env.RIT_BRANCH }}
+          fetch-depth: 0
+
```
fmacleal, line 63: suggestion:

I don't think this part is necessary, it won't impact which RIT branch will be used to execute the tests at the end. Because the container-action clones the repo and does the checkout again, [as you can see here](https://github.com/rsksmart/rootstock-integration-tests/blob/main/container-action/entrypoint.sh#L40), on the branch passed as parameter. Which you are missing to pass it on the next step. I suggest remove this step and pass the `RIT_BRANCH` as input to the action.

```suggestion
```

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
...
+          RSKJ_BRANCH="master"
+          RIT_BRANCH="${{ github.event.inputs.rit-branch || 'main' }}"
+          POWPEG_BRANCH="${{ github.event.inputs.powpeg-branch || 'master' }}"
+
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            RSKJ_BRANCH="${{ github.event.pull_request.head.ref }}"
+          elif [[ "${{ github.event_name }}" == "push" ]]; then
+            if [[ "${{ github.ref }}" == refs/heads/*-rc ]]; then
+              RSKJ_BRANCH="${{ github.ref }}"
+            fi
```
fmacleal, line 40: question:

This condition `"${{ github.ref }}" == refs/heads/*-rc` can be added as an AND in the previous if couldn't it? If yes, it will make it simpler, I suggest to add it there.

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
...
+          echo "RIT_BRANCH=$RIT_BRANCH" >> $GITHUB_ENV
+          echo "POWPEG_BRANCH=$POWPEG_BRANCH" >> $GITHUB_ENV
+
+      - name: Checkout rskj Repository
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 #v4.1.7
+        with:
+          repository: rsksmart/rskj
+          ref: ${{ env.RSKJ_BRANCH }}
+          fetch-depth: 1
```
fmacleal, line 55: suggestion:

I also think that this part is not necessary, same behaviour from `RIT_BRANCH`, it won't impact which RSKJ branch will be used to execute the tests at the end. Because the container-action clones the repo and does the checkout again, [as you can see here](https://github.com/rsksmart/rootstock-integration-tests/blob/main/container-action/entrypoint.sh#L18), on the branch passed as parameter. Which you are already passing it. So we are good on this.

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
...
+
```
fmacleal, line 56: suggestion:

I also think that this part is not necessary, same behaviour from `RIT_BRANCH`, it won't impact which RSKJ branch will be used to execute the tests at the end. Because the container-action clones the repo and does the checkout again, [as you can see here](https://github.com/rsksmart/rootstock-integration-tests/blob/main/container-action/entrypoint.sh#L18), on the branch passed as parameter. Which you are already passing it. So we are good on this.

```suggestion
```

.github/workflows/rit.yml
rmoreliovlabs: Addressed. It’s not strictly necessary to run tests on every PR separately from the destination branch. I’ve adjusted the workflow to focus on push events for the `master` branch and branches ending with `-rc`

.github/workflows/rit.yml
rmoreliovlabs: Solved, added rit-branch

.github/workflows/rit.yml
rmoreliovlabs: Solved

.github/workflows/rit.yml
rmoreliovlabs: Addressed!

.github/workflows/rit.yml
rmoreliovlabs: Good catch! solved

.github/workflows/rit.yml
```diff
@@ -0,0 +1,46 @@
+name: Rootstock Integration Tests
+
+on:
```
fmacleal, line 3: question:

I think the issue with the `pull_request` trigger was the source branch that was set to all with `**`. Don't we want to run it at least once for a PR opened inside our organization? Maybe add an if on the job `rootstock-integration-tests` with something like:

`    if: github.event.pull_request.head.repo.full_name == 'rsksmart/rskj'` 

Buuut, I don't see it as a big deal, because we can always run manually and we were already doing that triggering with a comment. So, to me, the way that it's set now it's very good and more than enough. 😃

fmacleal (approved):
Good job! I just have a question regarding your last change, but it doesn't prevent the approval. I think this way it will fill the initial requirements we have for this pipeline, we can always change if needed. :)

.github/workflows/rit.yml
Vovchyk: @rmoreliovlabs , I believe @bcodesido meant that it can be optimised a bit if that's being triggered automatically only for master / *-rc. For PRs created against other branches (aka feature branches), manual / on-demand triggering should be sufficient. This should make the trick, shouldn't it?:
```
  pull_request:
    types: [opened, reopened]
    branches:
      - "master"
      - "*-rc"
```

fmacleal (approved):

Vovchyk (dismissed):

fmacleal (dismissed):

Vovchyk:
@bcodesido , could you please check this final version? thanks

Vovchyk (approved):

bcodesido (approved):
LGTM

PR #2702, opened 2024-08-29
Create set and save SvpFundTransactionHashUnsigned methods

apancorb:
## Description
To save the spend transaction (unsigned) hash in the repository, we need to create new methods in BridgeStorageProvider: one for setting the variable and another one for saving it in the repository.

## Motivation and Context

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
@@ -80,6 +80,8 @@ public class BridgeStorageProvider {
     private boolean isSvpFundTxHashUnsignedSet = false;
     private Sha256Hash svpFundTxHashSigned;
     private boolean isSvpFundTxHashSignedSet = false;
+    private Sha256Hash svpSpendTxHashUnsigned;
+    private boolean isSvpSpendTxHashUnsignedSet = false;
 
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -594,6 +596,23 @@ private void saveSvpFundTxHashSigned() {
...
 
+    public void setSvpSpendTxHashUnsigned(Sha256Hash hash) {
+        this.svpSpendTxHashUnsigned = hash;
+        this.isSvpSpendTxHashUnsignedSet = true;
+    }
+
+    private void saveSvpSpendTxHashUnsigned() {
+        if (!activations.isActive(RSKIP419) || !isSvpSpendTxHashUnsignedSet) {
+            return;
+        }
```
apancorb, line 607: I am not sure about the purpose of the `isSvpSpendTxHashUnsignedSet` variable. From my understanding, we are never setting this variable back to false once we store it. So in a subsequent iteration this condition will always pass with whatever hash that is set. Is this expected?

Also shouldn't the if condition be using an AND operator instead of an OR operator? 

Maybe I am missing something here.

marcos-iov (approved):

PR #2714, opened 2024-09-09
Register fundtx change

julia-zack:
**Background**
Since the fund transaction will be treated as a pegout, its change will eventually be attempted to be registered through existing `registerBtcTransaction` method.

When that happen, and _after recognizing and processing it as a pegout_, we have to save its hash (now signed) in the corresponding storage entry, that meaning we have to add some extra checks to recognize this special transaction.

So, if validation process is ongoing, we will check if the received transaction is the fund transaction.

_We say that the validation process is ongoing if: proposedFederation exists, and validation period is not ended, i.e., currentRskBlock < proposedFederationCreationBlockNumber + svpPeriodDuration_

.
**IN THIS PR**
This pr creates a new private method, `processSvpFundTransactionHashSigned`, that:
- first checks the fund transaction unsigned hash is saved in the storage (if not, then do nothing)
- if it is, checks if the received btc tx without signatures matches the unsigned hash of the fundTx saved in svpFundTransactionUnsignedHash entry.
- if it does, the received tx hash is saved in the new `svpFundTransactionSignedHash` storage entry, and the value saved in svpFundTransactionUnsignedHash storage entry is removed


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -421,6 +424,21 @@ public void registerBtcTransaction(
         }
     }
 
+    private void processSvpFundTransactionHashSigned(BtcTransaction transaction) {
+        Optional<Sha256Hash> fundTransactionHashUnsignedOpt = provider.getSvpFundTxHashUnsigned();
+        if (!fundTransactionHashUnsignedOpt.isPresent()) {
+            return;
+        }
+        Sha256Hash fundTransactionHashUnsigned = fundTransactionHashUnsignedOpt.get();
```
apancorb, line 432: a bit less verbose could look something like this:

```suggestion
     Sha256Hash fundTransactionHashUnsigned = provider.getSvpFundTxHashUnsigned().orElse(null);
     if (fundTransactionHashUnsigned == null) {
        return;
     }
```
also you might consider using the `ifPresent` method which would look like:
```suggestion
        provider.getSvpFundTxHashUnsigned().ifPresent(fundTransactionHashUnsigned -> { ... });
```

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -421,6 +424,21 @@ public void registerBtcTransaction(
         }
     }
 
+    private void processSvpFundTransactionHashSigned(BtcTransaction transaction) {
```
apancorb, line 427: We could add log statements to trace the flow and help with debugging.

apancorb, line 427: Since this method is only called during an ongoing SVP period, can we safely call this `fundTx` instead of `transaction`?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -405,6 +405,9 @@ public void registerBtcTransaction(
                 case PEGOUT_OR_MIGRATION:
```
apancorb, line 405: I am wandering, the Bridge will consider the `fundTx` as a pegout since the `fundTx` will contain a change output to the current federation? Or how does it determine that this is a pegout? Not sure if it makes sense :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -421,6 +424,21 @@ public void registerBtcTransaction(
...
+        if (!fundTransactionHashUnsignedOpt.isPresent()) {
+            return;
+        }
+        Sha256Hash fundTransactionHashUnsigned = fundTransactionHashUnsignedOpt.get();
+
+        BtcTransaction transactionCopy = new BtcTransaction(networkParameters, transaction.bitcoinSerialize()); // this is needed to not remove signatures from the actual tx
+        BitcoinUtils.removeSignaturesFromTransactionWithP2shMultiSigInputs(transactionCopy);
+        if (transactionCopy.getHash().equals(fundTransactionHashUnsigned)) {
+            provider.setSvpFundTxHashSigned(transaction.getHash());
+            provider.setSvpFundTxHashUnsigned(null);
```
apancorb, line 438: Should we also call the `save` method after this? The way I understand this is since the value only live in memory, once the rsk transaction execution ends we will loose all the values in memory. So how do we persist this data before it ends? Maybe I am understanding this wrong :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: not really, because this method receives a btc transaction and then checks **if** it is the fund transaction.
Maybe i should change the method's name

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: because the sigHash was saved in the pegout index when it was created, thats how we recognize pegouts now :)
See [here](https://github.com/rsksmart/rskj/blob/56119ab694f4068c25d18125aaf2bd1f31b17f7a/rskj-core/src/main/java/co/rsk/peg/PegUtils.java#L50).

[This](https://github.com/rsksmart/rskj/blob/56119ab694f4068c25d18125aaf2bd1f31b17f7a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java#L1209) other method takes care of saving all the information when requesting a release, and [we call it](https://github.com/rsksmart/rskj/blob/56119ab694f4068c25d18125aaf2bd1f31b17f7a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java#L1005) after creating the fund transaction

Let me know if thats clear enough :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: mmm, i get your point, but im not sure jaja. I like it better the way it is now

julia-zack: i refactored it, take a look. Thanks :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
apancorb: Thanks, clear enough :)

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -803,6 +807,196 @@ private void assertEventWasEmittedWithExpectedData(byte[] expectedData) {
...
+
+            when(federationSupport.getProposedFederation()).thenReturn(Optional.empty());
+
+            // act
+            int activeFederationUtxosSizeBeforeRegisteringTx = federationSupport.getActiveFederationBtcUTXOs().size();
+            bridgeSupport.registerBtcTransaction(rskTx, svpFundTransaction.bitcoinSerialize(), btcBlockWithPmtHeight, pmtWithTransactions.bitcoinSerialize());
+            bridgeStorageProvider.save();
+
+            // assert
+            assertNewUtxoWasSaved(activeFederationUtxosSizeBeforeRegisteringTx);
```
apancorb, line 826: What would happen if the `fundTx` would not contain a change output to the current federation? I am assuming nothing would happen and normal flow should continue, but just double checking. 

Maybe what I am trying to say is, does it really matter that we check if a new UTXO was registered due to the `fundTx` being registered?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -803,6 +807,196 @@ private void assertEventWasEmittedWithExpectedData(byte[] expectedData) {
...
+            Sha256Hash parentTxHash = BitcoinTestUtils.createHash(1);
+            addInput(svpFundTransaction, parentTxHash);
+            addOutputChange(svpFundTransaction);
+            savePegoutIndex(svpFundTransaction);
+            saveSvpFundTransactionHashUnsigned(svpFundTransaction.getHash());
+
+            return svpFundTransaction;
+        }
+
+        private BtcTransaction createAndSavePegout() {
```
apancorb, line 931: Out of curiosity, for these private utility methods, are we sure they dont already exist? Or something similar?

apancorb (approved):
LGTM, good job! Left two last comments for your consideration

jeremy-then (approved):
LGTM.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
julia-zack: is a very good question, im adding a test without the change output. Thanks!

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
julia-zack: I checked and didn't found anything!

julia-zack: All transactions creation are done manually

PR #2716, opened 2024-09-10
Add methods to create a real valid chain with stored blocks to avoid using mocks

julia-zack:
To avoid keep using the `mockChainOfStoredBlocks` method, two new ones were created to recreate a valid chain that has:
- a valid partial merkle tree with the wanted transactions
- the block with that merkle root stored at wanted height
- a new chainhead stored block at wanted height

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
+++ b/rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
@@ -1,17 +1,19 @@
 package co.rsk.peg;
 
-import co.rsk.bitcoinj.core.BtcBlock;
-import co.rsk.bitcoinj.core.Sha256Hash;
-import co.rsk.bitcoinj.core.StoredBlock;
+import co.rsk.bitcoinj.core.*;
 import co.rsk.bitcoinj.store.BlockStoreException;
 import co.rsk.db.MutableTrieCache;
 import co.rsk.db.MutableTrieImpl;
+import co.rsk.peg.bitcoin.BitcoinTestUtils;
 import co.rsk.trie.Trie;
 import org.bouncycastle.util.encoders.Hex;
 import org.ethereum.core.Repository;
 import org.ethereum.db.MutableRepository;
 
 import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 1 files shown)


rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
```diff
@@ -22,6 +24,53 @@ public static Repository createRepository() {
...
+
+        // first create a block that has the wanted partial merkle tree
+        BtcBlock btcBlockWithPmt = createBtcBlockWithPmt(partialMerkleTree, networkParameters);
+        // store it on the chain at wanted height
+        StoredBlock storedBtcBlockThatHasTransactions = new StoredBlock(btcBlockWithPmt, BigInteger.ONE, btcBlockWithPmtHeight);
+        btcBlockStoreWithCache.put(storedBtcBlockThatHasTransactions);
+        btcBlockStoreWithCache.setMainChainBlock(btcBlockWithPmtHeight, btcBlockWithPmt.getHash());
+
+        // create and store a new chainHead at wanted chain height
+        Sha256Hash randomTransactionHash = Sha256Hash.of(Hex.decode("aa"));
```
jeremy-then, line 50: This is not really random. I would suggest to either rename it or make it really random.

julia-zack: i named it _random_ just because it is not related with the values received as inputs

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
```diff
@@ -22,6 +24,53 @@ public static Repository createRepository() {
...
     }
 
+    public static PartialMerkleTree createValidPmtWithTransactions(List<Sha256Hash> hashesToAdd, NetworkParameters networkParameters) {
+        byte[] bits = new byte[1];
+        bits[0] = 0x3f;
+
+        return new PartialMerkleTree(networkParameters, bits, hashesToAdd, 1);
+    }
+
+    public static void recreateChainAtHeightWithStoredBlockWithPmtAtHeight(
```
apancorb, line 34: nit: maybe a bit more redable?
```suggestion
    public static void recreateChainWithStoredBlockAndPmtAtHeight(
```

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
```diff
@@ -22,6 +24,53 @@ public static Repository createRepository() {
         return new MutableRepository(new MutableTrieCache(new MutableTrieImpl(null, new Trie())));
     }
 
+    public static PartialMerkleTree createValidPmtWithTransactions(List<Sha256Hash> hashesToAdd, NetworkParameters networkParameters) {
+        byte[] bits = new byte[1];
+        bits[0] = 0x3f;
```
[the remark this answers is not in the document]
apancorb: Agree, maybe a constant for the class

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
apancorb: Maybe using `otherTxHash` would clarify things

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
```diff
@@ -22,6 +24,53 @@ public static Repository createRepository() {
...
+        int btcBlockWithPmtHeight,
+        NetworkParameters networkParameters
+    ) throws BlockStoreException {
+
+        // first create a block that has the wanted partial merkle tree
+        BtcBlock btcBlockWithPmt = createBtcBlockWithPmt(partialMerkleTree, networkParameters);
+        // store it on the chain at wanted height
+        StoredBlock storedBtcBlockThatHasTransactions = new StoredBlock(btcBlockWithPmt, BigInteger.ONE, btcBlockWithPmtHeight);
+        btcBlockStoreWithCache.put(storedBtcBlockThatHasTransactions);
+        btcBlockStoreWithCache.setMainChainBlock(btcBlockWithPmtHeight, btcBlockWithPmt.getHash());
```
apancorb, line 47: What is the main purpose of `setMainChainBlock`?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
```diff
@@ -1,17 +1,19 @@
 package co.rsk.peg;
 
-import co.rsk.bitcoinj.core.BtcBlock;
-import co.rsk.bitcoinj.core.Sha256Hash;
-import co.rsk.bitcoinj.core.StoredBlock;
+import co.rsk.bitcoinj.core.*;
```
apancorb, line 3: nit: Maybe just using what we need is more explicit imo

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
julia-zack: i get the point, i just wanted to be clear about the fact that the block we want to store will be at one height, and the chain at other (equal or greater than the block height). But you are right, i will modify the name to make it more readable. Thanks :)

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
julia-zack: for that block to be the main chain at that height. 
so [this](https://github.com/rsksmart/rskj/blob/recreate_valid_block_stored_chain/rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java#L77) happens without mocking :)

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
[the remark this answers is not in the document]
julia-zack: this was modified to be more realistic :)

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTestUtil.java
```diff
@@ -22,6 +24,55 @@ public static Repository createRepository() {
...
+    public static PartialMerkleTree createValidPmtForTransactions(List<Sha256Hash> hashesToAdd, NetworkParameters networkParameters) {
+        byte[] relevantNodesBits = new byte[(int)Math.ceil(hashesToAdd.size() / 8.0)];
+        for (int i = 0; i < hashesToAdd.size(); i++) {
+            Utils.setBitLE(relevantNodesBits, i);
+        }
+
+        return PartialMerkleTree.buildFromLeaves(networkParameters, relevantNodesBits, hashesToAdd);
+    }
+
+    public static void recreateChainFromPmt(
```
marcos-iov, line 36: Should this return something? I guess the idea now is that it modifies the received `btcBlockStoreWithCache` object right? 

But is there any reason not to simply create a new one and return it?

julia-zack: because to create it you need a lot of parameters (repository, bridgeConstants, bridgeStorageProvider, activations, etc. See [here](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/main/java/co/rsk/peg/RepositoryBtcBlockStoreWithCache.java#L65)).
This way, in the tests we set the btcBlockStoreFactory and a new btcBlockStoreWithCache instance from it, using the parameters from the test

apancorb (approved):

marcos-iov (approved):

PR #2718, opened 2024-09-11
Create saveSvpSpendTxWaitingForSignatures method

apancorb:
## Description
For the pegnatories to be able to proceed with the spend transaction signing, they need to have access to it. Just as they do with the pegouts.

* Storage key added for svp spend tx waiting for signatures
* save logic with unit tests
* minor refactor to use safeSaveFromRepository for all svp save methods

## Motivation and Context


rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -628,11 +626,27 @@ private void saveSvpSpendTxHashUnsigned() {
...
+            svpSpendTxHashUnsigned,
+            BridgeSerializationUtils::serializeSha256Hash);
+    }
+
+    private void saveSvpSpendTxWaitingForSignatures() {
+        if (!activations.isActive(RSKIP419)) {
+            return;
+        }
+
+        if (svpSpendTxWaitingForSignatures == null ||
```
apancorb, line 640: Not sure if we should check for null here, since we would need a default value when there is no svp ongoing right?

marcos-iov: Probably not, once the transaction is fully signed we want to clear this storage value by setting it to null. As we do with other values

julia-zack: yes, careful here.
When there is no spend tx we should set the value to `new AbstractMap.SimpleEntry<>(null, null);`, and save it in the storage.
That's why we were not using the `safeSaveToRepository`, because we want the null values to be saved. And using the serializers throw a npe :)

julia-zack: yes, marcos is right
When there is no spend tx we should set the value to new AbstractMap.SimpleEntry<>(null, null);, and save it in the storage.

In that case, you could do what was being done in the other methods:
```
        byte[] data = Optional.ofNullable(svpSpendTxWaitingForSignatures)
            .map(BridgeSerializationUtils::serializeRskTxWaitingForSignatures)
            .orElse(null);
```

apancorb: Makes sense I have updated to allow null value of the entry to be saved, as done in the other methods. 

> When there is no spend tx we should set the value to new AbstractMap.SimpleEntry<>(null, null), and save it in the storage.

This will not be allowed since it would throw a NPE. If the reference is not null it will try to serialize it, and since it contains null references inside it will throw a NPE. This is the reason why we cannot allow to save an "empty entry" 

Let me know if this looks good

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -628,11 +626,29 @@ private void saveSvpSpendTxHashUnsigned() {
...
+        if (!activations.isActive(RSKIP419)) {
+            return;
+        }
+
+        boolean hasNullKeyOrValue = svpSpendTxWaitingForSignatures != null &&
+            (svpSpendTxWaitingForSignatures.getKey() == null || 
+             svpSpendTxWaitingForSignatures.getValue() == null);
+        if (hasNullKeyOrValue) {
+            throw new IllegalArgumentException(
+                String.format("Invalid svpSpendTxWaitingForSigantures: %s", svpSpendTxWaitingForSignatures));
```
julia-zack, line 645:
```suggestion
                String.format("Invalid svpSpendTxWaitingForSignatures: %s", svpSpendTxWaitingForSignatures));
```
just a little typo

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -834,6 +838,90 @@ void getSvpSpendTxHashUnsigned_whenHashIsCached_shouldReturnTheCachedHash() {
         }
     }
 
+    @Nested
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+    @Tag("save, set and get svp spend transaction waiting for signatures tests")
+    class SvpSpendTxWaitingForSignaturesTests {
+        private final Keccak256 rskTxHash = PegTestUtils.createHash3(1);
+        private final BtcTransaction svpSpendTx = new BtcTransaction(mainnetBtcParams);
+        private final String fieldName = "svpSpendTxWaitingForSignatures";
```
julia-zack, line 847: no big deal, but i found fieldName not very specific. Maybe `svpSpendTxWaitingForSignaturesField`? A bit redundant but more descriptive. Wdyt?

apancorb: Sounds good, just updated

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -834,6 +838,90 @@ void getSvpSpendTxHashUnsigned_whenHashIsCached_shouldReturnTheCachedHash() {
...
+            bridgeStorageProvider.save();
+
+            // Assert
+            byte[] actualSvpSpendTxWaitingForSignatures =
+                repository.getStorageBytes(bridgeAddress, SVP_SPEND_TX_WAITING_FOR_SIGNATURES.getKey());
+            assertNull(actualSvpSpendTxWaitingForSignatures);
+        }
+
+        @Test
+        void saveSvpSpendTxWaitingForSignatures_postLovell700AndEmptyEntry_shouldThrowIllegalArgumentException() {
```
julia-zack, line 873: wdyt about adding one test with just a null key and another one with just a null value?

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -834,6 +838,90 @@ void getSvpSpendTxHashUnsigned_whenHashIsCached_shouldReturnTheCachedHash() {
...
+            assertThrows(IllegalArgumentException.class, () -> bridgeStorageProvider.save());
+
+            // Assert
+            byte[] actualSvpSpendTxWaitingForSignatures =
+                repository.getStorageBytes(bridgeAddress, SVP_SPEND_TX_WAITING_FOR_SIGNATURES.getKey());
+            assertNull(actualSvpSpendTxWaitingForSignatures);
+        }
+
+        @Test
+        void saveSvpSpendTxWaitingForSignatures_postLovell700_shouldSaveInStorage() {
```
julia-zack, line 889:
```suggestion
        void saveSvpSpendTxWaitingForSignatures_postLovell700_withNonNullKeyAndValue_shouldSaveInStorage() {
```

apancorb: I think this implied

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
apancorb: Sure, just updated!

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -834,6 +838,122 @@ void getSvpSpendTxHashUnsigned_whenHashIsCached_shouldReturnTheCachedHash() {
...
+                repository.getStorageBytes(bridgeAddress, SVP_SPEND_TX_WAITING_FOR_SIGNATURES.getKey());
+            assertNull(actualSvpSpendTxWaitingForSignatures);
+        }
+
+        @Test
+        void saveSvpSpendTxWaitingForSignatures_postLovell700AndNullKeyInEntry_shouldThrowIllegalArgumentException() {
+            // Arrange
+            Map.Entry<Keccak256, BtcTransaction> svpSpendTxWaitingForSignatures =
+                new AbstractMap.SimpleEntry<>(null, svpSpendTx);
+            TestUtils.setInternalState(bridgeStorageProvider, svpSpendTxWaitingForSignaturesField, svpSpendTxWaitingForSignatures);
```
marcos-iov, line 893: Any reason to use this instead of `set` method directly?

apancorb: Created a setter method to update the entry.

jeremy-then (approved):
LGTM.

PR #2747, opened 2024-09-23
Register fundtx change

julia-zack:
Background
Since the fund transaction will be treated as a pegout, its change will eventually be attempted to be registered through existing `registerBtcTransaction` method.

When that happen, and after recognizing and processing it as a pegout, we have to update the svp fund transaction values (save the signed one in the corresponding storage entry, remove the unsigned hash from the corresponding storage entry), that meaning we have to add some extra checks to recognize this special transaction.

So, if validation process is ongoing, we will check if the received transaction is the fund transaction.

We say that the validation process is ongoing if proposedFederation exists and validation period is not ended, i.e., currentRskBlock < proposedFederationCreationBlockNumber + svpPeriodDuration

.
IN THIS PR
This pr creates a new private method, `updateSvpFundTransactionValuesIfPossible`, that:

-first checks the fund transaction unsigned hash is saved in the storage (if not, then do nothing)
-if it is, checks if the received btc tx without signatures matches the unsigned hash of the fundTx saved in `svpFundTransactionUnsignedHash` entry.
-if it does, the received tx is saved in the new `svpFundTransactionSigned` storage entry, and the value saved in `svpFundTransactionUnsignedHash` storage entry is removed


jeremy-then (approved):
LGTM.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -422,6 +425,38 @@ public void registerBtcTransaction(
...
+
+    private boolean isTheSvpFundTransaction(Sha256Hash svpFundTransactionHashUnsigned, BtcTransaction transaction) {
+        if (!transaction.hasWitness()) {
+            BtcTransaction transactionCopy = new BtcTransaction(networkParameters, transaction.bitcoinSerialize()); // this is needed to not remove signatures from the actual tx
+            BitcoinUtils.removeSignaturesFromTransactionWithP2shMultiSigInputs(transactionCopy);
+            return transactionCopy.getHash().equals(svpFundTransactionHashUnsigned);
+        }
+
+        return transaction.getHash().equals(svpFundTransactionHashUnsigned);
+    }
```
marcos-iov, line 449:
```suggestion
    private void updateSvpFundTransactionValuesIfPossible(BtcTransaction transaction) {
        if (!isTheSvpFundTransaction(transaction)) {
            return;
        }

        updateSvpFundTransactionValues(transaction);
    }

    private boolean isTheSvpFundTransaction(BtcTransaction transaction) {
        Optional<Sha256Hash> svpFundTransactionHashUnsigned = provider.getSvpFundTxHashUnsigned();
        if (!svpFundTransactionHashUnsigned.isPresent()) {
            return false;
        }
        
        Sha256Hash txHash = transaction.getHash();
        if (!transaction.hasWitness()) {
            BtcTransaction transactionWithoutSignatures = new BtcTransaction(networkParameters, transaction.bitcoinSerialize()); // this is needed to not remove signatures from the actual tx
            BitcoinUtils.removeSignaturesFromTransactionWithP2shMultiSigInputs(transactionCopy);
            txHash = transactionWithoutSignatures.getHash();
        }

        return txHash.equals(svpFundTransactionHashUnsigned);
    }
```

I think that simplifies things a bit. Only get the fund tx from storage where needed. And only do the hash comparisson once.

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -805,6 +809,235 @@ private void assertEventWasEmittedWithExpectedData(byte[] expectedData) {
                 .findFirst();
             assertTrue(data.isPresent());
         }
+
+        @Test
```
marcos-iov, line 813: I believe we have a specific class for registerBtcTransaction tests. `BridgeSupportRegisterBtcTransactionTest`. Should we move these tests there?

marcos-iov (commented):
Looking good, just a small suggestion

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
julia-zack: it makes sense actually. But im not sure 🙃

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: i don't super like the `isTheSvpFundTransaction` method also doing the `svpFundTxHashUnsigned` check.
And overall remember it would be removed soon and added to method being called from `updateCollections`. wdyt?

julia-zack: btw, the other suggestion about the txHash was done :) Thanks

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -422,6 +425,39 @@ public void registerBtcTransaction(
...
+        if (!svpFundTransactionHashUnsigned.isPresent()) {
+            return;
+        }
+
+        if (!isTheSvpFundTransaction(svpFundTransactionHashUnsigned.get(), transaction)) {
+            return;
+        }
+
+        updateSvpFundTransactionValues(transaction);
+    }
```
apancorb, line 439: You can take advantage of the Optional API and make this more concise and functional as follows:

```java
private void updateSvpFundTransactionValuesIfPossible(BtcTransaction transaction) {
    provider.getSvpFundTxHashUnsigned()
        .filter(hash -> isTheSvpFundTransaction(hash, transaction))
        .ifPresent(hash -> updateSvpFundTransactionValues(transaction));
}
````

File: rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java [hunk 4]
```diff
@@ -805,6 +809,235 @@ private void assertEventWasEmittedWithExpectedData(byte[] expectedData) {
...
+        void registerBtcTransaction_forSvpFundTransactionChange_whenSvpPeriodIsOngoing_shouldRegisterTransactionAndUpdateSvpFundTransactionValues() throws Exception {
+            // Arrange
+            BtcTransaction svpFundTransaction = arrangeSvpFundTransactionWithChange();
+
+            signInputs(svpFundTransaction); // a transaction trying to be registered should be signed
+            setUpForTransactionRegistration(svpFundTransaction);
+
+            // Act
+            int activeFederationUtxosSizeBeforeRegisteringTx = federationSupport.getActiveFederationBtcUTXOs().size();
+            bridgeSupport.registerBtcTransaction(rskTx, svpFundTransaction.bitcoinSerialize(), btcBlockWithPmtHeight, pmtWithTransactions.bitcoinSerialize());
```
apancorb, line 913: nit: consider wrapping long lines

apancorb (approved):
LGTM. left two comments for your consideration.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: nice! thanks

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: updated :)

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
nathanieliov: I agree with @marcos-iov

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
```diff
@@ -805,6 +809,260 @@ private void assertEventWasEmittedWithExpectedData(byte[] expectedData) {
...
+
+        private void addOutputChange(BtcTransaction transaction) {
+            // add output to the active fed
+            Script activeFederationP2SHScript = activeFederation.getP2SHScript();
+            transaction.addOutput(Coin.COIN.multiply(10), activeFederationP2SHScript);
+        }
+
+        private void savePegoutIndex(BtcTransaction pegout) {
+            // save sigHash in pegout index
+            Optional<Sha256Hash> inputSigHash = BitcoinUtils.getFirstInputSigHash(pegout);
```
nathanieliov, line 1009: There is a warning because the optional is being used without checking whether it is present.

nathanieliov (approved):
Well done. Just some small suggestions.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -406,6 +406,9 @@ public void registerBtcTransaction(
                 case PEGOUT_OR_MIGRATION:
                     logger.debug("[registerBtcTransaction] This is a peg-out or migration tx {}", btcTx.getHash());
                     processPegoutOrMigration(btcTx);
+                    if (svpIsOngoing()) {
+                        updateSvpFundTransactionValuesIfPossible(btcTx);
+                    }
```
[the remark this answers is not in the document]
julia-zack: I get the point, but i'd stick to YAGNI principle for now :)

rskj-core/src/test/java/co/rsk/peg/BridgeSupportTest.java
julia-zack: updated! Thanks

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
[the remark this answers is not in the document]
marcos-iov: That's an interesting point @nathanieliov. I think that in this case, the logic to process an svp transaction is not really part of the logic of processing a pegout or migration. It's part of registering a transaction in the Bridge. So, with that in mind, looks like this is the right place.

That's my view at least :)

marcos-iov (approved):

marcos-iov (approved):

PR #2775, opened 2024-09-30
Add missing empty p2sh input script to svp spend transaction

julia-zack:
The empty input script has to be added for every svp spend tx input.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -1083,13 +1083,15 @@ public class BridgeSupport {
         Script proposedFederationRedeemScript = proposedFederation.getRedeemScript();
         Script proposedFederationOutputScript = proposedFederation.getP2SHScript();
         addInputFromMatchingOutputScript(svpSpendTransaction, svpFundTxSigned, proposedFederationOutputScript);
-        addEmptyP2SHInputScript(svpSpendTransaction.getInput(0), proposedFederationRedeemScript);
+        svpSpendTransaction.getInput(0)
+            .setScriptSig(createBaseP2SHInputScriptThatSpendsFromRedeemScript(proposedFederationRedeemScript));
 
         Script flyoverRedeemScript =
             getFlyoverRedeemScript(bridgeConstants.getProposedFederationFlyoverPrefix(), proposedFederationRedeemScript);
         Script flyoverOutputScript = ScriptBuilder.createP2SHOutputScript(flyoverRedeemScript);
         addInputFromMatchingOutputScript(svpSpendTransaction, svpFundTxSigned, flyoverOutputScript);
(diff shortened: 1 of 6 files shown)


rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -440,6 +438,26 @@ void addInputFromOutputSentToScript_withMatchingOutputScript_shouldAddInputWithO
...
     }
 
+    @Test
+    void addP2SHEmptyInputScript_shouldAddExpectedScriptSigToInput() {
+        // arrange
+        Federation federation = P2shErpFederationBuilder.builder().build();
+        Script redeemScript = federation.getRedeemScript();
+
+        BtcTransaction transaction = new BtcTransaction(btcMainnetParams);
+        transaction.addInput(transaction.getHash(), 0, redeemScript);
```
marcos-iov, line 448: This line raises a few questions
- Is that method expecting a redeem script or a script sig? I'd debug this and check the transaction after adding the input
- If it's expecting a redeemScript, does it use it to build an empty scriptsig? Does this work with ERP type scripts? i.e. adding the flo op code
- Is this how we are creating the spend tx? If that is the case then no need to add the empty script sig, right?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1081,11 +1080,18 @@ private BtcTransaction createSvpSpendTransaction(BtcTransaction svpFundTxSigned,
     }
 
     private void addSvpSpendTransactionInputs(BtcTransaction svpSpendTransaction, BtcTransaction svpFundTxSigned, Federation proposedFederation) {
+        Script proposedFederationRedeemScript = proposedFederation.getRedeemScript();
         Script proposedFederationOutputScript = proposedFederation.getP2SHScript();
         addInputFromMatchingOutputScript(svpSpendTransaction, svpFundTxSigned, proposedFederationOutputScript);
-
-        Script flyoverProposedFederationOutputScript = getFlyoverScriptPubKey(bridgeConstants.getProposedFederationFlyoverPrefix(), proposedFederation.getRedeemScript());
-        addInputFromMatchingOutputScript(svpSpendTransaction, svpFundTxSigned, flyoverProposedFederationOutputScript);
+        svpSpendTransaction.getInput(0)
```
marcos-iov, line 1086: Should we also update the tests where the spend tx is created? Make sure it has the expected script sig after creation

julia-zack: i think thats covered with the test that tests the method, but there's no harm in adding it sure

julia-zack: updated

PR #2788, opened 2024-10-03
Refactor addSignature as previous step to create addSvpSpendTxSignature method

julia-zack:


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1676,173 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
+        provider.setSvpSpendTxWaitingForSignatures(null);
+    }
+
+    private void logMissingSignatures(BtcTransaction btcTx, Keccak256 rskTxHash, int neededSignatures, int federationSize) {
+        int missingSignatures = BridgeUtils.countMissingSignatures(btcContext, btcTx);
+        int signaturesCount = neededSignatures - missingSignatures;
+
+        logger.debug("Tx {} not yet fully signed. Requires {}/{} signatures but has {}",
+            rskTxHash, neededSignatures, federationSize, signaturesCount);
+    }
```
julia-zack, line 1671: this might seem unnecessary, but it would be needed when adding new method

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1676,173 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
+
+    private void logReleaseBtc(BtcTransaction btcTx, byte[] rskTxHashSerialized) {
+        logger.info("Tx fully signed {}. Hex: {}", btcTx, Bytes.of(btcTx.bitcoinSerialize()));
+        eventLogger.logReleaseBtc(btcTx, rskTxHashSerialized);
+    }
```
julia-zack, line 1676: this might seem unnecessary, but it would be needed when adding new method

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1570,15 +1571,48 @@ private void adjustBalancesIfChangeOutputWasDust(BtcTransaction btcTx, Coin sent
...
+            logger.warn("Expected {} signatures but received {}.", inputsSize, signaturesSize);
+            return true;
+        }
+        return false;
+    }
+
+    private void addReleaseSignatures(
+        BtcECKey federatorPublicKey,
+        List<byte[]> signatures,
+        byte[] rskTxHashSerialized,
```
apancorb, line 1611: Maybe it would be a good idea to replace this by the deserialized `Keccak256` hash

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1663,133 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
         return Optional.empty();
     }
 
+    private void logMissingSignatures(BtcTransaction btcTx, Keccak256 rskTxHash, int neededSignatures, int federationSize) {
+        int missingSignatures = BridgeUtils.countMissingSignatures(btcContext, btcTx);
+        int signaturesCount = neededSignatures - missingSignatures;
+
+        logger.debug("Tx {} not yet fully signed. Requires {}/{} signatures but has {}",
```
apancorb, line 1670: maybe this can be made at the `INFO` level. It seems interesting/important to me, wdyt?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1597,28 +1631,22 @@ public void addSignature(BtcECKey federatorPublicKey, List<byte[]> signatures, b
...
+        if (!BridgeUtils.hasEnoughSignatures(btcContext, btcTx) && logger.isDebugEnabled()) {
+            int neededSignatures = federation.getNumberOfSignaturesRequired();
+            logMissingSignatures(btcTx, rskTxHash, neededSignatures, federation.getSize());
+
+            return;
         }
 
-        processSigning(signingFederationMember, signatures, rskTxHash, btcTx, federation);
+        logReleaseBtc(btcTx, rskTxHashSerialized);
+        provider.getPegoutsWaitingForSignatures().remove(rskTxHash);
```
apancorb, line 1649: I put this only as a side note. We have to be careful with how we are mutating state here. Maybe something to analyze in the future.

rskj-core/src/main/java/co/rsk/peg/bitcoin/BitcoinUtils.java
```diff
@@ -95,4 +95,14 @@ public static Optional<TransactionOutput> searchForOutput(List<TransactionOutput
...
+
+    public static Sha256Hash generateSigHashForP2SHInput(BtcTransaction btcTx, int inputIndex) {
+        TransactionInput input = btcTx.getInput(inputIndex);
+        Optional<Script> redeemScript = extractRedeemScriptFromInput(input);
+        if (redeemScript.isEmpty()) {
+            throw new IllegalArgumentException("Couldn't extract redeem script from p2sh input");
+        }
+
+        return btcTx.hashForSignature(inputIndex, redeemScript.get(), BtcTransaction.SigHash.ALL, false);
+    }
```
apancorb, line 107: This can be made more functional, assuming that btcTx cannot be null:

```java
public static Sha256Hash generateSigHashForP2SHInput(BtcTransaction btcTx, int inputIndex) {
    return Optional.ofNullable(btcTx.getInput(inputIndex))
        .flatMap(input -> extractRedeemScriptFromInput(input))
        .map(redeemScript -> btcTx.hashForSignature(inputIndex, redeemScript, BtcTransaction.SigHash.ALL, false)) 
        .orElseThrow(() -> new IllegalArgumentException("Couldn't extract redeem script from p2sh input"));
}
```

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -140,7 +140,7 @@ void addSignature_fedPubKey_belongs_to_active_federation() throws Exception {
             createHash3(1).getBytes()
         );
 
-        verify(provider, times(1)).getPegoutsWaitingForSignatures();
+        verify(provider, times(0)).getPegoutsWaitingForSignatures();
```
apancorb, line 143:
```suggestion
        verify(provider, never()).getPegoutsWaitingForSignatures();
```

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -207,7 +207,7 @@ void addSignature_fedPubKey_belongs_to_retiring_federation() throws Exception {
...
```
apancorb, line 210:
```suggestion
        verify(provider, never()).getPegoutsWaitingForSignatures();
```

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
apancorb: Just for my knowledge, why is this never occurring now?

apancorb (commented):
Very nice refactor, left comments :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -998,7 +999,7 @@ private void logUpdateCollections(Transaction rskTx) {
     private boolean svpIsOngoing() {
```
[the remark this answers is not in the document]
julia-zack: super nice, thanks :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1570,15 +1571,48 @@ private void adjustBalancesIfChangeOutputWasDust(BtcTransaction btcTx, Coin sent
...
+        if (signatures == null || signatures.isEmpty()) {
+            return;
+        }
+
         Context.propagate(btcContext);
 
+        Keccak256 rskTxHash = new Keccak256(rskTxHashSerialized);
+        int signaturesSize = signatures.size();
+
+        BtcTransaction btcTx = provider.getPegoutsWaitingForSignatures().get(rskTxHash);
```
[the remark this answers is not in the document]
julia-zack: the current implementation will return a `new TreeMap<>()` if there are no pegoutsWFS, so the NPE is covered :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1570,15 +1571,48 @@ private void adjustBalancesIfChangeOutputWasDust(BtcTransaction btcTx, Coin sent
...
+        if (btcTx == null) {
+            logger.warn("No tx waiting for signature for hash {}. Probably fully signed already.", rskTxHash);
+            return;
+        }
+        if (notEnoughSignatures(btcTx.getInputs().size(), signaturesSize)) {
```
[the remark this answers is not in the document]
julia-zack: i agree, but i added it to avoid duplication when adding the svp spend tx signing

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: it uses both the serialized and the deserialized one, do you see any advantages in passing the deserialized one?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: i think it may add lot of logs. Have in mind that it will appear every time one fed signs a tx

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
julia-zack: because we are returning early when null / empty signatures :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1570,15 +1571,48 @@ private void adjustBalancesIfChangeOutputWasDust(BtcTransaction btcTx, Coin sent
...
      */
-    public void addSignature(BtcECKey federatorPublicKey, List<byte[]> signatures, byte[] rskTxHash) throws Exception {
+    public void addSignature(BtcECKey federatorBtcPublicKey, List<byte[]> signatures, byte[] rskTxHashSerialized) throws Exception {
+        if (signatures == null || signatures.isEmpty()) {
+            return;
+        }
+
         Context.propagate(btcContext);
 
+        Keccak256 rskTxHash = new Keccak256(rskTxHashSerialized);
```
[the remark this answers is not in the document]
julia-zack: nope, it throws an `IllegalArgumentException`. See the [implementation](https://github.com/rsksmart/rskj/blob/refactor_add_signature/rskj-core/src/main/java/co/rsk/crypto/Keccak256.java#L45).

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
apancorb: I think it should not add that many. It should only appear `n-1` times per tx. Notice with the pegout signed cache in place, addSignature should only get triggered once. I would find this info interesting as a fed node operator

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
apancorb: Maybe with the deserialized one you can do something like `getBytes()` or something. I am saying this to avoid doing it twice

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
apancorb: thanks :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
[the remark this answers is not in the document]
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1570,15 +1571,48 @@ private void adjustBalancesIfChangeOutputWasDust(BtcTransaction btcTx, Coin sent
      * The hash for the signature must be calculated with Transaction.SigHash.ALL and anyoneCanPay=false. The signature must be canonical.
      * If enough signatures were added, ask federators to broadcast the btc release tx.
      *
-     * @param federatorPublicKey   Federator who is signing
+     * @param federatorBtcPublicKey   Federator who is signing
      * @param signatures           1 signature per btc tx input
-     * @param rskTxHash            The id of the rsk tx
+     * @param rskTxHashSerialized            The id of the rsk tx
```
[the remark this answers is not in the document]
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: in that case we are doing `rskTxHashSerialized -> rskTxHash -> rskTxHash.getBytes()`
instead of just `rskTxHashSerialized`

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/bitcoin/BitcoinUtils.java
julia-zack: updated! Thanks

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1570,15 +1567,48 @@ private void adjustBalancesIfChangeOutputWasDust(BtcTransaction btcTx, Coin sent
...
+            return;
+        }
+        if (notEnoughSignatures(btcTx.getInputs().size(), signaturesSize)) {
+            return;
+        }
+
+        addReleaseSignatures(federatorBtcPublicKey, signatures, rskTxHashSerialized, btcTx);
+    }
+
+    private boolean notEnoughSignatures(int inputsSize, int signaturesSize) {
```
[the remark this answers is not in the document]
marcos-iov: There is also a general rule to make boolean methods/variables naming affirmative. To avoid possible double negation `if (! notEnoughSignatures)`

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
marcos-iov: Agree, it's a nice thing to focus on when we work on refactors next time

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1659,133 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
+                logger.warn("Input {} of tx {} already signed by this federator.", i, rskTxHash);
                 break;
             }
-        }
 
-        if(signed && activations.isActive(ConsensusRule.RSKIP326)) {
-            eventLogger.logAddSignature(federatorMember, btcTx, rskTxHash);
+            try {
+                int sigIndex = inputScript.getSigInsertionIndex(sighash, federatorBtcPublicKey);
+                inputScript = ScriptBuilder.updateScriptWithSignature(inputScript, txSigs.get(i).encodeToBitcoin(), sigIndex, 1, 1);
```
marcos-iov, line 1773: I'd consider using `getScriptSigWithSignature` method. Internally it calls `updateScriptWithSignature` but receives less params so no need to pass those `1, 1` value that no one knows what they are.

Double check just in case that the behavior is the same

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1659,133 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
-        List<TransactionSignature> txSigs = new ArrayList<>();
-        for (int i = 0; i < numInputs; i++) {
-            TransactionInput txIn = btcTx.getInput(i);
-            Script inputScript = txIn.getScriptSig();
-            List<ScriptChunk> chunks = inputScript.getChunks();
-            byte[] program = chunks.get(chunks.size() - 1).data;
-            Script redeemScript = new Script(program);
-            sighashes.add(btcTx.hashForSignature(i, redeemScript, BtcTransaction.SigHash.ALL, false));
+        List<Sha256Hash> sigHashes = new ArrayList<>();
+        for (int i = 0; i < btcTx.getInputs().size(); i++) {
```
marcos-iov, line 1683: Looks like we are iterating multiple times through the inputs, the sighashes, the signatures. Could this maybe be solved in a single loop?

For each input, get the sighash, get the signature, and add it?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1659,133 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
-                    } else if (retiringFederation != null && retiringFederation.hasBtcPublicKey(federatorBtcPublicKey)) {
-                        logger.debug("A member of the retiring federation is trying to sign a tx of the active one");
-                        return;
-                    }
-                    throw e;
-                }
-            } else {
-                logger.warn("Input {} of tx {} already signed by this federator.", i, new Keccak256(rskTxHash));
+            if (alreadySignedByThisFederator) {
+                logger.warn("Input {} of tx {} already signed by this federator.", i, rskTxHash);
```
marcos-iov, line 1767: Same here, logging the rsk tx has is probably not what was intended. I'd check this in the remaining places

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1659,133 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
             }
-        }
 
-        if(signed && activations.isActive(ConsensusRule.RSKIP326)) {
-            eventLogger.logAddSignature(federatorMember, btcTx, rskTxHash);
+            try {
+                int sigIndex = inputScript.getSigInsertionIndex(sighash, federatorBtcPublicKey);
+                inputScript = ScriptBuilder.updateScriptWithSignature(inputScript, txSigs.get(i).encodeToBitcoin(), sigIndex, 1, 1);
+                input.setScriptSig(inputScript);
+                logger.debug("Tx input {} for tx {} signed.", i, rskTxHash);
```
marcos-iov, line 1775: again

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1659,133 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
-            eventLogger.logReleaseBtc(btcTx, rskTxHash);
-        } else if (logger.isDebugEnabled()) {
-            int missingSignatures = BridgeUtils.countMissingSignatures(btcContext, btcTx);
-            int neededSignatures = federation.getNumberOfSignaturesRequired();
-            int signaturesCount = neededSignatures - missingSignatures;
-
-            logger.debug("Tx {} not yet fully signed. Requires {}/{} signatures but has {}",
-                    new Keccak256(rskTxHash), neededSignatures, getActiveFederationSize(), signaturesCount);
-        }
+        return signed;
```
marcos-iov, line 1788: Does this ever return false or it throws exception when something is wrong?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1659,133 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
+                logger.debug("Tx input {} for tx {} signed.", i, rskTxHash);
+                signed = true;
+            } catch (IllegalStateException e) {
+                Federation retiringFederation = getRetiringFederation();
+                if (getActiveFederation().hasBtcPublicKey(federatorBtcPublicKey)) {
+                    logger.debug("A member of the active federation is trying to sign a tx of the retiring one");
+                } else if (retiringFederation != null && retiringFederation.hasBtcPublicKey(federatorBtcPublicKey)) {
+                    logger.debug("A member of the retiring federation is trying to sign a tx of the active one");
+                }
+                throw e;
```
marcos-iov, line 1784: Where does this exception end up?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1570,15 +1567,48 @@ private void adjustBalancesIfChangeOutputWasDust(BtcTransaction btcTx, Coin sent
...
+
+    private boolean notEnoughSignatures(int inputsSize, int signaturesSize) {
+        if (inputsSize != signaturesSize) {
+            logger.warn("Expected {} signatures but received {}.", inputsSize, signaturesSize);
+            return true;
+        }
+        return false;
+    }
+
+    private void addReleaseSignatures(
```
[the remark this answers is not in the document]
julia-zack: the `addReleaseSignatures` was just added to distinguish it from the `addSvpSpendTxSignatures`. 
But I agree that the naming can be improved

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: yes, agree. I'll create a task on the backlog to have it in mind

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: i tried to, but it breaks behaviour

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -140,7 +140,7 @@ void addSignature_fedPubKey_belongs_to_active_federation() throws Exception {
             createHash3(1).getBytes()
         );
 
-        verify(provider, times(1)).getPegoutsWaitingForSignatures();
```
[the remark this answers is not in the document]
julia-zack: because now we are checking if signatures are null and `[returning early in that case](https://github.com/rsksmart/rskj/blob/ae68f5ef35e0f0fe7a6ae77a75f91c4c35a60836/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java#L1576)`.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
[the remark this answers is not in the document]
julia-zack: you are right actually, the hash is always valid.
Regarding the keccak suggestion, I think is just a matter of taste, because in the BridgeSupport implementation we are using both the keccak hash and the serialized hash

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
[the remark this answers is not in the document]
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: updated! Thanks for noticing it

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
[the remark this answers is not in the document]
marcos-iov: I think it's cleaner to receive the Keccack object, it ensures you will be working with a valid element. Receiving an array of bytes, could be anything. That would force you to do some validations since you can't be sure that the value was previously validated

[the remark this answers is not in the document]
julia-zack: make sense! I'll update it

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1659,135 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
+            Sha256Hash sigHash = sigHashes.get(i);
+
+            if (!federatorBtcPublicKey.verify(sigHash, decodedSignature)) {
                 logger.warn(
                     "Signature {} {} is not valid for hash {} and public key {}",
                     i,
-                    Bytes.of(sig.encodeToDER()),
-                    sighash,
+                    Bytes.of(decodedSignature.encodeToDER()),
+                    sigHashes,
```
marcos-iov, line 1720: Sure about this change?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1659,135 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
+                logger.warn("Input {} of tx {} already signed by this federator.", i, rskTxHash);
                 break;
             }
-        }
 
-        if(signed && activations.isActive(ConsensusRule.RSKIP326)) {
-            eventLogger.logAddSignature(federatorMember, btcTx, rskTxHash);
+            try {
+                int sigIndex = inputScript.getSigInsertionIndex(sighash, federatorBtcPublicKey);
+                Script outputScript = ScriptBuilder.createP2SHOutputScript(getRedeemScriptFromP2SHInputScript(inputScript));
```
marcos-iov, line 1774: I'd store the redeem script in a separate variable. To avoid all this chained calls

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1659,135 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
+                inputScript = outputScript.getScriptSigWithSignature(inputScript, txSigs.get(i).encodeToBitcoin(), sigIndex);
```
marcos-iov, line 1775: Maybe this should be a new variable? `inputScriptWithSignature`?

To avoid confusions, looks a bit strange like this

rskj-core/src/main/java/co/rsk/peg/bitcoin/BitcoinUtils.java
```diff
@@ -90,9 +90,23 @@ public static Script createBaseP2SHInputScriptThatSpendsFromRedeemScript(Script
         return outputScript.createEmptyInputScript(null, redeemScript);
     }
 
+    public static Script getRedeemScriptFromP2SHInputScript(Script inputScript) {
```
marcos-iov, line 93: tests?

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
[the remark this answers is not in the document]
marcos-iov: Thanks. Weird test anyway, not sure if that's what it is intended to test. At least it's not what the name says.

But we can review this at a later stage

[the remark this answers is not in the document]
marcos-iov: Thanks. Weird test anyway, not sure if that's what it is intended to test. At least it's not what the name says.

But we can review this at a later stage

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: thanks

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: makes sense

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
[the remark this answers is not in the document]
julia-zack: yep, all these tests are a bit weird. It will make sense to take a deeper look at them eventually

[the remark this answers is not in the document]
julia-zack: i created a task to do it

rskj-core/src/main/java/co/rsk/peg/bitcoin/BitcoinUtils.java
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -617,11 +617,12 @@ public void addSignature(Object[] args) throws VMException {
             }
             signatures.add(signatureByteArray);
         }
-        byte[] rskTxHash = (byte[]) args[2];
-        if (rskTxHash.length!=32) {
-            throw new BridgeIllegalArgumentException("Invalid rsk tx hash " + Bytes.of(rskTxHash));
+        byte[] rskTxHashSerialized = (byte[]) args[2];
+        if (rskTxHashSerialized.length!=32) {
```
apancorb, line 621:
```suggestion
        if (rskTxHashSerialized.length != 32) {
```

apancorb (approved):
LGTM

rskj-core/src/main/java/co/rsk/peg/bitcoin/BitcoinUtils.java
marcos-iov: Looking a bit more into this, we already have a method here `extractRedeemScriptFromInput`. Why not use that one passing the input directly? Instead of extracting the input script and then calling this

rskj-core/src/test/java/co/rsk/peg/bitcoin/BitcoinUtilsTest.java
```diff
@@ -467,7 +464,65 @@ void createBaseP2SHInputScriptThatSpendsFromRedeemScript_shouldCreateExpectedScr
...
+        // act
+        Sha256Hash sigHash = generateSigHashForP2SHTransactionInput(transaction, inputIndex);
+
+        // assert
+        Sha256Hash expectedSigHash = transaction.hashForSignature(inputIndex, redeemScript, BtcTransaction.SigHash.ALL, false);
+        assertEquals(expectedSigHash, sigHash);
+    }
+
+    @Test
+    void generateSigHashForP2SHInput_forEmptyInputScript_shouldThrowIllegalArgumentException() {
```
marcos-iov, line 511: A P2PKH input test case? An invalid inputIndex value?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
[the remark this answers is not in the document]
julia-zack: this was updated

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1635,119 +1657,137 @@ private Optional<Federation> getFederationFromPublicKey(BtcECKey federatorPublic
...
-                    Bytes.of(signatures.get(i))
-                );
-                return;
-            }
+        BtcECKey federatorBtcPublicKey = federatorMember.getBtcPublicKey();
+        List<TransactionSignature> txSigs;
+        try {
+            txSigs = getTransactionSignatures(federatorBtcPublicKey, sigHashes, signatures);
+        } catch (SignatureException e) {
+            logger.error("[processSigning] Unable to proceed with signing as the transaction signatures are incorrect.");
```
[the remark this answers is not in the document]
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/Bridge.java
julia-zack: updateddddd

rskj-core/src/main/java/co/rsk/peg/bitcoin/BitcoinUtils.java
julia-zack: could be, i find it weird for this method to return an optional. I think it should return the script or an exception, wdyt?

marcos-iov: I think the Optional gives a bit more flexibility, it get the redeem script if there is one. Then you can do all kinds of functional stuff with the optional return.

julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/Bridge.java
```diff
@@ -617,15 +618,13 @@ public void addSignature(Object[] args) throws VMException {
...
+        byte[] rskTxHashSerialized = (byte[]) args[2];
         try {
+            Keccak256 rskTxHash = deserializeRskTxHash(rskTxHashSerialized);
             bridgeSupport.addSignature(federatorPublicKey, signatures, rskTxHash);
-        } catch (BridgeIllegalArgumentException e) {
-            throw e;
-        } catch (Exception e) {
+        } catch (IllegalArgumentException e) {
+            throw new BridgeIllegalArgumentException("Invalid rsk tx hash " + Bytes.of(rskTxHashSerialized));
+        } catch (IOException e) {
```
marcos-iov, line 627: Is this change a good idea? I think all Bridge methods are capturing generic exceptions

marcos-iov: There might be unchecked exception thrown

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -2025,4 +2026,27 @@ void testGetUTXOsSentToAddresses_no_utxo_sent_to_given_address_after_RSKIP293()
 
         Assertions.assertTrue(foundUTXOs.isEmpty());
     }
+
+    @Test
```
marcos-iov, line 2030: Should these tests go in BridgeSerializationUtilsTest?

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
```diff
@@ -2025,4 +2026,27 @@ void testGetUTXOsSentToAddresses_no_utxo_sent_to_given_address_after_RSKIP293()
...
+
+        // act
+        Keccak256 rskTxHash = deserializeRskTxHash(rskTxHashSerialized);
+
+        // assert
+        assertEquals(expectedRskTxHash, rskTxHash);
+    }
+
+    @Test
+    void deserializeRskTxHash_withInvalidLength_throwsIllegalArgumentException() {
```
marcos-iov, line 2044: A test passing a null value?

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
julia-zack: yes, my bad. Thanks

rskj-core/src/test/java/co/rsk/peg/BridgeUtilsTest.java
julia-zack: good catch, thanks

julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/Bridge.java
julia-zack: updated

marcos-iov (approved):

PR #2801, opened 2024-10-11
Adding transient storage maps with TLOAD/TSTORE implementation

fmacleal:
This PR adds the basic implementation of the TLOAD/TSTORE opcodes with some DSL contracts tests. Still work in progress and more validations need to be added.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
```diff
@@ -406,4 +429,48 @@ private Optional<Keccak256> internalGetValueHash(byte[] key) {
...
+        if (value == null || value.length == 0) {
+            transientTrie.put(triekey, null);
+        } else {
+            transientTrie.put(triekey, value);
+        }
+    }
+
+    @Override
+    public void clearTransientStorage() {
+        this.transientTrie =  new MutableTrieImpl(new TrieStoreImpl(new HashMapDB()), new Trie());
```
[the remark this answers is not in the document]
fmacleal: Basically after every transaction execution. The thing is that this class it's initialized everytime a transaction starts anyway. Because this is used for the Storage during the transaction execution that might be rolled back and used by the storage. This clear will be called at end of a transaction.

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1967,9 +1967,16 @@ protected void executeOpcode() {
...
             case OpCodes.OP_SSTORE: doSSTORE();
                 break;
-            case OpCodes.OP_TLOAD: doTLOAD();
+            case OpCodes.OP_TLOAD:
+                if (!activations.isActive(RSKIP446)) {
+                    throw Program.ExceptionHelper.invalidOpCode(program);
+                }
+                doTLOAD();
                 break;
             case OpCodes.OP_TSTORE: doTSTORE();
```
asoto-iov, line 1976: I think it should go after the activation rule check.

rskj-core/src/test/java/co/rsk/vm/opcode/TransientStorageDslTest.java
```diff
@@ -45,23 +47,59 @@ void testTransientStorageOpcodesExecutionsWithRSKIPActivated() throws FileNotFou
...
+        String secondaryContractTransientStorageCreationTxName = "txTestTransientStorageOtherContract";
+        assertTransactionReceiptWithStatus(world, secondaryContractTransientStorageCreationTxName, "b02", true);
+
+        String checkingOpcodesTxName = "txTestTransientStorageNestedTransactionOtherContractDoesntShareMemory";
+        TransactionReceipt txReceipt = assertTransactionReceiptWithStatus(world, checkingOpcodesTxName, "b06", true);
+        Assertions.assertEquals(1, TransactionReceiptUtil.getEventCount(txReceipt, "ERROR", new String[]{"bytes32"}));
+    }
+
+    @Test
+    void testTransientStorageOpcodesExecutionFailsWithRSKIPDeactivated() throws FileNotFoundException, DslProcessorException {
```
asoto-iov, line 102: In the code the OP code call is done before the activation check, is the test working anyway?

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1327,30 +1327,30 @@ else if (oldValue != null && newValue.isZero()) {
...
+
+        program.stackPush(val);
+        // key could be returned to the pool, but transientStorageLoad semantics should be checked
         // to make sure storageLoad always gets a copy, not a reference.
         program.step();
     }
 
     protected void doTSTORE(){
-
+        DataWord address = program.stackPop();
```
asoto-iov, line 1347: About the naming of this variable, shouldn't it be "key" instead of address?

rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
```diff
@@ -25,19 +25,29 @@
...
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.math.BigInteger;
-import java.util.*;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Optional;
+import java.util.Set;
 
 public class MutableRepository implements Repository {
```
asoto-iov, line 52: I don't fully understand why the transient storage is implemented into the standard storage.

rskj-core/src/main/java/org/ethereum/core/Repository.java
```diff
@@ -27,7 +27,7 @@
 
 import java.math.BigInteger;
 
-public interface Repository extends RepositorySnapshot {
+public interface Repository extends RepositorySnapshot, TransientRepository {
```
asoto-iov, line 30: Why a repository is both standard repo and transient repo?

rskj-core/src/main/java/org/ethereum/core/TransientRepository.java
```diff
@@ -0,0 +1,40 @@
...
+import co.rsk.core.RskAddress;
+import org.ethereum.vm.DataWord;
+
+import javax.annotation.Nullable;
+
+public interface TransientRepository  {
+
+    void addTransientStorageRow(RskAddress addr, DataWord key, DataWord value);
+
+    void addTransientStorageBytes(RskAddress addr, DataWord key, byte[] value);
```
asoto-iov, line 31: is it needed these two methods? it seems that addTransientStorageBytes is only called from `addTransientStorageRow` and addTransientStorageRow it is only being used once in the code.

rskj-core/src/main/java/org/ethereum/db/TransientStorageRepositoryCreator.java
```diff
@@ -0,0 +1,36 @@
...
+import co.rsk.trie.TrieStoreImpl;
+import org.ethereum.core.Repository;
+import org.ethereum.datasource.HashMapDB;
+
+public class TransientStorageRepositoryCreator {
+
+    private TransientStorageRepositoryCreator() {
+    }
+
+    public static Repository createNewTransientStorage() {
```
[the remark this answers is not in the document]
fmacleal: Not used anymore, should have been deleted and I forgot it here during some rebases. Good catch,

rskj-core/src/test/java/co/rsk/vm/opcode/TransientStorageDslTest.java
fmacleal: It is working, because it fails for the other one that was fine. We can have a split test with a contract that does only TLOAD and other only TSTORE to catch these two situations, I will do.

rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
fmacleal: This is not the standard storage, it's the mutable storage that if a transaction is not rolled back it can be reverted. The transient storage it's also mutable and the lifecycle of it it's the same for the regular mutable storage. That's why it's implemented here, for the sake of simplicity and to be managed in the same way. It will get advantage of the whole lifecycle of the mutable storage that we already have.

rskj-core/src/main/java/org/ethereum/core/Repository.java
fmacleal: It's explained a bit above, if still not clear, let me know and I will try to explain more. Maybe we can have a call to discuss.

rskj-core/src/main/java/org/ethereum/core/TransientRepository.java
fmacleal: After follow your suggestion, is being used twice now. Let's keep it. :)

rskj-core/src/main/java/org/ethereum/db/MutableRepository.java
asoto-iov: ok, it is interesting point of view.

asoto-iov (approved):
Good work. 
My only concern is about the implementation of both transient and standard storage in the same classes. I think it could have some disadvantages. I would like to talk about it in a call but I think it is not a blocker to not continue with it.

PR #2808, opened 2024-10-16
Rename variables related to release amounts. Indicate satoshis or weis

marcos-iov:


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -840,33 +840,33 @@ public void releaseBtc(Transaction rskTx) throws IOException {
     }
 
     private void refundAndEmitRejectEvent(
-        co.rsk.core.Coin releaseRequestedValue,
+        co.rsk.core.Coin releaseRequestedValueInWeis,
         RskAddress senderAddress,
         RejectedPegoutReason reason
     ) {
         logger.trace(
             "[refundAndEmitRejectEvent] Executing a refund of {} to {}. Reason: {}",
```
nathanieliov, line 848: Do you think it is worth mentioning that the value being refunded is in Weis in this log?

nathanieliov (approved):
Just a small suggestion

PR #2838, opened 2024-10-31
Create updateSvpState method

julia-zack:
**Background**
In order to keep going with the validation process, in every `updateCollections` we will call a new validation-related method, `updateSvpState`, that will update the state of the svp process to see if the validation failed, or if conditions are met for creating the fund or the spend transaction.

 
IN THIS PR
Create a new `updateSvpState` method that will:
- If a proposed federation does not exist, do nothing.

- If a proposed federation exists and the validation period is ended, we can conclude that the validation failed.

-If a proposed federation exists, the validation period is not ended and the `svpFundTxHashUnigned` entry is empty, we have to proceed with the fund transaction creation and processing.

-If a proposed federation exists, the validation period is not ended and the `svpFundTxSigned` has a value stored, we will proceed with the spend tx creation.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -1032,19 +1032,15 @@ public class BridgeSupport {
             return;
         }
 
-        // if the unsigned fund tx hash is not present, we will proceed with the fund tx creation
-        // if it is, we keep waiting for the change to be registered.
-
-        // if the fund tx signed is present, then the change was registered,
+        // if the fund tx signed is present, then the fund transaction change was registered,
         // meaning we can create the spend tx.
-
-        // if none of those values are present, we keep waiting for the spend tx to be registered.
         Optional<BtcTransaction> svpFundTxSigned = provider.getSvpFundTxSigned();
         if (svpFundTxSigned.isPresent()) {
             processSvpSpendTransactionUnsigned(rskTxHash, proposedFederation, svpFundTxSigned.get());
             return;
         }
 
+        // if the unsigned fund tx hash is not present, meaning we can proceed with the fund tx creation
         Optional<Sha256Hash> svpFundTxHashUnsigned = provider.getSvpFundTxHashUnsigned();
         if (svpFundTxHashUnsigned.isEmpty()) {
             try {
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -1023,6 +1023,7 @@ public class BridgeSupport {
             return;
         }
 
+        Keccak256 rskTxHash = rskTx.getHash();
         // if the proposed federation exists and the validation period ended,
         // we can conclude that the svp failed
         Federation proposedFederation = proposedFederationOpt.get();
@@ -1040,14 +1041,14 @@ public class BridgeSupport {
         // if none of those values are present, we keep waiting for the spend tx to be registered.
         Optional<BtcTransaction> svpFundTxSigned = provider.getSvpFundTxSigned();
         if (svpFundTxSigned.isPresent()) {
-            processSvpSpendTransactionUnsigned(rskTx, proposedFederation, svpFundTxSigned.get());
+            processSvpSpendTransactionUnsigned(rskTxHash, proposedFederation, svpFundTxSigned.get());
             return;
         }
 
         Optional<Sha256Hash> svpFundTxHashUnsigned = provider.getSvpFundTxHashUnsigned();
         if (svpFundTxHashUnsigned.isEmpty()) {
             try {
-                processSvpFundTransactionUnsigned(rskTx, proposedFederation);
+                processSvpFundTransactionUnsigned(rskTxHash, proposedFederation);
             } catch (Exception e) {
                 logger.error("[updateSvpState] Error processing svp fund transaction unsigned.");
             }
@@ -1082,13 +1083,13 @@ public class BridgeSupport {
         return rskExecutionBlock.getNumber() <= validationPeriodEndBlock;
     }
 
-    private void processSvpFundTransactionUnsigned(Transaction rskTx, Federation proposedFederation) throws IOException, InsufficientMoneyException {
+    private void processSvpFundTransactionUnsigned(Keccak256 rskTxHash, Federation proposedFederation) throws IOException, InsufficientMoneyException {
         Coin spendableValueFromProposedFederation = bridgeConstants.getSpendableValueFromProposedFederation();
         BtcTransaction svpFundTransactionUnsigned = createSvpFundTransaction(proposedFederation, spendableValueFromProposedFederation);
 
         provider.setSvpFundTxHashUnsigned(svpFundTransactionUnsigned.getHash());
         PegoutsWaitingForConfirmations pegoutsWaitingForConfirmations = provider.getPegoutsWaitingForConfirmations();
-        settleReleaseRequest(pegoutsWaitingForConfirmations, svpFundTransactionUnsigned, rskTx.getHash(), spendableValueFromProposedFederation);
+        settleReleaseRequest(pegoutsWaitingForConfirmations, svpFundTransactionUnsigned, rskTxHash, spendableValueFromProposedFederation);
     }
 
     private BtcTransaction createSvpFundTransaction(Federation proposedFederation, Coin spendableValueFromProposedFederation) throws InsufficientMoneyException {
@@ -1121,10 +1122,8 @@ public class BridgeSupport {
         return sendRequest;
     }
 
-    private void processSvpSpendTransactionUnsigned(Transaction rskTx, Federation proposedFederation, BtcTransaction svpFundTxSigned) {
+    private void processSvpSpendTransactionUnsigned(Keccak256 rskTxHash, Federation proposedFederation, BtcTransaction svpFundTxSigned) {
         BtcTransaction svpSpendTransactionUnsigned = createSvpSpendTransaction(svpFundTxSigned, proposedFederation);
-
-        Keccak256 rskTxHash = rskTx.getHash();
         updateSvpSpendTransactionValues(rskTxHash, svpSpendTransactionUnsigned);
 
         Coin amountSentToActiveFed = svpSpendTransactionUnsigned.getOutput(0).getValue();
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -1017,7 +1017,44 @@ public class BridgeSupport {
         eventLogger.logUpdateCollections(sender);
     }
 
-    protected void processSVPFailure(Federation proposedFederation) {
+    protected void updateSvpState(Transaction rskTx) {
+        Optional<Federation> proposedFederationOpt = federationSupport.getProposedFederation();
+        if (proposedFederationOpt.isEmpty()) {
+            return;
+        }
+
+        // if the proposed federation exists and the validation period ended,
+        // we can conclude that the svp failed
+        Federation proposedFederation = proposedFederationOpt.get();
+        if (!validationPeriodIsOngoing(proposedFederation)) {
+            processSVPFailure(proposedFederation);
+            return;
+        }
+
+        // if the unsigned fund tx hash is not present, we will proceed with the fund tx creation
+        // if it is, we keep waiting for the change to be registered.
+
+        // if the fund tx signed is present, then the change was registered,
+        // meaning we can create the spend tx.
+
+        // if none of those values are present, we keep waiting for the spend tx to be registered.
+        Optional<BtcTransaction> svpFundTxSigned = provider.getSvpFundTxSigned();
+        if (svpFundTxSigned.isPresent()) {
+            processSvpSpendTransactionUnsigned(rskTx, proposedFederation, svpFundTxSigned.get());
+            return;
+        }
+
+        Optional<Sha256Hash> svpFundTxHashUnsigned = provider.getSvpFundTxHashUnsigned();
+        if (svpFundTxHashUnsigned.isEmpty()) {
+            try {
+                processSvpFundTransactionUnsigned(rskTx, proposedFederation);
+            } catch (Exception e) {
+                logger.error("[updateSvpState] Error processing svp fund transaction unsigned.");
+            }
+        }
+    }
+
+    private void processSVPFailure(Federation proposedFederation) {
         eventLogger.logCommitFederationFailure(rskExecutionBlock, proposedFederation);
         logger.warn(
             "[processSVPFailure] Proposed federation validation failed at block {}, so federation election will be allowed again.",
@@ -1034,22 +1071,18 @@ public class BridgeSupport {
 
     private boolean svpIsOngoing() {
         return federationSupport.getProposedFederation()
-            .map(Federation::getCreationBlockNumber)
-            .map(proposedFederationCreationBlockNumber ->
-                proposedFederationCreationBlockNumber + bridgeConstants.getFederationConstants().getValidationPeriodDurationInBlocks())
-            .filter(validationPeriodEndBlock -> rskExecutionBlock.getNumber() <= validationPeriodEndBlock)
+            .filter(this::validationPeriodIsOngoing)
             .isPresent();
     }
 
-    protected void processSvpFundTransactionUnsigned(Transaction rskTx) throws IOException, InsufficientMoneyException {
-        Optional<Federation> proposedFederationOpt = federationSupport.getProposedFederation();
-        if (proposedFederationOpt.isEmpty()) {
-            String message = "Proposed federation should be present when processing SVP fund transaction.";
-            logger.warn(message);
-            throw new IllegalStateException(message);
-        }
-        Federation proposedFederation = proposedFederationOpt.get();
+    private boolean validationPeriodIsOngoing(Federation proposedFederation) {
+        long validationPeriodEndBlock = proposedFederation.getCreationBlockNumber() +
+            bridgeConstants.getFederationConstants().getValidationPeriodDurationInBlocks();
 
+        return rskExecutionBlock.getNumber() <= validationPeriodEndBlock;
+    }
+
+    private void processSvpFundTransactionUnsigned(Transaction rskTx, Federation proposedFederation) throws IOException, InsufficientMoneyException {
         Coin spendableValueFromProposedFederation = bridgeConstants.getSpendableValueFromProposedFederation();
         BtcTransaction svpFundTransactionUnsigned = createSvpFundTransaction(proposedFederation, spendableValueFromProposedFederation);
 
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1017,7 +1017,41 @@ private void logUpdateCollections(Transaction rskTx) {
         eventLogger.logUpdateCollections(sender);
     }
 
-    protected void processSVPFailure(Federation proposedFederation) {
+    protected void updateSvpState(Transaction rskTx) {
```
apancorb, line 1020: I would include an `INFO` log per scenario that plays out. This should help a lot with reading logs later on.

apancorb, line 1020: If this method is only needed for testing, you can leave it package-private and avoid using `protected`

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1017,7 +1017,41 @@ private void logUpdateCollections(Transaction rskTx) {
...
+        Optional<Federation> proposedFederationOpt = federationSupport.getProposedFederation();
+        if (proposedFederationOpt.isEmpty()) {
+            return;
+        }
+
+        Keccak256 rskTxHash = rskTx.getHash();
```
apancorb, line 1026: nit: maybe this local var is redundant, we can directly call `.getHash()` from the param var. Feel free to leave as is.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1017,7 +1017,41 @@ private void logUpdateCollections(Transaction rskTx) {
...
+            return;
+        }
+
+        // if the unsigned fund tx hash is not present, meaning we can proceed with the fund tx creation
+        Optional<Sha256Hash> svpFundTxHashUnsigned = provider.getSvpFundTxHashUnsigned();
+        if (svpFundTxHashUnsigned.isEmpty()) {
+            try {
+                processSvpFundTransactionUnsigned(rskTxHash, proposedFederation);
+            } catch (Exception e) {
+                logger.error("[updateSvpState] Error processing svp fund transaction unsigned.");
```
apancorb, line 1049:
```suggestion
        logger.error("[updateSvpState] Error processing unsigned SVP fund transaction: {}", e.getMessage(), e);
```

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1017,7 +1017,41 @@ private void logUpdateCollections(Transaction rskTx) {
...
+        if (!validationPeriodIsOngoing(proposedFederation)) {
+            processSVPFailure(proposedFederation);
+            return;
+        }
+
+        // if the fund tx signed is present, then the fund transaction change was registered,
+        // meaning we can create the spend tx.
+        Optional<BtcTransaction> svpFundTxSigned = provider.getSvpFundTxSigned();
+        if (svpFundTxSigned.isPresent()) {
+            processSvpSpendTransactionUnsigned(rskTxHash, proposedFederation, svpFundTxSigned.get());
```
apancorb, line 1039: Why is `processSvpSpendTransactionUnsigned` not also wrapped in a try-catch block? Can this never throw an exception?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1017,7 +1017,41 @@ private void logUpdateCollections(Transaction rskTx) {
...
+            return;
+        }
+
+        // if the unsigned fund tx hash is not present, meaning we can proceed with the fund tx creation
+        Optional<Sha256Hash> svpFundTxHashUnsigned = provider.getSvpFundTxHashUnsigned();
+        if (svpFundTxHashUnsigned.isEmpty()) {
+            try {
+                processSvpFundTransactionUnsigned(rskTxHash, proposedFederation);
```
apancorb, line 1047: Would it make more sense to have the try-catch with the logging statement be present inside the method itself?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1017,7 +1017,41 @@ private void logUpdateCollections(Transaction rskTx) {
...
+            } catch (Exception e) {
+                logger.error("[updateSvpState] Error processing svp fund transaction unsigned.");
+            }
+        }
```
apancorb, line 1051: I was thinking on how we can change the order of these two logic blocks get executed. I think it would be easier to understand if the logic related to processing `svpFundTxHashUnsigned` happens first than the `svpFundTxSigned`. Maybe we can do something like this:

```java
// Check if the unsigned fund transaction hash is not present, and create the fund transaction if needed
Optional<Sha256Hash> svpFundTxHashUnsigned = provider.getSvpFundTxHashUnsigned();
if (svpFundTxHashUnsigned.isEmpty()) {
    try {
        processSvpFundTransactionUnsigned(rskTxHash, proposedFederation);
    } catch (Exception e) {
       ...
    }
   return;
}

// Check if the signed fund transaction is present, and create the spend transaction if available
Optional<BtcTransaction> svpFundTxSigned = provider.getSvpFundTxSigned();
if (svpFundTxSigned.isPresent()) {
    processSvpSpendTransactionUnsigned(rskTxHash, proposedFederation, svpFundTxSigned.get());
}
```

rskj-core/src/test/java/co/rsk/peg/BridgeSupportSvpTest.java
```diff
@@ -79,6 +81,11 @@ public class BridgeSupportSvpTest {
     private Federation activeFederation;
     private Federation proposedFederation;
 
+    private Sha256Hash svpFundTransactionHashUnsigned;
+    private BtcTransaction svpFundTransaction;
+    private Sha256Hash svpSpendTransactionHashUnsigned;
+    private BtcTransaction svpSpendTransaction;
```
apancorb, line 87: Maybe you can have all these be default constants to be reused by the tests.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: no no, it will be called from `updateCollections`.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: it shouldn't :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: remember that when we save the `svpFundTxSigned` we remove the `svpFundTxHashUnsigned`. So this order won't have the expected behavior

rskj-core/src/test/java/co/rsk/peg/BridgeSupportSvpTest.java
julia-zack: have in mind that in some cases we are testing the tx creations , so we cannot previously define them.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: updated, added the error message to the log

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1017,13 +1017,50 @@ private void logUpdateCollections(Transaction rskTx) {
...
+
+        // if the unsigned fund tx hash is not present, meaning we can proceed with the fund tx creation
+        Optional<Sha256Hash> svpFundTxHashUnsigned = provider.getSvpFundTxHashUnsigned();
+        if (svpFundTxHashUnsigned.isEmpty()) {
+            logger.info(
+                "[updateSvpState] Fund tx hash unsigned was not found, so fund tx creation will be processed."
+            );
+            processSvpFundTransactionUnsigned(rskTxHash, proposedFederation);
+        }
+    }
```
apancorb, line 1058: What would happen if it reaches the end of this method? Will this be considered an error?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
apancorb: Yes, but in the meantime it can be package-private. No big deal, resolving this.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1017,13 +1017,50 @@ private void logUpdateCollections(Transaction rskTx) {
         eventLogger.logUpdateCollections(sender);
     }
 
-    protected void processSVPFailure(Federation proposedFederation) {
+    protected void updateSvpState(Transaction rskTx) {
+        Optional<Federation> proposedFederationOpt = federationSupport.getProposedFederation();
+        if (proposedFederationOpt.isEmpty()) {
```
apancorb, line 1022: Maybe add a log here just for good measure.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1034,28 +1071,35 @@ private void allowFederationElectionAgain() {
...
+        Coin spendableValueFromProposedFederation = bridgeConstants.getSpendableValueFromProposedFederation();
+        try {
+            BtcTransaction svpFundTransactionUnsigned = createSvpFundTransaction(proposedFederation, spendableValueFromProposedFederation);
+            provider.setSvpFundTxHashUnsigned(svpFundTransactionUnsigned.getHash());
+            PegoutsWaitingForConfirmations pegoutsWaitingForConfirmations = provider.getPegoutsWaitingForConfirmations();
+            settleReleaseRequest(pegoutsWaitingForConfirmations, svpFundTransactionUnsigned, rskTxHash, spendableValueFromProposedFederation);
+        } catch (InsufficientMoneyException e) {
+            logger.error(
+                "[processSvpFundTransactionUnsigned] Insufficient funds for creating the fund transaction. Error message: {}",
+                e.getMessage()
```
apancorb, line 1095: I recommend logging the entire Throwable not just the message. That way we can trace what happened. 
```suggestion
                "[processSvpFundTransactionUnsigned] Insufficient funds for creating the fund transaction",
                e
```

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1034,28 +1071,35 @@ private void allowFederationElectionAgain() {
...
+            );
+        } catch (IOException e) {
+            logger.error(
+                "[processSvpFundTransactionUnsigned] IOException getting the pegouts waiting for confirmations. Error message: {}",
+                e.getMessage()
```
apancorb, line 1100: same here

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: nope, probably we are just waiting for the fund tx change to be registered

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
nathanieliov: What if we add it at the `DEBUG` level? I think that could be fine at that level.

julia-zack: thats a good option, thanks! Will update

julia-zack: Updated

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1034,28 +1072,35 @@ private void allowFederationElectionAgain() {
...
-            throw new IllegalStateException(message);
-        }
-        Federation proposedFederation = proposedFederationOpt.get();
+    private boolean validationPeriodIsOngoing(Federation proposedFederation) {
+        long validationPeriodEndBlock = proposedFederation.getCreationBlockNumber() +
+            bridgeConstants.getFederationConstants().getValidationPeriodDurationInBlocks();
 
-        Coin spendableValueFromProposedFederation = bridgeConstants.getSpendableValueFromProposedFederation();
-        BtcTransaction svpFundTransactionUnsigned = createSvpFundTransaction(proposedFederation, spendableValueFromProposedFederation);
+        return rskExecutionBlock.getNumber() <= validationPeriodEndBlock;
```
nathanieliov, line 1083: Why is the block `validationPeriodEndBlock` still considered ongoing? Should not this logic be equal to the migration age? [There](https://github.com/rsksmart/rskj/blob/a13a0458f3db13928a5f15d4048503e7cb5f5c2b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java#L1266), we are not considering the end block an ongoing state.

nathanieliov: I would suggest going with the same logic as for migration to be consistent with the meaning of the end block of an ongoing process.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
apancorb: That seems strange, do you remember what it said?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: makes sense

julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: not really

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
apancorb: Should we log that scenario?

julia-zack: i don't think so. We won't be sure what the actual scenario is

apancorb (approved):
LGTM, nice job

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
apancorb: Just noting here, having the entire stacktrace enable us to know what exactly happened which will gives us a better insight to what occurred during runtime

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
nathanieliov: Nice!

nathanieliov (approved):
LGTM

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1017,13 +1017,51 @@ private void logUpdateCollections(Transaction rskTx) {
...
+            );
+            processSVPFailure(proposedFederation);
+            return;
+        }
+
+        Keccak256 rskTxHash = rskTx.getHash();
+        // if the fund tx signed is present, then the fund transaction change was registered,
+        // meaning we can create the spend tx.
+        Optional<BtcTransaction> svpFundTxSigned = provider.getSvpFundTxSigned();
+        if (svpFundTxSigned.isPresent()) {
```
marcos-iov, line 1043: Should we also be checking that there is no spend tx created? Just in case.

I know there should only be one at a time so if both are found perhaps we have a problem

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1017,13 +1017,51 @@ private void logUpdateCollections(Transaction rskTx) {
         eventLogger.logUpdateCollections(sender);
     }
 
-    protected void processSVPFailure(Federation proposedFederation) {
+    protected void updateSvpState(Transaction rskTx) {
```
marcos-iov, line 1020: To be honest, I'd find it easier to follow if the logic was written in chronological order. I know it requires more validations but we can probably make use of auxiliary methods. The way it's written now, for every new condition that appears in the code, you need to keep in mind the previous condition that was already validated a few lines above and the protocol logic about one having one SVP related value in storage at once.

I'd go with a very straight forward approach 

``` 
if (!proposedFederationExists) return;

if (validationPeriodIsOngoing) {
     if (shouldCreateSvpFundTx()) {
        createFundTx()
        return;
      }
     if (shouldCreateSvpSpendTx()) {
        createSpendTx()
        return;
      }

      validationFailed();
}
```
What do you think? I think that's the only 3 cases we need to cover

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1034,28 +1072,35 @@ private void allowFederationElectionAgain() {
...
 
-        Coin spendableValueFromProposedFederation = bridgeConstants.getSpendableValueFromProposedFederation();
-        BtcTransaction svpFundTransactionUnsigned = createSvpFundTransaction(proposedFederation, spendableValueFromProposedFederation);
+        return rskExecutionBlock.getNumber() < validationPeriodEndBlock;
+    }
 
-        provider.setSvpFundTxHashUnsigned(svpFundTransactionUnsigned.getHash());
-        PegoutsWaitingForConfirmations pegoutsWaitingForConfirmations = provider.getPegoutsWaitingForConfirmations();
-        settleReleaseRequest(pegoutsWaitingForConfirmations, svpFundTransactionUnsigned, rskTx.getHash(), spendableValueFromProposedFederation);
+    private void processSvpFundTransactionUnsigned(Keccak256 rskTxHash, Federation proposedFederation) {
```
marcos-iov, line 1086: Maybe call this `create` rather than `process`?

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
marcos-iov: I think you need to get rid if the `{}`
```
log("an error ocurred", e)
```
I think, not 100% sure

rskj-core/src/test/java/co/rsk/peg/BridgeSupportSvpTest.java
```diff
@@ -152,13 +173,13 @@ void setUp() {
...
+        void updateSvpState_whenSvpFundTxHashUnsigned_shouldLogValidationFailureAndClearValue() {
             // arrange
-            Sha256Hash svpFundTxHashUnsigned = BitcoinTestUtils.createHash(1);
-            bridgeStorageProvider.setSvpFundTxHashUnsigned(svpFundTxHashUnsigned);
+            svpFundTransactionHashUnsigned = BitcoinTestUtils.createHash(1);
+            bridgeStorageProvider.setSvpFundTxHashUnsigned(svpFundTransactionHashUnsigned);
 
             // act
-            bridgeSupport.processSVPFailure(proposedFederation);
+            bridgeSupport.updateSvpState(rskTx);
```
marcos-iov, line 182: Would it be possible to test by calling `updateCollections` directly? Or does that require a complex setup?

It would make the test super complete and `updateSvpState` could be private

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: makes sense, will update

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: i named it `process` to be consistent with `processPegoutOrMigration` method. But no big deal, I can change it :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: that was it. Thanks guys :)

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julia-zack: updated

PR #2840, opened 2024-11-07
Make updateSvpState private. Call it from updateCollections

julia-zack:
Call `updateSvpState` from `updateCollections`.
Make `updateSvpState` private and adapt tests.


rskj-core/src/test/java/co/rsk/peg/BridgeSupportSvpTest.java
```diff
@@ -129,6 +130,10 @@ void setUp() {
             logs
         );
 
+        ECKey key = RskTestUtils.getEcKeyFromSeed("key");
+        RskAddress address = new RskAddress(key.getAddress());
+        when(rskTx.getSender(any())).thenReturn(address); // to not throw when calling update collections
```
apancorb, line 135: Should we explicitly say what this param should be?

apancorb (approved):
LGTM

rskj-core/src/test/java/co/rsk/peg/BridgeSupportSvpTest.java
julia-zack: updated

PR #2842, opened 2024-11-07
Do not set last retired fed p2sh script in new commitFederation impl

julia-zack:
Last-retired-federation script shouldn't be set in new commitFederation implementation, since we will set the to-be-active-federation after its validation.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
@@ -702,8 +702,9 @@ public class FederationSupportImpl implements FederationSupport {
         clearPendingFederationVoting();
 
         if (activations.isActive(RSKIP186)) {
-            setNewActiveFederationCreationBlockHeight();
-            preserveLastRetiredFederationScript();
+            // since we are creating the to-be-active-fed in this block,
+            // its creation block height is this block number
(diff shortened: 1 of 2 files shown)


jeremy-then (approved):
LGTM.

apancorb (approved):
LGTM

nathanieliov (approved):

rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java
```diff
@@ -733,12 +734,11 @@ private FederationChangeResponseCode commitPendingFederation(PendingFederation c
         // set proposed federation
         Federation proposedFederation = buildFederationFromPendingFederation(currentPendingFederation);
         provider.setProposedFederation(proposedFederation);
+        setNewActiveFederationCreationBlockHeight();
```
marcos-iov, line 737: Should this be done at this point? Shouldn't we set this value after the proposed federation is validated? If the validation fails we'd end up with an invalid value here, right?

julia-zack: we discussed this when developing phase1, the idea was that the activation period remain the same. But yes it makes sense to do it after the validation.

julia-zack: updated

marcos-iov (approved):

PR #2844, opened 2024-11-11
fmacleal/integrate tstorage in transaction execution

fmacleal:
## Description

In this MR, 

We are adding the TLOAD and TSTORE opcodes in the transaction execution context, validating that the transient storage it's using the memory context correctly between different contract calls. More specifically, if we use `delegatecall`, `callcode`, `call` and `staticcall` the values stored between these different contracts subcalls are storing and loading the values as defined by the [EIP-1153 ](https://eips.ethereum.org/EIPS/eip-1153#specification). This way, we will ensure that no contract it's accessing the memory from the other regarding the context call used.

Besides this, I have addressed some sonar issues regarding the use of switchcase in the opcodes classes.

## Types of changes

## Checklist:

* **Other information**:


asoto-iov (approved):

rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -1361,6 +1361,10 @@ protected void doTLOAD(){
     protected void doTSTORE(){
         //TODO: Gas cost calculation will be done here and also shared contexts verifications for
         // different types of calls
+        if (program.isStaticCall() && program.getActivations().isActive(RSKIP91)) {
```
Vovchyk, line 1364: do you think we should still check for `RSKIP91`? I guess it was only needed when `STATICCALL` opcode was introduced

fmacleal: Yes, makes sense. At this height, staticcall would be for sure enabled. I will remove the verification of it.

Vovchyk (approved):
lgtm

PR #2846, opened 2024-11-14
Handling transaction revert in eth_estimateGas

rmoreliovlabs:
## Description
This PR refactors the `eth_estimateGas` RPC call on RSK to handle reverted transactions similarly to Geth. Previously, the `eth_estimateGas` call on RSK did not return an error when a transaction was reverted. This change adds a missing check to ensure that when a transaction is reverted, the `eth_estimateGas` call returns an error, aligning with Geth’s behavior. Check this documentation for more info: https://www.notion.so/iovlabs/Analysis-of-eth_estimateGas-Behavior-on-RSKJ-Node-for-Reverted-Transactions-13cc132873f980c0aa87de78ed0f84a0

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
@@ -146,7 +146,7 @@ public class EthModule
             } else {
                 res = callConstant(args, block);
             }
-            handleTransactionRevert(res);
+            handleTransactionRevertIfHappens(res);
             hReturn = HexUtils.toUnformattedJsonHex(res.getHReturn());
 
             return hReturn;
@@ -181,7 +181,7 @@ public class EthModule
             );
 
             ProgramResult res = executor.getResult();
-            handleTransactionRevert(res);
+            handleTransactionRevertIfHappens(res);
 
             estimation = internalEstimateGas(executor.getResult());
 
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -357,4 +347,22 @@ private ProgramResult callConstantWithState(CallArguments args, Block executionB
                 hexArgs.getFromAddress()
         );
     }
+
+    private void handleTransactionRevert(ProgramResult res) {
```
fmacleal, line 351: nitpic:

Renaming to something that doesn't mean that it will always be handled. Something like `handleTransactionRevertIfHappens`.

fmacleal (commented):
Good job! It was indeed a tricky issue to find. My suggestion would be to add a test to validate this revert. You could use the contract you created to debug. If you add it, it would be awesome. 🙂

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
rmoreliovlabs: thanks for the feedback! renamed the method.

fmacleal (approved):
LGTM. Good job.

Vovchyk (approved):

PR #2903, opened 2024-12-23
Remove validationPeriodIsOngoing method

julia-zack:
`validationPeriodIsOngoing` and `isSvpOngoing` were very similar, so no need to have them both.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -1029,7 +1029,7 @@ public class BridgeSupport {
         // if the proposed federation exists and the validation period ended,
         // we can conclude that the svp failed
         Federation proposedFederation = proposedFederationOpt.get();
-        if (!validationPeriodIsOngoing(proposedFederation)) {
+        if (!isSvpOngoing()) {
             processSvpFailure(proposedFederation);
             return;
         }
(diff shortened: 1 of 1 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1079,13 +1079,12 @@ private void allowFederationElectionAgain() {
     }
 
     private boolean isSvpOngoing() {
```
apancorb, line 1081: If you want to make it more functional:

```java
private boolean isSvpOngoing() {
    return federationSupport
            .getProposedFederation()
            .map(proposedFederation -> 
                rskExecutionBlock.getNumber() <
                proposedFederation.getCreationBlockNumber() +
                bridgeConstants.getFederationConstants().getValidationPeriodDurationInBlocks()
            )
            .orElse(false);
}
```

apancorb (approved):

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
nathanieliov: Nice!

julia-zack: very nice! Updated

marcos-iov (approved):

PR #2911, opened 2025-01-02
Replacing shouldSave federation by a tracker class

julianlen:
Created a new class FederationTracker that tracks whether the federation must be saved (if it was modified)


julianlen:
An alternative is to create a new Tracker each time FederationStorageProvider fetches a new (old or pending) federation. If the federation should be saved, FederationStorageProvider may call a special constructor that sets the flag "modified" (previously shouldBeSaved).

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
```diff
@@ -0,0 +1,25 @@
+package co.rsk.peg.federation;
+
+public abstract class FederationTracker<T> {
+    private T federation;
+    private boolean modified = false;
+    public boolean isModified() {
```
julia-zack, line 6:
```suggestion
    private boolean modified = false;
    
    public boolean isModified() {
```

julia-zack: just an extra blank line to improve readability :D

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
```diff
@@ -0,0 +1,25 @@
...
+        return this.modified;
+    }
+
+    public T getFederation() {
+        return this.federation;
+    }
+    public boolean isFederationPresent() {
```
julia-zack, line 13:
```suggestion
    }
    
    public boolean isFederationPresent() {
```

same

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
```diff
@@ -0,0 +1,25 @@
...
+        return this.federation != null;
+    }
+
+    public void setFederationWithTracking(T aFederation) {
```
julia-zack, line 17:
```suggestion
    public void set(T aFederation) {
```

i find the other name a bit redundant, wdyt?

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -166,24 +159,24 @@ public Federation getOldFederation(FederationConstants federationConstants, Acti
             }
         );
 
+        oldFederationTracker.replaceFederation(oldFederation);
         return oldFederation;
```
julia-zack, line 163:
```suggestion
        oldFederationTracker.replaceFederation(oldFederation);
        return oldFederationTracker.getFederation();
```

?

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
```diff
@@ -0,0 +1,25 @@
...
+        this.federation = aFederation;
+        this.modified = true;
+    }
+
+    public void replaceFederation(T aFederation) {
+        this.federation = aFederation;
+    }
```
julia-zack, line 24: shouldn't we set `modified = true` here too?

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
julia-zack, line 1: maybe we can avoid the _federation_ word in the method's names, wdyt? We already know is a _Federation_ Tracker

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -197,13 +190,13 @@ public PendingFederation getPendingFederation() {
                 }
             );
 
+        pendingFederationTracker.replaceFederation(pendingFederation);
         return pendingFederation;
```
julia-zack, line 194:
```suggestion
        pendingFederationTracker.replaceFederation(pendingFederation);
        return pendingFederationTracker.get();
```

?

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
julianlen: I wanted to clarify somehow that "modified" is being set, and differentiate it from "replace"

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
julianlen: This replaces the case where the variable `this.oldFederation`  is overridden with the oldFederation obtained from the storage, the variable "shouldBeSaved" remained false. This situation happens when the oldFederation is absent or the oldFederation hasn't been modified so far

julia-zack: ooookey makes sense

julia-zack: thanks :)

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
julia-zack: `setNew` ? wdyt?

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
jeremy-then: I agree.

jeremy-then (dismissed):
LGTM.

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -146,13 +139,13 @@ public void setNewFederation(Federation federation) {
 
     @Override
     public Federation getOldFederation(FederationConstants federationConstants, ActivationConfig.ForBlock activations) {
-        if (oldFederation != null || shouldSaveOldFederation) {
-            return oldFederation;
+        if (oldFederationTracker.isPresent() || oldFederationTracker.isModified()) {
```
nathanieliov, line 142: Isn't the second condition here useless? What would be the case where `oldFederationTracker` is empty but `isModified`?  It would return null, is it what we expect?

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
```diff
@@ -23,21 +23,14 @@
...
     private List<UTXO> oldFederationBtcUTXOs;
     private Federation newFederation;
-    private Federation oldFederation;
-    private boolean shouldSaveOldFederation = false;
-
-    private PendingFederation pendingFederation;
-    private boolean shouldSavePendingFederation = false;
-
+    private final FederationTracker<Federation> oldFederationTracker = new FederationTrackerImpl();
+    private final FederationTracker<PendingFederation> pendingFederationTracker = new PendingFederationTrackerImpl();
```
apancorb, line 30: Maybe move them above to the top next with the other final instance vars

rskj-core/src/main/java/co/rsk/peg/federation/FederationTrackerImpl.java
```diff
@@ -0,0 +1,4 @@
+package co.rsk.peg.federation;
+
+public class FederationTrackerImpl extends FederationTracker<Federation> {
```
apancorb, line 3: There is no need for these two impl classes. You can leave `FederationTracker` as a normal class with the generics as it is.

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
```diff
@@ -0,0 +1,27 @@
+package co.rsk.peg.federation;
+
+public abstract class FederationTracker<T> {
```
apancorb, line 3: Not sure if we could use an observer pattern here. Would it be possible as things are now? Would it be too much work? This way we avoid using the `modified` flag.

rskj-core/src/main/java/co/rsk/peg/federation/FederationStorageProviderImpl.java
julianlen: It is, not clear the background but it is the expected behavior. If the method 'setFederation' is called passing null as a federation, it should keep the flag "isModified" as it was

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
julianlen: done!

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
```diff
@@ -0,0 +1,27 @@
+package co.rsk.peg.federation;
+
+public class FederationTracker<T> {
```
apancorb, line 3: We could have `T extends Federation` just to make this more explicit. We will need to have PendingFederation extend Federation. Maybe we can look more into this.

rskj-core/src/main/java/co/rsk/peg/federation/FederationTracker.java
apancorb: I think before tacking this we need to understand the modified flag behavior. Is it only here as an optimization 	that prevents redundant repository calls when the state is accessed multiple times? Is it due to high gas costs? Can we just do a simple read-write strategy? Then we can see how to approach this, but ideally we do not want to add modified flags or any unnecessary code.

julianlen: agree

nathanieliov: Very interesting discussion. Agree with @apancorb

jeremy-then (approved):
LGTM.

PR #2922, opened 2025-01-09
Adding min and max ranges to check the value provided by the Oracle's

asoto-iov:
## Description
This pull request introduces a configurable valid range (minValidPrice and maxValidPrice) for the BTC price retrieved from the oracle. The system will validate the price against this range before updating it. If the price is outside the range, it will not be updated, and a warning will be logged.

 ## Motivation and Context
The motivation behind this change is to improve the robustness of the system by preventing updates with incorrect or extreme BTC prices. By enforcing a valid range, we can ensure that only reliable and expected values are used, minimizing potential risks and maintaining system integrity.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/config/mining/StableMinGasPriceSystemConfig.java
fmacleal, line 1: issue:

There is a test class for this one [`StableMinGasPriceSystemConfigTest`](https://github.com/rsksmart/rskj/blob/1bf9d53a883d254c46e117804e32c874337890ca/rskj-core/src/test/java/co/rsk/config/mining/StableMinGasPriceSystemConfigTest.java#L31). Maybe would be good also add validations for these new properties there? I see that the json in this class isn't updated.

rskj-core/src/main/java/co/rsk/mine/gas/provider/StableMinGasPriceProvider.java
```diff
@@ -148,4 +151,13 @@ private Optional<Long> fetchPriceSync() {
 
         return Optional.empty();
     }
+
+    @VisibleForTesting
+    boolean isIntoRange(long value) {
```
fmacleal, line 156: nitpick:

This is more a personal opinion, but I am not used in see `isInto` for these situations. I vouch to rename it maybe with a better explanation like `valueBetweenConfiguredRange`. Because to let it clear that it's a configuration.

fmacleal (changes requested):
Good job! I have a few comments though, let me know what you think.

fmacleal (approved):
LGTM

Thanks for have followed the suggestions. 🙂

rmoreliovlabs (approved):

PR #2935, opened 2025-01-16
Create Federation Change Integration Test With Migration Phase

apancorb:
## Description
This new integration test will validate that when a federation changes, the UTXOs will be moved and migrated as expected after Lovell is activated. 

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -161,9 +186,54 @@ void whenAllActivationsArePresentAndFederationChanges_shouldCreateCommitAndActiv
...
+
+    @Test
+    void whenAllActivationsArePresentAndFederationChanges_shouldMigrateAllFunds() throws Exception {
+        // Arrange
+    
+        var activations = ActivationConfigsForTest.all().forBlock(0);
+        setUpFederationChange(activations);
+        // Create a default original federation using the list of UTXOs
+        var originalFederation = createOriginalFederation(
+            FederationType.P2SH_ERP, ORIGINAL_FEDERATION_KEYS, activations);
```
nathanieliov, line 203: The most realistic would be to use `PedingFederation.buildFederation` method and let it choose the federation type based on the activations as the logic of the method does.

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -161,9 +186,54 @@ void whenAllActivationsArePresentAndFederationChanges_shouldCreateCommitAndActiv
...
+        // Move blockchain until the migration phase
+        activateMigration(activations);
+        // Migrate
+        migrateUTXOs();
+
+        assertNewAndOldFederationsHaveExpectedAddress(
+            newFederation.getAddress(), originalFederation.getAddress());
+        assertMigrationHasStarted();
+        verifySigHashes(activations);
+        verifyPegoutTransactionCreatedEventWasEmitted(activations);
```
nathanieliov, line 228: Let's add an initial check checking the state before the migration so you can verify that before the migration, there was no pegout request, waiting for confirmation or waiting for signature.

Wdyt?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -414,8 +552,130 @@ private void assertNewAndOldFederationsHaveExpectedAddress(
...
+    private void verifyPegouts(ActivationConfig.ForBlock activations) throws Exception {
+        var activeFederation = federationStorageProvider.getNewFederation(
+            BRIDGE_CONSTANTS.getFederationConstants(), activations);
+        var retiringFederation = federationStorageProvider.getOldFederation(
+            BRIDGE_CONSTANTS.getFederationConstants(), activations);
+
+        for (PegoutsWaitingForConfirmations.Entry pegoutEntry : bridgeStorageProvider.getPegoutsWaitingForConfirmations().getEntries()) {
+            var pegoutBtcTransaction = pegoutEntry.getBtcTransaction();
+            for (TransactionInput input : pegoutBtcTransaction.getInputs()) {
+                // Each input should contain the right scriptsig
```
nathanieliov, line 627:
```suggestion
                // Each input should contain the right scriptSig
```

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
julia-zack: thats a very good point

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -414,8 +552,130 @@ private void assertNewAndOldFederationsHaveExpectedAddress(
...
+        }
+    }
+    
+    private void verifyPegouts(ActivationConfig.ForBlock activations) throws Exception {
+        var activeFederation = federationStorageProvider.getNewFederation(
+            BRIDGE_CONSTANTS.getFederationConstants(), activations);
+        var retiringFederation = federationStorageProvider.getOldFederation(
+            BRIDGE_CONSTANTS.getFederationConstants(), activations);
+
+        for (PegoutsWaitingForConfirmations.Entry pegoutEntry : bridgeStorageProvider.getPegoutsWaitingForConfirmations().getEntries()) {
```
nathanieliov, line 624: Not sure about this. I think we should assert against constant values. I mean, for example, based on the initial conditions of the test, we should know the number of migration pegouts that should be created and assert that that number of pegouts is precisely what gets created. Then, check each one was created as expected. Wdyt?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -161,9 +186,54 @@ void whenAllActivationsArePresentAndFederationChanges_shouldCreateCommitAndActiv
...
+            newFederation.getAddress(), originalFederation.getAddress());
+        assertMigrationHasStarted();
+        verifySigHashes(activations);
+        verifyPegoutTransactionCreatedEventWasEmitted(activations);
+        verifyPegouts(activations);
+        
+        // Move blockchain until the end of the migration phase
+        endMigration(activations);
+
+        assertMigrationHasEnded(newFederation);
```
nathanieliov, line 234: Shouldn't you check also that  the utxos were migrated as expected using `assertUTXOsReferenceMovedFromNewToOldFederation`?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
julia-zack: but the original federation is the current federation, not the new one

nathanieliov: I know, but the `buildFederation` already has the logic to build the proper federation type according to the activations, so you can create the current federation from it if you want to. The idea is to only need to rely on the activations to create the correct current federation.

julia-zack: i think is kind of confusing since the new federation is built from that method

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
apancorb: `assertMigrationHasNotStarted()` verifies that there is no pegouts in the queue. I believe it already does this.

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
apancorb: The UTXOs for the original federation are randomly generated. So the size may vary per test. It was already done in the legacy tests here: https://github.com/rsksmart/rskj/blob/feature/powpeg_validation_protocol-phase4/rskj-core/src/test/java/co/rsk/peg/federation/PowpegMigrationTest.java

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
apancorb: Yes, let me add that

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -414,8 +556,130 @@ private void assertNewAndOldFederationsHaveExpectedAddress(
...
+        assertFalse(bridgeStorageProvider.getPegoutsWaitingForConfirmations().getEntries().isEmpty());
+    }
+
+    private void assertMigrationHasEnded(Federation newFederation) {
+        // New active federation still there, retiring federation no longer there
+        assertEquals(newFederation, bridgeSupport.getActiveFederation());
+        assertNull(bridgeSupport.getRetiringFederationAddress());
+    }
+    
+    private void assertLastRetiredFederationP2SHMatchesWithOriginalFederation(FederationType federationType, Federation originalFederation, ActivationConfig.ForBlock activations) {
```
julia-zack, line 575:
```suggestion
    private void assertLastRetiredFederationP2SHScriptMatchesWithOriginalFederationOne(FederationType federationType, Federation originalFederation, ActivationConfig.ForBlock activations) {
```
wdyt?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -414,8 +556,130 @@ private void assertNewAndOldFederationsHaveExpectedAddress(
         assertEquals(expectedOldFederationAddress, bridgeSupport.getRetiringFederationAddress());
     }
     
-    private void assertFundsWereNotMigrated() throws Exception {
+    private void assertMigrationHasNotStarted() throws Exception {
```
nathanieliov, line 559: I think this method is missing some assertion to match what the method's name means about what it is doing. I think, to assert the migration has not started, it should check:

- Current block is behind `fedActivationAge + fundsMigrationAgeBegin`
- No existing retiring fed utxos has been transferred to the new fed. For this, it should check retiring fed utxos still the same, and there is no active fed utxo.

Wdyt?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
nathanieliov: We've been removing/getting away from random behavior from tests. I think, we should discuss if we should keep this being random or make it deterministic.

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -161,9 +186,58 @@ void whenAllActivationsArePresentAndFederationChanges_shouldCreateCommitAndActiv
...
+        // Since Lovell is activated we will commit the proposed federation
+        commitProposedFederation(activations);
+        // Move blockchain until the activation phase
+        activateNewFederation(activations);
+
+        assertUTXOsReferenceMovedFromNewToOldFederation(
+            originalUTXOs, activations);
+        assertNewAndOldFederationsHaveExpectedAddress(
+            newFederation.getAddress(), originalFederation.getAddress());
+        assertMigrationHasNotStarted();
```
julia-zack, line 221: all this part is basically the first test right? maybe we should keep just this one, that tests the whole process. Wdyt?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -318,17 +393,58 @@ private void commitProposedFederation(ActivationConfig.ForBlock activations) {
...
+        var updateCollectionsTx = buildUpdateCollectionsTx();
+        bridgeSupport.updateCollections(updateCollectionsTx);
+        bridgeSupport.save();
+    }
+
+    private void migrateUTXOs() throws Exception {
+        // Migrate while there are still utxos to migrate
+        var remainingUTXOs = federationStorageProvider.getOldFederationBtcUTXOs();
+        while (!remainingUTXOs.isEmpty()) {
+            var updateCollectionsTx = buildUpdateCollectionsTx();
```
julia-zack, line 442:
```suggestion
        // Migrate while there are still utxos to migrate
        var remainingUTXOs = federationStorageProvider.getOldFederationBtcUTXOs();
        var updateCollectionsTx = buildUpdateCollectionsTx();
        while (!remainingUTXOs.isEmpty()) {
```

it always created the same tx right?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -414,8 +556,130 @@ private void assertNewAndOldFederationsHaveExpectedAddress(
...
+
+    private void verifyPegoutTxSigHashIndex(BtcTransaction pegoutTx, ActivationConfig.ForBlock activations) {
+        var lastPegoutSigHash = BitcoinUtils.getFirstInputSigHash(pegoutTx);
+        assertTrue(lastPegoutSigHash.isPresent());
+
+        if (activations.isActive(ConsensusRule.RSKIP379)) {
+            assertTrue(bridgeStorageProvider.hasPegoutTxSigHash(lastPegoutSigHash.get()));
+        } else {
+            assertFalse(bridgeStorageProvider.hasPegoutTxSigHash(lastPegoutSigHash.get()));
+        }
```
julia-zack, line 677: i'd try to avoid booleans in tests. I think its better to have both `assertPegoutTxSigHashesAreSaved` and `assertPegoutTxSigHashesAreNotSaved`, wdyt?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -161,9 +186,58 @@ void whenAllActivationsArePresentAndFederationChanges_shouldCreateCommitAndActiv
...
+        activateMigration(activations);
+        // Migrate
+        migrateUTXOs();
+
+        assertNewAndOldFederationsHaveExpectedAddress(
+            newFederation.getAddress(), originalFederation.getAddress());
+        assertMigrationHasStarted();
+        verifySigHashes(activations);
+        verifyPegoutTransactionCreatedEventWasEmitted(activations);
+        verifyPegouts(activations);
```
julia-zack, line 233: wdyt about having the calls for these three methods inside the `assertMigrationHasStarted`?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
apancorb: Okay this idea makes sense and I agree. I will create essentially a test as you describe that tests the entire federation change using the current activations.

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
apancorb: Let us keep it as is since this will be used independently for different tests to come. In the next PR I will migrate all the legacy tests and it will make sense there.

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
apancorb: Nice catch , thanks!

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
apancorb: Done

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
apancorb: I agree, in this case it makes sense

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
apancorb: I have added the first bullet point. I believe the second is already tested in `assertUTXOsReferenceMovedFromNewToOldFederation`. Is that correct?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -388,10 +474,36 @@ private List<UTXO> createRandomUTXOs(Address owner) {
...
+            federation.getRedeemScript();
+    }
+   
+    private static Script getFederationDefaultP2SHScript(Federation federation) {
+        return federation instanceof ErpFederation ?
+            ((ErpFederation) federation).getDefaultP2SHScript() :
+            federation.getP2SHScript();
+    }
+
+    private Transaction buildUpdateCollectionsTx() {
```
julia-zack, line 490: would it make sense to have this a global static variable? Is always building the same tx

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -414,8 +526,138 @@ private void assertNewAndOldFederationsHaveExpectedAddress(
...
         // Pegouts waiting for confirmations should be empty
         assertTrue(bridgeStorageProvider.getPegoutsWaitingForConfirmations().getEntries().isEmpty());
     }
+     
+    private void assertMigrationHasStarted() throws Exception {
+        // Pegouts waiting for confirmations should not be empty
+        assertFalse(bridgeStorageProvider.getPegoutsWaitingForConfirmations().getEntries().isEmpty());
+    }
+
+    private void assertMigrationHasEnded(Federation newFederation) {
```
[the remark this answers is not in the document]
apancorb: For consistency sake, let us keep the naming as it is.

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
apancorb: Yes we can add this detail.

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -126,49 +148,72 @@ private enum FederationType {
...
         commitProposedFederation(activations);
+        // Next federation creation block height should be as expected
+        assertLastRetiredFederationP2SHScriptMatchesWithOriginalFederation(
+            FederationType.P2SH_ERP, originalFederation, activations);
         // Move blockchain until the activation phase
         activateNewFederation(activations);
-
-        // Assert
+   
         assertUTXOsReferenceMovedFromNewToOldFederation(originalUTXOs, activations);
```
julia-zack, line 191: this should be being checked right after committing the proposed fed

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -126,49 +148,72 @@ private enum FederationType {
...
-        assertNewAndOldFederationsHaveExpectedAddress(newFederation.getAddress(), originalFederation.getAddress());
-        assertFundsWereNotMigrated();
+        assertNewAndOldFederationsHaveExpectedAddress(
```
julia-zack, line 192: same here

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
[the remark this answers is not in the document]
julia-zack: well, I think is confusing that the `assertMigrationHasStarted` checks for the migration transaction and the `assertMigrationHasEnded` checks for the live federation

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -126,49 +148,72 @@ private enum FederationType {
...
     private BridgeSupport bridgeSupport;
 
     @Test
-    void whenAllActivationsArePresentAndFederationChanges_shouldCreateCommitAndActiveNewFed() throws Exception {
+    void whenAllActivationsArePresentAndFederationChanges_shouldSuccesfullyChangeFederation() throws Exception {
         // Arrange
+   
         var activations = ActivationConfigsForTest.all().forBlock(0);
-        setUpFederationChange(activations);
+        setUp(activations);
```
julia-zack, line 171: you are receiving the `activations` in a lot of methods. Why don't set it as a global variable?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -126,49 +148,72 @@ private enum FederationType {
...
-        // Act
+        // Act & Assert
+   
         // Create pending federation using the new federation keys
         var newFederation = createPendingFederation(NEW_FEDERATION_KEYS, activations); 
         commitPendingFederation();
         // Since Lovell is activated we will commit the proposed federation
         commitProposedFederation(activations);
+        // Next federation creation block height should be as expected
+        assertLastRetiredFederationP2SHScriptMatchesWithOriginalFederation(
```
julia-zack, line 186: missing the `nextFederationCreationBlockHeight` assertion

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -414,8 +527,138 @@ private void assertNewAndOldFederationsHaveExpectedAddress(
...
+
+                Optional<Federation> spendingFederationOptional = Optional.empty();
+                if (inputStandardRedeemScript.equals(getFederationDefaultRedeemScript(activeFederation))) {
+                    spendingFederationOptional = Optional.of(activeFederation);
+                } else if (retiringFederation != null &&
+                    inputStandardRedeemScript.equals(getFederationDefaultRedeemScript(retiringFederation))) {
+                    spendingFederationOptional = Optional.of(retiringFederation);
+                } else {
+                    fail("Pegout scriptsig does not match any Federation");
+                }
```
julia-zack, line 628:
```suggestion
                Optional<Federation> spendingFederation = getSpendingFederation();
                assertTrue(spendingFederation.isPresent())
                
                private Optional<Federation> getSpendingFederation() {
                      if (inputStandardRedeemScript.equals(getFederationDefaultRedeemScript(activeFederation))) {
                               return activeFederation;
                      }
                      
                       if (retiringFederation != null &&
                    inputStandardRedeemScript.equals(getFederationDefaultRedeemScript(retiringFederation))) {
                               return retiringFederation;
                    }
                }
```

something like that, wdyt?

julia-zack: btw, are we gonna have tests with pegouts from the retiring fed?

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
```diff
@@ -126,49 +148,72 @@ private enum FederationType {
             ECKey.fromPublicOnly(Hex.decode("02be5d357d62be7b2d42de0343d1297129a0a8b5f6b8bb8c46eefc9504db7b56e1")),
             ECKey.fromPublicOnly(Hex.decode("032706b02f64b38b4ef7c75875aaf65de868c4aa0d2d042f724e16924fa13ffa6c"))
         ));
+    private static final Transaction UPDATE_COLLECTIONS = buildUpdateCollectionsTx();
```
julia-zack, line 151:
```suggestion
    private static final Transaction UPDATE_COLLECTIONS_TX = buildUpdateCollectionsTx();
```

a nit one

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
apancorb: Will update

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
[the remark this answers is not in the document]
apancorb: Alright, will change

rskj-core/src/test/java/co/rsk/peg/federation/FederationChangeIT.java
julia-zack: we could set them in each test, but nevermind :) sounds good

PR #2972, opened 2025-01-30
Return correct output script in ErpFederation when fed type is segwit compatible

julia-zack:


rskj-core/src/test/java/co/rsk/peg/federation/P2shP2wshErpFederationTest.java
julia-zack, line 1: This test copies the [P2shErpFederationTest](https://github.com/rsksmart/rskj/blob/master/rskj-core/src/test/java/co/rsk/peg/federation/P2shErpFederationTest.java)

marcos-iov (approved):

PR #2979, opened 2025-02-05
Handle exception when removing signatures from a tx with non p2sh-legacy-multisig inputs

julia-zack:
`fed:LOVELL-7.0.0-rc`


marcos-iov (approved):

josedahlquist (approved):

jeremy-then (approved):
LGTM.

PR #2984, opened 2025-02-06
Fixing validation of tx size according the max size allowed for a tx

fmacleal:
## Description

Before, we were allowing a transaction length bigger than the max allowed. 
This commit fixes it and add a test to validate.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/test/java/co/rsk/net/messages/MessageTest.java
```diff
@@ -307,23 +314,50 @@ void encodeDecodeTransactionsMessage() {
...
             Transaction tx1 = txs.get(k);
             Transaction tx2 = newmessage.getTransactions().get(k);
 
-            Assertions.assertEquals(tx1.getHash(), tx2.getHash());
+            assertEquals(tx1.getHash(), tx2.getHash());
         }
     }
 
+    @Test
+    void encodeDecodeTransactionsMessageWithHigherSize() throws IOException {
```
nagarev, line 335: I would suggest a name like this: `encodeDecodeTransactionMessage_withSizeBiggerThanMaximum_returnsEmptyTransactions` or something like that. I know this class has really strange names, but maybe we can start using something similar to this for new tests. WDYT?

nagarev (dismissed):
Really loved the refactor! beautiful new test! I only left a minor comment for you to consider and decide.

Vovchyk (approved):

rskj-core/src/test/java/co/rsk/net/messages/MessageTest.java
fmacleal:
```suggestion
    void encodeDecodeTransactionMessage_withSizeBiggerThanMaximum_returnsEmptyTransactions() throws IOException {
```

Vovchyk (approved):

PR #3013, opened 2025-02-28
Rollback the custom regtest federation private keys to the default ones

julianlen:
BridgeRegTestConstants is now a singleton. BridgeRegTestConstants stores the instance of FederationRegTestConstants in a global variable. Since all the tests share this variable, the test will fail if they try to use different Regtest federation private keys.

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
@@ -3,6 +3,7 @@ package co.rsk.peg;
 import static co.rsk.peg.BridgeSupportTestUtil.createRepository;
 import static co.rsk.peg.PegTestUtils.createBaseInputScriptThatSpendsFromTheFederation;
 import static co.rsk.peg.PegTestUtils.createHash3;
+import static co.rsk.peg.federation.FederationTestUtils.REGTEST_FEDERATION_PRIVATE_KEYS;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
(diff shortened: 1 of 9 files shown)


nathanieliov (approved):

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -57,11 +56,14 @@
 
 class BridgeSupportAddSignatureTest {
     private static final RskAddress bridgeAddress = PrecompiledContracts.BRIDGE_ADDR;
-    private static final List<BtcECKey> federationECKeys = BitcoinTestUtils.getBtcEcKeysFromSeeds(
-        new String[]{"fa91", "fa92", "fa93"}, true
+    private static final List<BtcECKey> REGTEST_FEDERATION_PRIVATE_KEYS = Arrays.asList(
+        BtcECKey.fromPrivate(Hex.decode("45c5b07fc1a6f58892615b7c31dca6c96db58c4bbc538a6b8a22999aaa860c32")),
```
nathanieliov, line 60: You should get these private keys from the `FederationTestUtils`

julianlen: Yep, and found some others tests where I didn't delete the constant

nathanieliov (approved):

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
```diff
@@ -57,11 +57,9 @@
 
 class BridgeSupportAddSignatureTest {
     private static final RskAddress bridgeAddress = PrecompiledContracts.BRIDGE_ADDR;
-    private static final List<BtcECKey> federationECKeys = BitcoinTestUtils.getBtcEcKeysFromSeeds(
-        new String[]{"fa91", "fa92", "fa93"}, true
-    );
+    private static final List<BtcECKey> REGTEST_FEDERATION_PUBLIC_KEYS = REGTEST_FEDERATION_PRIVATE_KEYS.stream().map(BtcECKey::getPubKey).map(BtcECKey::fromPublicOnly).toList();
```
julia-zack, line 60: would it make sense to have this in `FederationTestUtils` too?

julia-zack (approved):
left a comment but lgtm :)

rskj-core/src/test/java/co/rsk/peg/BridgeSupportAddSignatureTest.java
julianlen: Even when used only in one place?

julia-zack: nope, in that case it does not make sense :)

marcos-iov: Can't you just call BridgeRegTestConstants.getGenesisFederationPublicKeys?

marcos-iov (approved):

PR #3037, opened 2025-03-17
Add release outpoints values structure to bridge storage provider

julia-zack:
**Background**
To be able to sign segwit-compatible transactions, we need to save the releases outpoints values in the storage.
New entries will be added at runtime and added to the existing ones in storage.

To avoid consuming too much space in storage, each entry will have its own storage key, and its outpoints as the saved value.

_Note: we shouldn't be allowed to call any of these methods pre RSKIP305_

**Requirements**
-Create **set** and **get** methods to set and get specific `releaseTxHash` outpoints values.
-Create **save** method that adds new values to the storage.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
@@ -261,7 +261,7 @@ public class BridgeStorageProvider {
             return;
         }
 
-        if (releaseTxHash == null || outpointsValues == null) {
+        if (releaseTxHash == null || outpointsValues == null || outpointsValues.isEmpty()) {
             throw new IllegalArgumentException(
                 String.format("Invalid release outpoints values entry, has hash %s and coins list %s", releaseTxHash, outpointsValues)
             );
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
@@ -19,6 +19,7 @@
 package co.rsk.peg;
 
 import static co.rsk.peg.BridgeStorageIndexKey.*;
+import static java.util.Objects.isNull;
 import static org.ethereum.config.blockchain.upgrades.ConsensusRule.*;
 
 import co.rsk.bitcoinj.core.*;
@@ -63,6 +64,7 @@ public class BridgeStorageProvider {
     private ReleaseRequestQueue releaseRequestQueue;
     private PegoutsWaitingForConfirmations pegoutsWaitingForConfirmations;
     private SortedMap<Keccak256, BtcTransaction> pegoutsWaitingForSignatures;
+    private SortedMap<Sha256Hash, List<Coin>> releasesOutpointsValues;
 
     private HashMap<Sha256Hash, Long> btcTxHashesToSave;
 
@@ -239,6 +241,69 @@ public class BridgeStorageProvider {
         }
     }
 
+    public Optional<List<Coin>> getReleaseOutpointsValues(Sha256Hash releaseTxHash) {
+        if (!activations.isActive(RSKIP305)) {
+            throw new IllegalStateException("Can't call this method before RSKIP305 activation.");
+        }
+
+        return Optional.ofNullable(releasesOutpointsValues)
+            // first search in cache
+            .map(map -> map.get(releaseTxHash))
+            // search in storage
+            .or(() -> Optional.ofNullable(
+                    repository.getStorageBytes(contractAddress, getStorageKeyForReleaseOutpointsValues(releaseTxHash)))
+                .map(BridgeSerializationUtils::deserializeOutpointsValues)
+            );
+    }
+
+    public void setReleaseOutpointsValues(Sha256Hash releaseTxHash, List<Coin> outpointsValues) {
+        if (!activations.isActive(RSKIP305)) {
+            return;
+        }
+
+        if (releaseTxHash == null || outpointsValues == null) {
+            throw new IllegalArgumentException(
+                String.format("Invalid release outpoints values entry, has hash %s and coins list %s", releaseTxHash, outpointsValues)
+            );
+        }
+
+        if (releaseTxHashStorageKeyAlreadyExists(releaseTxHash)) {
+            throw new IllegalArgumentException("Release tx hash key already exists in storage.");
+        }
+
+        if (releasesOutpointsValues == null) {
+            releasesOutpointsValues = new TreeMap<>();
+        }
+        releasesOutpointsValues.put(releaseTxHash, outpointsValues);
+    }
+
+    private boolean releaseTxHashStorageKeyAlreadyExists(Sha256Hash releaseTxHash) {
+        byte[] data = repository.getStorageBytes(
+            contractAddress,
+            getStorageKeyForReleaseOutpointsValues(releaseTxHash)
+        );
+
+        return data != null;
+    }
+
+    private void saveReleasesOutpointsValues() {
+        if (!activations.isActive(RSKIP305)) {
+            return;
+        }
+
+        if (isNull(releasesOutpointsValues)) {
+            return;
+        }
+
+        releasesOutpointsValues.forEach(
+            (releaseTxHash, outpointsValues) -> safeSaveToRepository(
+                getStorageKeyForReleaseOutpointsValues(releaseTxHash),
+                outpointsValues,
+                BridgeSerializationUtils::serializeOutpointsValues
+            )
+        );
+    }
+
     public SortedMap<Keccak256, BtcTransaction> getPegoutsWaitingForSignatures() throws IOException {
         if (pegoutsWaitingForSignatures != null) {
             return pegoutsWaitingForSignatures;
@@ -251,7 +316,7 @@ public class BridgeStorageProvider {
         return pegoutsWaitingForSignatures;
     }
 
-    public void savePegoutsWaitingForSignatures() {
+    private void savePegoutsWaitingForSignatures() {
         if (pegoutsWaitingForSignatures == null) {
             return;
         }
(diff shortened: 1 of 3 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -239,6 +240,54 @@ public void savePegoutsWaitingForConfirmations() {
...
+            BridgeSerializationUtils::deserializeReleasesOutpointsValues
+        );
+
+        return Optional.ofNullable(releasesOutpointsValues)
+            .filter(map -> !map.isEmpty())
+            .map(Collections::unmodifiableSortedMap);
+    }
+
+    public void setReleaseOutpointsValues(Sha256Hash releaseTxHash, List<Coin> outpointsValues) {
+        if (releaseTxHash == null || outpointsValues == null) {
```
jeremy-then, line 263: Do we need the `if (!activations.isActive(RSKIP305))` check here? I know that the `save...` and `get...` methods already have it and that it might be enough. But this one could set `releasesOutpointsValues` to a map that will never be used.

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -239,6 +240,54 @@ public void savePegoutsWaitingForConfirmations() {
...
+            RELEASES_OUTPOINTS_VALUES,
+            BridgeSerializationUtils::deserializeReleasesOutpointsValues
+        );
+
+        return Optional.ofNullable(releasesOutpointsValues)
+            .filter(map -> !map.isEmpty())
+            .map(Collections::unmodifiableSortedMap);
+    }
+
+    public void setReleaseOutpointsValues(Sha256Hash releaseTxHash, List<Coin> outpointsValues) {
```
jeremy-then, line 262: Do we need the `if (!activations.isActive(RSKIP305))` check here? I know that the `save...` and `get...` methods already have it and that it might be enough. But this one could set `releasesOutpointsValues` to a map that can be added records but could never be used.

julia-zack: it sure doesn't hurt, ill add it

julia-zack: updated

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -1409,6 +1379,216 @@ private void assertNoSVPValues() {
...
+            // Act
+            Optional<SortedMap<Sha256Hash, List<Coin>>> actualReleasesOutpointsValues = bridgeStorageProvider.getReleasesOutpointsValues();
+
+            // Assert
+            assertTrue(actualReleasesOutpointsValues.isPresent());
+            assertEquals(releasesOutpointsValuesMap, actualReleasesOutpointsValues.get());
+        }
+
+        @Test
+        void getReleasesOutpointsValue_whenEntryIsInStorageAndAnotherIsSet_shouldReturnTheSetEntry() {
```
nathanieliov, line 1506: We're missing some test cases, for instance, when entries are already in the storage, and new entries are set and then saved, it should add the new ones to the map, and then when calling the get, it should return the old and new ones in the resulted map, right?

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -239,6 +240,58 @@ public void savePegoutsWaitingForConfirmations() {
...
+        if (releaseTxHash == null || outpointsValues == null) {
+            throw new IllegalArgumentException(
+                String.format("Invalid release outpoints values entry, has hash %s and coins list %s", releaseTxHash, outpointsValues)
+            );
+        }
+
+        if (releasesOutpointsValues == null) {
+            releasesOutpointsValues = new TreeMap<>();
+        }
+        releasesOutpointsValues.put(releaseTxHash, outpointsValues);
```
nathanieliov, line 276: We should discuss if this method is correct, as it is now; if you don't call `getReleasesOutpointsValues` first, and there are saved entries, then you will end up overriding the saved entries with a new map.

julia-zack: mmm i dont think so. This method is setting the entries just in the memory, they wont disappear if they are saved in storage.

julia-zack: you are soooooo right!!!! Thanks!!

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
julia-zack: updated!

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
julia-zack: updated!

julia-zack: (This method was actually correct, since its for setting releases in execution time.
The one that was wrong was the method for saving)

rskj-core/src/test/java/co/rsk/peg/BridgeStorageProviderTest.java
```diff
@@ -1008,55 +1015,20 @@ void saveSvpSpendTxWaitingForSignatures_preLovell700_shouldNotSaveInStorage() {
...
-
-            // Assert
-            byte[] actualSvpSpendTxWaitingForSignatures = repository.getStorageBytes(
-                bridgeAddress,
-                SVP_SPEND_TX_WAITING_FOR_SIGNATURES.getKey()
+        private static Stream<Arguments> invalidEntryArgs() {
+            return Stream.of(
+                Arguments.of(null, null),
+                Arguments.of(spendTxCreationHash, null),
+                Arguments.of(null, svpSpendTx)
```
julia-zack, line 1022: replaced separated tests with parameterized one

jeremy-then (approved):
LGTM

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -239,6 +241,64 @@ public void savePegoutsWaitingForConfirmations() {
...
+            return;
+        }
+
+        if (releaseTxHash == null || outpointsValues == null) {
+            throw new IllegalArgumentException(
+                String.format("Invalid release outpoints values entry, has hash %s and coins list %s", releaseTxHash, outpointsValues)
+            );
+        }
+
+        if (releasesOutpointsValues == null) {
```
jeremy-then, line 273: I think I understand what @nathanieliov is saying.

We are checking that `if (releasesOutpointsValues == null)` when we haven't potentially get the `releasesOutpointsValues` from the storage. If we don't ensure we get it from the storage and set it in this class state before trying to reassign it with `releasesOutpointsValues = new TreeMap<>();` and then we save this new map, we could override other values that might be present in the storage already.

The following is just an idea, take your time to analyze this case and let us know:

```suggestion
        if (releasesOutpointsValues == null) {
        
             SortedMap<Sha256Hash, List<Coin>> savedReleasesOutpointsValues = safeGetFromRepository(
            RELEASES_OUTPOINTS_VALUES,
            BridgeSerializationUtils::deserializeReleasesOutpointsValues
        );
        
        if (savedReleasesOutpointsValues == null) {
            releasesOutpointsValues = new TreeMap<>();
        } else {
          releasesOutpointsValues = savedReleasesOutpointsValues;
        }
        
   }
```

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -239,6 +241,69 @@ public void savePegoutsWaitingForConfirmations() {
...
+                .map(BridgeSerializationUtils::deserializeOutpointsValues)
+            );
+    }
+
+    public void setReleaseOutpointsValues(Sha256Hash releaseTxHash, List<Coin> outpointsValues) {
+        if (!activations.isActive(RSKIP305)) {
+            return;
+        }
+
+        if (releaseTxHash == null || outpointsValues == null) {
```
nathanieliov, line 264: I think we should check outpointValues not to be empty

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
```diff
@@ -239,6 +241,69 @@ public void savePegoutsWaitingForConfirmations() {
...
+        }
+
+        if (releaseTxHashStorageKeyAlreadyExists(releaseTxHash)) {
+            throw new IllegalArgumentException("Release tx hash key already exists in storage.");
+        }
+
+        if (releasesOutpointsValues == null) {
+            releasesOutpointsValues = new TreeMap<>();
+        }
+        releasesOutpointsValues.put(releaseTxHash, outpointsValues);
```
nathanieliov, line 277: Do you think you should create a copy of the outpointValues? Instead of using the reference to avoid further modifications of the list once it is set. Wdyt?

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
julia-zack: good one, thanks

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
julia-zack: but the idea is for it to be modifiable until it is saved. To be able to add several values in a row

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
julia-zack: updated

rskj-core/src/main/java/co/rsk/peg/BridgeStorageProvider.java
nathanieliov: I mean the method param `outpointsValues` if you don't create a copy, the caller can modify the list later, and it would also impact the `releasesOutpointsValues`. It is a good practice to make list unmmutable/copy when you receive it by method param.

nathanieliov (approved):

PR #3038, opened 2025-03-17
Print warning message when we find any disabled network upgrade

rmoreliovlabs:
## Description
We are printing a warning message when running the node so we know which NetworkUpgrade are disabled at the point of the node running.

## Motivation and Context

We had an issue where the node of a client was creating a hard fork due to client not having the latest network upgrade `LOVELL700` enabled / added to their config.

We spent a lot of time trying to investigate the issue and we found out it was a configuration issue. If have noticed the configuration issue before we wouldn't spend a lot of time debugging the problem.

## How Has This Been Tested?

We just ran a node from scratch. At the beginning since the block is at height of 0, all the existing network upgrades will be shown, even if they are added to the rsk config, because they will not be active due to the best block height.

Finally, the same node was restarted, so we were able to see how the number of network upgrades printed as disabled were decrease as the best block height was increasing.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/RskContext.java
+++ b/rskj-core/src/main/java/co/rsk/RskContext.java
@@ -93,7 +93,6 @@ import co.rsk.trie.MultiTrieStore;
 import co.rsk.trie.TrieStore;
 import co.rsk.trie.TrieStoreImpl;
 import co.rsk.util.RskCustomCache;
-import co.rsk.util.SystemUtils;
 import co.rsk.validators.*;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
@@ -337,7 +336,6 @@ public class RskContext implements NodeContext, NodeBootstrapper {
 
         if (blockchain == null) {
             blockchain = getBlockChainLoader().loadBlockchain();
-            SystemUtils.printDisabledNetworkUpgrades(logger, blockchain, getRskSystemProperties().getActivationConfig());
         }
 
         return blockchain;
--- a/rskj-core/src/main/java/co/rsk/RskContext.java
+++ b/rskj-core/src/main/java/co/rsk/RskContext.java
@@ -93,6 +93,7 @@ import co.rsk.trie.MultiTrieStore;
 import co.rsk.trie.TrieStore;
 import co.rsk.trie.TrieStoreImpl;
 import co.rsk.util.RskCustomCache;
+import co.rsk.util.SystemUtils;
 import co.rsk.validators.*;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
@@ -336,6 +337,7 @@ public class RskContext implements NodeContext, NodeBootstrapper {
 
         if (blockchain == null) {
             blockchain = getBlockChainLoader().loadBlockchain();
+            SystemUtils.printDisabledNetworkUpgrades(logger, blockchain, getRskSystemProperties().getActivationConfig());
         }
 
         return blockchain;
--- a/rskj-core/src/main/java/co/rsk/util/SystemUtils.java
+++ b/rskj-core/src/main/java/co/rsk/util/SystemUtils.java
@@ -24,6 +24,7 @@ import org.ethereum.config.blockchain.upgrades.NetworkUpgrade;
 import org.ethereum.core.Blockchain;
 import org.slf4j.Logger;
 
+import javax.annotation.Nonnull;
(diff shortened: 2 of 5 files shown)


Vovchyk (changes requested):
in general, the PR lgtm, but some tests are failing

Vovchyk (approved):

rskj-core/src/main/java/co/rsk/util/SystemUtils.java
```diff
@@ -69,4 +72,19 @@ private static Stream<Pair<String, String>> runtimePropsStream() {
...
+    private static List<NetworkUpgrade> getDisabledNetworkUpgrades(Blockchain blockchain, ActivationConfig activationConfig) {
+        final var latestBlock = blockchain.getBestBlock();
+
+        return Arrays.stream(NetworkUpgrade.values())
+                .filter(networkUpgrade -> !activationConfig.containsNetworkUpgrade(networkUpgrade) || !activationConfig.isActive(networkUpgrade, latestBlock.getNumber()))
+                .toList();
+    }
+
+    public static void printDisabledNetworkUpgrades(Logger logger,  Blockchain blockchain, ActivationConfig activationConfig) {
+        final var disabledNetworkUpgrades = getDisabledNetworkUpgrades(blockchain, activationConfig);
```
fmacleal, line 85: suggestion:

Just for safety, would be good to validate the inputs and throws an `IllegalArgumentException` just in case.

Something like:
```
    if (logger == null || blockchain == null || activationConfig == null) {
        throw new IllegalArgumentException("Logger, Blockchain, and ActivationConfig must not be null");
    }
```

rskj-core/src/main/java/co/rsk/util/SystemUtils.java
```diff
@@ -69,4 +72,19 @@ private static Stream<Pair<String, String>> runtimePropsStream() {
...
+        final var latestBlock = blockchain.getBestBlock();
+
+        disabledNetworkUpgrades.forEach(disabledNetworkUpgrade -> logger.warn("WARNING: Network upgrade {} is DISABLED. Best block number is: {}.", disabledNetworkUpgrade.name(), latestBlock.getNumber()));
```
fmacleal, line 88: thought:

I am wondering now, that maybe we could indeed log some info if no network upgrade is disabled and a summary of amount of network upgrade disabled. Let me know your opinion, I think it would pass a good info.

```
    if (disabledNetworkUpgrades.isEmpty()) {
        logger.info("All network upgrades are active.");
    } else {
        logger.warn("Total number of disabled network upgrades: {}", disabledNetworkUpgrades.size());
        disabledNetworkUpgrades.forEach(disabledNetworkUpgrade -> 
            logger.warn("WARNING: Network upgrade {} is DISABLED. Best block number is: {}.", 
                disabledNetworkUpgrade.name(), latestBlockNumber)
        );
    }
```

fmacleal (commented):
Good job! 

I have a few comments and suggestions, let me know your opinion.

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -336,6 +337,7 @@ public synchronized Blockchain getBlockchain() {
 
         if (blockchain == null) {
             blockchain = getBlockChainLoader().loadBlockchain();
+            SystemUtils.printDisabledNetworkUpgrades(logger, blockchain, getRskSystemProperties().getActivationConfig());
```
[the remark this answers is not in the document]
rmoreliovlabs: No specific reason to put it there, it was goign to executed once since the `blockchain` is instanced once and we need to print network upgrade information.

I tried to put it there, but I will need to cast it, because it that class is using a `NodeContext` instead of the `RskContext` and seem to be intended.

However, I will update it to be there, let me konw how it looks like @fmacleal

rskj-core/src/main/java/co/rsk/util/SystemUtils.java
rmoreliovlabs: In sake of simplicity added the `@NonNull` annotation

fmacleal: Indeed, way better and simpler. Thanks!

rskj-core/src/main/java/co/rsk/util/SystemUtils.java
rmoreliovlabs: Addressed

fmacleal (approved):
Thank you for have replied and addressed so quickly!
🙏

PR #3052, opened 2025-03-26
Throw exception for RLP.getNextElementIndex when pos is not correct

rmoreliovlabs:
## Description
Throw exception for `RLP.getNextElementIndex` when pos is not correct and when next element index if out of range for [post + 1, payload.length).

## Motivation and Context
Back then the security team found a few vulnerabilities in the RLP class. Most of the findings were fixed but the case with `RLP.getNextElementIndex` wasn’t. Even though the security team doesn’t believe that this can be exploited, they recommend to tackle it now to prevent potential misuse of the method in future.

This is an example caught by their fuzzer:
```
class RLPTest {
    ...
    
    @Test
    void testGetNextElementIndex() {
        byte[] payload = Base64.decodeBase64("/g==");
        RLP.getNextElementIndex(payload, 0);
    }
    ...
}
```

and this is an example of the test that produces smaller position than initial:

```
  public void testGetNextElementIndexSmaller() {
      byte[] maliciousPayload = new byte[]{
        0, 0, 0, 0, // empty pad 
        (byte) 0xb8,
        (byte) 0xff, // lengthOfLength = 8
        (byte) 0xff, (byte) 0xff, (byte) 0xff, (byte)0xF4, (byte)0x00, (byte)0x00, (byte)0x00, (byte)0x00,
        (byte) 0x01
      };

      int pos = 5;
      int nextIndex = RLP.getNextElementIndex(maliciousPayload, pos);

      System.out.println("nextIndex = " + nextIndex); // nextIndex = 2
      assertTrue(nextIndex > pos);
  }
```

## How Has This Been Tested?

Just added unit tests

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -71,6 +77,8 @@
  * @since 01.04.2014
  */
 public class RLP {
+    private static final Logger logger = LoggerFactory.getLogger(RLP.class);
```
Vovchyk, line 80: not sure we really need one more logger in this class.. RLP seems to be kinda data entity, like String class. Maybe FrameCodec is a better place for logging this?

rskj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -191,43 +199,62 @@ public static BigInteger decodeBigInteger(byte[] data, int index) {
...
-            return pos + 1;
+            nextElementIndex = pos + 1;
         }
-        return -1;
+
+        if (pos + 1 <= nextElementIndex && nextElementIndex < payload.length) {
+            return nextElementIndex;
+        }
+
+        throw new RLPException("Next element index out of next pos and payload length range");
```
Vovchyk, line 257: so `-1` isn't legitimate value anymore? then shouldn't be thrown an exception in [here](https://github.com/rsksmart/rskj/pull/3052/files#diff-12b253aa5c996cabe09f5ab95a1f6a4123ed6faf5fc56518fcf549644cb0d47bR211) too?

rskj-core/src/main/java/org/ethereum/util/RLP.java
rmoreliovlabs: This change was made due to a SonarQube complain, throwing back the error is not an option here because some methods needs to return somethnig back instead of throwing the error

rskj-core/src/main/java/org/ethereum/util/RLP.java
rmoreliovlabs: There are other conditions where the `-1` will be returned, according to the context, we just want to cover specific 2 conditions, you can see nextElementIndex is itinitialized as `-1` because of that

rskj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -191,43 +199,62 @@ public static BigInteger decodeBigInteger(byte[] data, int index) {
...
-            return pos + 1 + length;
+            nextElementIndex = pos + 1 + length;
         }
         if ((payload[pos] & 0xFF) == OFFSET_SHORT_ITEM) {
-            return pos + 1;
+            nextElementIndex = pos + 1;
         }
         if ((payload[pos] & 0xFF) < OFFSET_SHORT_ITEM) {
-            return pos + 1;
+            nextElementIndex = pos + 1;
```
asoto-iov, line 250: With these changes we are not breaking the if check, so we ware always checking of the if conditionals. I think it would be better to avoid it. 
We can avoid id by using else if or extracting this part of get the nextElementIndex to a functio nand continue with the other checks later.

rmoreliovlabs: That makes sense! Updating the PR

asoto-iov (approved):

rskj-core/src/main/java/org/ethereum/util/RLP.java
Vovchyk: hmm, so SonarQube requires the RLP class to have a logger?

rskj-core/src/main/java/org/ethereum/util/RLP.java
Vovchyk: there's only one usage of this method in FrameCodec. As I can see, you already replaced one return of -1 with throwing the RLPException error. Wouldn't it make sense to replace the other one?

rskj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -191,43 +199,57 @@ public static BigInteger decodeBigInteger(byte[] data, int index) {
         return BigIntegers.fromUnsignedByteArray(bytes);
     }
 
+    private static void assertLengthOfLength(byte[] payload,byte lengthOfLength) {
+        if (lengthOfLength >= payload.length) {
+            throw new RLPException("lengthOfLength cannot be higher than payload length");
+        }
+    }
+
     public static int getNextElementIndex(byte[] payload, int pos) {
```
fmacleal, line 208: thought:

I know isn't part at all from the ticket, but, I see a big value if we refactor the chain of if-else for at least a switch, where the input of the switch would be the firstByte, like this:

```
    int firstByte = payload[pos] & 0xFF;
...
```

But I struggled a bit to do that without use new features of java 21, so...it would be a nice to have but not obligatory.

fmacleal (dismissed):
Good job, apart of the logger that I am still not sure...it's all good.

Nice that you have added the tests suggested by security.

rskj-core/src/main/java/org/ethereum/util/RLP.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/org/ethereum/util/RLP.java
rmoreliovlabs: We cannot do that, due to the greater / less than conditions

rskj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -191,43 +195,55 @@ public static BigInteger decodeBigInteger(byte[] data, int index) {
...
 
         if (pos >= payload.length) {
-            return -1;
+            throw new RLPException("pos out of payload length range");
```
Vovchyk, line 207: maybe we should also check if `pos` isn't negative?

rskj-core/src/main/java/org/ethereum/util/RLP.java
```diff
@@ -191,43 +195,55 @@ public static BigInteger decodeBigInteger(byte[] data, int index) {
         return BigIntegers.fromUnsignedByteArray(bytes);
     }
 
+    private static void assertLengthOfLength(byte[] payload,byte lengthOfLength) {
+        if (lengthOfLength >= payload.length) {
```
Vovchyk, line 199: it seems that `lengthOfLength` means a size of the length value in bytes, not a position inside the `payload` array. Perhaps it should be checked that `(pos + lengthOfLength)` is within boundaries. Plus it'd make sense to also check if `lengthOfLength` isn't negative?

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/util/RLP.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/org/ethereum/util/RLP.java
rmoreliovlabs: Addressed

fmacleal (approved):
LGTM

Vovchyk (approved):

PR #3056, opened 2025-04-01
deleted the PendingFederation constructor to replace it for the Builder

julianlen:
Option 1 is the option that requires fewer changes. The steps are 

1. It keeps the federation member creation.
2. It uses those members and extract their btc public keys
3. It uses the builder with those keys


jeremy-then (approved):
LGTM.

PR #3062, opened 2025-04-07
feat(snap): add snap rate limiter

Vovchyk:
This feature adds for a snap server a configurable limit on a max number of requests in a queue from one peer

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/core/bc/BlockChainFlusher.java
```diff
@@ -100,7 +100,7 @@ private synchronized void flush() {
     }
 
     private void flushAll() {
-        Metric metric = profiler.start(Profiler.PROFILING_TYPE.BLOCKCHAIN_FLUSH);
+        Metric metric = profiler.start(Profiler.MetricKind.BLOCKCHAIN_FLUSH);
```
rmoreliovlabs, line 103: Why are we changing `PROFILING_TYPE` to `MetricKind`?

Can we add the the reasoning in the PR description?

rskj-core/src/main/java/co/rsk/net/sync/SyncMessageHandler.java
```diff
@@ -64,12 +69,24 @@ public void run() {
...
+            Metric metric = null;
             try {
                 job = jobQueue.take();
 
+                metric = profiler.start(job.getMetricKind());
+
+                if (maxSenderJobs != null) {
+                    Peer sender = job.getSender();
+                    NodeID senderId = sender.getPeerNodeID();
+                    long jobCount = 1 + jobQueue.stream().filter(j -> j.getSender().getPeerNodeID().equals(senderId)).count();
```
rmoreliovlabs, line 81: Wondering if there is a way to keep the count in a map by senderId so we avoid filtering + counting in the loop. Also not sure how many jobs could be there. If it is possible to have many jobs then that might be a problem. 

Do you think it is worth to think about adding a map where the key is the sender id and value is the count for this purpose?

rmoreliovlabs: Souldn't that logic only work for snap?

rmoreliovlabs (changes requested):
Could we clarify that there are many changes related to a renaming of an enum?

rskj-core/src/main/java/co/rsk/core/bc/BlockChainFlusher.java
Vovchyk: the name itself was changed to reflect its meaning - it's type of metric, not type of profiling. In addition to that it was changed from upper snake case to pascal one to align with other enums naming - Profiler.PROFILING_TYPE isn't an enum value, it's an enum name

rmoreliovlabs: Sounds good, please add that to the PR description. Thanks

rmoreliovlabs (approved):

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
```diff
@@ -142,7 +146,20 @@ public SnapshotProcessor(Blockchain blockchain,
 
         this.checkHistoricalHeaders = checkHistoricalHeaders;
         this.parallel = isParallelEnabled;
-        this.thread = new Thread(new SyncMessageHandler("SNAP/server", requestQueue, listener) {
+        this.requestQueue = new LinkedBlockingQueue<>() {
+            @Override
+            public boolean offer(@Nonnull SyncMessageHandler.Job job) {
```
asoto-iov, line 151: I like this, but I’m afraid this function might end up a bit hidden. Maybe it’s just a matter of preference, but I would do this override in a new class that extends LinkedBlockingQueue.

asoto-iov:
looks good to me, good job!

fmacleal (commented):
Good job! 

I have a few comments, let me know your thoughts.

asoto-iov (approved):

rskj-core/src/test/java/co/rsk/net/SnapshotProcessorTest.java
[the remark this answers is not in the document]
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
```diff
@@ -204,17 +209,12 @@ public void processSnapStatusRequest(Peer sender, SnapStatusRequestMessage reque
...
-                @Override
-                public void run() {
-                    processSnapStatusRequestInternal(sender, requestMessage);
-                }
-            });
-        } catch (InterruptedException e) {
-            logger.warn("SnapStatusRequestMessage processing was interrupted", e);
-            Thread.currentThread().interrupt();
-        }
+        offerJob(new SyncMessageHandler.Job(sender, requestMessage, Profiler.MetricKind.SNAP_STATUS_REQUEST) {
```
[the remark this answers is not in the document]
rmoreliovlabs: Addressed by @Vovchyk

rskj-core/src/main/java/co/rsk/metrics/profilers/Profiler.java
```diff
@@ -12,7 +29,7 @@ public interface Profiler {
      * Depending on what is actually being profiled, new categories can be added or
      * categories not needed can be removed
      */
-    enum PROFILING_TYPE {
+    enum MetricKind {
```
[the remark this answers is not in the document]
rmoreliovlabs: Addressed by @Vovchyk

PR #3069, opened 2025-04-09
Add step to download test results from pipeline

rmoreliovlabs:
## Description

Added a step to upload the test results as artifacts in github actions, so dev can take a look to them for further reviews

Also we added retention days of one week to these artifacts

## Motivation and Context

Sometimes it is difficult to use Gtihub Actions UI to look for test failures.

Right now the possible options we have to see test failures are:

1. We can look for the keyword `FAILED` in the logs:

Screenshot 2025-04-10 at 10 19 14 AM

2. We can click on Explain Error

https://github.com/user-attachments/assets/da03d968-27ba-4d49-8928-b6f70269da80

## How Has This Been Tested?

Yes, we made some tests fail in purpose, in order to see the reports generated

Screenshot 2025-04-10 at 6 00 14 PM

Screenshot 2025-04-10 at 5 58 57 PM

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -99,6 +99,14 @@ jobs:
           path: |
             rskj-core/build/reports/
 
+      - name: Upload HTML Test Report
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 #v4.6.2
+        with:
+          name: smell-test-html-test-report
+          path: |
+            rskj-core/build/reports/
+
       - name: Run SonarQube analysis
         env:
           GH_EVENT: ${{ github.event_name }}
@@ -259,6 +267,14 @@ jobs:
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 #v4.6.2
         with:
           name: test-reports
+          path: |
+            rskj-core/build/reports/
+
+      - name: Upload HTML Test Report
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 #v4.6.2
+        with:
+          name: unit-tests-java17-html-test-report
           path: |
             rskj-core/build/reports/
 
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -100,9 +100,10 @@ jobs:
             rskj-core/build/reports/
 
       - name: Upload HTML Test Report
-        if: always()
+        if: failure()
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 #v4.6.2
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -99,6 +99,14 @@ jobs:
           path: |
             rskj-core/build/reports/
 
+      - name: Upload HTML Test Report
```
Vovchyk, line 102: wondering if we should always upload results or only if the tests are failing?

Vovchyk (changes requested):

fmacleal (approved):
Apart the comment from @Vovchyk, I think the result is pretty good and more than enough.

Let's just agree with the rest of the team if we should publish it always or only when it fails.

.github/workflows/build_and_test.yml
nathanieliov: To me, it makes sense only when the tests failed

nathanieliov (approved):

.github/workflows/build_and_test.yml
rmoreliovlabs: Addressed

rmoreliovlabs:
> Apart the comment from @Vovchyk, I think the result is pretty good and more than enough.
> 
> Let's just agree with the rest of the team if we should publish it always or only when it fails.

Done, we decided to only upload it when there are test failures

.github/workflows/build_and_test.yml
```diff
@@ -99,6 +99,14 @@ jobs:
...
+        if: failure()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 #v4.6.2
+        with:
+          name: smell-test-html-test-report
+          path: |
+            rskj-core/build/reports/
```
asoto-iov, line 108: Just a small comment about these files, How many days will these files be available?
I believe the default is 90. If that's correct, we could reduce it using the` retention-days `property to avoid overloading the storage with unnecessary files.

fmacleal (dismissed):
LGTM.

Well done.

.github/workflows/build_and_test.yml
rmoreliovlabs: That makes sense, do you know which would be correct? I'm not very familiar to the  Github Action rules

lucasvuotto: I believe that one week should be enough. The value should be useful for the devs, not for us. Nevertheless, it's GitHub's storage, not ours, so _we_ don't face issues with the storage use. :D

rmoreliovlabs: WIll add retention-days of one week (7 days) then

fmacleal (approved):

nathanieliov (approved):

PR #3071, opened 2025-04-09
feat(snap): add JMX profiling

Vovchyk:
this adds a new `Profiler` implementation - `JmxProfiler`. Jmx is a java tech, which can also be used to export custom metrics (besides existing ones like memory usage, etc.) and works quite well in tandem with the Prometheus/Grafana integration. More info on that [here](https://fndomoraes.medium.com/exporting-custom-metrics-from-jmx-to-prometheus-265226f1219c)

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/metrics/jmx/JmxMetricMBean.java
```diff
@@ -0,0 +1,29 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package co.rsk.metrics.jmx;
+
+import co.rsk.metrics.profilers.Profiler;
+
+import java.math.BigInteger;
+
+public interface JmxMetricMBean {
```
rmoreliovlabs, line 25: Why do we need an interface?

rmoreliovlabs (changes requested):
PR LGTM overall

Could we add PR description and evidence on how this looks like?

rskj-core/src/main/java/co/rsk/metrics/jmx/JmxMetricMBean.java
Vovchyk: it's just a convention (more info [here](https://docs.oracle.com/javase/tutorial/jmx/mbeans/standard.html)). Also MBeanServer won't allow to register a "bean" concrete implementation without it being implementing such interface

rmoreliovlabs: Good to know

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -127,7 +131,15 @@ public abstract class SystemProperties {
     protected SystemProperties(ConfigLoader loader) {
         try {
             this.configFromFiles = loader.getConfig();
-            if(logger.isTraceEnabled()){
+            if (isSystemProfilingEnabled(this.configFromFiles)) {
```
asoto-iov, line 134: minor thing but I wold extract this profiling specific logic to a initProfiling (kind of) method. Just to avoid this constructor to grow in the future with similar new features.

asoto-iov (approved):

rskj-core/src/main/java/co/rsk/metrics/jmx/JmxMetric.java
```diff
@@ -0,0 +1,95 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package co.rsk.metrics.jmx;
+
+import co.rsk.metrics.profilers.Profiler;
+
+import java.math.BigInteger;
+
+public class JmxMetric implements JmxMetricMBean {
```
fmacleal, line 25: question:

If we created a specific interface for Jmx, why do we have a `Metric` interface then? Couldn't we delete it or will this metric implement that one as well?

rskj-core/src/main/java/co/rsk/metrics/profilers/impl/JmxProfiler.java
```diff
@@ -0,0 +1,85 @@
...
+    public void stop(Metric metric) {
+        if (metric instanceof MetricImpl) {
+            JmxMetric jmxMetric = jmxMetrics.get(((MetricImpl) metric).getKind());
+            if (jmxMetric != null) {
+                jmxMetric.updateDuration(((MetricImpl) metric).getDuration());
+            }
+        }
+    }
+
+    private static class MetricImpl implements Metric {
```
fmacleal, line 68: comment:

Ok, now I see the `MetricImpl`, why is it private then? Why have an implementation here and not extract it so it can be used by other metrics and this way be reused?

thought:
What prevents us to make these objects and method implementation be in the `Metric` interface as default method? Or, at least, the `getKind()` and `getDuration()` methods could be there.

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -600,6 +607,14 @@ public boolean shouldPrintSystemInfo() {
         return getBoolean(PROPERTY_PRINT_SYSTEM_INFO, false);
     }
 
+    public boolean isSystemProfilingEnabled() {
+        return isSystemProfilingEnabled(this.configFromFiles);
+    }
```
fmacleal, line 612: comment:

Not used, is really needed?

rskj-core/src/main/java/co/rsk/metrics/jmx/JmxMetric.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+        return this.duration;
+    }
+
+    @Override
+    public synchronized BigInteger getCumulativeDuration() {
+        return this.cumulativeDuration;
+    }
+
+    public synchronized void updateDuration(long duration) {
```
fmacleal, line 62: suggestion:

I know that the metrics calculations are quite straightforward, but I think it would be helpful having a test class validating these calculations and so we can assure that all of them are being updated accordingly and giving us assertive values. WDYT?

rskj-core/src/test/java/co/rsk/metrics/profilers/impl/JmxProfilerTest.java
fmacleal, line 1: suggestion:

Add a test to validate the `RuntimeException` if a wrong metric kind is passed.

rskj-core/src/main/java/co/rsk/metrics/profilers/impl/JmxProfiler.java
```diff
@@ -0,0 +1,85 @@
...
+
+public class JmxProfiler implements Profiler {
+
+    private final Map<MetricKind, JmxMetric> jmxMetrics = new HashMap<>();
+
+    public void register(@Nonnull MBeanServer mbs) {
+        Objects.requireNonNull(mbs);
+
+        for (MetricKind kind : MetricKind.values()) {
+            JmxMetric jmxMetric = new JmxMetric(kind, MetricAggregate.MAX);
```
fmacleal, line 41: question:

Why the `MetricAggregate.MAX` was passed as default? Are these metric aggregates configurable? Where do we configure them?

fmacleal (commented):
Nice metrics addition, interesting and useful to be expanded.

I have a few comments though, mainly, where are we using the JMX metric in the other [PR](https://github.com/rsksmart/rskj/pull/3062/files). Do we have an example documentation to use it?

Would be nice to having tests covering this use.

rskj-core/src/main/java/co/rsk/metrics/jmx/JmxMetric.java
Vovchyk: this is a bit different thing required by the built-in JMX framework. Reynold asked about something similar in [here](https://github.com/rsksmart/rskj/pull/3071#discussion_r2035513194).

rskj-core/src/main/java/co/rsk/metrics/profilers/impl/JmxProfiler.java
Vovchyk: I'd say just because it is specific to the JmxProfiler implementation. No need to expose this to outside, as it should be transparent to a user. Maybe similar concept is a logger. We don't know its implication and in fact it depends on the configuration

rskj-core/src/main/java/co/rsk/metrics/profilers/impl/JmxProfiler.java
Vovchyk: no, it's not configurable as of now. I experimented with different options and left MAX as a default option. But yes, in future, that can be made configurable if needed. Just didn't want to overload this PR. For me, for the time being, MetricAggregate.MAX is just sufficient enough. WDYT?

PR #3076, opened 2025-04-11
Update docker release process

rmoreliovlabs:
## Description

In this PR we are improving our pipeline:

1. We are adding an extra check to the `docker-verification.yml > Check Docker container built` step to look for the `IMPORTED_BEST` char sequence in the logs, the will let us know that the node is actually syncing.
2. We are splitting the old `docker-release.yml` into 2:
  * One to update the docker image edge tag when there is a commit in master
  * The other one to update remaning docker tags (such as latest) when there is a tag created in github.

## Motivation and Context

A partner reported a situation they were observing: their nodes running on docker were stalling on 2,589,174 and 3,344,303 blocks for testnet and mainnet respectively.

The docker image was using Java Temurin 21, that was upgraded from Java Temurin 17 on previous weeks, by applying a Dependabot suggestion.

The partner shared a docker-compose.yml file with us  to try to reproduce the error by ourselves.

We detected the some issues where one of them was: `the docker pipeline wasn't validating the built image`.

That is why we are doing these changes, to validate the node is running properly and avoid updating docker latest tag with master code, instead wait for the latest tag generation.

## How Has This Been Tested?

We were able to test `docker-verifivcation.yml`

Screenshot 2025-04-11 at 5 19 42 PM

We tested the docker image tag generation with own docker account and it worked:

1. For the `edge` tag:

https://github.com/rmoreliovlabs/rskj/actions/runs/14452682585/job/40528857581

Screenshot 2025-04-14 at 2 17 51 PM

2. For the `latest` tag:

https://github.com/rmoreliovlabs/rskj/actions/runs/14452990471/job/40529809401

Screenshot 2025-04-14 at 2 51 24 PM

## Types of changes

## Checklist:

* **Other information**:


.github/workflows/docker-verification.yml
```diff
@@ -28,9 +28,4 @@ jobs:
...
       - name: Check Docker container built
         run: |
-          if [ $(docker inspect -f '{{.State.Running}}' rskj-container) = "true" ]; then
-            echo "Container is running properly"
-          else
-            echo "Container failed to start"
-            exit 1
-          fi
\ No newline at end of file
+          ./validate-docker-container.sh
```
fmacleal, line 31: comment:

I know that it's a good practice to add this validation in the script, but recently, every time I am adding a script in the pipelines it have been highlighted that this could be exploited  in a PR substituting the script. Explained a bit more [here](https://cycode.com/blog/github-actions-vulnerabilities/).

I would say that maybe for this case, since the script it's straightforward and the pipeline it's just for that I would suggest to just copy the script content here in this part. Wouldn't hurt to have this validation here, we don't plan to change it much. :)

.github/workflows/docker-release-tags-to-latest.yml
```diff
@@ -26,7 +24,6 @@ jobs:
           tags: |
             type=ref,event=branch
             type=ref,event=tag
-            type=edge
```
fmacleal, line 29: question:

Will it overwrite the latest with this version? 🤔 
Just checking.

fmacleal (changes requested):
Good job!

I have just a suggestion and some question, apart of that. I think it's nice solution.
Initially I thought and put the two releases in the same pipeline, but then, I think it's really better split. So, I liked this approach.

Also, good script to validate the synching. Simple and efficient, well done.

.github/workflows/docker-verification.yml
rmoreliovlabs: The script does not work properly when copying and pasting the code. This part specifically:

```
        docker logs rskj-container > logs.txt

        count=$(grep -c "IMPORTED_BEST" logs.txt)
```

is not getting populated, not sure what is happening, probably someone from devops could help here

fmacleal: I might take a look later, but meanwhile, would you mind to reach plataform team to help you?

validate-docker-container.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [ "$(docker inspect -f '{{.State.Running}}' rskj-container)" = "true" ]; then
+    max_attempts=20
+    attempts=0
+    count=0
+
+    while [ $attempts -lt $max_attempts ]; do
+        docker logs rskj-container > logs.txt
+
+        count=$(grep -c "IMPORTED_BEST" logs.txt)
```
lucasvuotto, line 11:
```suggestion
        count=$(docker logs rskj-container | grep -c "IMPORTED_BEST")
```

.github/workflows/docker-verification.yml
rmoreliovlabs: Thanks @lucasvuotto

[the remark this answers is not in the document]
rmoreliovlabs: Addressed

validate-docker-container.sh
rmoreliovlabs: Adderessed

validate-docker-container.sh
```diff
@@ -0,0 +1,34 @@
...
+
+        if [ $count -gt 0 ]; then
+        echo "✅ Found 'IMPORTED_BEST' block in logs!"
+        break
```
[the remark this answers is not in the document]
rmoreliovlabs: Addressed

validate-docker-container.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+
+if [ "$(docker inspect -f '{{.State.Running}}' rskj-container)" = "true" ]; then
```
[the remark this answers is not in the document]
rmoreliovlabs: Addressed

validate-docker-container.sh
lucasvuotto, line 1: I really like that you merged everything in the script! There is only a minor but important nit in this case: the commands in `run:` are run with `set -e` set, which means that the script will abort if any command fails. The same can be achieved here by adding `set -e` in the line right after `#!/bin/sh`.

Also, if the container is stopped at the end of the script run, this becomes very friendly to be run locally too. You can achieve that with a construction like this

```
container_id=$(docker run -d --name rskj-container rsksmart/rskj:latest --regtest)
trap 'docker stop "$container_id"' EXIT INT QUIT TERM
```

lucasvuotto (dismissed):

validate-docker-container.sh
rmoreliovlabs: Let me try that thanks!

validate-docker-container.sh
```diff
@@ -0,0 +1,41 @@
...
+done
+
+if [ $count -le 0 ]; then
+    echo "❌ 'IMPORTED_BEST' block not found in logs after $max_attempts attempts"
+    exit 1
+fi
+
+echo "Container is running properly"
+
+trap 'docker stop "$container_id"' EXIT INT QUIT TERM
```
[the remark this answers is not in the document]
rmoreliovlabs: Ooh, got it, thanks for catch, addressed

lucasvuotto (dismissed):

fmacleal (approved):
Cool!
LGTM!

.github/workflows/docker-verification.yml
```diff
@@ -1,36 +1,64 @@
 name: Docker image validation
 
 on:
-  pull_request:
+  pull_request_target:
```
lucasvuotto, line 4: If my reading on `pull_request_target` is correct, this won't be ran with the new code, but with that of the base branch of the PR, meaning that it won't be possible to test that modifications of `Dockerfile` don't break the container, defeating the purpouse of this workflow altogether. Is this the intended way?

fmacleal: Yes, totally makes sense.

We should definitely revisit this and rollback this part. I don't see risk since we are blocking this for forks and also the `pull_request` doesn't have secret access. So we are good here. Can you rollback this part @rmoreliovlabs?

fmacleal (changes requested):
Rollback the part and use `pull_request` instead of `pull_request_target` in the `docker-verification.yml`.

.github/workflows/docker-verification.yml
rmoreliovlabs: Sure

fmacleal (approved):
LGTM

PR #3079, opened 2025-04-14
Added base class to implement stateOverride featue into eth_call

asoto-iov:
## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,8 +1,8 @@
 # mining
-/rskj-core/src/main/java/co/rsk/mine/                                       @rsksmart/rsk-mining
-/rskj-core/src/main/java/co/rsk/rpc/modules/mnr/                            @rsksmart/rsk-mining
-/rskj-core/src/main/java/co/rsk/validators/ProofOfWorkRule.java             @rsksmart/rsk-mining
-/rskj-core/src/main/java/co/rsk/validators/Rskip92MerkleProofValidator.java @rsksmart/rsk-mining
+/rskj-core/src/main/java/co/rsk/mine/                                       @rsksmart/rsk-core
+/rskj-core/src/main/java/co/rsk/rpc/modules/mnr/                            @rsksmart/rsk-core
+/rskj-core/src/main/java/co/rsk/validators/ProofOfWorkRule.java             @rsksmart/rsk-core
+/rskj-core/src/main/java/co/rsk/validators/Rskip92MerkleProofValidator.java @rsksmart/rsk-core
 
 # powpeg
 /rskj-core/src/main/java/co/rsk/peg/                                        @rsksmart/rsk-fed


rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
```diff
@@ -0,0 +1,133 @@
...
+
+    public RskAddress getMovePrecompileToAddress() {
+        return movePrecompileToAddress;
+    }
+
+    public void setMovePrecompileToAddress(RskAddress movePrecompileToAddress) {
+        this.movePrecompileToAddress = movePrecompileToAddress;
+    }
+
+    public Repository applyToRepository(Repository repository) {
```
rmoreliovlabs, line 79: shouldn't this be like an account repository? should we be explicit?

Thinking that probably we can move this to another class, probably we could use an existing utility class

rmoreliovlabs (changes requested):

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -133,6 +134,14 @@ public Map<String, Object> bridgeState() throws IOException, BlockStoreException
         return state.stateToMap();
     }
 
+    public String call(CallArgumentsParam argsParam, BlockIdentifierParam bnOrId, List<AccountOverride> accountOverrideList) {
```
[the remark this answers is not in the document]
asoto-iov: Inside the AccountOverride object I place the address value, so a map it is not needed.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
asoto-iov: I think here it is a good place to do it. There is no other places where state override is happening, also if the class disappears the method should do aswell. 
Move this to another class would be to mix different business logic, and create a new one for this simple method I think it is not worthy. 

> shouldn't this be like an account repository? should we be explicit? 
I don't get this sorry

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
```diff
@@ -0,0 +1,133 @@
...
+        if (address == null) {
```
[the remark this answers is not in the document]
asoto-iov: It is more optimal for this simple case to use the if null

rmoreliovlabs (approved):

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
```diff
@@ -0,0 +1,133 @@
...
+import java.math.BigInteger;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Objects;
+
+public class AccountOverride {
+    private BigInteger balance;
+    private Long nonce;
+    private byte[] code;
+    private Map<DataWord, DataWord> state;
```
fmacleal, line 17: thought:

I am just wondering if this Map should be `DataWord` of ir the value should be of other type. But, I guess that this can easily be changed in the next ticket I am afraid.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
```diff
@@ -0,0 +1,133 @@
...
+    private Map<DataWord, DataWord> stateDiff;
+    private RskAddress address;
+    //TODO movePrecompile to address
+    private RskAddress movePrecompileToAddress;
```
fmacleal, line 21: question: 

What is this movePrecompileToAddress?

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
fmacleal: suggestion:

Honestly, I would say that it would be good indeed move this logic outside this model. I think the way it is, it's violating the SRP (Single Responsibility Principle). The model should have methods that are related with calculation from the model. For example, validate the state, or check if the balance is higher than zero, that the nonce has a valid value and so on.

I would suggest to create an interface to follow a bit the DIP (Dependency Inversion Principle) where the logic that will use the values in this model will be implemented in a concrete class that implements this interface. For example:

```
public interface StateOverrideApplier {
  Repository applyToRepository(Repository repository, AccountOverride accountOverride);
}
```

Creating an implementation of this `StateOverrideApplier` that could be in the class that will use this interface. That way, if we want to change the logic of this implementation, it would be separated and this model would follow the Open Closed principle (being closed to change but open for expansion). It would be easier to test as well, since you could create a test class only for the implementation.

Also, let's say that we want to have a StateOverride for another thing which is not the Trie for example, it would be just create the new implementation and use it wherever is needed.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
```diff
@@ -0,0 +1,133 @@
...
+
+    public BigInteger getBalance() {
+        return balance;
+    }
+
+    public void setBalance(BigInteger balance) {
+        this.balance = balance;
```
fmacleal, line 28: question:

Can the balance have any value?  Can it be null? Lower than zero?

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public Long getNonce() {
+        return nonce;
+    }
+
+    public void setNonce(Long nonce) {
+        this.nonce = nonce;
```
fmacleal, line 36: question:

Can the nonce have any value? Can it be null? Lower than zero?

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
```diff
@@ -0,0 +1,133 @@
...
+    public void setStateDiff(Map<DataWord, DataWord> stateDiff) {
+        this.stateDiff = stateDiff;
+    }
+
+    public RskAddress getAddress() {
+        return address;
+    }
+
+    public void setAddress(RskAddress address) {
+        this.address = address;
```
fmacleal, line 68: question:

Can the address have any value? Can it be null?

fmacleal (changes requested):
Good job.

I have a few comments.
I am also wondering if we could have a simple test class for the `AccountOverride` if some of the set needs some logic, we could have tests for this.

If we agree in the segregation of the business logic from the model, I would also suggest to create a test class for this. Very dummy, but in a way that we could expand in the next PR.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
asoto-iov: I think so

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
asoto-iov: Yep it could be null. About the negative the transaction execution would fail, but it is something the user should expect.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
asoto-iov: Nope, there is an if condition to make it fail during the apply but if you prefer I can add it to the constructor and make the validation there.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
asoto-iov: I actually started thinking along the same lines but decided to keep it simple at first to avoid over-engineering, since it’s a small and localized use case.

That said, I agree with your points — I’ll refactor it and implement the StateOverrideApplier as you suggested. Thanks

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
asoto-iov: It is the type used in the storage aswell that's why I'm using it, what is the problem with it?

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
asoto-iov: It is the missing feature that we were talking during the planing about the moving a precompiled to the given address. 
The implementation is more complex so we decided to tackle it separately.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
fmacleal: Would be interesting to be sure. Otherwise we might have weird behavior. 😅 
I am not sure that it can be negative, never seen a a balance below zero in a call transaction.
If isn't possible, we should verify here if balance is below zero and throw exception in case it happens.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
fmacleal: I would say that these validations should go to the model, no?
Do you think we should put it elsewhere and let it fail in a later state?

rskj-core/src/test/java/co/rsk/rpc/modules/eth/DefaultStateOverrideApplierTest.java
fmacleal, line 1: praise:

Nicee tests. Really liked. Congrats. 😄

rmoreliovlabs (approved):

fmacleal (dismissed):
Cool! Good job!

rskj-core/src/main/java/co/rsk/rpc/modules/eth/subscribe/AccountOverride.java
fmacleal: Addressed in current commits.

rskj-core/src/main/resources/reference.conf
```diff
@@ -406,6 +406,8 @@ sync {
 
 rpc {
     callGasCap = 50000000
+    # Allow StateOverride in eth_call
+    allowStateOverride = false
```
Vovchyk: suggestion (to be a little bit more specific): `allowCallStateOverride`

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -134,24 +144,49 @@ public Map<String, Object> bridgeState() throws IOException, BlockStoreException
...
     public String call(CallArgumentsParam argsParam, BlockIdentifierParam bnOrId) {
+        return call(argsParam, bnOrId, Collections.emptyList());
+    }
+
+    public String call(CallArgumentsParam argsParam, BlockIdentifierParam bnOrId, List<AccountOverride> accountOverrideList) {
+
+        boolean shouldPerformStateOverride = accountOverrideList != null && !accountOverrideList.isEmpty();
+
+        if (shouldPerformStateOverride && !allowStateOverride) {
+            throw new UnsupportedOperationException("State Override is not supported");
```
Vovchyk: maybe `throw invalidParamError("State override is not allowed");`?

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
```diff
@@ -152,6 +152,9 @@ public PrecompiledContracts(RskSystemProperties config,
         this.remascConfig = new RemascConfigFactory(RemascContract.REMASC_CONFIG).createRemascConfig(config.netName());
     }
 
+    public boolean precompiledContactExists(RskAddress address) {
```
Vovchyk, line 155: why didn't you want to make use of `PrecompiledContracts#getContractForAddress()`? It respects activations, handles `null` address - everything that we care about

Vovchyk (changes requested):

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
fmacleal: suggestion:

Agreed, it's safer use the existing method, since a contract might not be active and still exist.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -134,24 +144,49 @@ public Map<String, Object> bridgeState() throws IOException, BlockStoreException
...
+            mutableRepository = new MutableRepository(new TrieStoreImpl(new HashMapDB()), result.getFinalState());
+        }
+
+        if (shouldPerformStateOverride) {
+            if (mutableRepository == null) {
+                mutableRepository = (MutableRepository) repositoryLocator.snapshotAt(block.getHeader());
+            }
+            for (AccountOverride accountOverride : accountOverrideList) {
+                if (precompiledContracts.precompiledContactExists(accountOverride.getAddress())) {
+                    throw new InvalidParameterException("Precompiled contracts can not ve overridden");
```
fmacleal: fix:

There is a typo here. It should be: `Precompiled contracts can not be overridden`

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -134,24 +144,49 @@ public Map<String, Object> bridgeState() throws IOException, BlockStoreException
...
+                }
+                stateOverrideApplier.applyToRepository(mutableRepository, accountOverride);
+            }
+        }
```
fmacleal: suggestion:

The complexity of this chain of if stayed too big, it's very hard to read and follow the flow. I suggest to break it down in smaller methods with meaningful names. With the help of some AI agent we get to this refactor:

```
    public String call(CallArgumentsParam argsParam, BlockIdentifierParam bnOrId,
            List<AccountOverride> accountOverrideList) {
        validateStateOverrideSupport(accountOverrideList);

        ExecutionBlockRetriever.Result result = executionBlockRetriever.retrieveExecutionBlock(bnOrId.getIdentifier());
        Block block = result.getBlock();

        MutableRepository repository = prepareRepository(block, result.getFinalState(), accountOverrideList);
        CallArguments callArgs = argsParam.toCallArguments();

        String hReturn = null;
        try {
            ProgramResult programResult = executeCall(callArgs, block, repository);
            handleTransactionRevertIfHappens(programResult);
            hReturn = HexUtils.toUnformattedJsonHex(programResult.getHReturn());
            return hReturn;
        } finally {
            LOGGER.debug("eth_call(): {}", hReturn);
        }
    }

    private void validateStateOverrideSupport(List<AccountOverride> accountOverrideList) {
        boolean shouldPerformStateOverride = accountOverrideList != null && !accountOverrideList.isEmpty();
        if (shouldPerformStateOverride && !allowStateOverride) {
            throw new UnsupportedOperationException("State Override is not supported");
        }
    }

    private MutableRepository prepareRepository(Block block, Trie finalState,
            List<AccountOverride> accountOverrideList) {
        MutableRepository repository = createRepositoryFromFinalState(finalState);

        if (shouldApplyStateOverrides(accountOverrideList)) {
            repository = ensureMutableRepository(repository, block);
            applyStateOverrides(repository, accountOverrideList);
        }

        return repository;
    }

    private MutableRepository createRepositoryFromFinalState(Trie finalState) {
        if (finalState != null) {
            return new MutableRepository(new TrieStoreImpl(new HashMapDB()), finalState);
        }
        return null;
    }

    private boolean shouldApplyStateOverrides(List<AccountOverride> accountOverrideList) {
        return accountOverrideList != null && !accountOverrideList.isEmpty();
    }

    private MutableRepository ensureMutableRepository(MutableRepository repository, Block block) {
        if (repository == null) {
            return (MutableRepository) repositoryLocator.snapshotAt(block.getHeader());
        }
        return repository;
    }

    private void applyStateOverrides(MutableRepository repository, List<AccountOverride> accountOverrideList) {
        for (AccountOverride accountOverride : accountOverrideList) {
            validatePrecompiledContractOverride(accountOverride);
            stateOverrideApplier.applyToRepository(repository, accountOverride);
        }
    }

    private void validatePrecompiledContractOverride(AccountOverride accountOverride) {
        if (precompiledContracts.precompiledContactExists(accountOverride.getAddress())) {
            throw new InvalidParameterException("Precompiled contracts can not be overridden");
        }
    }

    private ProgramResult executeCall(CallArguments callArgs, Block block, MutableRepository repository) {
        return repository != null ? callConstant(callArgs, block, repository) : callConstant(callArgs, block);
    }
```

rskj-core/src/test/java/co/rsk/rpc/modules/eth/EthModuleTest.java
```diff
@@ -102,14 +108,198 @@ void callSmokeTest() {
...
+        RepositorySnapshot snapshot = new MutableRepository(new TrieStoreImpl(new HashMapDB()), new Trie());
+        when(repositoryLocator.snapshotAt(any())).thenReturn(snapshot);
+
+        ReversibleTransactionExecutor executor = mock(ReversibleTransactionExecutor.class);
+        when(executor.executeTransaction(any(),eq(block), any(), any(), any(), any(), any(), any(), any()))
+                .thenReturn(executorResult);
+
+        EthModule eth = new EthModule(
+                null,
+                anyByte(),
```
fmacleal: suggestion:

Substitute this for a real byte, this is supposed to be used in assertions. Substitute by any byte like: `(byte) 0` or `(byte) 1`. Or, create an auxiliar method to generate a method for you.

rskj-core/src/test/java/co/rsk/rpc/modules/eth/EthModuleTest.java
```diff
@@ -102,14 +108,198 @@ void callSmokeTest() {
...
+    @Test
+    void testCall_whenCalledWithAccountOverrideListButAccountOverrideIsNotAllowed_throwsExceptionAsExpected() {
+        // Given
+        CallArgumentsParam callArgumentsParam = TransactionFactoryHelper.toCallArgumentsParam(new CallArguments());
+        BlockIdentifierParam blockIdentifierParam = new BlockIdentifierParam("latest");
+        List<AccountOverride> accountOverrideList = List.of(new AccountOverride(TestUtils.generateAddress("test")));
+
+        EthModule eth = new EthModule(
+                null,
+                anyByte(),
```
fmacleal: fix:

same comment from above

rskj-core/src/test/java/co/rsk/rpc/modules/eth/EthModuleTest.java
```diff
@@ -102,14 +108,198 @@ void callSmokeTest() {
...
+                .thenReturn(blockResultMock);
+        when(blockResultMock.getBlock()).thenReturn(blockMock);
+        when(blockResultMock.getFinalState()).thenReturn(new Trie());
+
+        PrecompiledContracts precompiledContractsMock = mock(PrecompiledContracts.class);
+        when(precompiledContractsMock.precompiledContactExists(address)).thenReturn(true);
+
+        EthModule eth = new EthModule(
+                null,
+                anyByte(),
```
fmacleal: fix:

same comment from above

rskj-core/src/test/java/co/rsk/rpc/modules/eth/EthModuleTest.java
```diff
@@ -102,14 +108,198 @@ void callSmokeTest() {
...
+        RepositorySnapshot snapshot = new MutableRepository(new TrieStoreImpl(new HashMapDB()), new Trie());
+        when(repositoryLocator.snapshotAt(any())).thenReturn(snapshot);
+
+        ReversibleTransactionExecutor executor = mock(ReversibleTransactionExecutor.class);
+        when(executor.executeTransaction(any(),eq(block), any(), any(), any(), any(), any(), any(), any()))
+                .thenReturn(executorResult);
+
+        EthModule eth = new EthModule(
+                null,
+                anyByte(),
```
fmacleal: fix:

same comment from above

rskj-core/src/test/java/co/rsk/rpc/modules/eth/EthModuleTest.java
fmacleal, line 1: praise:

Good tests here, I just suggest to separate the tests adding the parts. `//given //when //then` so it's clear the setup of the test, when you are calling and the result expected.

rskj-core/src/test/java/org/ethereum/rpc/Web3ImplUnitTest.java
fmacleal, line 1: comment:

I don't get the changes here, is it because you have autoformat set?

rskj-core/src/test/java/org/ethereum/vm/PrecompiledContractTest.java
```diff
@@ -353,4 +355,40 @@ void blake2fTestOnNotActivatedHardFork() {
...
+        // Given
+        RskAddress address = TestUtils.generateAddress("0000000000000000000000000000000000000000");
+
+        // When
+        boolean result = precompiledContracts.precompiledContactExists(address);
+
+        // Then
+        assertFalse(result);
+    }
+
```
fmacleal: suggestion:

Once you remove the method you created from the PrecompiledContracts class, these tests can be removed as well.

fmacleal (commented):
Godo job! 

I liked from the tests added and the changes seemed to be what we agreed, congrats.
I have added some comments.

rskj-core/src/main/resources/reference.conf
nagarev: Done!

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
nagarev: Done!

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
nagarev: I'll update that!

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
nagarev: Done!

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
nagarev: I did a similar refactor to that. Thanks!

rskj-core/src/test/java/co/rsk/rpc/modules/eth/EthModuleTest.java
nagarev: Done in every place, including the already existing test.

fmacleal (approved):
LGTM, thanks for have addressed the comments. Good job!

fmacleal (dismissed):

fmacleal (approved):
Perfect, I think these changes are more than enough to cover the negative balance and nonce.

rskj-core/src/main/java/org/ethereum/vm/PrecompiledContracts.java
VaWheel: should this method be removed as not used anywhere?

PR #3111, opened 2025-05-06
Implement log index calculation relative to the block

asoto-iov:
Refactors log index calculation to be block-relative instead of transaction-relative.

## Description
This PR implements a new way to calculate the log index relative to the block, instead of relative to each transaction.  

Previously, the log index was reset for each transaction in a block, like this:  
**Block 0:**
- Tx0:
  - log0: index 0  
  - log1: index 1  
- Tx1:
  - log0: index 0  
- Tx2:
  - log0: index 0  
  - log1: index 1  
  - log2: index 2  

With the new implementation, log indexing is continuous throughout the block:  
**Block 0:**
- Tx0:
  - log0: index 0  
  - log1: index 1  
- Tx1:
  - log0: index 2  
- Tx2:
  - log0: index 3  
  - log1: index 4  
  - log2: index 5  
 
## Motivation and Context
This change increases compatibility with Ethereum and Geth clients, which expect log indexes to be continuous within a block.  

It affects RPC methods that return logs, such as:
- `eth_getLogs`
- `eth_getTransactionReceipt`
- `debug_traceBlockByNumber`
- `debug_traceTransaction`

## How Has This Been Tested?
Unit and integration tests were added/updated to ensure the new log index behavior is correct and consistent across blocks and transactions. Compatibility was also verified against expected Ethereum behavior.

## Types of changes

## Checklist:

* **Other information**:


rmoreliovlabs (approved):
According to the analysis the solution looks good to me.

Vovchyk (changes requested):

Vovchyk:
lgtm, mostly minor things

rskj-core/src/jmh/java/co/rsk/jmh/runners/BenchmarkLogIndexRunner.java
```diff
@@ -0,0 +1,18 @@
...
+import co.rsk.jmh.web3.LogIndexBenchmark;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.CommandLineOptionException;
+import org.openjdk.jmh.runner.options.Options;
+
+public class BenchmarkLogIndexRunner {
+
+    public static void main(String[] args) throws CommandLineOptionException, RunnerException {
+        Options opt = OptionsHelper.createE2EBuilder(args, "results.csv")
```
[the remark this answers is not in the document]
asoto-iov: sure

rskj-core/src/jmh/java/co/rsk/jmh/web3/plan/LogIndexPlan.java
```diff
@@ -0,0 +1,67 @@
...
+import org.web3j.protocol.core.DefaultBlockParameterNumber;
+import org.web3j.protocol.core.methods.request.EthFilter;
+
+import java.util.Collections;
+
+@State(Scope.Benchmark)
+public class LogIndexPlan extends BasePlan{
+
+    private EthFilter blockRangeAndTopicFilter;
+    private EthFilter bockHashFilter;
```
fmacleal, line 36: suggestion:

typo in the variable name. Probably it should be `blockHashFilter`.

rskj-core/src/main/java/co/rsk/rpc/modules/debug/trace/call/CallTracer.java
```diff
@@ -86,17 +87,31 @@ public JsonNode traceTransaction(@Nonnull String transactionHash, @Nonnull Trace
         txInfo.setTransaction(tx);
 
         ProgramTraceProcessor programTraceProcessor = new ProgramTraceProcessor();
-        this.blockExecutor.traceBlock(programTraceProcessor, VmConfig.LIGHT_TRACE, block, parent.getHeader(), false, false);
-
+        BlockResult blockResult = this.blockExecutor.traceBlock(programTraceProcessor, VmConfig.LIGHT_TRACE, block, parent.getHeader(), false, false);
+        if(blockResult == null || blockResult.getExecutedTransactions() != null && blockResult.getExecutedTransactions().isEmpty()) {
+            logger.error("Unsuccessful block execution for transaction: {}", transactionHash);
+            return null;
```
fmacleal, line 93: suggestion:

Maybe we could create and throw a proper exception here? Will we return an empty json? Wouldn't be better to return a proper JSON error?

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -589,6 +594,17 @@ public void execute(@Nonnull Runnable command) {
         return result;
     }
 
+    private List<TransactionReceipt> updateReceipts(List<TransactionReceipt> receipts) {
```
fmacleal, line 597: suggestion:

Does it worth we use immutability here and create a new list with the copy of the receipts to avoid side effects?

fmacleal (commented):
Good job!

I have few suggestions or questions.

Apart the commented part,  in general, I would suggest also breakdown or extract some loop logics with meaningful methods names to improve readability.

But this is more nitpick, no need to add. 😅

rskj-core/src/main/java/co/rsk/rpc/modules/debug/trace/call/CallTracer.java
fmacleal: Makes sense, let's keep like this then.

fmacleal (dismissed):
LGTM

Vovchyk (approved):

fmacleal (approved):
LGTM

PR #3123, opened 2025-05-13
Add get/set/save method for weiTransferredToUnionBridge

nathanieliov:
- Add get/set/save method for weiTransferredToUnionBridge
- Update save method to store weiTransferredToUnionBridge when value exists
- Add tests for get/set weiTransferredToUnionBridge methods

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeStorageProvider.java
+++ b/rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeStorageProvider.java
@@ -10,5 +10,7 @@ public interface UnionBridgeStorageProvider {
     Optional<RskAddress> getAddress(ActivationConfig.ForBlock activations);
     void setLockingCap(Coin lockingCap);
     Optional<Coin> getLockingCap(ActivationConfig.ForBlock activations);
+    Optional<co.rsk.core.Coin> getWeisTransferredToUnionBridge(ActivationConfig.ForBlock activations);
+    void setWeisTransferredToUnionBridge(co.rsk.core.Coin weisTransferred);
     void save(ActivationConfig.ForBlock activations);
 }
--- a/rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeStorageProviderImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeStorageProviderImpl.java
@@ -1,6 +1,7 @@
 package co.rsk.peg.union;
 
 import static java.util.Objects.isNull;
+import static java.util.Objects.nonNull;
 
 import co.rsk.bitcoinj.core.Coin;
 import co.rsk.core.RskAddress;
@@ -18,6 +19,7 @@ public class UnionBridgeStorageProviderImpl implements UnionBridgeStorageProvide
 
     private RskAddress unionBridgeAddress;
     private Coin unionBridgeLockingCap;
+    private co.rsk.core.Coin weisTransferredToUnionBridge;
 
     public UnionBridgeStorageProviderImpl(StorageAccessor bridgeStorageAccessor) {
         this.bridgeStorageAccessor = bridgeStorageAccessor;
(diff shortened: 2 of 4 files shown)


rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeStorageProviderImpl.java
```diff
@@ -44,6 +45,14 @@ public void save(ActivationConfig.ForBlock activations) {
                 BridgeSerializationUtils::serializeCoin
             );
         }
+
+        if (!isNull(weiTransferredToUnionBridge)) {
```
jeremy-then, line 49: I think that using `if (nonNull(weiTransferredToUnionBridge)) {` is better here since you are already using that package.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeStorageProviderImpl.java
```diff
@@ -91,4 +100,26 @@ public Optional<Coin> getLockingCap(ForBlock activations) {
             ))
         );
     }
+
+    @Override
+    public void setWeiTransferredToUnionBridge(co.rsk.core.Coin weiTransferred) {
+        if (weiTransferred != null && weiTransferred.compareTo(co.rsk.core.Coin.ZERO) < 0) {
```
jeremy-then, line 106:
```suggestion
        if (nonNull(weiTransferred) && weiTransferred.compareTo(co.rsk.core.Coin.ZERO) < 0) {
```

Since you are already using that package in this file.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeStorageProviderImpl.java
```diff
@@ -91,4 +100,26 @@ public Optional<Coin> getLockingCap(ForBlock activations) {
...
+            throw new IllegalArgumentException("Wei transferred to Union Bridge cannot be negative");
+        }
+        this.weiTransferredToUnionBridge = weiTransferred;
+    }
+
+    @Override
+    public Optional<co.rsk.core.Coin> getWeiTransferredToUnionBridge(ForBlock activations) {
+        if (!activations.isActive(ConsensusRule.RSKIP502)) {
+            return Optional.empty();
```
jeremy-then, line 115: Probably no big deal but, should this throw an exception instead? This should never be attempted to be called prior to RSKIP502.

jeremy-then (approved):
LGTM. Left a couple of optional comments.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeStorageProviderImpl.java
nathanieliov: This comment is also related to Marcos' comment. Let's discuss this further to find an appropriate solution.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeStorageProvider.java
```diff
@@ -10,5 +10,7 @@ public interface UnionBridgeStorageProvider {
     Optional<RskAddress> getAddress(ActivationConfig.ForBlock activations);
     void setLockingCap(Coin lockingCap);
     Optional<Coin> getLockingCap(ActivationConfig.ForBlock activations);
+    Optional<co.rsk.core.Coin> getWeisTransferredToUnionBridge(ActivationConfig.ForBlock activations);
+    void setWeisTransferredToUnionBridge(co.rsk.core.Coin weisTransferred);
```
marcos-iov, line 14: Idea: if instead of a setter method we have 2 separate methods `increase` and `decrease`? Feel like that is less error prone and gives us control on how we manipulate this value. A setter method let's set any value you want any time

nathanieliov: It looks good to me to have some constraints in the storage layer to reduce the risk of making a mistake with a critical amount of balance. This implies a refactor that I would like to add once we have the `no-rskip-check` refactor done. Wdyt?

PR #3125, opened 2025-05-13
Make P2shP2wshErpFederationBuilder to have 20 members

julia-zack:


rskj-core/src/main/java/co/rsk/peg/federation/StandardMultisigFederation.java
```diff
@@ -47,9 +44,4 @@ public Script getRedeemScript() {
 
         return redeemScript;
     }
-
-    private void validateRedeemScriptSize() {
-        Script redeemScript = this.getRedeemScript();
-        ScriptValidations.validateScriptSigElementSize(redeemScript);
-    }
```
julia-zack, line 54: When des/erializing a p2sh-p2wsh erp segwit fed, we just des/serialize the standard multisig part. 
When having 20 members, this validation was throwing an error since a fed with 20 members will have a redeem script that exceeds the script sig elements allowed size.

So this had to be removed in order to be able to deserialize the p2sh-p2wsh erp segwit fed

rskj-core/src/test/java/co/rsk/peg/federation/StandardMultisigFederationTest.java
```diff
@@ -85,10 +85,10 @@ void createInvalidFederation_aboveMaxScriptSigSize() {
...
         NetworkParameters networkParameters = federation.getBtcParams();
         FederationArgs federationArgs = new FederationArgs(newMembers, creationTime, creationBlockNumber, networkParameters);
-        ScriptCreationException exception =
-            assertThrows(ScriptCreationException.class,
-                () -> FederationFactory.buildStandardMultiSigFederation(federationArgs));
-        assertEquals(ABOVE_MAX_SCRIPT_ELEMENT_SIZE, exception.getReason());
+        // this is intentional since the serialization of federations just serializes the standard part
+        // when adding segwit, this thrown an exception since the p2sh-p2wsh erp federation with 20 members
+        // exceeded the redeem script size limit. So the decision was to remove the size limit check.
+        assertDoesNotThrow(() -> FederationFactory.buildStandardMultiSigFederation(federationArgs));
```
nathanieliov, line 91: I think this comment should go into the `des/serialization` method unit test instead. This border case should be tested for `co.rsk.peg.BridgeSerializationUtils#deserializeFederationAccordingToVersion`  method, passing 
a `STANDARD_MULTISIG_FEDERATION` and  a `P2SH_P2WSH_ERP_FEDERATION` federation with 20 members.

Besides that, what if `deserializeStandardMultisigFederationWithDeserializer` method do not assume that any fed can be an `standardMultiSig` and return just the fed args? By doing that, we don't need to remove the validation size.

PR #3127, opened 2025-05-14
Add `requestUnionRbtc` method to `UnionBridgeSupport`

nathanieliov:
Add `requestUnionRbtc` method to `UnionBridgeSupport`

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
@@ -26,8 +26,12 @@ public class UnionBridgeSupportImpl implements UnionBridgeSupport {
     private final UnionBridgeStorageProvider storageProvider;
     private final SignatureCache signatureCache;
 
-    public UnionBridgeSupportImpl(ActivationConfig.ForBlock activations, UnionBridgeConstants constants, UnionBridgeStorageProvider storageProvider,
-        SignatureCache signatureCache) {
+    public UnionBridgeSupportImpl(
+        ActivationConfig.ForBlock activations,
+        UnionBridgeConstants constants,
+        UnionBridgeStorageProvider storageProvider,
+        SignatureCache signatureCache
+    ) {
         this.activations = activations;
         this.constants = constants;
         this.storageProvider = storageProvider;
@@ -89,11 +93,10 @@ public class UnionBridgeSupportImpl implements UnionBridgeSupport {
     }
 
     private boolean isUnauthorizedCaller(Transaction tx) {
-        // Check if the caller is isUnauthorizedCaller to set a new bridge contract address
         AddressBasedAuthorizer authorizer = constants.getChangeAuthorizer();
         boolean isUnauthorizedCaller = !authorizer.isAuthorized(tx, signatureCache);
         if (isUnauthorizedCaller) {
-            String baseMessage = String.format("Caller is not authorized to update union bridge contract address. Caller address: %s", tx.getSender());
+            String baseMessage = String.format("Caller is not authorized to execute this method. Caller address: %s", tx.getSender());
             logger.warn(LOG_PATTERN, "isUnauthorizedCaller", baseMessage);
         }
 
@@ -144,6 +147,60 @@ public class UnionBridgeSupportImpl implements UnionBridgeSupport {
         logger.info("[{}] Union Locking Cap has been increased. New value: {}",
             INCREASE_LOCKING_CAP_TAG, newCap.value);
         return UnionResponseCode.SUCCESS;
+    }
+
+    @Override
+    public UnionResponseCode requestUnionRbtc(Transaction tx, co.rsk.core.Coin amount) {
+        final String REQUEST_UNION_RBTC_TAG = "requestUnionRbtc";
+
+        if (isCallerNoUnionBridgeContractAddress(tx)) {
+            return UnionResponseCode.UNAUTHORIZED_CALLER;
+        }
+
+        if (isInvalidAmount(amount)) {
+            return UnionResponseCode.INVALID_VALUE;
+        }
+
+        storageProvider.setWeisTransferredToUnionBridge(amount);
+        logger.info("[{}] Amount requested by the union bridge has been transferred. Amount Requested: {}.", REQUEST_UNION_RBTC_TAG, amount);
+        return UnionResponseCode.SUCCESS;
+    }
+
+    private boolean isCallerNoUnionBridgeContractAddress(Transaction tx) {
+        RskAddress unionBridgeContractAddress = getUnionBridgeContractAddress();
+        boolean isCallerNoUnionBridgeContractAddress = !tx.getSender(signatureCache).equals(unionBridgeContractAddress);
+        if (isCallerNoUnionBridgeContractAddress) {
+            String baseMessage = String.format("Caller is not the Union Bridge Contract Address. Caller address: %s, Union Bridge Contract Address: %s", tx.getSender(), unionBridgeContractAddress);
+            logger.warn(LOG_PATTERN, "isCallerNoUnionBridgeContractAddress", baseMessage);
+        }
+        return isCallerNoUnionBridgeContractAddress;
+    }
+
+    private boolean isInvalidAmount(co.rsk.core.Coin amountRequested) {
+        boolean isAmountNullOrGreaterThanZero = amountRequested == null || amountRequested.compareTo(co.rsk.core.Coin.ZERO) < 1;
+        if (isAmountNullOrGreaterThanZero) {
+            logger.warn(
+                "[{isInvalidAmount}] Amount requested cannot be negative or zero. Amount requested: {}", amountRequested);
+            return true;
+        }
+
+        co.rsk.core.Coin lockingCap = co.rsk.core.Coin.fromBitcoin(storageProvider.getLockingCap(activations)
+            .orElse(constants.getInitialLockingCap()));
+
+        co.rsk.core.Coin previousAmountRequested = storageProvider.getWeisTransferredToUnionBridge(activations)
+            .orElse(co.rsk.core.Coin.ZERO);
+
+        co.rsk.core.Coin newAmountRequested = previousAmountRequested.add(amountRequested);
+        boolean doesNewAmountAndPreviousAmountRequestedSurpassLockingCap =
+            newAmountRequested.compareTo(lockingCap) > 0;
+        if (doesNewAmountAndPreviousAmountRequestedSurpassLockingCap) {
+            logger.warn(
+                "[{isInvalidAmount}] New amount request + previous amount requested cannot be greater than the Union Locking Cap. Previous amount requested: {}. New amount request: {} . Union Locking Cap: {}", previousAmountRequested, newAmountRequested, lockingCap
+            );
+            return true;
+        }
+
+        return false;
     }
 
     private boolean isInvalidLockingCap(Coin newCap) {
--- a/rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
+++ b/rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
@@ -59,15 +59,15 @@ public class UnionBridgeSupportImpl implements UnionBridgeSupport {
         logger.info("[{}] Setting new union bridge contract address: {}", SET_UNION_BRIDGE_ADDRESS_TAG, unionBridgeContractAddress);
 
         // Check if the network is MAINNET as the contract address can only be set in testnet or regtest
-        if (isEnvironmentDisabled()) {
+        if (!isEnvironmentEnabled()) {
             return UnionResponseCode.ENVIRONMENT_DISABLED;
         }
 
-        if (isUnauthorizedCaller(tx)) {
+        if (!isAuthorizedCaller(tx)) {
             return UnionResponseCode.UNAUTHORIZED_CALLER;
         }
 
-        if (isInvalidAddress(unionBridgeContractAddress)) {
+        if (!isValidAddress(unionBridgeContractAddress)) {
             return UnionResponseCode.INVALID_VALUE;
         }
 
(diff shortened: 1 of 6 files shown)


jeremy-then (approved):
LGTM.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -146,6 +149,60 @@ public UnionResponseCode increaseLockingCap(Transaction tx, Coin newCap) {
...
 
+    @Override
+    public UnionResponseCode requestUnionRbtc(Transaction tx, co.rsk.core.Coin amount) {
+        final String REQUEST_UNION_RBTC_TAG = "requestUnionRbtc";
+
+        if (isCallerNoUnionBridgeContractAddress(tx)) {
+            return UnionResponseCode.UNAUTHORIZED_CALLER;
+        }
+
+        if (isInvalidAmount(amount)) {
```
marcos-iov, line 160: To be honest I find this negative method names hard to follow. Positive logic is much more easy to read and follow

`if (!isValidAmount) return`

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -89,11 +93,10 @@ private boolean isEnvironmentDisabled() {
     }
 
     private boolean isUnauthorizedCaller(Transaction tx) {
-        // Check if the caller is isUnauthorizedCaller to set a new bridge contract address
         AddressBasedAuthorizer authorizer = constants.getChangeAuthorizer();
         boolean isUnauthorizedCaller = !authorizer.isAuthorized(tx, signatureCache);
         if (isUnauthorizedCaller) {
-            String baseMessage = String.format("Caller is not authorized to update union bridge contract address. Caller address: %s", tx.getSender());
+            String baseMessage = String.format("Caller is not authorized to execute this method. Caller address: %s", tx.getSender());
```
marcos-iov, line 99: Why make it generic? Isn't this for a specific authorizer?

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -146,6 +149,60 @@ public UnionResponseCode increaseLockingCap(Transaction tx, Coin newCap) {
...
+        if (isCallerNoUnionBridgeContractAddress) {
+            String baseMessage = String.format("Caller is not the Union Bridge Contract Address. Caller address: %s, Union Bridge Contract Address: %s", tx.getSender(), unionBridgeContractAddress);
+            logger.warn(LOG_PATTERN, "isCallerNoUnionBridgeContractAddress", baseMessage);
+        }
+        return isCallerNoUnionBridgeContractAddress;
+    }
+
+    private boolean isInvalidAmount(co.rsk.core.Coin amountRequested) {
+        boolean isAmountNullOrGreaterThanZero = amountRequested == null || amountRequested.compareTo(co.rsk.core.Coin.ZERO) < 1;
+        if (isAmountNullOrGreaterThanZero) {
```
marcos-iov, line 181: So is a zero value allowed or not? Log message says one thing, code says another. Is there a test?

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -146,6 +149,60 @@ public UnionResponseCode increaseLockingCap(Transaction tx, Coin newCap) {
...
+            logger.warn(
+                "[{isInvalidAmount}] Amount requested cannot be negative or zero. Amount requested: {}", amountRequested);
```
marcos-iov, line 183: `{isInvalidAmount}`?

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -146,6 +149,60 @@ public UnionResponseCode increaseLockingCap(Transaction tx, Coin newCap) {
...
+            return true;
+        }
+
+        co.rsk.core.Coin lockingCap = co.rsk.core.Coin.fromBitcoin(storageProvider.getLockingCap(activations)
+            .orElse(constants.getInitialLockingCap()));
```
marcos-iov, line 188: Isn't it simpler to call getLockingCap()?

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -146,6 +149,60 @@ public UnionResponseCode increaseLockingCap(Transaction tx, Coin newCap) {
...
+
+        co.rsk.core.Coin previousAmountRequested = storageProvider.getWeisTransferredToUnionBridge(activations)
```
marcos-iov, line 190: Shouldn't the storage return 0 if there is no value?

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -146,6 +149,60 @@ public UnionResponseCode increaseLockingCap(Transaction tx, Coin newCap) {
...
+            .orElse(co.rsk.core.Coin.ZERO);
+
+        co.rsk.core.Coin newAmountRequested = previousAmountRequested.add(amountRequested);
+        boolean doesNewAmountAndPreviousAmountRequestedSurpassLockingCap =
+            newAmountRequested.compareTo(lockingCap) > 0;
+        if (doesNewAmountAndPreviousAmountRequestedSurpassLockingCap) {
+            logger.warn(
+                "[{isInvalidAmount}] New amount request + previous amount requested cannot be greater than the Union Locking Cap. Previous amount requested: {}. New amount request: {} . Union Locking Cap: {}", previousAmountRequested, newAmountRequested, lockingCap
```
marcos-iov, line 198: `[{isInvalidAmount}]` that's a strange format

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -146,6 +149,60 @@ public UnionResponseCode increaseLockingCap(Transaction tx, Coin newCap) {
...
+        co.rsk.core.Coin lockingCap = co.rsk.core.Coin.fromBitcoin(storageProvider.getLockingCap(activations)
+            .orElse(constants.getInitialLockingCap()));
+
+        co.rsk.core.Coin previousAmountRequested = storageProvider.getWeisTransferredToUnionBridge(activations)
+            .orElse(co.rsk.core.Coin.ZERO);
+
+        co.rsk.core.Coin newAmountRequested = previousAmountRequested.add(amountRequested);
+        boolean doesNewAmountAndPreviousAmountRequestedSurpassLockingCap =
+            newAmountRequested.compareTo(lockingCap) > 0;
+        if (doesNewAmountAndPreviousAmountRequestedSurpassLockingCap) {
```
marcos-iov, line 196: Is the corner case when exactly the locking cap value is requested tested?

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: Because this validation is not exclusive to the update union bridge contract address method, this validation is being used for updating the locking cap as well, and will be used for pausing the system too. 

The same authorizer is being used for each of the union methods that need authorization to be called.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: Refactored

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: The variable name was wrong. Updated. And yes, there is a ParameterizedTest that includes zero amount.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: updated

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: Yes, it is. The issue is that currently `getLockingCap ` checks the RSKIP and returns an optional value, so it implies checking if the value `isPresent` or adding an extra `else` returning the default value. I would opt for using an else statement, understanding that this is temporary until we eliminate the activations.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: That change is included in the refactor to eliminate activations and remove the optionals. There is already a specific task for the refactoring of the get/set weisTransferredToUnionBridge storage provider methods.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: Yes, there is. However, I added a new one to account for scenarios where weisTransferred already has a value, and a new amount is sent such that the sum of the two values equals the locking cap.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: Updated

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -81,36 +81,39 @@ public UnionResponseCode setUnionBridgeContractAddressForTestnet(@Nonnull Transa
         return UnionResponseCode.SUCCESS;
     }
 
-    private boolean isEnvironmentDisabled() {
+    private boolean isEnvironmentEnabled() {
         String currentNetworkId = constants.getBtcParams().getId();
 
-        boolean isEnvironmentDisable = currentNetworkId.equals(NetworkParameters.ID_MAINNET);
-        if (isEnvironmentDisable) {
+        boolean isMainnet = currentNetworkId.equals(NetworkParameters.ID_MAINNET);
```
marcos-iov: Instead of checking the environment that is not enabled, I'd a check to validate that the environment is either testnet or regtest.

It's a bit subtle, but we know in which environment this is enabled. For any other environment it's disabled. So the method returns `false` by default instead of `true`

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -81,36 +81,39 @@ public UnionResponseCode setUnionBridgeContractAddressForTestnet(@Nonnull Transa
...
+        boolean isAuthorized = authorizer.isAuthorized(tx, signatureCache);
+        if (!isAuthorized) {
             String baseMessage = String.format("Caller is not authorized to execute this method. Caller address: %s", tx.getSender());
-            logger.warn(LOG_PATTERN, "isUnauthorizedCaller", baseMessage);
+            logger.warn(LOG_PATTERN, "isAuthorizedCaller", baseMessage);
+            return false;
         }
 
-        return isUnauthorizedCaller;
+        return true;
```
marcos-iov: Similar here, I think it's better that the default return by the method is `false`. Only return true if the caller is authorized

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -81,36 +81,39 @@ public UnionResponseCode setUnionBridgeContractAddressForTestnet(@Nonnull Transa
...
-        if (isInvalidAddress) {
+        boolean isValidAddress = unionBridgeContractAddress != null && !unionBridgeContractAddress.equals(EMPTY_ADDRESS);
+        if (!isValidAddress) {
             String baseMessage = "Union Bridge Contract Address cannot be null or empty";
-            logger.warn(LOG_PATTERN, "isInvalidAddress", baseMessage);
+            logger.warn(LOG_PATTERN, "isValidAddress", baseMessage);
+            return false;
         }
-        return isInvalidAddress;
+        return true;
```
marcos-iov: same here

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -166,26 +169,25 @@ public UnionResponseCode requestUnionRbtc(Transaction tx, co.rsk.core.Coin amoun
...
-    private boolean isInvalidAmount(co.rsk.core.Coin amountRequested) {
-        boolean isAmountNullOrGreaterThanZero = amountRequested == null || amountRequested.compareTo(co.rsk.core.Coin.ZERO) < 1;
-        if (isAmountNullOrGreaterThanZero) {
+    private boolean isValidAmount(co.rsk.core.Coin amountRequested) {
+        boolean isAmountNullOrLessThanZero = amountRequested == null || amountRequested.compareTo(co.rsk.core.Coin.ZERO) < 1;
+        if (isAmountNullOrLessThanZero) {
             logger.warn(
-                "[{isInvalidAmount}] Amount requested cannot be negative or zero. Amount requested: {}", amountRequested);
-            return true;
+                "[isValidAmount] Amount requested cannot be negative or zero. Amount requested: {}", amountRequested);
```
marcos-iov, line 181: Message says the amount cannot be 0 but the validation allows it

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
```diff
@@ -195,35 +197,35 @@ private boolean isInvalidAmount(co.rsk.core.Coin amountRequested) {
...
             logger.warn(
-                "[isInvalidLockingCap] Attempted value tries to increase Union Locking Cap above its limit. Value attempted: {} . maxLockingCapIncreasedAllowed: {}",
+                "[isValidLockingCap] Attempted value tries to increase Union Locking Cap above its limit. Value attempted: {} . maxLockingCapIncreasedAllowed: {}",
                 newCap.value, maxLockingCapIncreaseAllowed.value);
-            return true;
+            return false;
         }
 
-        return false;
+        return true;
```
marcos-iov, line 223: same as other validation methods

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: What if we simply return the result of the validation? For instance:

```
private boolean isEnvironmentTestnetOrRegtest() {
        String currentNetworkId = constants.getBtcParams().getId();

        boolean isTestnetOrRegtest = currentNetworkId.equals(NetworkParameters.ID_TESTNET) || currentNetworkId.equals(NetworkParameters.ID_REGTEST);
        if (!isTestnetOrRegtest) {
            String baseMessage = String.format("Union Bridge Contract Address can only be set in Testnet and RegTest environments. Current network: %s", currentNetworkId);
            logger.warn(LOG_PATTERN, "isEnvironmentTestnetOrRegtest", baseMessage);
        }
        return isTestnetOrRegtest;
    }
```

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: Updated. Now is returning the result of the validation variable.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: Updated. Now is returning the result of the validation variable.

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: Variable name updated

rskj-core/src/main/java/co/rsk/peg/union/UnionBridgeSupportImpl.java
nathanieliov: I would say this does not apply to this method. There are two checks, each with its own log message and I cannot return true before having checked both.

marcos-iov (approved):

PR #3152, opened 2025-06-02
Remove standard federation size validation

marcos-iov:
Remove standard federation script size validation on creation

## Description
Federations are always serialized and deserialized as standard multisig federations. Because the standard multisig information is the important data to keep in storage. ERP related information is obtained from the Bridge constants.

This is also true for P2WSH-P2SH federations, but these federation can now have up to 20 members. So the script size validation done in the creation of standard multisig federations needs to be moved to `getRedeemScript` method. Basically, no validation is done when creating the federation, only whn getting the redeem script

## Types of changes

## Checklist:

* **Other information**:


jeremy-then (approved):
LGTM.

PR #3155, opened 2025-06-03
feat(snap): add caching for status request messages

Vovchyk:
## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
+++ b/rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
@@ -97,6 +97,9 @@ public class SnapshotProcessor implements InternalService, SnapProcessor {
     private volatile Boolean isRunning;
     private final Thread thread;
 
+    // Cache for last processed snap status request
+    private CachedSnapStatusData lastSnapStatusCache;
+
     @SuppressWarnings("java:S107")
     public SnapshotProcessor(Blockchain blockchain,
                              TrieStore trieStore,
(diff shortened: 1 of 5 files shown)


rmoreliovlabs (dismissed):

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
```diff
@@ -783,4 +814,42 @@ public void stop() {
...
+        }
+
+        public LinkedList<BlockDifficulty> getDifficulties() {
+            return difficulties;
+        }
+
+        public long getTrieSize() {
+            return trieSize;
+        }
+    }
```
fmacleal: suggestion:

Extract this to a separated file inside the package `co.rsk.net.messages` which contains a lot of class defined for this.

Instead of making a static class, I suggest use the new [record](https://medium.com/@houssemmedine.drissi/class-vs-record-difference-between-class-and-record-in-java-75d7ea6da145) type from recent java versions. It avoid the boiler plate of get methods and constructor and also it's immutable, already have the toString, equals, hashCode and so on. And is interesting for these cases of holding data. 

This whole static class would stay just like this:

```java
public record CachedSnapStatusData(Keccak256 blockHash ,
                                   LinkedList<Block> blocks,
                                   LinkedList<BlockDifficulty> difficulties,
                                   long trieSize) {
}
```

But you would also have to change the get to access the fields directly in the way that records generates, which doesn't have the get. It would be `snapStatusCache.blocks()` instead of snapStatusCache.getBlocks()`.

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
```diff
@@ -223,17 +226,41 @@ public void run() {
...
+            logger.debug("Using cached snap status data for block hash: {}", snapStatusCache.getBlockHash());
+            SnapStatusResponseMessage responseMessage = new SnapStatusResponseMessage(
+                    requestMessage.getId(),
+                    snapStatusCache.getBlocks(),
+                    snapStatusCache.getDifficulties(),
+                    snapStatusCache.getTrieSize()
+            );
+            sender.sendMessage(responseMessage);
+            return;
+        }
```
fmacleal: suggestion:

This could be extracted to a method and then we could check an optional and return if present. Something like:

```
private Optional<SnapStatusResponseMessage> getFromCache(
    CachedSnapStatusData cache, 
    Block checkpointBlock, 
    String requestId
) {
    if (cache != null && cache.blockHash().equals(checkpointBlock.getHash())) {
        logger.debug("Using cached snap status data for block hash: {}", cache.blockHash());
        return Optional.of(new SnapStatusResponseMessage(
            requestId,
            cache.blocks(),
            cache.difficulties(),
            cache.trieSize()
        ));
    }
    return Optional.empty();
}
```

Then we in the method would stay simpler to read:

```
    Optional<SnapStatusResponseMessage> cachedResponse = getFromCache(
        this.lastSnapStatusCache, 
        checkpointBlock, 
        requestMessage.getId()
    );
    if (cachedResponse.isPresent()) {
        sender.sendMessage(cachedResponse.get());
        return;
    }
```

rskj-core/src/test/java/co/rsk/net/SnapshotProcessorTest.java
```diff
@@ -622,6 +622,59 @@ void givenErrorRLPData_thenOnStateChunkErrorIsCalled() {
         verify(snapSyncState, times(1)).submitRequest(any(PeerSelector.class), any(RequestFactory.class));
     }
 
+    @Test
```
fmacleal, line 625: praise:

Cool! Good test! 🙂

fmacleal (approved):
LGTM, thanks for have followed the suggestions.

fmacleal (approved):

PR #3184, opened 2025-06-10
Adding consensus rules for rskip516 for ec secp256k1 curve

fmacleal:
Adding consensus rules for RSKIP516, planned to be activated in the next RSKj hardfork.

## Motivation and Context
We need the consensus rule in order o implement the logic associated to it.

## How Has This Been Tested?
We added test classes for the new consensus rule

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ConsensusRule.java
+++ b/rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ConsensusRule.java
@@ -105,7 +105,8 @@ public enum ConsensusRule {
     RSKIP453("rskip453"),
     RSKIP454("rskip454"),
     RSKIP459("rskip459"),
-    RSKIP460("rskip460")
+    RSKIP460("rskip460"),
+    RSKIP516("rskip516"), //  Addition of precompiled contracts for add and mul operations on the secp256k1 curve
     ;
 
     private final String configKey;
--- a/rskj-core/src/main/resources/config/testnet2.conf
+++ b/rskj-core/src/main/resources/config/testnet2.conf
@@ -14,7 +14,8 @@ blockchain.config {
         fingerroot500 = 4015800,
         arrowhead600 = 4927100,
         arrowhead631 = -1,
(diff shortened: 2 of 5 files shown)


rskj-core/src/main/java/co/rsk/core/types/bytes/Bytes.java
```diff
@@ -106,6 +106,14 @@ static boolean equalBytes(Bytes b1, Bytes b2) {
      * @return the wrapped by this instance byte array.
      */
     byte[] asUnsafeByteArray();
+
+    static void arraycopy(byte[] orig, int srcPos, byte[] dest, int destPos, int length) {
```
Vovchyk: seems this one duplicates the system one - `System.arraycopy(...)`

rmoreliovlabs (approved):

rskj-core/src/main/java/co/rsk/core/types/bytes/Bytes.java
Vovchyk: just an idea: if check for null needed, can be like this - Optional.ofNullable(orig).map()

Vovchyk (changes requested):

rskj-core/src/main/java/co/rsk/core/types/bytes/Bytes.java
rmoreliovlabs: This was added by me in a previous PR. I remember in the past when using some java 8+ features, that we don't want to use these features in critical place having a lot of calls by the node. Since this will be part of smart contract executions I didn't add it there.

Just to let know the reason why we did not add optionals for this method

fmacleal: Ok, changing the base branch, this code doesn't exist anymore. Maybe we could discuss this in slack and close this conversation?

Because, it indeed was added by another PR, my change for consensus rule it's way simpler. You can check it again now.

Vovchyk (approved):

PR #3208, opened 2025-06-19
Add temporary Trie Store creation in RskContext

rmoreliovlabs:
## Description
Add temporary Trie Store creation in RskContext

## Motivation and Context
Considering the snap sync protocol working theory, we will use a Temporary Trie Store to save the snap sync chunk messages and validate it. One we the validation succeeds , we can move all the data from the Temporary Trie Store to the Actual Trie Store

In this PR we are adding the changes to support the tmp trie in the SnapshotProcessor constructor

## How Has This Been Tested?

No, but test were updated properly and they are passing successfully.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -2094,12 +2095,24 @@ private SnapshotProcessor getSnapshotProcessor() {
...
 
+    private synchronized TrieStore getTtmpSnapSyncTrieStore() {
+        checkIfNotClosed();
+
+        if (tmpSnapSyncTrieStore == null) {
+            tmpSnapSyncTrieStore = buildAbstractTrieStore(Paths.get(getRskSystemProperties().databaseDir()).resolve(SnapshotProcessor.TMP_TRIE_DIR_NAME));
+        }
+
+        return tmpSnapSyncTrieStore;
+    }
```
fmacleal: suggestion:

This method is basically the method `getTrieStore` with the path being the difference. WDYT in re-use this method passing the different paths for each trie store? 

You could even have a method to resolve this path, which would abstract this complexity of path.

nitpick:
Try to break down this `Paths.get` you are adding a lot of functions chained and it's making the readability suffer.

issue:
typo in the name of the method, there is two `T`s for the method name.

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
```diff
@@ -74,8 +74,11 @@ public class SnapshotProcessor implements InternalService {
     public static final int BLOCK_CHUNK_SIZE = 400;
     public static final int BLOCKS_REQUIRED = 6000;
     public static final long CHUNK_ITEM_SIZE = 1024L;
+    public static final String TMP_TRIE_DIR_NAME = "snapSyncTmpTrie";
```
fmacleal: suggestion:
I think we might have an issue if this name is fixed, maybe we should have some timestamp on it, to avoid conflicts? What would happen if this folder already exists? Will we have an error?

 This name could be defined in the `RSKContext` and passed below, or even create a DTO using `record` with the path and the TrieStore, something like this:

```
public record TmpSnapshotTrieStore(TrieStore trieStore, Path tmpSnapshotPath) implements AutoCloseable {

    @Override
    public void close() {
        try {
            trieStore.dispose();
        } catch (Exception e) {
            throw new RuntimeException("Error closing TrieStore", e);
        }
    }
}
```

With that, we could have this path managed in the context and we would have access here on this class.

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
```diff
@@ -132,9 +137,13 @@ public SnapshotProcessor(Blockchain blockchain,
                       int maxSenderRequests,
                       boolean checkHistoricalHeaders,
                       boolean isParallelEnabled,
-                      @Nullable SyncMessageHandler.Listener listener) {
+                      @Nullable SyncMessageHandler.Listener listener,
+                      TrieStore tmpTrieStore,
+                      String databaseDir) {
```
fmacleal: comment:

If we have some kind of record( like a DTO) it would be a single object, since they are related, I think it makes more sense.

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
rmoreliovlabs: I don't think we need this, we are using the same mechanism for the current trie store it will work the same way. Also I don't think it is a good idea to have a different folder because that means we will not be able to resume the where we left the snap sync if the node is turned off or stopped by any reason. We can always use `--reset` flag is we want to start over.

Basically we will be following the same logic and use cases of the current trie store implementation

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
rmoreliovlabs: We could probably use database dir for other purpose later, I think it is a general purpose variable not only meant to be specifically used with the trie store.

For the future, I think I would suggest reading refacotring the whole parameters itself in the constructor I don't think it is a good idea at all to have more than 4 or 5 arguments in a function

rskj-core/src/main/java/co/rsk/RskContext.java
rmoreliovlabs: Regarding the suggestion, the variables and the path are different, and basically that is what the function is doing. 

The function that is actually solving containing the common behavior is `buildAbstractTrieStore`. `getTrieStore()` and `getTtmpSnapSyncTrieStore` only set different path and different variables, but they both use `buildAbstractTrieStore`

Regarding the nitpick, split the operation into multiple variables.

Ragarding the issue, addressed

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
fmacleal: Is never a good idea pass a parameter for a future use, I would rather to be more concise and just have the variables we really need. 

If you will need this and this parameter it's only for the temporary trie, I vouch we create an object like I suggested above. Otherwise this database dir will stay lost and can be used for other purposes, which is not the case here neither we want that.

I would like to stop us to just keep adding parameters not following good practices just because.

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -2094,12 +2095,26 @@ private SnapshotProcessor getSnapshotProcessor() {
...
+
+        if (tmpSnapSyncTrieStore == null) {
+            final var databasePath = Paths.get(getRskSystemProperties().databaseDir());
+            final var tmpDatabasePath = databasePath.resolve(SnapshotProcessor.TMP_TRIE_DIR_NAME);
+            tmpSnapSyncTrieStore = buildAbstractTrieStore(tmpDatabasePath);
+        }
+
+        return tmpSnapSyncTrieStore;
+    }
+
```
fmacleal: suggestion:

This method is practically the same from the `getTrieStore`, you could refactor this method to receive the database path and re-use the method.

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
fmacleal: We can discuss this in the daily, I would like to explore the idea and check what can happen if we have a conflicting name.

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
fmacleal: Ok, I double checked and noticed that we actually don't have databasedir, I thought it was present in the `SnapshotProcessor`. Makes sense to have this here then.

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -175,6 +178,7 @@ public class RskContext implements NodeContext, NodeBootstrapper {
     private org.ethereum.db.BlockStore blockStore;
     private NetBlockStore netBlockStore;
     private TrieStore trieStore;
+    private KeyValueDataSource tmpSnapSyncKeyValueDataSource;
```
Vovchyk: as we only need this tmp data source while snap is active, maybe we could create and store that instance not globally but, let's say, in `SnapSyncState`? The class responsible for creation / initialisation could be `KeyValueDataSourceUtils`

rmoreliovlabs:
> pls check the sonar issues

This will be merged in the [fix-snap-010](https://github.com/rsksmart/rskj/tree/fix-snap-010), the implementation will remove those sonar issues

rskj-core/src/main/java/co/rsk/RskContext.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -1330,6 +1330,12 @@ public synchronized void close() {
             logger.trace("wallet closed.");
         }
 
+        if (snapshotProcessor != null) {
```
Vovchyk, line 1333: do we need this? I believe it's being disposed in the node runner

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
```diff
@@ -73,8 +83,11 @@ public class SnapshotProcessor implements InternalService, SnapProcessor {
     public static final int BLOCK_CHUNK_SIZE = 400;
     public static final int BLOCKS_REQUIRED = 6000;
     public static final long CHUNK_ITEM_SIZE = 1024L;
+    public static final String TMP_NODES_DIR_NAME = "snapSyncTmpNodes";
     private final Blockchain blockchain;
     private final TrieStore trieStore;
+    private KeyValueDataSource tmpSnapSyncKeyValueDataSource;
```
Vovchyk: can it be SnapSyncState where we store this db instance, as the lifecycle of it is limited by the time of snap syncing? On the other hand, SnapshotProcessor stays alive until node shutdown. Alternatively, you could manage lifetime of tmpSnapSyncKeyValueDataSource in this class and close it right after snap finishes..

Vovchyk (changes requested):
pls check the sonar warnings

rskj-core/src/main/java/co/rsk/RskContext.java
rmoreliovlabs: We need to close the key value datasource when the node is closed

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
rmoreliovlabs: I think snapshotprocessor is fine because we should be able to keep the connection alive until snapshot processor stops working. We need to consider that when a node is fully sync, if it stays behind, snap sync will happen again, so we don't need to recreate the folder structure again and start the connection again

rskj-core/src/main/java/co/rsk/RskContext.java
Vovchyk: but it already gets closed. SnapshotProcessor implements InternalService, which is being managed, as any other service, by the node runner.

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
Vovchyk: this is the point. I don't think we need to preserve temporary db open for the entire node life. This db is only needed  while snap syncing is active, and this is basically a few hours. Afterwards, we do not need that tmp db. Also I believe the idea was to move the temporary db and substitute the main one in the "unitrie" folder after finishing snap and making all the checks - but for that you'd need to close the db, wouldn't you?

rmoreliovlabs: Addressed

PR #3214, opened 2025-06-24
Consider segwit output script in peg utils legacy

julia-zack:


rskj-core/src/main/java/co/rsk/peg/PegUtilsLegacy.java
julia-zack, line 1: BitcoinWrapperImpl from powpeg-node uses this impl for isValidPeginTx. See [here](https://github.com/rsksmart/powpeg-node/blob/master/src/main/java/co/rsk/federate/bitcoin/BitcoinWrapperImpl.java#L336)

julia-zack: a tx from the last retired fed to the new active (both segwit) will prob be rejected as a pegin since it would be from a multisig. We could register it manually but

PR #3231, opened 2025-06-27
Add tmp trie store implementation in snapshot processor

rmoreliovlabs:
## Description
Add tmp trie store implementation in snapshot processor

For validating if the trie structure is correct within the **Temporary Datasource:**

1. We need to confirm every node exists in the trie.
2. We need to make sure every children referenced from the parent node exists in the DB.
3. Sinc all messages are coming in chunks, we should consider that not all children might be in the same chunk. So if you don’t find a child in a chunk, we should check if we can find it in the DB.
4. If we don’t find a `hash` in the trie or the DB, then we can say the trie is invalid.
5. If all the checks are fine, then we move the node to the Actual Trie and delete it from the Temporary Datasource

When the validation is successful, we can move the content from the Temporary Datasource to the actual Trie Store. We need to make sure the temporary entries are deleted after data is moved.

## Motivation and Context
Considering the snap sync protocol working theory, we will use a Temporary Trie Store to save the snap sync chunk messages and validate it. One we the validation succeeds , we can move all the data from the Temporary Datasource to the Actual Trie Store

In this PR we are adding the changes to store snap chunks in temporarydatasource and when the snap sync finishes we move it to the Actual Trie Store and clean the Temporary Datasource.

## How Has This Been Tested?

Screenshot 2025-06-26 at 1 23 53 PM

Screenshot 2025-07-04 at 5 02 20 PM

Screenshot 2025-07-04 at 5 00 47 PM

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
```diff
@@ -78,7 +79,7 @@ public class SnapshotProcessor implements InternalService {
     private final Blockchain blockchain;
     private final TrieStore trieStore;
     private final KeyValueDataSource tmpSnapSyncKeyValueDataSource;
-    public static final int TMP_NODES_SIZE_KEY = -1;
+    private static final int TMP_NODES_SIZE_KEY = -1;
```
fmacleal: question:

why this value is `-1`? What is this value used for? Is it a seed? If it's a seed, wouldn't be better to be a random value?

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
fmacleal, line 1: issue:

The value `BLOCK_NUMBER_CHECKPOINT` in the begin of this class is never used. Please, delete it.

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
```diff
@@ -649,12 +650,23 @@ private void processOrderedStateChunkResponse(SnapSyncState state, Peer peer, Sn
...
             for (int i = 0; i < trieElements.size(); i++) {
                 final RLPElement trieElement = trieElements.get(i);
                 byte[] value = trieElement.getRLPData();
                 nodes.add(TrieDTO.decodeFromSync(value));
+
+                tmpSnapSyncKeyValueDataSource.put(ByteUtil.intToBytes(existingNodesSize + i), value);
             }
             nodes.get(0).setLeftHash(firstNodeLeftHash);
+
+            tmpSnapSyncKeyValueDataSource.put(ByteUtil.intToBytes(TMP_NODES_SIZE_KEY), ByteUtil.intToBytes(nodes.size() + existingNodesSize));
```
fmacleal: question:

What are you saving at this part? After you save all the tries you save the size of the nodes you have in the for plus the size of nodes you have saved?

rskj-core/src/main/java/co/rsk/trie/TrieDTOInOrderRecoverer.java
```diff
@@ -25,35 +25,51 @@
...
 
     private TrieDTOInOrderRecoverer() {
         throw new UnsupportedOperationException("This is a utility class and cannot be instantiated");
     }
 
-    public static Optional<TrieDTO> recoverTrie(TrieDTO[] trieCollection, Consumer<? super TrieDTO> processTrieDTO) {
-        Optional<TrieDTO> result = recoverSubtree(trieCollection, 0, trieCollection.length - 1, processTrieDTO);
-        result.ifPresent(processTrieDTO);
-        return result;
+    public static RecoverSubtreeResponse recoverTrie(Function<Integer, TrieDTO> getNodeFn, int trieCollectionSize, BiConsumer<? super TrieDTO, Integer> processTrieDTO) {
```
fmacleal: suggestion:

Maybe `getNodeFn` could have the index associated to it. Like `getNodeFnByIndex`.

rskj-core/src/main/java/co/rsk/trie/TrieDTOInOrderRecoverer.java
```diff
@@ -25,35 +25,51 @@
...
+        final var response = recoverSubtree(getNodeFn, 0, trieCollectionSize - 1, processTrieDTO);
+        final var result = response.node();
+        result.ifPresent(node -> processTrieDTO.accept(node, response.index()));
+
+        return response;
     }
 
-    private static Optional<TrieDTO> recoverSubtree(TrieDTO[] trieCollection, int start, int end, Consumer<? super TrieDTO> processTrieDTO) {
+    private static RecoverSubtreeResponse recoverSubtree(Function<Integer, TrieDTO> getNodeFn, int start, int end, BiConsumer<? super TrieDTO, Integer> processTrieDTO) {
```
fmacleal: comment:

I know it kinda make sense these changes, but it became very complex to get what was done. You have a function that applies, where the logic isn't in this class and a BiConsumer that is applied here that also isn't defined here.

Could it be refactored to let it clear? Or Some explanation of the logic that isn't here.

fmacleal (commented):
Good job!

I just have a few comments and questions.

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
rmoreliovlabs: I don't know about that variable but seems to be used

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
rmoreliovlabs: this is where the size of the nodes are stored, since we are storing things in a key / value map, we don't have the size of the nodes and RocksDB might create other keys in it's inner workings. We want that size to iterate all the nodes in the map because the key is the index of the node as the nodes were in a list

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
rmoreliovlabs: I am storing the size of the nodes when every chunk arrives. Each chunk comes with a new list of nodes to be added, so we need to keep the size of the nodes up to date

rskj-core/src/main/java/co/rsk/trie/TrieDTOInOrderRecoverer.java
rmoreliovlabs: sure

rskj-core/src/main/java/co/rsk/trie/TrieDTOInOrderRecoverer.java
rmoreliovlabs: I am using the same pattern as previously with the only difference that I am specifying how to get nodes and small change is sending also the index to be stored in the TrieDTO.

Of course I will add some documentation around it

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
fmacleal: But if it's a key value map, why not have it set as a random final value instead of a fix -1? The key is fixed, if a new key clashes with this one, it would fail in RocksDB instead of overwrite no?

The -1 it's predictable, maybe it can easily be changed by someone trying to exploit this no?
If I understood, this is the key where you will store the size of all nodes saved, is it?

rskj-core/src/main/java/co/rsk/net/SnapshotProcessor.java
fmacleal: From yesterday it wasn't, maybe after the rebase it started to be used.

PR #3262, opened 2025-07-15
Remove unclelist

rmoreliovlabs:
## Description

1. Get rid of uncleList from `SuperBlockFields`
2. Remove logic that populates the `uncleList` in the miner server
3. Rremove logic from consensus rules related to `uncleList`

## Motivation and Context

Due to recent changes in the Superchain spec, we don’t need `uncleList` to be part of super block fields any more, so it should be removed and everything else should be cleared up of it

It matters because of new requirements

## How Has This Been Tested?
Just running the node and updating existing tests

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
@@ -186,13 +186,12 @@ public class BlockExecutor {
         header.setTxExecutionSublistsEdges(result.getTxEdges());
 
         if (activationConfig.isActive(RSKIP481, block.getNumber())) {
-            Trio<Block, List<BlockHeader>, SuperBridgeEvent> superParentAndUnclesAndBridgeEvent =
+            final var superParentAndBridgeEvent =
                     FamilyUtils.findSuperParentAndUnclesAndBridgeEvent(blockStore, receiptStore, block, result.getTransactionReceipts());
 
             SuperBlockFields superBlockFields = makeSuperBlockFields(
-                    superParentAndUnclesAndBridgeEvent.getFirst(),
-                    superParentAndUnclesAndBridgeEvent.getMiddle(),
-                    superParentAndUnclesAndBridgeEvent.getLast()
+                    superParentAndBridgeEvent.superParent(),
+                    superParentAndBridgeEvent.superBridgeEvent()
             );
             block.setSuperChainFields(superBlockFields);
         }
@@ -201,15 +200,13 @@ public class BlockExecutor {
         profiler.stop(metric);
     }
 
-    private static SuperBlockFields makeSuperBlockFields(Block superParent, List<BlockHeader> uncles, SuperBridgeEvent event) {
+    private static SuperBlockFields makeSuperBlockFields(Block superParent, SuperBridgeEvent event) {
         Bytes superParentHash = superParent == null ? null : Bytes.of(superParent.getHash().getBytes());
         long superParentBlockNumber = superParent == null ? 0 : superParent.getSuperBlockFields().getBlockNumber() + 1;
-        List<BlockHeader> uncleList = superParent == null ? Collections.emptyList() : uncles;
 
         return new SuperBlockFields(
                 superParentHash,
                 superParentBlockNumber,
-                uncleList,
                 event
         );
     }
--- a/rskj-core/src/main/java/co/rsk/core/bc/FamilyUtils.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/FamilyUtils.java
@@ -39,6 +39,8 @@ import static java.lang.Math.max;
  */
 public class FamilyUtils {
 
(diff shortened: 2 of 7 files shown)


rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -191,7 +191,6 @@ private void fill(Block block, BlockResult result) {
 
             SuperBlockFields superBlockFields = makeSuperBlockFields(
                     superParentAndUnclesAndBridgeEvent.getFirst(),
-                    superParentAndUnclesAndBridgeEvent.getMiddle(),
```
fmacleal, line 194: suggestion:

Rename this field to remove the uncle reference. Maybe call it `superParentAndBridgeEvent`

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
fmacleal, line 1: issue:

I see that you are still using the method `FamilyUtils::findSuperParentAndUnclesAndBridgeEvent`. Shouldn't we change this method to [stop add the uncles](https://github.com/rsksmart/rskj/blob/60b5bce0a56cf2d7ba7ecc9f88f98476e1662c26/rskj-core/src/main/java/co/rsk/core/bc/FamilyUtils.java#L189) in the list?

It seems to me a waste of resources looking for uncles that won't be used to create the superBlock.

After remove it, shouldn't we also rename the method to `findSuperParentAndBridgeEvent`?

rskj-core/src/test/java/org/ethereum/rpc/Web3ImplLogsTest.java
```diff
@@ -847,7 +847,7 @@ void getLogsTwiceFromBlockchainWithCallContractAndFilterBySecondTopic() throws E
     @Test
     void getLogsFromBlockchainWithEventInContractCreationReturnsAsExpectedWithBlockHashFilter() throws Exception {
         addEventInContractCreation();
-        final String blockHash = "0xdd4b7155e16fa7d78305159ece846dcfc2cceaf249bd033715bf0c45af98d429";
+        final String blockHash = "0x810f136f3185c9b2583dc62472c09d2567c388d56af03442ca6ee99c679dfccf";
```
fmacleal, line 850: question:

why did we have to change this? Did the hash changed after you removed the uncles?

rmoreliovlabs: Yes

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
rmoreliovlabs: addressed

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
rmoreliovlabs: Good catch! Addressed

rskj-core/src/main/java/co/rsk/core/bc/FamilyUtils.java
```diff
@@ -39,6 +39,8 @@
  */
 public class FamilyUtils {
 
+    public record FindSuperParentAndAncestorsResponse (Block superParent, SuperBridgeEvent superBridgeEvent) {}
```
fmacleal: question:

Why the ancestors in the name if we don't have uncles but only super parent and superBridgeEvent?
Are the ancestors different from the parents?

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -186,13 +186,12 @@ private void fill(Block block, BlockResult result) {
         header.setTxExecutionSublistsEdges(result.getTxEdges());
 
         if (activationConfig.isActive(RSKIP481, block.getNumber())) {
-            Trio<Block, List<BlockHeader>, SuperBridgeEvent> superParentAndUnclesAndBridgeEvent =
+            final var superParentAndBridgeEvent =
                     FamilyUtils.findSuperParentAndUnclesAndBridgeEvent(blockStore, receiptStore, block, result.getTransactionReceipts());
```
fmacleal, line 190: suggestion:

The method does still have `Uncles` in the name although we don't have uncles anymore.

rskj-core/src/main/java/co/rsk/core/bc/FamilyUtils.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
rmoreliovlabs: Addressed

fmacleal (approved):

PR #3277, opened 2025-08-07
Add super block difficulty calculation

rmoreliovlabs:
Add super block difficulty calculation

## Description

Superchain Difficulty Retarget Function

Let m = 168. This is called the retargenting interval (don’t confuse it with the epoch). It is
defined to occur approximately once per week if the hashrate doesn’t change.

Let Te = 7*24*60*60 = 604800 (number of seconds in one week) be the expected retarget interval.

Let j be the superblock number. When (j % m == 0) the computation of a new difficulty will occur.

Let superBlock[j] be the superblock header with super block number j.

The algorithm to compute the new difficulty is the following:

Set Ta = (superBlock[j-m].timeStamp-superBlock[j].timeStamp). This is the actual time it took to
mine m superblocks.

Let C = Te / Ta . The value C is the correction factor.

newSuperDifficulty = oldSuperDifficulty * C

To avoid using floating point or biginteger arithmetic for the division by Te, the previous equation
is computed as:

C128 = 128 * Te / Ta

newSuperDifficulty = ( oldSuperDifficulty * C128 ) / 128

Where C128 is computed using 32-bit or 64-bit integer arithmetic. C128 is the factor C scaled by
128. Biginteger division by 128 can be computed with a right bit shift.

This new difficulty will affect blocks after block j. The difficulty of block j will not depend on the
retargeting result, so the superDifficulty field does not depend on the superblock timestamp.
The retarget algorithm is similar to Bitcoin’s. We define the maximum and minimum difficulty
adjustments as:
- Maximum C128 = 192 (50% increase in difficulty)
- Minimum C128 = 64 (50% decrease in difficulty)

This is to avoid unbounded difficulty [rising attacks](https://arxiv.org/pdf/1312.7013).

The minimum is not defined as the inverse of the maximum, such as in Bitcoin (4X and 0.25X).

## Motivation and Context
Superchain feature is the motivation for this change. We want to compute super blocks difficulty in a faster way keeping reliability.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/validators/BlockDifficultyRule.java
```diff
@@ -48,7 +51,12 @@ public boolean isValid(BlockHeader header, BlockHeader parent) {
             logger.warn("BlockDifficultyRule - block or parent are null");
             return false;
         }
-        BlockDifficulty calcDifficulty = difficultyCalculator.calcDifficulty(header, parent);
+
+        if (header.isSuper().orElse(false)) {
+            return true;
+        }
```
rmoreliovlabs: We need to check how we will validate super blocks, because the difficulty will be different when we reach to the retargeting interval

rskj-core/src/main/java/org/ethereum/validator/DifficultyRule.java
```diff
@@ -36,14 +37,20 @@ public class DifficultyRule extends DependentBlockHeaderRule {
...
         this.difficultyCalculator = difficultyCalculator;
+        this.blockStore = blockStore;
     }
 
     @Override
     public boolean validate(BlockHeader header, BlockHeader parent) {
-        BlockDifficulty calcDifficulty = difficultyCalculator.calcDifficulty(header, parent);
+        if (header.isSuper().orElse(false)) {
+            return true;
+        }
```
rmoreliovlabs: We need to check how we will validate super blocks, because the difficulty will be different when we reach to the retargeting interval

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -48,6 +58,27 @@ public BlockDifficulty calcDifficulty(BlockHeader header, BlockHeader parentHead
         return getBlockDifficulty(header, parentHeader);
     }
 
+    private Block getBlockForSuperBlockRetargetingInterval(BlockStore blockStore, BlockHeader blockHeader) {
+        var retargetingBlock = blockStore.getBlockByHash(blockHeader.getParentHash().getBytes());
+
+        for (int count = 1; count < constants.getSuperBlockRetargetingInterval().intValue() - 1; count++) {
```
nagarev: is there any reason for starting at "1" in the for loop and the subtraction on the upper bound?

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -48,6 +58,27 @@ public BlockDifficulty calcDifficulty(BlockHeader header, BlockHeader parentHead
...
+            retargetingBlock = blockStore.getBlockByHash(retargetingBlock.getParentHash().getBytes());
+        }
+
+        return retargetingBlock;
+    }
+
+    public BlockDifficulty calcSuperDifficulty(BlockStore blockStore, BlockHeader header, BlockHeader parentHeader) {
+        var shouldUpdateDifficulty = header.getNumber() % constants.getSuperBlockRetargetingInterval().longValue() == 0;
+        
+        if(shouldUpdateDifficulty) {
```
nagarev:
```suggestion
        if (shouldUpdateDifficulty) {
```

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -103,4 +148,38 @@ private static BlockDifficulty calcDifficultyWithTimeStamps(
...
     }
+
+    private static BlockDifficulty calcSuperDifficultyWithTimeStamps(
+            BlockDifficulty oldDifficulty,
+            BigInteger correctionFactor,
+            BigInteger superBlockDifficultyCorrectionFactor,
+            BigInteger superBlockDifficultyMinCorrectionFactor,
+            BigInteger superBlockDifficultyMaxCorrectionFactor) {
+
+        if(correctionFactor.compareTo(superBlockDifficultyMinCorrectionFactor) < 0) {
```
nagarev:
```suggestion
        if (correctionFactor.compareTo(superBlockDifficultyMinCorrectionFactor) < 0) {
```

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -103,4 +148,38 @@ private static BlockDifficulty calcDifficultyWithTimeStamps(
...
+            var newSuperDifficulty = oldDifficulty
+                    .multiply(superBlockDifficultyMinCorrectionFactor)
+                    .asBigInteger()
+                    .divide(superBlockDifficultyCorrectionFactor);
+
+            return new BlockDifficulty(newSuperDifficulty);
+        }
+
+        if(correctionFactor.compareTo(superBlockDifficultyMaxCorrectionFactor) > 0) {
```
nagarev:
```suggestion
        if (correctionFactor.compareTo(superBlockDifficultyMaxCorrectionFactor) > 0) {
```

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -103,4 +148,38 @@ private static BlockDifficulty calcDifficultyWithTimeStamps(
...
+            var newSuperDifficulty = oldDifficulty
+                    .multiply(superBlockDifficultyMaxCorrectionFactor)
+                    .asBigInteger()
+                    .divide(superBlockDifficultyCorrectionFactor);
+
+            return new BlockDifficulty(newSuperDifficulty);
+        }
+
+        var newSuperDifficulty = oldDifficulty
```
nagarev: maybe we could extract this operations to a new function that returns BlockDifficulty, and receives the correction factors, something like this:

```java
private BlockDifficulty getSuperDifficulty(oldDifficulty, correctionFactor, difficultyCorrectionFactor) {
    var newSuperDifficulty = oldDifficulty
                    .multiply(correctionFactor)
                    .asBigInteger()
                    .divide(difficultyCorrectionFactor);

     return new BlockDifficulty(newSuperDifficulty);
}
```

And then call that function inside the if blocks. WDYT of that?

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
rmoreliovlabs: Yes because we already started getting the retargeting block above in line 62

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
rmoreliovlabs: Addressed

nagarev (approved):
LGTM

rskj-core/src/main/java/co/rsk/RskContext.java
```diff
@@ -1142,7 +1142,7 @@ public synchronized BlockValidationRule getBlockValidationRule() {
                                     new PrevMinGasPriceRule(),
                                     new BlockParentNumberRule(),
                                     blockTimeStampValidationRule,
-                                    new BlockDifficultyRule(getDifficultyCalculator()),
+                                    new BlockDifficultyRule(getDifficultyCalculator(), getBlockStore()),
```
fmacleal: **issue (blocking)**: 

I don't think we need to change the current `BlockDifficultyRule`. The idea would be to add a specific `supperBlockDifficulty` field in the `SuperBlockFields` [class](https://github.com/rsksmart/rskj/blob/a825c101a35392f4ceb8b65a8cbc7bb1d19a0a40/rskj-core/src/main/java/co/rsk/core/bc/SuperBlockFields.java#L30) that would hold the information for the superBlock difficulty. And then in the `BlockExecutor` when we are filling the `superBlockFields` we should calculate the supper difficulty and fill it in the `superBlockFields`. The good thing in do that in this place it's because you will have the superParent already. 

Also, any validation regarding the supper difficulty, I would recommend to be separated from the regular difficulty. Would probably be better create a `BlockSuperDifficultyRule` for this, to be implemented in the class `SuperBlockRule` [here](https://github.com/rsksmart/rskj/blob/a825c101a35392f4ceb8b65a8cbc7bb1d19a0a40/rskj-core/src/main/java/co/rsk/validators/SuperBlockRule.java#L41).


**chore(blocking)**: 
As per the comment above, all the `getBlockStore()` here should be removed and this class unchanged.

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -36,7 +38,15 @@ public DifficultyCalculator(ActivationConfig activationConfig, Constants constan
         this.constants = constants;
     }
 
-    public BlockDifficulty calcDifficulty(BlockHeader header, BlockHeader parentHeader) {
+    public BlockDifficulty calcDifficulty(BlockStore blockStore, BlockHeader header, BlockHeader parentHeader) {
```
fmacleal: **issue(blocking)**: 

The idea is not substitute the regular block difficulty. The fields from the superblock will be independent and will be in the `superChainDataHash` as stated in the Superblocks document R3. 

**suggestion(blocking)**:
Move all this logic here that calculate a `supperDifficulty` to a specific SuperBlockDifficultyCalculator. That way we will have an independent class that deals with the Superchain difficulty.

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -48,6 +58,27 @@ public BlockDifficulty calcDifficulty(BlockHeader header, BlockHeader parentHead
...
 
+    private Block getBlockForSuperBlockRetargetingInterval(BlockStore blockStore, BlockHeader blockHeader) {
+        var retargetingBlock = blockStore.getBlockByHash(blockHeader.getParentHash().getBytes());
+
+        for (int count = 1; count < constants.getSuperBlockRetargetingInterval().intValue() - 1; count++) {
+            retargetingBlock = blockStore.getBlockByHash(retargetingBlock.getParentHash().getBytes());
+        }
+
+        return retargetingBlock;
+    }
```
fmacleal: **suggestion(blocking)**: 

Since this shouldn't be here, you probably need to move this to `BlockExecutor` [class](https://github.com/rsksmart/rskj/blob/663c8550a0f6786e0611c6709c0751695017d28e/rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java#L189). Once there, you will have in the `FamilyUtil::findSuperParentAndUnclesAndBridgeEvent` the super block already. You use it to get the previous difficulty to be used in the `superBlockFields` adding the difficulty.

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -61,6 +92,20 @@ private BlockDifficulty getBlockDifficulty(
...
     }
 
+    private BlockDifficulty getSuperBlockDifficulty(BlockHeader curBlockHeader, BlockHeader retargetingBlockHeader) {
+        var currentTimestamp = curBlockHeader.getTimestamp() - retargetingBlockHeader.getTimestamp();
+        var correctionFactor = constants.getSuperBlockExpectedRetargetIntervalInSeconds()
+                .divide(BigInteger.valueOf(currentTimestamp))
+                .multiply(constants.getSuperBlockDifficultyCorrectionFactor());
+
+        return calcSuperDifficultyWithTimeStamps(
+                retargetingBlockHeader.getDifficulty(),
```
fmacleal: **issue(blocking)**:

From what I understood, this is the difficulty from the block itself, and not from the superblock. We shouldn't use it. You should get the difficulty here from the `SuperBlockField`. A superblock, it's always a regular block, but a regular block might not be a superblock. What will state this is the presence of the `superBlockFields` inside this regular block.

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -103,4 +148,35 @@ private static BlockDifficulty calcDifficultyWithTimeStamps(
...
+
+        if (correctionFactor.compareTo(superBlockDifficultyMinCorrectionFactor) < 0) {
+            return computeSuperBlockDifficulty(oldDifficulty, superBlockDifficultyMinCorrectionFactor, superBlockDifficultyCorrectionFactor);
+        }
+
+        if (correctionFactor.compareTo(superBlockDifficultyMaxCorrectionFactor) > 0) {
+            return computeSuperBlockDifficulty(oldDifficulty, superBlockDifficultyMaxCorrectionFactor, superBlockDifficultyCorrectionFactor);
+        }
+
+        return computeSuperBlockDifficulty(oldDifficulty, correctionFactor, superBlockDifficultyCorrectionFactor);
```
fmacleal: **suggestion**:

Is it really necessary this indirection to call a method in three different places just to pass the correctionFactor differently? Why not just update the correction factor according the logic and do the math here? 🤔

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
```diff
@@ -103,4 +148,35 @@ private static BlockDifficulty calcDifficultyWithTimeStamps(
...
+
+    private static BlockDifficulty computeSuperBlockDifficulty(
+            BlockDifficulty oldDifficulty,
+            BigInteger correctionFactor,
+            BigInteger superBlockDifficultyCorrectionFactor
+    ) {
+        var newSuperDifficulty = oldDifficulty
+                .multiply(correctionFactor)
+                .asBigInteger()
+                .divide(superBlockDifficultyCorrectionFactor);
```
fmacleal: **suggestion**:

One of the advantages in the algorithm is the shift right instead of use division, it has a better performance. 

Maybe we should just use `BigInteger::shiftRight(7)` instead of divide.

rskj-core/src/main/java/co/rsk/mine/BlockToMineBuilder.java
```diff
@@ -255,7 +255,7 @@ private BlockHeader createHeader(
                 .setSuperBlockResolver(isSuperResolver)
                 .build();
 
-        newHeader.setDifficulty(difficultyCalculator.calcDifficulty(newHeader, newBlockParentHeader));
+        newHeader.setDifficulty(difficultyCalculator.calcDifficulty(blockStore, newHeader, newBlockParentHeader));
```
fmacleal: **chore**:

Remove once we change to not mix the difficulty from Superblock with the regular Superblock.

rskj-core/src/main/java/co/rsk/validators/BlockDifficultyRule.java
fmacleal: As stated above, it shouldn't be done here. The concept of difficulty validation from the block should not be mixed with the concept from the superblock.

rskj-core/src/main/java/co/rsk/validators/BlockDifficultyRule.java
fmacleal, line 1: **chore(blocking)**:

We probably need to revert the changes here, since there is no need to add anything from the superblock.

rskj-core/src/main/java/co/rsk/validators/BlockDifficultyRule.java
fmacleal: **question**:

I didn't quite get the intention here. If it's a `superblock` we would skip any validation? Couldn't this cause an issue in the chain accepting blocks not valid and maybe even a fork?

rskj-core/src/main/java/org/ethereum/config/Constants.java
```diff
@@ -61,6 +61,13 @@ public class Constants {
     private static final int MINIMUM_DIFFICULTY = 131072;
     private static final int TESTNET_AND_DEVNET_DURATION_LIMIT = 14;
     private static final int REGTEST_DURATION_LIMIT = 10;
+    private static final BigInteger SUPER_BLOCK_DIFFICULTY_RETARGETING_INTERVAL = BigInteger.valueOf(168L);
+    private static final BigInteger SUPER_BLOCK_EXPECTED_RETARGETING_INTERVAL_IN_SECONDS = BigInteger.valueOf(604800); // 1 week
+    private static final BigInteger SUPER_BLOCK_DIFFICULTY_CORRECTION_FACTOR = BigInteger.valueOf(128);
+    private static final BigInteger SUPER_BLOCK_DIFFICULTY_MIN_CORRECTION_FACTOR = BigInteger.valueOf(64);
+    private static final BigInteger SUPER_BLOCK_DIFFICULTY_MAX_CORRECTION_FACTOR = BigInteger.valueOf(192);
+    private static final BigInteger REGTEST_SUPER_BLOCK_DIFFICULTY_RETARGETING_INTERVAL = BigInteger.valueOf(3L);
+    private static final BigInteger REGTEST_SUPER_BLOCK_EXPECTED_RETARGETING_INTERVAL_IN_SECONDS = BigInteger.valueOf(10);
```
fmacleal: **thought**: 

I am wondering if we could add some of these parameters as config, maybe worth to check it internally with the union bridge team. That way, if we want to adjust the superchain, we wouldn't need to release a new version of the node. Also, we could have different configs for each network, mainnet, testnet and regtest.

rskj-core/src/main/java/org/ethereum/validator/DifficultyRule.java
fmacleal: **question:**

I didn't quite get the intention here. If it's a superblock we would skip any validation? Couldn't this cause an issue in the chain accepting blocks not valid and maybe even a fork?

rskj-core/src/test/java/co/rsk/core/DifficultyCalculatorTest.java
fmacleal, line 1: **praise**:

Nice tests. Once we create a specific SuperblockDifficultyCalculator we could use it here and it would be simpler, probably.

fmacleal (changes requested):
There are changes necessary, we don't want to mix the `SuperblockDifficulty` with the regular `BlockDifficulty`. All the comments are related with that and we will make this change more concise and associated to the superchain only without change the canonical chain behaviour. If you want, we can have a call to discuss better the matter.

rskj-core/src/main/java/co/rsk/RskContext.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
rmoreliovlabs: We will execute the same math operations, and the parameter is the only thing that changes, I think we will unnecessarily have duplicated the math operations

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
rmoreliovlabs: Addressded

rskj-core/src/main/java/co/rsk/mine/BlockToMineBuilder.java
rmoreliovlabs: Done

rskj-core/src/main/java/co/rsk/validators/BlockDifficultyRule.java
rmoreliovlabs: Done

rskj-core/src/main/java/co/rsk/validators/BlockDifficultyRule.java
rmoreliovlabs: Done

rmoreliovlabs: This is not needed anymore

rskj-core/src/main/java/org/ethereum/config/Constants.java
rmoreliovlabs: Yeah, let's check with them, this is an easy change in case this is something that could be modifiable from the config files

rskj-core/src/main/java/org/ethereum/validator/DifficultyRule.java
rmoreliovlabs: Not needed anymore

rskj-core/src/test/java/co/rsk/core/DifficultyCalculatorTest.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
```diff
@@ -187,12 +191,17 @@ private void fill(Block block, BlockResult result) {
...
 
         if (activationConfig.isActive(RSKIP481, block.getNumber())) {
-            final var superParentAndBridgeEvent =
-                    FamilyUtils.findSuperParentAndUnclesAndBridgeEvent(blockStore, receiptStore, block, result.getTransactionReceipts());
+            final var superParentAndBridgeEvent = FamilyUtils.findSuperParentAndBridgeEventAndSuperDifficulty(
+                    blockStore,
+                    receiptStore,
+                    block,
+                    result.getTransactionReceipts(),
+                    superDifficultyCalculator);
```
fmacleal: **question**:

Why is it needed to pass the calculator to the class that finds the superParent? Wouldn't be better to just use the found superParent here and calculate the difficulty afterwards here after we have the `FindSuperParentAndBridgeEventResponse`? You will have the `superParent` and the block `here`. 

Passing it below it's adding even more dependencies in the method `findSuperParentAndBridgeEventAndSuperDifficulty` no?

rskj-core/src/main/java/co/rsk/core/bc/FamilyUtils.java
```diff
@@ -175,17 +176,18 @@ public static Set<Keccak256> getFamily(BlockStore store, long blockNumber, Kecca
...
-            List<TransactionReceipt> blockReceipts) {
+            List<TransactionReceipt> blockReceipts,
+            SuperDifficultyCalculator superDifficultyCalculator) {
         Pair<Block, List<Block>> superParentAndAncestors = findSuperParentAndAncestors(blockStore, block.getHeader());
         Block superParent = superParentAndAncestors.getLeft();
         List<Block> ancestors = superParentAndAncestors.getRight();
 
         if (superParent == null) {
-            return new FindSuperParentAndBridgeEventResponse(null,  null);
+            return new FindSuperParentAndBridgeEventResponse(null,  null, BlockDifficulty.ONE);
```
fmacleal: **question**:

Why are we configuring the `superParentDifficulty` for such a lower value? Can't this cause any issue of generate a superBlock before the expected?

rskj-core/src/main/java/co/rsk/validators/SuperBlockRule.java
```diff
@@ -97,6 +100,13 @@ protected boolean areSuperBlockFieldsValid(Block block, SuperBlockFields superBl
             if (superBlockFields.getBlockNumber() != superParent.getSuperBlockFields().getBlockNumber() + 1) {
                 return false;
             }
+
```
fmacleal, line 106: **suggestion**: 

I am missing some logs here, could you add some for when the difficulty are not valid?

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
fmacleal: Yes, but I didn't suggest to duplicate the math operations, I suggested to have a single math operation but we calculate separately the correctionFactor and use the calculated correction factor in the math operation. With that, we wouldn't have indirections and would be clearer that is being doing without have to call three times the method just because the changed correction factor.

```
var finalCorrectionFactor = constants.getSuperBlockExpectedRetargetIntervalInSeconds()
                .divide(BigInteger.valueOf(currentTimestamp));

if (correctionFactor.compareTo(superBlockDifficultyMinCorrectionFactor) < 0) {
   finalCorrectionFactor = superBlockDifficultyMinCorrectionFactor;
}

if (correctionFactor.compareTo(superBlockDifficultyMaxCorrectionFactor) > 0) {
   finalCorrectionFactor = superBlockDifficultyMaxCorrectionFactor;
}
...
 var newSuperDifficulty =  oldDifficulty
                .multiply(correctionFactor)
                .asBigInteger();

        return new BlockDifficulty(newSuperDifficulty);
```

rskj-core/src/main/java/co/rsk/core/DifficultyCalculator.java
fmacleal: **issue**:
Now you are dividing twice, because you shifted to the right (which is the same than divide) and then you are dividing again by the superBlockDifficultyCorrectionFactor. The initial intention would be to substitute the division by the shift right.

image

rskj-core/src/main/java/org/ethereum/config/Constants.java
```diff
@@ -61,6 +61,14 @@ public class Constants {
     private static final int MINIMUM_DIFFICULTY = 131072;
     private static final int TESTNET_AND_DEVNET_DURATION_LIMIT = 14;
     private static final int REGTEST_DURATION_LIMIT = 10;
+    private static final BigInteger SUPER_BLOCK_DIFFICULTY_RETARGETING_INTERVAL = BigInteger.valueOf(168L);
+    private static final BigInteger SUPER_BLOCK_EXPECTED_RETARGETING_INTERVAL_IN_SECONDS = BigInteger.valueOf(604800); // 1 week
+    private static final BigInteger SUPER_BLOCK_DIFFICULTY_CORRECTION_FACTOR = BigInteger.valueOf(128);
+    private static final BigInteger SUPER_BLOCK_DIFFICULTY_MIN_CORRECTION_FACTOR = BigInteger.valueOf(64);
+    private static final BigInteger SUPER_BLOCK_DIFFICULTY_MAX_CORRECTION_FACTOR = BigInteger.valueOf(192);
+    private static final Integer SUPER_BLOCK_DIFFICULTY_CORRECTION_FACTOR_SHIFT = 7;
```
fmacleal, line 68: **issue**:

The SUPER_BLOCK_DIFFICULTY_CORRECTION_FACTOR and SUPER_BLOCK_DIFFICULTY_CORRECTION_FACTOR_SHIFT are basically the same thing. We don't need both, divide by 128 it's the same than shift right by 7. I suggested the shift because it's a cheaper operation, keeping it, we should remove the other.

rskj-core/src/test/java/co/rsk/core/SuperDifficultyCalculatorTest.java
fmacleal, line 1: **suggestion*:

Would nice to add a few more test scenarios, covering a bit more of edge cases.

rskj-core/src/main/java/co/rsk/core/bc/BlockExecutor.java
rmoreliovlabs: It makes sense, I didn't know this was only meant to look for parent.

It is not super clear for me since it also have super bridger events and super uncles in the past

Addressed the comment

rskj-core/src/main/java/co/rsk/core/bc/FamilyUtils.java
rmoreliovlabs: The default value of a Super Block is ONE, and also that is the value for the first super block when we do not have a super parent. If it is zero, any number multiplied by zero will result zero

rskj-core/src/main/java/co/rsk/validators/SuperBlockRule.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/org/ethereum/config/Constants.java
rmoreliovlabs: if the correction factor will use always a number that is power of 2, then this will be fine, but if we would like to use any numer, we can't use shift operation. Let me know if you think it is a good idea to rely only on numbers that are power of 2

rskj-core/src/test/java/co/rsk/core/SuperDifficultyCalculatorTest.java
rmoreliovlabs: sounds good

rskj-core/src/main/java/org/ethereum/config/Constants.java
fmacleal: Personally, I don't have a strong opinion, I am just suggesting the power of 2 as it's stated in the requirements document. And it seems that it's the way that Sergio thought in do it. so, I suggest to remove the 128 and use the power of 2 now. If we need to change this correction factor in the future, we can revisit.

Since we are calculating it frequently, it's better to stick with the cheapest operation.

rmoreliovlabs: Sounds good let's stick to the most convenient operation for now

PR #3289, opened 2025-08-19
Adding validations to state override

nagarev:
## Description

Adding validations to limit some state override parameters

## Types of changes

## Checklist:

* **Other information**:

N/A

### DIFF
--- a/rskj-core/src/main/java/co/rsk/rpc/modules/eth/AccountOverride.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/modules/eth/AccountOverride.java
@@ -34,6 +34,10 @@ import static org.ethereum.rpc.exception.RskJsonRpcRequestException.unimplemente
 
 public class AccountOverride {
 
+    public static final int MAX_ACCOUNT_OVERRIDES = 1;
+    public static final int MAX_CODE_SIZE = 1;
+    public static final int MAX_STATE_CHANGES = 1;
+
     private BigInteger balance;
     private Long nonce;
     private byte[] code;
@@ -76,6 +80,9 @@ public class AccountOverride {
     }
 
     public void setCode(byte[] code) {
+        if (code.length > MAX_CODE_SIZE) {
+            throw invalidParamError("Code length in bytes exceeded. Max " + MAX_CODE_SIZE);
+        }
         this.code = code;
     }
 
@@ -84,6 +91,9 @@ public class AccountOverride {
     }
 
     public void setState(Map<DataWord, DataWord> state) {
+        if (state.size() > MAX_STATE_CHANGES) {
+            throw invalidParamError("Number of state changes exceeded. Max " + MAX_STATE_CHANGES);
+        }
         this.state = state;
     }
 
@@ -105,33 +115,33 @@ public class AccountOverride {
 
     public AccountOverride fromAccountOverrideParam(AccountOverrideParam accountOverrideParam) {
 
-        if (accountOverrideParam.getMovePrecompileToAddress() != null) {
-            this.setMovePrecompileToAddress(accountOverrideParam.getMovePrecompileToAddress().getAddress());
+        if (accountOverrideParam.movePrecompileToAddress() != null) {
+            this.setMovePrecompileToAddress(accountOverrideParam.movePrecompileToAddress().getAddress());
         }
 
-        if (accountOverrideParam.getBalance() != null) {
-            this.setBalance(HexUtils.stringHexToBigInteger(accountOverrideParam.getBalance().getHexNumber()));
+        if (accountOverrideParam.balance() != null) {
+            this.setBalance(HexUtils.stringHexToBigInteger(accountOverrideParam.balance().getHexNumber()));
         }
 
-        if (accountOverrideParam.getNonce() != null) {
-            this.setNonce(HexUtils.jsonHexToLong(accountOverrideParam.getNonce().getHexNumber()));
+        if (accountOverrideParam.nonce() != null) {
+            this.setNonce(HexUtils.jsonHexToLong(accountOverrideParam.nonce().getHexNumber()));
         }
 
-        if (accountOverrideParam.getCode() != null) {
-            this.setCode(accountOverrideParam.getCode().getRawDataBytes());
+        if (accountOverrideParam.code() != null) {
+            this.setCode(accountOverrideParam.code().getRawDataBytes());
         }
 
-        if (accountOverrideParam.getState() != null) {
+        if (accountOverrideParam.state() != null) {
             Map<DataWord, DataWord> newState = new HashMap<>();
-            for (Map.Entry<HexDataParam, HexDataParam> entry : accountOverrideParam.getState().entrySet()) {
+            for (Map.Entry<HexDataParam, HexDataParam> entry : accountOverrideParam.state().entrySet()) {
                 newState.put(entry.getKey().getAsDataWord(),entry.getValue().getAsDataWord());
             }
             this.setState(newState);
         }
 
-        if (accountOverrideParam.getStateDiff() != null) {
+        if (accountOverrideParam.stateDiff() != null) {
             Map<DataWord, DataWord> newStateDiff = new HashMap<>();
-            for (Map.Entry<HexDataParam, HexDataParam> entry : accountOverrideParam.getStateDiff().entrySet()) {
+            for (Map.Entry<HexDataParam, HexDataParam> entry : accountOverrideParam.stateDiff().entrySet()) {
                 newStateDiff.put(entry.getKey().getAsDataWord(),entry.getValue().getAsDataWord());
             }
             this.setStateDiff(newStateDiff);
--- a/rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
+++ b/rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
@@ -152,14 +152,15 @@ public class EthModule
     }
 
     public String call(CallArgumentsParam argsParam, BlockIdentifierParam bnOrId, List<AccountOverride> accountOverrideList) {
-        boolean shouldPerformStateOverride = accountOverrideList != null && !accountOverrideList.isEmpty();
 
-        validateStateOverrideAllowance(shouldPerformStateOverride);
+        if (accountOverrideList != null) {
+            validateAccountOverride(accountOverrideList);
+        }
 
         ExecutionBlockRetriever.Result result = executionBlockRetriever.retrieveExecutionBlock(bnOrId.getIdentifier());
         Block block = result.getBlock();
 
-        MutableRepository mutableRepository = prepareRepository(result, block, shouldPerformStateOverride, accountOverrideList);
+        MutableRepository mutableRepository = prepareRepository(result, block, accountOverrideList);
 
         CallArguments callArgs = argsParam.toCallArguments();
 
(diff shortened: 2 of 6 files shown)


rskj-core/src/main/java/org/ethereum/rpc/parameters/AccountOverrideParam.java
```diff
@@ -31,47 +31,9 @@
...
-    }
-
-    public Map<HexDataParam, HexDataParam> getStateDiff() {
-        return stateDiff;
-    }
-
-    public HexAddressParam getMovePrecompileToAddress() {
-        return movePrecompileToAddress;
-    }
+public record AccountOverrideParam(HexNumberParam balance, HexNumberParam nonce, HexDataParam code,
```
rmoreliovlabs, line 34: nit: Any reason why we are changing this class to a record in this PR? If possible I would separate this refactor into a different PR to avoid having many file changes and avoid confusion in the review.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/AccountOverride.java
```diff
@@ -42,11 +42,17 @@ public class AccountOverride {
     private RskAddress address;
     private RskAddress movePrecompileToAddress;
 
-    public AccountOverride(RskAddress address) {
+    private final int maxOverridableCodeSize;
+    private final int maxStateOverrideChanges;
+
+    public AccountOverride(RskAddress address, int maxOverridableCodeSize, int maxStateOverrideChanges) {
```
rmoreliovlabs: nit: In general, I think it is better idea to separate validations in a different place rather within the DTO, including adding logic inside, but since this was already developed and already has logic inside. In future refactor I would move all the validations and logic out of this class.

rmoreliovlabs (dismissed):
2 small nits on my end, LGTM!

rskj-core/src/main/java/org/ethereum/rpc/parameters/AccountOverrideParam.java
nagarev: I decided to apply that here since I was already modifying the class, but yes, it enlarges the PR for sure.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/AccountOverride.java
fmacleal: Yes, makes sense, I think it actually should be in the `DefaultStateOverrideApplier` class. We already perform some validations there, not sure if those are enough. If they are, I am not against in remove these validations from this DTO here.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -152,14 +161,16 @@ public String call(CallArgumentsParam argsParam, BlockIdentifierParam bnOrId) {
     }
 
     public String call(CallArgumentsParam argsParam, BlockIdentifierParam bnOrId, List<AccountOverride> accountOverrideList) {
-        boolean shouldPerformStateOverride = accountOverrideList != null && !accountOverrideList.isEmpty();
+        boolean shouldApplyStateOverride = shouldApplyStateOverride(accountOverrideList);
 
-        validateStateOverrideAllowance(shouldPerformStateOverride);
+        if (shouldApplyStateOverride) {
+            validateAccountOverride(accountOverrideList);
+        }
```
fmacleal: **suggestion**:

Why don't merge these two parts and methods in a single one? They are completely related. The boolean `shouldApplyStateOverride` could receive the result from the validation if the `AccountOverride` should be applied. Something like:

```
        boolean shouldApplyStateOverride = validateAccountOverride(accountOverrideList);

        ExecutionBlockRetriever.Result result = executionBlockRetriever.retrieveExecutionBlock(bnOrId.getIdentifier());
        Block block = result.getBlock();
...
```

And then, in the `validateAccountOverride` you could check if it should be applied or not:

```
    private boolean validateAccountOverride(List<AccountOverride> accountOverrideList) {
        boolean shouldApplyStateOverride = accountOverrideList != null && !accountOverrideList.isEmpty();
        
        if(shouldApplyStateOverride) {
            if (!allowCallStateOverride) {
                throw invalidParamError("State override is not allowed");
            }

            if (accountOverrideList.size() > maxAccountOverrides) {
                throw invalidParamError("Number of account overrides exceeded. Max " + maxAccountOverrides);
            }
        }
        return shouldApplyStateOverride;
    }
```

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -203,6 +208,20 @@ private MutableRepository prepareRepository(ExecutionBlockRetriever.Result resul
...
+
+        if (accountOverrideList.size() > maxAccountOverrides) {
+            throw invalidParamError("Number of account overrides exceeded. Max " + maxAccountOverrides);
+        }
+    }
+
+    private boolean shouldApplyStateOverride(List<AccountOverride> accountOverrideList) {
+        return accountOverrideList != null && !accountOverrideList.isEmpty();
+    }
+
```
fmacleal: **comment**:

Already explained my suggestion above.

rskj-core/src/main/java/org/ethereum/rpc/parameters/AccountOverrideParam.java
fmacleal: **praise**:

Although it enlarges the PR, it reduces a lot of code. Well done. :) 
I prefer record.

rskj-core/src/main/java/org/ethereum/rpc/Web3Impl.java
```diff
@@ -417,7 +417,12 @@ public String eth_call(CallArgumentsParam args, BlockRefParam blockRefParam) {
     @Override
     public String eth_call(CallArgumentsParam args, BlockRefParam blockRefParam, Map<HexAddressParam, AccountOverrideParam> accParam) {
         List<AccountOverride> accountOverrideList = accParam.entrySet().stream()
-                .map(entry -> new AccountOverride(entry.getKey().getAddress()).fromAccountOverrideParam(entry.getValue())).toList();
+                .map(entry -> new AccountOverride(
+                        entry.getKey().getAddress(),
+                        ethModule.getMaxOverridableCodeSize(),
+                        ethModule.getMaxStateOverrideChanges())
+                        .fromAccountOverrideParam(entry.getValue()))
```
fmacleal: **thought**:

It's weird we have to pass the limits to a the object all the time, the DTO indeed shouldn't have any logic as @rmoreliovlabs suggested. We probably should have a ticket for this improvement.

rskj-core/src/main/resources/reference.conf
```diff
@@ -407,6 +407,12 @@ rpc {
     callGasCap = 50000000
     # Allow StateOverride in eth_call
     allowCallStateOverride = false
+    # Maximum amount of account overrides on eth_call's state override
+    maxAccountOverrides = 0
+    # Maximum size (in bytes) of code to be overridden on eth_call's state override
+    maxOverridableCodeSize = 0
+    # Maximum amount of state changes allowed on eth_call's state override
+    maxStateOverrideChanges = 0
```
fmacleal, line 416: **question**:

why is the default zero? shouldn't we set as reference an usable number that people will actually need to have it set? If we set it to zero, wouldn't mean that we wouldn't be able to override any account with any code no?

rskj-core/src/main/java/org/ethereum/rpc/parameters/HexNumberParam.java
```diff
@@ -65,7 +65,8 @@ public String toString() {
     }
 
     public static boolean isHexNumberLengthValid(String hex) {
-        return hex != null && hex.length() <= MAX_HEX_NUM_LEN;
+        int maxLength = (HexUtils.hasHexPrefix(hex)) ? MAX_HEX_NUM_LEN : HEX_NUM_BYTE_LENGTH * 2;
+        return hex != null && hex.length() <= maxLength;
```
fmacleal, line 69: **question**:

What's the impact in now starts to check the prefix? Wouldn't this break any thing? I see that this it's used by the `getBlock` in the class `Web3InformationRetriever` and if someone it's used to get hex block without the prefix? Wouldn't this be an issue? 🤔

rskj-core/src/main/java/co/rsk/rpc/modules/eth/AccountOverride.java
nagarev: The only thing I see regarding not validating inside the DTO (a topic that doesn't have a unique answer in general, and is more a matter of opinion) is that the validations inside `DefaultOverrideApplier` (the only applier that we have right now) could end up being duplicated in whatever other applier we decide to implement on the future. But I'm fine moving the validation logic there if both of you consider it better.

rskj-core/src/main/resources/reference.conf
nagarev: I decided to go with `0` as this feature is disabled by default. We can totally add usable defaults if we define some.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
```diff
@@ -109,7 +110,8 @@ public EthModule(
             ActivationConfig activationConfig,
             PrecompiledContracts precompiledContracts,
             boolean allowCallStateOverride,
-            StateOverrideApplier stateOverrideApplier) {
+            StateOverrideApplier stateOverrideApplier,
+            int maxAccountOverrides) {
```
fmacleal, line 114: **comment**:
We definitely need to improve this configuration management, doesn't make sense we have to change the constructor parameters for every new parameter needed. Why can't we just pass the config object? 

Also, why the `allowCallStateOverride` is disabled by default? Isn't possible to configure it at all? 🤔

rskj-core/src/test/java/co/rsk/rpc/modules/eth/DefaultStateOverrideApplierTest.java
```diff
@@ -77,47 +79,172 @@ private void populateRepository(Repository repository, RskAddress address) {
...
+        // Given
+        AccountOverride accountOverride = new AccountOverride(null);
+
+        // When
+        RskJsonRpcRequestException exception = assertThrows(RskJsonRpcRequestException.class, () -> {
+            stateOverrideApplier.applyToRepository(repository, accountOverride);
+        });
+
+        // Then
+        assertEquals(-32602, exception.getCode());
```
fmacleal, line 92: **question**:

Where did this negative code came from? Is it the same returned by Ethereum?

rskj-core/src/test/java/co/rsk/rpc/modules/eth/DefaultStateOverrideApplierTest.java
```diff
@@ -77,47 +79,172 @@ private void populateRepository(Repository repository, RskAddress address) {
...
+        assertEquals("Address cannot be null", exception.getMessage());
```
fmacleal, line 93: **thought**:

Although we are checking that the correct exception was thrown, I would say that compare the exception message it's a bit too much specific. If we change the error message, test will fail and it doesn't make much sense IMHO.

Ideally, we could have different codes for different errors, buuut, I saw the class RskJsonRpcRequestException::invalidParamError` and it would be tricky to refactor it now. 

Let's go like this for now, but let's keep in mind that test that specific doesn't bring much value and we could avoid in the future.

rskj-core/src/test/java/co/rsk/rpc/modules/eth/EthModuleTest.java
```diff
@@ -215,6 +218,48 @@ void testCall_whenCalledWithAccountOverrideListButAccountOverrideIsNotAllowed_th
         assertEquals("State override is not allowed", exception.getMessage());
     }
 
+    @Test
+    void testCall_whenCalledWithAccountOverrideListButAccountOverridesAreHigherThanMaxAllowed_throwsExceptionAsExpected() {
```
fmacleal, line 222: **praise**:

Good test! :)

fmacleal (commented):
Good job, it improved a lot. I have a few comments, but nothing big.

rskj-core/src/main/resources/reference.conf
fmacleal: But I still think we need to define a default value in the reference, I am not comfortable in release it with 0. It's basically release a feature that can't be used. Is it just because we can't use this right now? Do you have a rough estimative of what could be some good values to put as default?

rskj-core/src/test/java/co/rsk/rpc/modules/eth/DefaultStateOverrideApplierTest.java
nagarev: Yup, they're part of the JSON-RPC specification.

rskj-core/src/main/java/co/rsk/rpc/modules/eth/EthModule.java
nagarev: Totally agree with you, the configuration part can (and should) be improved a lot!

Regarding the default state, we can define the minimums and leave it allowed by default. 💪🏼

fmacleal (approved):
LGTM

PR #3310, opened 2025-09-16
Support BitcoinJ Network Constants

rmoreliovlabs:
Support BitcoinJ Network Constants

## Description
Right now we are using always `RegTestParams` no matter the which network we selected. Now we want to select right parameters for the right network which could be `mainnet, testnet, testnet2 or regtest`

## Motivation and Context
Right now we are using always `RegTestParams` no matter the which network we selected. Now we want to select right parameters for the right network which could be `mainnet, testnet, testnet2 or regtest`

## How Has This Been Tested?
We updated the tests and ran the node in different networks to make sure different network parameters were being set.

## Types of changes

## Checklist:

* **Other information**:


dlealmac (approved):
I approve this PR.

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -207,6 +212,26 @@ public Constants getNetworkConstants() {
         return constants;
     }
 
+    public NetworkParameters getBitcoinjNetworkConstants() {
```
dlealmac, line 216: @rmoreliovlabs please, do we need to add unit test for this method?

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -207,6 +212,26 @@ public Constants getNetworkConstants() {
...
+        if (networkParameters == null) {
+            switch (netName()) {
+                case "main":
```
dlealmac: @rmoreliovlabs do we have a motivation to not use constants for the network names?

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
rmoreliovlabs: Addressed

dlealmac (approved):
I approve this PR.

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -110,6 +114,14 @@ public abstract class SystemProperties {
...
     public static final String PROPERTY_PERSIST_BLOOMS_CACHE_SNAPSHOT = "cache.blooms.persist-snapshot";
 
+    private final Map<String, NetworkName> NETWORK_NAME_MAP = Map.of(
+            NetworkName.MAINNET.getName(), NetworkName.MAINNET,
+            NetworkName.TESTNET.getName(), NetworkName.TESTNET,
+            NetworkName.TESTNET2.getName(), NetworkName.TESTNET2,
+            NetworkName.DEVNET.getName(), NetworkName.DEVNET,
+            NetworkName.REGTEST.getName(), NetworkName.REGTEST
+    );
+
```
fmacleal: **suggestion**:

With the suggestion from above, we can eliminate this map and minimise the changes in this class.

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -181,23 +194,24 @@ public ActivationConfig getActivationConfig() {
 
     public Constants getNetworkConstants() {
         if (constants == null) {
-            switch (netName()) {
-                case "main":
+            final var networkName = NETWORK_NAME_MAP.get(netName());
```
fmacleal: **suggestion**:

With my proposal, this part would stay like this:


```suggestion
            final var networkName =  NetworkName.getNetworkByName(netName());
```

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
```diff
@@ -207,6 +221,27 @@ public Constants getNetworkConstants() {
         return constants;
     }
 
+    public NetworkParameters getBitcoinjNetworkConstants() {
+        if (networkParameters == null) {
+            final var networkName = NETWORK_NAME_MAP.get(netName());
```
fmacleal: **suggestion:**

With my proposal, this part would stay like this:

```suggestion
            final var networkName =  NetworkName.getNetworkByName(netName());
```

rskj-core/src/main/java/org/ethereum/config/NetworkName.java
```diff
@@ -0,0 +1,16 @@
...
+    private final String name;
+
+    NetworkName(String name) {
+        this.name = name;
+    }
+
+    public String getName() {
+        return name;
+    }
+
```
fmacleal, line 20: **suggestion**:

WDYT in move the `NETWORK_NAME_MAP` to a function here, adding a static function that would do the conversion. That way, all the logic regarding the mapping of the network stays in the enum. And we don't need this mapping there in the `SystemProperties` class.  I have also added an UNKNOWN option for when someone pass an unknown name, this will trigger the `RuntimeException` when someone pass a wrong name to this enum.

With my suggestion, this enum would be like this:

```suggestion
    MAINNET("main"), TESTNET("testnet"), TESTNET2("testnet2"), DEVNET("devnet"), REGTEST("regtest"), UNKNOWN("unknown");

    private final String name;

    NetworkName(String name) {
        this.name = name;
    }

    public String getName() {
        return name;
    }

    public static NetworkName getNetworkByName(String nameToFind) {
        return Arrays.stream(NetworkName.values())
                .filter( network -> network.name.toLowerCase().equals(nameToFind.toLowerCase()))
                .findFirst()
                .orElse(UNKNOWN);
    }
```

rskj-core/src/test/java/co/rsk/config/RskSystemPropertiesTest.java
```diff
@@ -371,4 +374,78 @@ void checkLastSessionPeersFilePathProperty(){
         assertEquals("/dbdir/lastPeers.properties", testSystemProperties.getLastKnewPeersFilePath().toString());
     }
 
+    @Test
```
fmacleal, line 377: **praise**:

Nice tests! Congrats on adding them. 🙂

rskj-core/src/main/java/org/ethereum/config/NetworkName.java
rmoreliovlabs: Addressed with a simplified version

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
rmoreliovlabs: Addressed

rskj-core/src/main/java/org/ethereum/config/SystemProperties.java
rmoreliovlabs: Addressed

PR #3312, opened 2025-09-19
Make filter id less predictable

rmoreliovlabs:
Make filter id less predictable

## Description
Make filter id less predictable

## Motivation and Context
The Ethereum RPC includes the ability to filter for on-chain events in several different ways. One of these is a *continuous* filter that can be queried for more information that contains a plethora of commands. The idea is to create a filter that can be queried multiple times or contains a websocket connection that outputs the log event data as new information is posted into the blocks via the `eth_newFilter` RPC. Once done with the filter, `eth_uninstallFilter` can be called to delete it.

When making the calls, it's important to know the filter ID being used - this unique identifier can be used to query for the events via HTTP requests. Additionally, this ID is what is used in order to delete the filter later. 

This ID is effectively a secret on public RPCs. In other implementations of the EVM RPC endpoints, such as GETH, this is cryptographically secure randomly generated value that looks like a UUID. In Rootstock, this is a sequential and incrementing value.

## Impact 
Since this ID is somewhat sensitive to the operations, this creates one major issue and two smaller issues on users of these endpoints from RPC providers:
- Skipping Events
- Denial of Service
- Information disclosure

The most impactful part of this vulnerability is the risk of the filter itself. `eth_getFilterChanges` is used to get new events on a given filter but only gets the new ones. Once an event has been seen by the filter, it will not be returned on the filter again. So, if an attacker *knows* the ID that they're looking for, they can observe the events and predict the legitimate application for viewing them. 

Off-chain infrastructure viewing events is used in a multitude of places including bridges, logging and monitoring and many multi-step operations that require off-chain infrastructure to do, such as signing data. By preventing these applications from seeing the data, it introduces an uptime risk on Rootstock. For instance, a bridge from Rootstock to EVM could be forced to drop transfers by using up the events being observed themselves. Or, an attacker finds the ID of an application used for logging/monitoring by a target and they effectively *disable* it using this attack. Both of these uptime risks have serious end user impact and protocol impact. 

The second issue is around `eth_uninstallFilter`. To delete a filter, an off-chain application calls `eth_uninstallFilter`. If this ID is known then a different user could call this to delete the filter for the application. In most real world scenarios it is likely that the end-user has an application that handles the filter being removed because they are on a timer.

The final issue is around information disclosure. By viewing the filter logs of other applications, you can see potentially sensitive information. In the case of MEV, it's an arms race -- being able to see what your competitors are doing would give you a competitive edge. For protocols using this for logging/monitoring, this may expose hidden invariants among other things.

## How Has This Been Tested?

https://github.com/user-attachments/assets/0377b67f-f812-4df1-92af-837251e7ee9b

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/rpc/FilterManager.java
```diff
@@ -39,18 +40,35 @@ public class FilterManager {
...
-    private static final long FILTER_CLEANUP_PERIOD = Duration.ofMinutes(1).toMillis(); // 1 minute in milliseconds
+    private final long FILTER_TIMEOUT;
+    private final long FILTER_CLEANUP_PERIOD;
 
     private final Object filterLock = new Object();
-    private final AtomicInteger filterCounter = new AtomicInteger(1);
 
     @GuardedBy("filterLock")
-    private final Map<Integer, Filter> installedFilters = new HashMap<>();
+    private final Map<String, Filter> installedFilters = new HashMap<>();
```
fmacleal: **suggestion**:

Since you are changing here, maybe we could use `ConcurrentHashMap` and stop to use this `filterLock` plus the annotation `@GuardedBy`. WDYT? I was reading about this and seems that we should default to `ConcurrentHashMap` when we want a map thread safe and with non null keys. Probably the `filterLock` was interesting because we also had to increment the `AtomicInteger filterCounter`. Since you deleted it, makes sense to remove this annotation as well and stick with the known and safe `ConcurrentHashMap`. WDYT?

rskj-core/src/main/java/org/ethereum/rpc/FilterManager.java
```diff
@@ -39,18 +40,35 @@ public class FilterManager {
...
+            }
+
+            @Override
+            public void onPendingTransactionsReceived(List<Transaction> transactions) {
+                newPendingTx(transactions);
+            }
+        });
+    }
+
+    public FilterManager(Ethereum eth, long filterTimeoutInMs, long filterCleanupPeriodInMs) {
```
fmacleal, line 66: **question**:

Have you created this constructor for testing purposes?

rmoreliovlabs: Yes

rskj-core/src/main/java/org/ethereum/rpc/FilterManager.java
rmoreliovlabs: makes sense, addressed

fmacleal (approved):
LGTM

PR #3325, opened 2025-09-26
Refactor traceFilter method for improved performance and reliability

dlealmac:
## Description
This PR refactors the traceFilter method in TraceModuleImpl to address performance issues and improve reliability.

Key Changes:
Early Termination: Implemented early termination when enough traces are collected, reducing unnecessary block processing
Forward Processing: Changed from reverse to forward block iteration for better performance and intuitive behavior
Parameter Validation: Added validation for block ranges and count limits to prevent invalid requests
Null Safety: Added proper null handling for count and after parameters with sensible defaults
Resource Protection: Introduced MAX_TRACES_PER_REQUEST constant (10,000) to prevent performance issues
Enhanced Logging: Added comprehensive debug logging for better observability and troubleshooting

## Motivation and Context
Slowness was observed under certain conditions.

## How Has This Been Tested?
Unit Testing: All 14 unit tests in TraceModuleImplTest are passing after the refactoring, covering basic functionality, contract operations, and trace filtering scenarios. Fixed test methods that were missing required count parameters.
Postman API Testing: Tested with various parameter combinations including optional parameters, edge cases (invalid block ranges, count limits, pagination), and performance scenarios. Verified early termination works correctly and all error conditions are properly handled with descriptive messages.

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -54,6 +53,8 @@ public class TraceModuleImpl implements TraceModule {
     private static final String PENDING_BLOCK = "pending";
     private static final Logger logger = LoggerFactory.getLogger("web3");
 
+    private static final int MAX_TRACES_PER_REQUEST = 10000;
```
rmoreliovlabs: nit: Should this be configurable? if so we should use the `.conf` files instead. We also have a `Constants` file for more internal configurations. Not sure if any of these options are ok for this work.

This is because we kind of try to have all constants in one place

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -129,34 +130,54 @@ public JsonNode traceBlock(String blockArgument) throws Exception {
 
     @Override
     public JsonNode traceFilter(TraceFilterRequest traceFilterRequest) {
-        List<List<TransactionTrace>> blockTracesGroup = new ArrayList<>();
+        int count = traceFilterRequest.getCount() != null ? traceFilterRequest.getCount() : MAX_TRACES_PER_REQUEST;
+        int after = traceFilterRequest.getAfter() != null ? traceFilterRequest.getAfter() : 0;
+
+        if (count > MAX_TRACES_PER_REQUEST) {
+            throw RskJsonRpcRequestException.invalidParamError("Count value too big. Maximum " + MAX_TRACES_PER_REQUEST + " traces allowed.");
+        }
```
rmoreliovlabs: nit: In order to use the features from newer java version I would refactor it this way:
```
final var count = Optional.ofNullable(traceFilterRequest).map(TraceFilterRequest::getCount).orElse(MAX_TRACES_PER_REQUEST);

Optional.of(count).filter(c -> c > MAX_TRACES_PER_REQUEST).orElseThrow(() -> RskJsonRpcRequestException.invalidParamError("Count value too big. Maximum " + MAX_TRACES_PER_REQUEST + " traces allowed."));

final var after = Optional.ofNullable(traceFilterRequest).map(TraceFilterRequest::getAfter).orElse(0);
```

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -129,34 +130,54 @@ public JsonNode traceBlock(String blockArgument) throws Exception {
...
 
+        List<TransactionTrace> allTraces = new ArrayList<>();
         Block fromBlock = getBlockByTagOrNumber(traceFilterRequest.getFromBlock(), traceFilterRequest.getFromBlockNumber());
-        Block block = getBlockByTagOrNumber(traceFilterRequest.getToBlock(), traceFilterRequest.getToBlockNumber());
+        Block toBlock = getBlockByTagOrNumber(traceFilterRequest.getToBlock(), traceFilterRequest.getToBlockNumber());
+        toBlock = toBlock == null ? blockchain.getBestBlock() : toBlock;
```
rmoreliovlabs: nit: Probably use optionals?

```
toBlock = Optional.ofNullable(toBlock).orElseGet(BlockChain::getBestBlock);
```

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -129,34 +130,54 @@ public JsonNode traceBlock(String blockArgument) throws Exception {
...
+            if (allTraces.size() >= totalNeeded) {
+                logger.debug("traceFilter: Early termination at block {} with {} traces (needed {})",
+                        currentBlock.getNumber(), allTraces.size(), totalNeeded);
+                break;
+            }
 
-        if (traceFilterRequest.getAfter() != null) {
-            txTraceStream = txTraceStream.skip(traceFilterRequest.getAfter());
+            currentBlock = this.blockchain.getBlockByNumber(currentBlock.getNumber() + 1);
         }
```
rmoreliovlabs: nit: Can we create a function in a utility to paginate the blocks? feels like this is something we can reuse in another moment. You can create a file PaginationUtil probably and abstact the parameters needed

rmoreliovlabs (approved):
Remember to add some tests

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: for some reason this change makes overall performance much worse, I chose not to include it

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: I analyzed some options, but since the ``after`` parameter still needs to be refactored, I think it's better not to add complexity at this point.

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: done, thank you

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: done, thank you

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
rmoreliovlabs: Completely agree

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
rmoreliovlabs: Would be good to know the reason why, but I'm fine with the current changes

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -129,34 +133,59 @@ public JsonNode traceBlock(String blockArgument) throws Exception {
...
-        }
+        while (currentBlock != null && toBlock != null && currentBlock.getNumber() <= toBlock.getNumber()) {
+            List<TransactionTrace> builtTraces = buildBlockTraces(currentBlock, traceFilterRequest);
 
-        Collections.reverse(blockTracesGroup);
+            allTraces.addAll(builtTraces);
+            processedBlocks++;
 
-        Stream<TransactionTrace> txTraceStream = blockTracesGroup.stream().flatMap(Collection::stream);
+            if (allTraces.size() >= totalNeeded) {
```
italo-sampaio: Is this early termination really needed? I'm thinking two things:
1 - When will this happen in practice?
2 - Won't this return a valid, but incomplete response (meaning: some traces will be missing)?

italo-sampaio (approved):
LGTM, just left a question

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
italo-sampaio: 3 things actually:
3 - isn't the first check (`if (count > maxTracesPerRequest)`) enough to solve the performance issue?

italo-sampaio: After discussing offline, we agreed that this is really what we want, marking as resolved

dlealmac: @italo-sampaio thank you for comments!

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -129,34 +133,59 @@ public JsonNode traceBlock(String blockArgument) throws Exception {
 
     @Override
     public JsonNode traceFilter(TraceFilterRequest traceFilterRequest) {
-        List<List<TransactionTrace>> blockTracesGroup = new ArrayList<>();
+        if (traceFilterRequest == null) {
+            throw RskJsonRpcRequestException.invalidParamError("Invalid trace_filter parameters.");
+        }
+
+        int maxTracesPerRequest = config.rpcTraceMaxTracesPerRequest();
```
fmacleal: **suggestion**:

Add this as a global variable in the constructor, to avoid it being initialised every time we call `traceFilter`. If you notice, the `TraceModuleImpl` class is initialised only once by the RSKContext. Adding this parameter that never changes in the constructor, save a bit of resources.

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -129,34 +133,59 @@ public JsonNode traceBlock(String blockArgument) throws Exception {
...
 
-        while (fromBlock != null && block != null && block.getNumber() >= fromBlock.getNumber()) {
-            List<TransactionTrace> builtTraces = buildBlockTraces(block, traceFilterRequest);
+        int processedBlocks = 0;
+        int totalNeeded = after + count;
+        Block currentBlock = fromBlock;
 
-            blockTracesGroup.add(builtTraces);
+        logger.debug("traceFilter: Starting processing from block {} to block {}, skipCount={}, limitCount={}",
+                fromBlock != null ? fromBlock.getNumber() : -1, toBlock != null ? toBlock.getNumber() : -1, after, count);
```
fmacleal: **question**:

What's the gain in put -1 here when the number is not defined? Since it's the primitive value, the initial value will be 0 no?

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -129,34 +133,59 @@ public JsonNode traceBlock(String blockArgument) throws Exception {
...
+        List<TransactionTrace> allTraces = new ArrayList<>();
         Block fromBlock = getBlockByTagOrNumber(traceFilterRequest.getFromBlock(), traceFilterRequest.getFromBlockNumber());
-        Block block = getBlockByTagOrNumber(traceFilterRequest.getToBlock(), traceFilterRequest.getToBlockNumber());
+        Block toBlock = getBlockByTagOrNumber(traceFilterRequest.getToBlock(), traceFilterRequest.getToBlockNumber());
+        toBlock = toBlock == null ? blockchain.getBestBlock() : toBlock;
 
-        block = block == null ? blockchain.getBestBlock() : block;
+        if (fromBlock != null && toBlock != null && fromBlock.getNumber() > toBlock.getNumber()) {
+            throw RskJsonRpcRequestException.invalidParamError("fromBlock cannot be greater than toBlock");
+        }
```
fmacleal: **suggestion**:

Extract this validation logic to a separate and private method, just to improve readability.

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -129,34 +133,59 @@ public JsonNode traceBlock(String blockArgument) throws Exception {
...
+        int totalNeeded = after + count;
+        Block currentBlock = fromBlock;
 
-            blockTracesGroup.add(builtTraces);
+        logger.debug("traceFilter: Starting processing from block {} to block {}, skipCount={}, limitCount={}",
+                fromBlock != null ? fromBlock.getNumber() : -1, toBlock != null ? toBlock.getNumber() : -1, after, count);
 
-            block = this.blockchain.getBlockByHash(block.getParentHash().getBytes());
-        }
+        while (currentBlock != null && toBlock != null && currentBlock.getNumber() <= toBlock.getNumber()) {
```
fmacleal: **nitpick**
Why validate `toBlock` if you already set it as bestBlock if no parameter is passed? Doesn't it mean that it will never be null?

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -129,34 +133,59 @@ public JsonNode traceBlock(String blockArgument) throws Exception {
...
-        if (traceFilterRequest.getCount() != null) {
-            txTraceStream = txTraceStream.limit(traceFilterRequest.getCount());
+        Stream<TransactionTrace> txTraceStream = allTraces.stream();
+
+        if (after > 0) {
+            txTraceStream = txTraceStream.skip(after);
         }
 
-        List<TransactionTrace> traces = txTraceStream.collect(Collectors.toList());
+        List<TransactionTrace> traces = txTraceStream.limit(count).collect(Collectors.toList());
```
fmacleal: **suggestion**:

Sublist is more efficient than Java streams, you could use that to apply the pagination. Something like:

```
    int fromIndex = Math.min(after, allTraces.size());
    int toIndex = Math.min(after + count, allTraces.size());
    List<TransactionTrace> traces = allTraces.subList(fromIndex, toIndex);
```

rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceModuleImplTest.java
```diff
@@ -791,6 +792,63 @@ void executedContractWithDelegateCallToNonExistentContract() throws Exception {
...
+        traceFilterRequest.setCount(Integer.MAX_VALUE);
+
+        RskJsonRpcRequestException exception = Assertions.assertThrows(
+            RskJsonRpcRequestException.class,
+            () -> traceModule.traceFilter(traceFilterRequest)
+        );
+
+        Assertions.assertTrue(exception.getMessage().contains("Count value too big"));
+        Assertions.assertTrue(exception.getMessage().contains("Maximum"));
+        Assertions.assertTrue(exception.getMessage().contains("traces allowed"));
```
fmacleal: **nitpick**:

Use the full power of hamcrest and evaluate the whole object at once instead of multiple times:

```
        assertThat(exception.getMessage(), CoreMatchers.allOf(
                containsString("Count value too big"),
                containsString("Maximum"),
                containsString("traces allowed")));
```

rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceModuleImplTest.java
```diff
@@ -791,6 +792,63 @@ void executedContractWithDelegateCallToNonExistentContract() throws Exception {
...
+        traceFilterRequest.setFromBlock("0x2"); // Block 2
+        traceFilterRequest.setToBlock("0x1");  // Block 1
+        traceFilterRequest.setCount(10);
+
+        RskJsonRpcRequestException exception = Assertions.assertThrows(
+            RskJsonRpcRequestException.class,
+            () -> traceModule.traceFilter(traceFilterRequest)
+        );
+
+        Assertions.assertEquals("fromBlock cannot be greater than toBlock", exception.getMessage());
```
fmacleal: **nitpick**:

I rather to use the hamcrest matcher than the equals:

```
        assertThat(exception.getMessage(), containsString("fromBlock cannot be greater than toBlock"));
```

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: done, thank you!

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: fromBlock can be null for block values that doesn't exist. So it is to indicate that we do not have getNumber() from fromBlock because fromBlock is null. Does that makes sense?

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: done, thank you!

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: good catch, done, thank you!

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: done, thank you!

rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceModuleImplTest.java
dlealmac: done, thank you!

rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceModuleImplTest.java
dlealmac: done, thank you!

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -64,20 +65,25 @@ public class TraceModuleImpl implements TraceModule {
     private final ExecutionBlockRetriever executionBlockRetriever;
 
     private final SignatureCache signatureCache;
+    private final RskSystemProperties config;
```
fmacleal: **suggestion**:

Unused variable. You can remove.
```suggestion
```

fmacleal (dismissed):
Awesome, thank you for the changes. Approving now, just missing a small thing to remove the unused variable.

fmacleal (approved):
LGTM

PR #3467, opened 2026-02-27
Enforce RSKIP-144 Dependency on RSKIP-351

meldsun0:
## Description
RSKIP-351 introduces Block Header Version 1 and the Block Header Extension structure.

RSKIP-144 does not define an independent header field. Instead, it introduces txExecutionSublistsEdges, which is encoded inside the Block Header Extension, explicitly relying on the structure defined in RSKIP-351.

Therefore, without RSKIP-351 (i.e., without Header Version 1 and the Block Header Extension), there is no valid encoding location for the RSKIP-144 field.

## Motivation and Context
Allowing RSKIP-144 to be active while RSKIP-351 is inactive creates an inconsistent and invalid state.

## How Has This Been Tested?
- Test cases that previously enabled 144 without 351 have been updated. 
-  Code paths have been reviewed to ensure there is no execution path that allows RSKIP-144 logic to run without RSKIP-351 being active. 

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ActivationConfig.java
```diff
@@ -107,6 +107,22 @@ public boolean isActive(ConsensusRule consensusRule, long blockNumber) {
         return 0 <= activationHeight && activationHeight <= blockNumber;
     }
 
+    /**
+     * Returns {@code true} if all given {@link ConsensusRule} instances
+     * are active at the specified block number.
+     * <p>If no rules are provided, this method returns {@code false}.
```
ilanolkies, line 113:
```suggestion
     * If no rules are provided, this method returns {@code false}.
```

What is the html?

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ActivationConfig.java
```diff
@@ -107,6 +107,22 @@ public boolean isActive(ConsensusRule consensusRule, long blockNumber) {
...
+     *
+     * @param number the block number to evaluate
+     * @param consensusRules the rules to check, in evaluation order
+     * @return {@code true} if all rules are active and at least one rule
+     *         is provided; {@code false} otherwise
+     */
+    public boolean isActive(long number, ConsensusRule... consensusRules) {
```
ilanolkies:
```suggestion
    public boolean areActive(long number, ConsensusRule... consensusRules) {
```

I would put the consensus rules first, as `isActive`

ilanolkies (commented):
It is not clear for me if the changes are complete for the objective. I was expecting `txExecutionSublistsEdges` to be removed from `BlockHeaderV0`, or to behave as void. Also, it is not clear for me how passing RSKIP351 activation everywhere fixes the issue... what if someone activates one of the two by mistake? the code will run as if none were activated without halt

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ActivationConfig.java
meldsun0: - Varargs (...) must be the last parameter.
- I will rename the method anyway.

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ActivationConfig.java
meldsun0: <p> is used to explicitly define paragraph breaks in JavaDoc

wlaclaw (dismissed):
Looks good in general

meldsun0:
> It is not clear for me if the changes are complete for the objective. I was expecting `txExecutionSublistsEdges` to be removed from `BlockHeaderV0`, or to behave as void. Also, it is not clear for me how passing RSKIP351 activation everywhere fixes the issue... what if someone activates one of the two by mistake? the code will run as if none were activated without halt

This PR is not meant to modify txExecutionSublistsEdges or BlockHeaderV0. That is outside its scope.

The goal is only to enforce the dependency between RSKIP144 and RSKIP351 by:

1.  Fixing tests that allowed RSKIP144 without RSKIP351.
2. Adding an explicit check so RSKIP144 behavior runs only if RSKIP351 is also enabled.

If only one of them is activated by mistake, the dependency condition is not satisfied, and the guarded code will not execute.

A more complete solution for handling RSKIP dependencies will be addressed in a separate PR.

wlaclaw (dismissed):

rskj-core/src/main/java/co/rsk/net/handler/TxPendingValidator.java
```diff
@@ -73,7 +73,7 @@ public TxPendingValidator(Constants constants, ActivationConfig activationConfig
     public TransactionValidationResult isValid(Transaction tx, Block executionBlock, @Nullable AccountState state) {
         long executionBlockNumber = executionBlock.getNumber();
         ActivationConfig.ForBlock activations = activationConfig.forBlock(executionBlockNumber);
-        BigInteger gasLimit = activations.isActive(ConsensusRule.RSKIP144)
+        BigInteger gasLimit = activations.isActive(ConsensusRule.RSKIP351) && activations.isActive(ConsensusRule.RSKIP144)
```
nagarev: I think that I'm missing something here, is there any reason we're not using the new `areActive` method here?

rskj-core/src/main/java/co/rsk/validators/ValidTxExecutionSublistsEdgesRule.java
```diff
@@ -26,7 +26,8 @@ public ValidTxExecutionSublistsEdgesRule(ActivationConfig activationConfig) {
 
     @Override
     public boolean isValid(Block block) {
-        if (!activationConfig.isActive(ConsensusRule.RSKIP144, block.getHeader().getNumber())) {
+        if (!(activationConfig.isActive(ConsensusRule.RSKIP351, block.getHeader().getNumber())
```
nagarev: Same question that in previous comment. I may be missing something for sure 🤔

rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
```diff
@@ -210,7 +210,8 @@ private BlockHeader decodeHeader(RLPList rlpHeader, boolean compressed, boolean
 
         if ((!isBlockHeaderCompressionEnabled(blockNumber) || !compressed)
             && rlpHeader.size() > r
-            && activationConfig.isActive(ConsensusRule.RSKIP144, blockNumber)) {
+            && activationConfig.isActive(ConsensusRule.RSKIP351, blockNumber)
```
nagarev: Here I would ask the same question as before. Additionally, using that method would make the code comply with sonar here.

nagarev: There's two more instances of the same situation in this file at [348](https://github.com/rsksmart/rskj/pull/3467/changes#diff-3a2320e737a23ebd96e3aa514f2aa76458067708aa32715c22964ca7ee97a7daR348) and [394](https://github.com/rsksmart/rskj/pull/3467/changes#diff-3a2320e737a23ebd96e3aa514f2aa76458067708aa32715c22964ca7ee97a7daR394).

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -177,7 +178,7 @@ private boolean init() {
         }
 
         long txGasLimit = GasCost.toGas(tx.getGasLimit());
-        long gasLimit = activations.isActive(RSKIP144) ? sublistGasLimit : GasCost.toGas(executionBlock.getGasLimit());
+        long gasLimit = (activations.isActive(RSKIP351) && activations.isActive(RSKIP144)) ? sublistGasLimit : GasCost.toGas(executionBlock.getGasLimit());
```
nagarev, line 181: same

nagarev (commented):
I really like the changes, I had those stupid questions, I would like the answer here as to document for future reference, as I couldn't figure out what those were done that way 🤔.

meldsun0:
> I really like the changes, I had those stupid questions, I would like the answer here as to document for future reference, as I couldn't figure out what those were done that way 🤔.

Suggested  changes have been applied when possible. Good catch!

nagarev (approved):
LGTM! Thanks for taking the time to update that when possible beside the amount of tests to be updated. 💪🏼

fmacleal (approved):
LGTM

PR #3470, opened 2026-03-02
Added test to validate config "rpc.logs.maxBlocksToQuery"

wlaclaw:
## Motivation and Context
Need to check if current configuration limits are working for eth_getLogs RPC method

## Types of changes
- new test case introduced

## LogFilterTest.java Notice
`LogFilterTest.java` had non unix  EOL format. I updated it but all newlines delimiters changed and it is basically all file as a diff.

### DIFF
--- a/rskj-core/src/test/java/org/ethereum/rpc/Web3ImplLogsTest.java
+++ b/rskj-core/src/test/java/org/ethereum/rpc/Web3ImplLogsTest.java
@@ -17,8 +17,10 @@
  */
 package org.ethereum.rpc;
 
+import com.typesafe.config.ConfigValueFactory;
 import co.rsk.blockchain.utils.BlockGenerator;
 import co.rsk.config.TestSystemProperties;
+import co.rsk.jsonrpc.JsonRpcError;
 import co.rsk.core.Coin;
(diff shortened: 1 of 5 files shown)


dlealmac (dismissed):
LGTM!

italo-sampaio (dismissed):
LGTM!

rskj-core/src/test/java/org/ethereum/rpc/Web3ImplLogsTest.java
```diff
@@ -887,6 +893,35 @@ void getLogsThrowsExceptionWhenBlockHashIsUsedCombinedWithToBlock() {
         Assertions.assertThrows(RskJsonRpcRequestException.class, () -> web3.eth_getLogs(fr));
     }
 
+    @Test
```
nagarev, line 896: Awesome tests and refactorings! only nitpick is that we're trying to add comments (and structure) test in the // Given, // When, // Then format when possible, can you try to modify this so it matches that style? Thanks a lot!

wlaclaw: Done

PR #3472, opened 2026-03-06
Added tests for buildAmountTo and simplified those methods repeated

julianlen:
## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,919 @@
...
+                .withScriptPubKey(federationOutputScript)
+                .withValue(MINIMUM_PEGIN_TX_VALUE)
+                .buildMany(numberOfUtxos, i -> createHash(i + 1));
+            setUpWallet(federationUTXOs);
+        }
+
+        @Test
+        void buildAmountTo_whenRSKIP201IsNotActive_shouldCreateReleaseTxWithBtcVersion1() {
+            // Arrange
+            setUpActivations(PAPYRUS_ACTIVATIONS);
```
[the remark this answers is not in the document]
julianlen: done

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -84,4 +87,55 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
     public static void assertBtcTxVersionIs2(BtcTransaction releaseTransaction) {
         assertEquals(BTC_TX_VERSION_2, releaseTransaction.getVersion());
     }
+
+    public static void assertDestinationAddress(List<TransactionOutput> releaseTransactionOutputs,
+                                                Address expectedDestinationAddress) {
+        for (TransactionOutput output : releaseTransactionOutputs) {
+            Address destinationAddress = output.getScriptPubKey().getToAddress(BTC_MAINNET_PARAMS);
```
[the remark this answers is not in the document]
julianlen: should I add it as parameter?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildBatchedPegoutsTest.java
```diff
@@ -123,7 +123,7 @@ void buildBatchedPegouts_whenNoPegoutRequests_shouldThrowIllegalStateException()
         void buildBatchedPegouts_whenRSKIP201IsNotActive_shouldCreateBatchedPegoutsTxWithBtcVersion1() {
             // Arrange
             setUpActivations(PAPYRUS_ACTIVATIONS);
-            ReleaseTransactionBuilder releaseTransactionBuilder = setupWalletAndCreateReleaseTransactionBuilder(federationUTXOs);
+            ReleaseTransactionBuilder releaseTransactionBuilder = createReleaseTransactionBuilder();
```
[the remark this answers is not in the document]
julianlen: done

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -84,4 +87,55 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+        Coin expectedChangeAmount = inputTotalAmount.subtract(expectedSentPegoutAmount);
+
+        if (isDust(expectedChangeAmount)) {
+            Coin amountToGetNonDustValue = MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT.subtract(expectedChangeAmount);
+            expectedSentPegoutAmount = expectedSentPegoutAmount.subtract(amountToGetNonDustValue);
+            expectedChangeAmount = MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT;
+        }
+
+        int expectedNumberOfOutputs = 1;
+        assertEquals(expectedNumberOfOutputs, releaseTransactionChangeOutputs.size());
```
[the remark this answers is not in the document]
julianlen: there is no scenario with more than one output with these tests

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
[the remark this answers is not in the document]
julia-zack: i think it makes sense yes

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -85,4 +88,55 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+
+    public static void assertDestinationAddress(List<TransactionOutput> releaseTransactionOutputs,
+                                                Address expectedDestinationAddress) {
+        for (TransactionOutput output : releaseTransactionOutputs) {
+            Address destinationAddress = output.getScriptPubKey().getToAddress(BTC_MAINNET_PARAMS);
+            assertEquals(expectedDestinationAddress, destinationAddress);
+        }
+    }
+
+    public static void assertReleaseTxHasChangeAndPegoutsAmountWithFeesProperly(BtcTransaction releaseTransaction,
```
julia-zack: i dont understand this name

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -85,4 +88,55 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+                                                                                List<TransactionOutput> releaseTransactionChangeOutputs,
+                                                                                Coin expectedSentPegoutAmount) {
+        Coin inputTotalAmount = releaseTransaction.getInputSum();
+        Coin expectedChangeAmount = inputTotalAmount.subtract(expectedSentPegoutAmount);
+
+        if (isDust(expectedChangeAmount)) {
+            Coin amountToGetNonDustValue = MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT.subtract(expectedChangeAmount);
+            expectedSentPegoutAmount = expectedSentPegoutAmount.subtract(amountToGetNonDustValue);
+            expectedChangeAmount = MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT;
+        }
```
julia-zack: i think we shouldn't have an `if` in an asserting utility function. We should know exactly what we are asserting

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julia-zack, line 1: why creating a `ReleaseTransactionBuilderBuildAmountToTest` ? shouldn't we have just a `ReleaseTransactionBuilderTest` class and test different methods there?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,1385 @@
...
+    private static final Context BTC_CONTEXT = new Context(BTC_MAINNET_PARAMS);
+
+    private static final Coin MINIMUM_PEGOUT_TX_VALUE = BRIDGE_MAINNET_CONSTANTS.getMinimumPegoutTxValue();
+    private static final Coin MINIMUM_PEGIN_TX_VALUE = BRIDGE_MAINNET_CONSTANTS.getMinimumPeginTxValue(IRIS_ACTIVATIONS);
+
+    private static final Coin HIGH_FEE_PER_KB = Coin.valueOf(1_000_000);
+    private static final Coin DUSTY_AMOUNT_SEND_REQUESTED = MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT.minus(Coin.SATOSHI);
+    private static final Coin THOUSAND_SATOSHIS = Coin.valueOf(1000);
+    private static final int RECIPIENT_ADDRESS_KEY_OFFSET = 2100;
+    private static final Address RECIPIENT_ADDRESS = createRecipientAddress();
```
julia-zack, line 57: for example, you'll be repeating all (or almost all) these things if you have a different class for each method right?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,1385 @@
...
+        }
+    }
+
+    private void setUpActivations(ActivationConfig.ForBlock activations) {
+        this.activations = activations;
+    }
+
+    private void setUpFeePerKb(Coin feePerKb) {
+        this.feePerKb = feePerKb;
+    }
```
julia-zack, line 428: i think these methods dont make sense. Do it directly in the general setup function.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,1385 @@
...
+    protected Script federationOutputScript;
+    protected Script federationRedeemScript;
+    protected Wallet wallet;
+
+    private ActivationConfig.ForBlock activations;
+    private Coin feePerKb;
+
+    @BeforeEach
+    void setUp() {
+        setUpActivations(IRIS_ACTIVATIONS);
```
julia-zack, line 72: we should set `allActivations` as the default, so when adding new tests they are set with last behavior. Set specific activations when needed

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,1385 @@
...
+            List<TransactionInput> pegoutInputs = pegoutTransaction.getInputs();
+            int numberOfUtxos = 1;
+            assertEquals(numberOfUtxos, pegoutInputs.size());
+            assertReleaseTxInputsHasProperFormatAndBelongsToStandardMultisigFederation(
+                pegoutTransaction,
+                federationRedeemScript,
+                federationUTXOs
+            );
+            assertPegoutTxHasPegoutAndChangeOutputs(pegoutTransaction, MINIMUM_PEGOUT_TX_VALUE);
+            assertSelectedUtxosBelongToTheInputs(amountToResult.selectedUTXOs(), pegoutInputs);
```
julia-zack: you are repeating these assertions in every test. Can't you abstract them to a single method that performs all validations?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,1385 @@
...
+        }
+
+        @Test
+        void buildAmountTo_whenChangeIsOneSatoshi_shouldCreatePegoutTxDecrementingFirstOutputAndSettingNonDustChange() {
```
julia-zack: whats the point of this test? you already have `buildAmountTo_whenChangeIsDust_shouldCreatePegoutTxDecrementingFirstOutputAndSettingNonDustChange`

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,1385 @@
...
+                MINIMUM_PEGOUT_TX_VALUE);
+
+            // Assert
+            assertBuildResultResponseCode(COULD_NOT_ADJUST_DOWNWARDS, amountToResult);
+            assertNull(amountToResult.btcTx());
+            assertNull(amountToResult.selectedUTXOs());
+        }
+
+        @Test
+        void buildAmountTo_whenEstimatedFeeIsHighAndUtxosAreEnough_shouldCreatePegoutTx() {
```
julia-zack: i dont fully understand this test, "high fees" is kind of arbitrary right? 
I think this scenario does not differ with one with "low fees" and less utxos total amount

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,1385 @@
...
+            BuildResult amountToResult = releaseTransactionBuilder.buildAmountTo(RECIPIENT_ADDRESS, MINIMUM_PEGOUT_TX_VALUE);
+
+            // Assert
+            assertBuildResultResponseCode(INSUFFICIENT_MONEY, amountToResult);
+            assertNull(amountToResult.btcTx());
+            assertNull(amountToResult.selectedUTXOs());
+        }
+
+        @Test
+        void buildAmountTo_whenRSKIP201IsNotActive_shouldCreatePegoutTxWithBtcVersion1() {
```
julia-zack: was there a psh-erp fed before rskip201?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,1385 @@
...
+                federationRedeemScript,
+                federationUTXOs
+            );
+
+            assertPegoutTxHasPegoutAndChangeOutputs(pegoutTransaction, MINIMUM_PEGOUT_TX_VALUE);
+            assertSelectedUtxosBelongToTheInputs(amountToResult.selectedUTXOs(), pegoutInputs);
+        }
+
+        @Test
+        void buildAmountTo_whenMultipleUtxosCanCoverAmount_shouldCreatePegoutTx() {
```
julia-zack: why these two tests differ?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,1385 @@
...
+        @Test
+        void buildAmountTo_whenWalletHasExactFundsForPegoutRequest_shouldCreatePegoutTxWithNoChangeOutput() {
+            // Arrange
+            federationUTXOs = List.of(
+                UTXOBuilder.builder()
+                    .withScriptPubKey(federationOutputScript)
+                    .withValue(MINIMUM_PEGOUT_TX_VALUE)
+                    .build()
+            );
+            ReleaseTransactionBuilder releaseTransactionBuilder = setupWalletAndCreateReleaseTransactionBuilder(federationUTXOs);
```
julia-zack: i have the feeling that you are setting a general setup "before each" test but then repeating this code block in almost every test.
Consider removing the `BeforeEach` from the general setup function, and modifying it to receive needed things as params. Then call it in every test.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,1385 @@
...
+            BuildResult amountToResult = releaseTransactionBuilder.buildAmountTo(RECIPIENT_ADDRESS, DUSTY_AMOUNT_SEND_REQUESTED);
+
+            // Assert
+            assertBuildResultResponseCode(DUSTY_SEND_REQUESTED, amountToResult);
+            assertNull(amountToResult.btcTx());
+            assertNull(amountToResult.selectedUTXOs());
+        }
+
+        @Test
+        void buildAmountTo_whenEstimatedFeeIsTooHigh_shouldReturnCouldNotAdjustDownwards() {
```
julia-zack: same here, i don't get why this test is different than `buildAmountTo_whenUtxosAreMinimumNonDustValue_shouldReturnCouldNotAdjustDownwards`

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,1385 @@
...
+            BuildResult amountToResult = releaseTransactionBuilder.buildAmountTo(RECIPIENT_ADDRESS, MINIMUM_PEGOUT_TX_VALUE);
+
+            // Assert
+            assertBuildResultResponseCode(INSUFFICIENT_MONEY, amountToResult);
+            assertNull(amountToResult.btcTx());
+            assertNull(amountToResult.selectedUTXOs());
+        }
+
+        @Test
+        void buildAmountTo_whenRSKIP201IsNotActive_shouldCreatePegoutTxWithBtcVersion1() {
```
julia-zack: please remember that we discussed that having a test with a segwit fed and PAPYRUS activations does not make sense

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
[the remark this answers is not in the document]
julianlen: done https://github.com/rsksmart/rskj/commit/cfc0417adeeca09d50a9faf9ac297a2273ed5164

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
julianlen: done https://github.com/rsksmart/rskj/commit/7bcc65be004c2a9e3d68767eb508a8d41a8a4b31

but this will be improved when changing all assertions

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: The thing is that buildAmountTo is specific for individual pegouts. This is deactivated after IRIS

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: so may be it doesn't make sense to have any test in this file with p2sh-p2wsh, right?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: rskip201 -> iris300
RSKIP353 (p2sh) -> hop401

then, no

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: Actually, rskip271 -> batchedPegout -> hop 400, therefore it seems just the standard multisig used this method

julianlen: done

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/c55acd2fc54faa630215013b69c5a470add37ce2

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: We are testing two edge cases for a dust change. If the change is 1 satoshi, it takes 2699 satoshis from the first output, while if the change is 2699 satoshis, it takes 1 satoshi from the first output

julianlen: I can change the name of the test:

buildAmountTo_whenChangeIsMaxDustValue_shouldCreatePegoutTxDecrementingFirstOutputAndSettingNonDustChange

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: The fees are high enough, arbitrary, tho.

We have 2 scenarios:
1. buildAmountTo_whenEstimatedFeeIsTooHigh_shouldReturnCouldNotAdjustDownwards 
2. buildAmountTo_whenEstimatedFeeIsHighAndUtxosAreEnough_shouldCreatePegoutTx

The second one is to demonstrate that the covering of the fees depends strictly on the utxos available. 

Although we can just keep the first one, the second one is kind of reflected in the rest of the tests, testing the happy path.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: they differ in the number of inputs

julianlen: we are testing a btc tx building. At least having one test with 1 input and another with multiples I think it's enough, wdyt?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: makes sense

julianlen: done https://github.com/rsksmart/rskj/commit/1c62e1ebd4027c1b15bf45c1338f7f07a2fa6c5e

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: This's another (unrealistic tho) edge case

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: As we discussed, we are going to create another PR for this

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: same

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: I'll see if there're some repeating assretions to abstract htem to a single method

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: We decided to clarify buildAmountTo_whenEstimatedFeeIsTooHigh_shouldReturnCouldNotAdjustDownwards with buildAmountTo_whenEstimatedFeeIsTooHighAndNotEnoughUtxos_shouldReturnCouldNotAdjustDownwards

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: we discussed it and it's okay

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
julianlen: To avoid this I separated in 2 methods:

`assertPegoutsAndChangeValuesWhenOriginalChangeIsDust`
`assertPegoutsAndChangeValuesWhenOriginalChangeIsNonDust`

both called from

`assertPegoutTxHasPegoutAndChangeOutputsWhenOriginalChangeIsDust`
`assertBatchedPegoutsTxHasPegoutAndChangeOutputsWhenOriginalChangeIsDust`
`assertPegoutTxHasPegoutAndChangeOutputsWhenOriginalChangeIsNonDust`
`assertBatchedPegoutsTxHasPegoutAndChangeOutputsWhenOriginalChangeIsNonDust`

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: I'll keep it to avoid having `this.feePerKb = feePerKb;` everywhere

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/ce38dc2c14f1433412029818cb9f95612e2d98bb

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -87,4 +86,79 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+    public static void assertPegoutsAndChangeValuesWhenOriginalChangeIsDust(BtcTransaction releaseTransaction,
+                                                                            List<TransactionOutput> releaseTransactionChangeOutputs,
+                                                                            Coin expectedSentPegoutAmount) {
+        Coin inputTotalAmount = releaseTransaction.getInputSum();
+        Coin expectedChangeAmount = inputTotalAmount.subtract(expectedSentPegoutAmount);
+        assertTrue(isDust(expectedChangeAmount));
+
+        Coin amountToGetNonDustValue = MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT.subtract(expectedChangeAmount);
+        expectedSentPegoutAmount = expectedSentPegoutAmount.subtract(amountToGetNonDustValue);
+        expectedChangeAmount = MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT;
```
julia-zack: please don't re-declare variables 🙏

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+import co.rsk.peg.federation.StandardMultiSigFederationBuilder;
+import co.rsk.test.builders.UTXOBuilder;
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.List;
+import org.ethereum.config.blockchain.upgrades.ActivationConfig;
+import org.ethereum.config.blockchain.upgrades.ActivationConfigsForTest;
+import org.ethereum.core.Repository;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
```
julia-zack: check sonar warning here

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+
+class ReleaseTransactionBuilderBuildAmountToTest {
+
+    private static final int EXPECTED_NUMBER_OF_CHANGE_OUTPUTS = 1;
+    private static final ActivationConfig.ForBlock IRIS_ACTIVATIONS = ActivationConfigsForTest.iris300().forBlock(0);
+    private static final ActivationConfig.ForBlock PAPYRUS_ACTIVATIONS = ActivationConfigsForTest.papyrus200().forBlock(0);
+
+    private static final BridgeConstants BRIDGE_MAINNET_CONSTANTS = BridgeMainNetConstants.getInstance();
+    private static final NetworkParameters BTC_MAINNET_PARAMS = BRIDGE_MAINNET_CONSTANTS.getBtcParams();
+    private static final Context BTC_CONTEXT = new Context(BTC_MAINNET_PARAMS);
```
julia-zack: i think this is only being used in the `setUpWallet` method, you can move the declaration there.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+
+    private static final Coin MINIMUM_PEGOUT_TX_VALUE = BRIDGE_MAINNET_CONSTANTS.getMinimumPegoutTxValue();
+    private static final Coin MINIMUM_PEGIN_TX_VALUE = BRIDGE_MAINNET_CONSTANTS.getMinimumPeginTxValue(IRIS_ACTIVATIONS);
+
+    private static final Coin HIGH_FEE_PER_KB = Coin.valueOf(1_000_000);
+    private static final Coin DUSTY_AMOUNT_SEND_REQUESTED = MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT.minus(Coin.SATOSHI);
+    private static final Coin THOUSAND_SATOSHIS = Coin.valueOf(1000);
```
julia-zack: same for this, only being used in one place

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+    private static final int RECIPIENT_ADDRESS_KEY_OFFSET = 2100;
```
julia-zack: same here

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+    private static final Address RECIPIENT_ADDRESS = createRecipientAddress();
```
julia-zack, line 57:
```suggestion
    private static final Address RECIPIENT_ADDRESS = BtcECKey.fromPrivate(BigInteger.valueOf(2100)).toAddress(BTC_MAINNET_PARAMS);
```
i think this should be enough

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+    private static final int EXPECTED_NUMBER_OF_CHANGE_OUTPUTS = 1;
+    private static final ActivationConfig.ForBlock IRIS_ACTIVATIONS = ActivationConfigsForTest.iris300().forBlock(0);
+    private static final ActivationConfig.ForBlock PAPYRUS_ACTIVATIONS = ActivationConfigsForTest.papyrus200().forBlock(0);
+
+    private static final BridgeConstants BRIDGE_MAINNET_CONSTANTS = BridgeMainNetConstants.getInstance();
+    private static final NetworkParameters BTC_MAINNET_PARAMS = BRIDGE_MAINNET_CONSTANTS.getBtcParams();
+    private static final Context BTC_CONTEXT = new Context(BTC_MAINNET_PARAMS);
+
+    private static final Coin MINIMUM_PEGOUT_TX_VALUE = BRIDGE_MAINNET_CONSTANTS.getMinimumPegoutTxValue();
+    private static final Coin MINIMUM_PEGIN_TX_VALUE = BRIDGE_MAINNET_CONSTANTS.getMinimumPeginTxValue(IRIS_ACTIVATIONS);
```
julia-zack, line 53: careful here. For example, the test `buildAmountTo_whenRSKIP201IsNotActive_shouldCreatePegoutTxWithBtcVersion1` is using papyrus activations but iris pegin tx value. Is that correct?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+    void buildAmountTo_whenRSKIP201IsNotActive_shouldCreatePegoutTxWithBtcVersion1() {
+        // Arrange
+        setUpActivations(PAPYRUS_ACTIVATIONS);
+        int numberOfUtxos = 10;
+        federationUTXOs = UTXOBuilder.builder()
+            .withScriptPubKey(federationOutputScript)
+            .withValue(MINIMUM_PEGIN_TX_VALUE)
+            .buildMany(numberOfUtxos, i -> createHash(i + 1));
+        setUpWallet(federationUTXOs);
+        ReleaseTransactionBuilder releaseTransactionBuilder = setupWalletAndCreateReleaseTransactionBuilder(federationUTXOs);
```
julia-zack: ? you are setting the wallet twice

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+        BuildResult amountToResult = releaseTransactionBuilder.buildAmountTo(RECIPIENT_ADDRESS, MINIMUM_PEGOUT_TX_VALUE);
+
+        // Assert
+        assertBuildResultResponseCode(INSUFFICIENT_MONEY, amountToResult);
+        assertNull(amountToResult.btcTx());
+        assertNull(amountToResult.selectedUTXOs());
+    }
+
+    @Test
+    void buildAmountTo_whenRSKIP201IsNotActive_shouldCreatePegoutTxWithBtcVersion1() {
```
julia-zack, line 97: this name is not descriptive enough, you are asserting a lot more things than just the btc tx version

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julia-zack: check other places where this happens too

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+        BuildResult amountToResult = releaseTransactionBuilder.buildAmountTo(RECIPIENT_ADDRESS, MINIMUM_PEGOUT_TX_VALUE);
+
+        // Assert
+        assertBuildResultResponseCode(SUCCESS, amountToResult);
+        BtcTransaction pegoutTransaction = amountToResult.btcTx();
+        assertBtcTxVersionIs1(pegoutTransaction);
+
+        List<TransactionInput> pegoutInputs = pegoutTransaction.getInputs();
+        int numberOfUtxosExpected = 1;
+        assertEquals(numberOfUtxosExpected, pegoutInputs.size());
```
julia-zack: aren't you testing this in `assertSelectedUtxosBelongToTheInputs` ?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+        assertReleaseTxInputsHasProperFormatAndBelongsToStandardMultisigFederation(
+            pegoutTransaction,
+            federationRedeemScript,
+            federationUTXOs
+        );
+        assertPegoutTxWithPegoutAndChangeOutputsWhenOriginalChangeIsNonDust(pegoutTransaction, MINIMUM_PEGOUT_TX_VALUE);
```
julia-zack: please rename this method, is confusing

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+        int expectedNumberOfChangeOutputs = 0;
+        assertPegoutTxOutputAndChangeOutputsNumbers(pegoutTransaction, expectedNumberOfPegoutOutputs, expectedNumberOfChangeOutputs);
+
+        List<TransactionOutput> pegoutTransactionOutputs = pegoutTransaction.getOutputs();
+        assertDestinationAddress(pegoutTransactionOutputs, RECIPIENT_ADDRESS, BTC_MAINNET_PARAMS);
+
+        assertReleaseTxWithNoChangeHasPegoutsAmountWithFeesProperly(pegoutTransaction, requestedAmount);
+    }
+
+    private List<TransactionOutput> getPegoutOutputs(BtcTransaction pegoutTransaction) {
```
julia-zack:
```suggestion
    private List<TransactionOutput> getUserOutputs(BtcTransaction pegoutTransaction) {
```
or something like that, easier to compare `ChangeOutputs` and `UserOutputs`. They are all the pegout outputs.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+        return pegoutTransaction.getOutputs().stream()
+            .filter(this::isPegoutOutput)
+            .toList();
+    }
+
+    private boolean isPegoutOutput(TransactionOutput pegoutOutput) {
+        return !isFederationOutput(pegoutOutput);
```
julia-zack: mmmm i dont like this comparison. Better check against `RECIPIENT_ADDRESS`.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+        assertSelectedUtxosBelongToTheInputs(amountToResult.selectedUTXOs(), pegoutInputs);
+    }
+
+    @Test
+    void buildAmountTo_whenSingleUtxoCanCoverAmount_shouldCreatePegoutTx() {
+        // Arrange
+        int numberOfUtxos = 10;
+        federationUTXOs = UTXOBuilder.builder()
+            .withScriptPubKey(federationOutputScript)
+            .withValue(MINIMUM_PEGIN_TX_VALUE)
```
julia-zack, line 127: no big deal, but i'd change this value with a more random one. For example, use `Coin.COIN`

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+            pegoutTransaction,
+            federationRedeemScript,
+            federationUTXOs
+        );
+        assertPegoutTxWithPegoutAndChangeOutputsWhenOriginalChangeIsNonDust(pegoutTransaction, MINIMUM_PEGOUT_TX_VALUE);
+        assertSelectedUtxosBelongToTheInputs(amountToResult.selectedUTXOs(), pegoutInputs);
+    }
+
+    @Test
+    void buildAmountTo_whenMultipleUtxosCanCoverAmount_shouldCreatePegoutTx() {
```
julia-zack:
```suggestion
    void buildAmountTo_whenMultipleUtxosAreNeededToCoverAmount_shouldCreatePegoutTx() {
```

i find this name a bit more descriptive, wdyt?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+
+        List<TransactionInput> pegoutInputs = pegoutTransaction.getInputs();
+        int numberOfUtxosExpected = 1;
+        assertEquals(numberOfUtxosExpected, pegoutInputs.size());
+        assertReleaseTxInputsHasProperFormatAndBelongsToStandardMultisigFederation(
+            pegoutTransaction,
+            federationRedeemScript,
+            federationUTXOs
+        );
+        assertPegoutTxWithOnlyPegoutOutputs(pegoutTransaction, MINIMUM_PEGOUT_TX_VALUE);
```
julia-zack:
```suggestion
        assertPegoutTxOnlyHasUserOutputs(pegoutTransaction, MINIMUM_PEGOUT_TX_VALUE);
```

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+
+    private static final BridgeConstants BRIDGE_MAINNET_CONSTANTS = BridgeMainNetConstants.getInstance();
+    private static final NetworkParameters BTC_MAINNET_PARAMS = BRIDGE_MAINNET_CONSTANTS.getBtcParams();
+    private static final Context BTC_CONTEXT = new Context(BTC_MAINNET_PARAMS);
+
+    private static final Coin MINIMUM_PEGOUT_TX_VALUE = BRIDGE_MAINNET_CONSTANTS.getMinimumPegoutTxValue();
+    private static final Coin MINIMUM_PEGIN_TX_VALUE = BRIDGE_MAINNET_CONSTANTS.getMinimumPeginTxValue(IRIS_ACTIVATIONS);
+
+    private static final Coin HIGH_FEE_PER_KB = Coin.valueOf(1_000_000);
+    private static final Coin DUSTY_AMOUNT_SEND_REQUESTED = MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT.minus(Coin.SATOSHI);
```
julia-zack:
```suggestion
    private static final Coin DUST_VALUE = MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT.minus(Coin.SATOSHI);
```

no big deal, just a suggestion

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+            pegoutTransaction,
+            federationRedeemScript,
+            federationUTXOs
+        );
+        assertPegoutTxWithPegoutAndChangeOutputsWhenOriginalChangeIsDust(pegoutTransaction, MINIMUM_PEGOUT_TX_VALUE);
+        assertSelectedUtxosBelongToTheInputs(amountToResult.selectedUTXOs(), pegoutInputs);
+    }
+
+    @Test
+    void buildAmountTo_whenChangeIsNonDustForOneSatoshi_shouldCreatePegoutTxWithNoModificationInTheValues() {
```
julia-zack:
```suggestion
    void buildAmountTo_whenChangeIsMinNonDustValue_shouldCreatePegoutTxWithNoModificationInTheValues() {
```

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+        BuildResult amountToResult = releaseTransactionBuilder.buildAmountTo(RECIPIENT_ADDRESS, MINIMUM_PEGOUT_TX_VALUE);
+
+        // Assert
+        assertBuildResultResponseCode(SUCCESS, amountToResult);
+        BtcTransaction pegoutTransaction = amountToResult.btcTx();
+        assertBtcTxVersionIs2(pegoutTransaction);
+
+        List<TransactionInput> pegoutInputs = pegoutTransaction.getInputs();
+        int numberOfUtxos = 1;
+        assertEquals(numberOfUtxos, pegoutInputs.size());
```
julia-zack: not sure if, generally speaking, this check adds value or its mostly noise

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+        );
+        ReleaseTransactionBuilder releaseTransactionBuilder = setupWalletAndCreateReleaseTransactionBuilder(federationUTXOs);
+
+        // Act
+        BuildResult amountToResult = releaseTransactionBuilder.buildAmountTo(RECIPIENT_ADDRESS, MINIMUM_PEGOUT_TX_VALUE);
+
+        // Assert
+        assertBuildResultResponseCode(SUCCESS, amountToResult);
+        BtcTransaction pegoutTransaction = amountToResult.btcTx();
+        assertBtcTxVersionIs2(pegoutTransaction);
```
julia-zack: can we abstract this to a method? 
is being checked in almost every test, except for the ones that return an unsuccessful result and the only one that belongs to papyrus activations.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,621 @@
...
+        Coin requestedAmount
+    ) {
+        int expectedNumberOfPegoutOutputs = 1;
+        assertPegoutTxOutputAndChangeOutputsNumbers(pegoutTransaction, expectedNumberOfPegoutOutputs, EXPECTED_NUMBER_OF_CHANGE_OUTPUTS);
+
+        List<TransactionOutput> pegoutTransactionChangeOutputs = getChangeOutputs(pegoutTransaction);
+        assertDestinationAddress(pegoutTransactionChangeOutputs, federationAddress, BTC_MAINNET_PARAMS);
+
+        List<TransactionOutput> pegoutOutputs = getPegoutOutputs(pegoutTransaction);
+        assertDestinationAddress(pegoutOutputs, RECIPIENT_ADDRESS, BTC_MAINNET_PARAMS);
```
julia-zack:
```
        int expectedNumberOfPegoutOutputs = 1;
        assertPegoutTxOutputAndChangeOutputsNumbers(pegoutTransaction, expectedNumberOfPegoutOutputs, EXPECTED_NUMBER_OF_CHANGE_OUTPUTS);

        List<TransactionOutput> pegoutTransactionChangeOutputs = getChangeOutputs(pegoutTransaction);
        assertDestinationAddress(pegoutTransactionChangeOutputs, federationAddress, BTC_MAINNET_PARAMS);

        List<TransactionOutput> pegoutOutputs = getPegoutOutputs(pegoutTransaction);
        assertDestinationAddress(pegoutOutputs, RECIPIENT_ADDRESS, BTC_MAINNET_PARAMS);
```

this part is duplicated. Abstract it to an `assertOutputs` method. 
Then check directly the dust/non-dust related parts in specific tests.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -87,4 +86,79 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+            Address destinationAddress = output.getScriptPubKey().getToAddress(networkParameters);
+            assertEquals(expectedDestinationAddress, destinationAddress);
+        }
+    }
+
+    public static void assertPegoutsAndChangeValuesWhenOriginalChangeIsDust(BtcTransaction releaseTransaction,
+                                                                            List<TransactionOutput> releaseTransactionChangeOutputs,
+                                                                            Coin expectedSentPegoutAmount) {
+        Coin inputTotalAmount = releaseTransaction.getInputSum();
+        Coin expectedChangeAmount = inputTotalAmount.subtract(expectedSentPegoutAmount);
```
julia-zack:
```suggestion
        Coin originalChangeAmount = inputTotalAmount.subtract(expectedSentPegoutAmount);
```
is NOT the expected change amount, but the original one. You are actually expecting it to be different from that amount.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -87,4 +86,79 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+                                                NetworkParameters networkParameters) {
+        for (TransactionOutput output : releaseTransactionOutputs) {
+            Address destinationAddress = output.getScriptPubKey().getToAddress(networkParameters);
+            assertEquals(expectedDestinationAddress, destinationAddress);
+        }
+    }
+
+    public static void assertPegoutsAndChangeValuesWhenOriginalChangeIsDust(BtcTransaction releaseTransaction,
+                                                                            List<TransactionOutput> releaseTransactionChangeOutputs,
+                                                                            Coin expectedSentPegoutAmount) {
```
julia-zack:
```suggestion
                                                                            Coin requestedAmount) {
```

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -87,4 +86,79 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+            .map(TransactionOutput::getValue)
+            .reduce(Coin::add)
+            .orElse(Coin.ZERO);
+    }
+
+    private static boolean isDust(Coin expectedChangeAmount) {
+        return expectedChangeAmount.compareTo(MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT) < 0;
+    }
+
+    public static void assertReleaseTxWithNoChangeHasPegoutsAmountWithFeesProperly(BtcTransaction releaseTransaction,
```
julia-zack: please rename this method, is not clear

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
julianlen: done https://github.com/rsksmart/rskj/commit/c2da885f43b6221ece9a193c5493824031d4fe80

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: 👍

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/203506e2a70245a860f800b027e3a9c1204cc3c5

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/203506e2a70245a860f800b027e3a9c1204cc3c5

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: `createRecipientAddress()` is clearer, isn't it?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: right, done

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/5f345de4050841e57af9e1100977512ebd4430fe

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: Here we are asserting that the pegoutTx is created correctly but particularly that the version is 1 buildAmountTo_whenRSKIP201IsNotActive_shouldCreatePegoutTxSuccesfullyWithBtcVersion1 would be more descriptive?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: not the number of inputs, just belonging

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/a8ca26bbaf761408e905d9c91bce1897056626c7

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/cd132cc9f2edce20d002248939d0189abcbbb4a0

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: wouldn't it better to leverage these tests and user an "edge" case like the min pegin value?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: changed to whenSingleUtxoCoversRequestedAmount and whenMultipleUtxosCoverRequestedAmount

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: Here we are checking more than "only has user outputs".

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/e938bb1565f0d0523e9db58f2d6c6771079700ba

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/e08368cd8fe0ed5ec82b63506daeb851f889e58e

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: I guess the namings were confusing. I changed it to

int expectedNumberOfInputs = 1;
assertEquals(expectedNumberOfInputs, pegoutInputs.size());

I think having the control about how many inputs we expect the tx to have is not noise

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
julianlen: got it, done https://github.com/rsksmart/rskj/commit/74e8989d17890f90fab9919ccf6d5ae48579138b

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
julianlen: done https://github.com/rsksmart/rskj/commit/74e8989d17890f90fab9919ccf6d5ae48579138b

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
julianlen: assertReleaseTxWithOnlyUserOutputsAmounts?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julia-zack: no big deal at all, but I think having a one-line method for declaring a static final constant is confusing. But feel free to leave it as is

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julia-zack: good point yes. Wdyt about parameterizing the tests a bit to use some different values? For example minimum pegin, 1btc, a big utxo, etc.

And this makes me think, shouldn't we test some scenarios with utxos with different values? Not sure if its worth it for this method (since its kind of deprecated), but for the others

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julia-zack: oh my bad, i wanted to point out that calling them **user** output is way more accurate than **pegout** output

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julia-zack: expectedNumberOfInputs is a much better name, thanks

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -87,4 +86,78 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+
+    public static void assertUserAndChangeOutputsValuesWhenOriginalChangeIsDust(BtcTransaction releaseTransaction,
+                                                                                List<TransactionOutput> releaseTransactionChangeOutputs,
+                                                                                Coin requestedAmount) {
+        Coin inputTotalAmount = releaseTransaction.getInputSum();
+        Coin originalChangeAmount = inputTotalAmount.subtract(requestedAmount);
+        assertTrue(isDust(originalChangeAmount));
+
+        Coin amountToGetNonDustValue = MIN_NON_DUST_VALUE_FOR_P2SH_OUTPUT_SCRIPT.subtract(originalChangeAmount);
+        requestedAmount = requestedAmount.subtract(amountToGetNonDustValue);
```
julia-zack:
```suggestion
        Coin amountToSend = requestedAmount.subtract(amountToGetNonDustValue);
```

is not the requestedAmount indeed, but the actually sent one

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -87,4 +86,78 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+                                                                                   List<TransactionOutput> releaseTransactionChangeOutputs,
+                                                                                   Coin requestedAmount) {
+        Coin inputTotalAmount = releaseTransaction.getInputSum();
+        Coin expectedChangeAmount = inputTotalAmount.subtract(requestedAmount);
+        assertUserAndChangeOutputsValues(releaseTransaction, releaseTransactionChangeOutputs, requestedAmount, expectedChangeAmount);
+    }
+
+    private static void assertUserAndChangeOutputsValues(BtcTransaction releaseTransaction,
+                                                         List<TransactionOutput> releaseTransactionChangeOutputs,
+                                                         Coin requestedAmount,
```
julia-zack:
```suggestion
    private static void assertUserAndChangeOutputsValues(BtcTransaction releaseTransaction,
                                                         List<TransactionOutput> releaseTransactionChangeOutputs,
                                                         Coin amountToSend,
```

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -87,4 +86,78 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+                                                Address expectedDestinationAddress,
+                                                NetworkParameters networkParameters) {
+        for (TransactionOutput output : releaseTransactionOutputs) {
+            Address destinationAddress = output.getScriptPubKey().getToAddress(networkParameters);
+            assertEquals(expectedDestinationAddress, destinationAddress);
+        }
+    }
+
+    public static void assertUserAndChangeOutputsValuesWhenOriginalChangeIsDust(BtcTransaction releaseTransaction,
+                                                                                List<TransactionOutput> releaseTransactionChangeOutputs,
```
julia-zack, line 100: wdyt about receiving just the releaseTransaction as a param? You can get the change and user outputs from there.
Thats less error prone, since you could pass _oneReleaseTx, anotherReleaseTxChangeOutputs_ as params by mistake.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -87,4 +86,78 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+    }
+
+    private static void assertUserAndChangeOutputsValues(BtcTransaction releaseTransaction,
+                                                         List<TransactionOutput> releaseTransactionChangeOutputs,
+                                                         Coin requestedAmount,
+                                                         Coin expectedChangeAmount) {
+        Coin changeOutputsAmount = getChangeOutputsAmount(releaseTransactionChangeOutputs);
+        assertEquals(expectedChangeAmount, changeOutputsAmount);
+
+        Coin userOutputsAmount = releaseTransaction.getOutputSum().subtract(changeOutputsAmount);
```
julia-zack, line 127: wdyt about calculating the user outputs amount checking the outputs and then assert it matches `releaseTransaction.getOutputSum().subtract(changeOutputsAmount)` ?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
```diff
@@ -87,4 +86,78 @@ public static void assertBtcTxVersionIs1(BtcTransaction releaseTransaction) {
...
+        Coin releaseTransactionFees = releaseTransaction.getFee();
+        Coin userOutputsAndFeesAmount = releaseTransactionFees.add(userOutputsAmount);
+        assertEquals(requestedAmount, userOutputsAndFeesAmount);
+        Coin inputTotalAmount = releaseTransaction.getInputSum();
+        assertEquals(inputTotalAmount, userOutputsAndFeesAmount.add(changeOutputsAmount));
+    }
+
+    private static Coin getChangeOutputsAmount(List<TransactionOutput> outputs) {
```
julia-zack:
```suggestion
    private static Coin getOutputsAmount(List<TransactionOutput> outputs) {
```

more accurate, since you are receiving just a list of outputs

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julia-zack: what i meant was to direclty do `feePerKb = value` in the beforeEach general setup function. But no big deal, is okay this way

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julia-zack: leave it as is, is okay

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,614 @@
...
+        wallet = BridgeUtils.getFederationSpendWallet(
+            btcContext,
+            federation,
+            utxos,
+            true,
+            bridgeStorageProvider
+        );
+    }
+
+    private ReleaseTransactionBuilder setupWalletAndCreateReleaseTransactionBuilder(List<UTXO> utxos) {
```
julia-zack: no need to pass the utxos right? You are using the global `federationUTXOs` variable
```suggestion
    private ReleaseTransactionBuilder setupWalletAndCreateReleaseTransactionBuilder() {
```

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,614 @@
...
+    protected ReleaseTransactionBuilder createReleaseTransactionBuilder() {
+        return new ReleaseTransactionBuilder(
+            BTC_MAINNET_PARAMS,
+            wallet,
+            federationFormatVersion,
+            federationAddress,
+            feePerKb,
+            activations
+        );
+    }
```
julia-zack: actually, 

```suggestion
    private void setupWalletAndCreateReleaseTransactionBuilder() {
        setUpWallet();
        releaseTransactionBuilder = new ReleaseTransactionBuilder(
            BTC_MAINNET_PARAMS,
            wallet,
            federationFormatVersion,
            federationAddress,
            feePerKb,
            activations
        );
    }
```

and use federationUTXOs variable directly in `setUpWallet`.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
```diff
@@ -0,0 +1,614 @@
...
+        List<TransactionInput> pegoutInputs = pegoutTransaction.getInputs();
+        int expectedNumberOfInputs = 1;
+        assertEquals(expectedNumberOfInputs, pegoutInputs.size());
+        assertReleaseTxInputsHasProperFormatAndBelongsToStandardMultisigFederation(
+            pegoutTransaction,
+            federationRedeemScript,
+            federationUTXOs
+        );
+        assertPegoutTxWithUserAndChangeOutputsWhenOriginalChangeIsNonDust(pegoutTransaction, requestedAmount);
+        assertSelectedUtxosBelongToTheInputs(amountToResult.selectedUTXOs(), pegoutInputs);
```
julia-zack: you are repeating this -almost- exact code block in -almost- every test. Consider doing something like:

```
assertPegoutWithOriginalChange {
        assertPegout();
        assertPegoutTxWithUserAndChangeOutputsWhenOriginalChangeIsNonDust();
}

assertPegout() {
         assertSuccessfulResponseCode();
         assertBtcTxVersion();
         assertNumberOfInputs(expectedNumberOfInputs, pegoutInputs);
         assertReleaseTxInputsHasProperFormatAndBelongsToStandardMultisigFederation(whatever);
         assertSelectedUtxosBelongToTheInputs(whatever);
}
```


this may seem just an "aesthetic" change, but it improves readability a lot

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/87423e3209444412f2c586ea4a955682c2bfdf80

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/203506e2a70245a860f800b027e3a9c1204cc3c5

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: done https://github.com/rsksmart/rskj/commit/6cc5dba987aa9f35f8337fe1aedfc0be4f7dbca6

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
julianlen: https://github.com/rsksmart/rskj/commit/ce38dc2c14f1433412029818cb9f95612e2d98bb

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: this'll be figured out when improving the assertions

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: https://github.com/rsksmart/rskj/commit/ac706a96505df6fb8e9fa1fce259e9cd8398d9fa

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: https://github.com/rsksmart/rskj/commit/c55acd2fc54faa630215013b69c5a470add37ce2

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: it'll be handled in assertions improvemnt

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: already added a task for this

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: yes, this will be managed in assertions improvement

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
julianlen: yes, this will be managed in assertions improvement

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
jeremy-then: Actually, the suggestion might introduce some other issues.

It's true that we can extract the change from it. The request value? Probably not that easy. 

For example, we are using this `assertUserAndChangeOutputsValuesWhenOriginalChangeIsDust` function from `ReleaseTransactionBuilderBuildBatchedPegoutsTest`, in `assertBatchedPegoutsTxHasPegoutAndChangeOutputsWhenOriginalChangeIsDust`.

This is called by a test like `buildBatchedPegouts_whenOriginalChangeIsMaxDustValue_shouldCreateBatchedPegoutsTxDecrementingFirstOutputAndSettingNonDustChange`, that uses `releaseTransactionBuilder.buildBatchedPegouts`.

This way, fees would be already extracted from the outputs. Meaning, we would probably not be able to get the original `requested` amount.

---

Something simpler would be to simply use the `MINIMUM_PEGOUT_TX_VALUE` value directly, since right now, this function is being passed that value exactly. But, that would fix this function to use this value and not any other in the future.

---

Right now, even though the function might look it it could be simplify (and I think it could but complicating other things), it is ready to scale to other scenarios.

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderAssertions.java
jeremy-then: Good suggestion. I’d rather not fold that into this PR to keep scope tight, the existing assertions already cover our intent here. OK to follow up separately?

rskj-core/src/test/java/co/rsk/peg/ReleaseTransactionBuilderBuildAmountToTest.java
jeremy-then: I suggest to keep it as it is now.

marcos-iov (approved):

PR #3475, opened 2026-03-12
Default properties for eth_getLogs

wlaclaw:
This PR added default safeguard limiters for RPC `eth_getLogs` function.

### DIFF
--- a/rskj-core/src/main/resources/config/devnet.conf
+++ b/rskj-core/src/main/resources/config/devnet.conf
@@ -50,3 +50,16 @@ wallet {
 }
 
 targetgaslimit = 6800000
+
+
+rpc {
+    logs {
+	# Reasonable limit to do not overload your node
+	# You could lower it down to 1k as a precaution
+        maxBlocksToQuery = 5000
+	# Ballpark size esitmation for 20k logs is about 10Mb
+	# With gzip compression it could be less than 1Mb
+        maxLogsToReturn = 20000
+    }
+}
+
--- a/rskj-core/src/main/resources/config/devnet.conf
+++ b/rskj-core/src/main/resources/config/devnet.conf
@@ -51,15 +51,3 @@ wallet {
 
 targetgaslimit = 6800000
 
-
-rpc {
-    logs {
-	# Reasonable limit to do not overload your node
-	# You could lower it down to 1k as a precaution
-        maxBlocksToQuery = 5000
-	# Ballpark size esitmation for 20k logs is about 10Mb
-	# With gzip compression it could be less than 1Mb
-        maxLogsToReturn = 20000
-    }
-}
-
--- a/rskj-core/src/main/resources/config/main.conf
+++ b/rskj-core/src/main/resources/config/main.conf
@@ -102,15 +102,3 @@ wallet {
     accounts = []
 }
 
-
-rpc {
-    logs {
-	# Reasonable limit to do not overload your node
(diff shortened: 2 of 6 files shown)


rskj-core/src/main/resources/config/devnet.conf
```diff
@@ -49,3 +49,16 @@ wallet {
...
 
 targetgaslimit = 6800000
+
+
+rpc {
+    logs {
+	# Reasonable limit to do not overload your node
+	# You could lower it down to 1k as a precaution
+        maxBlocksToQuery = 5000
+	# Ballpark size esitmation for 20k logs is about 10Mb
```
fmacleal: **nitpick**:

Typo in `estimation`.


```suggestion
	# Ballpark size estimation for 20k logs is about 10Mb
```

rskj-core/src/main/resources/config/devnet.conf
```diff
@@ -49,3 +49,16 @@ wallet {
 }
 
 targetgaslimit = 6800000
+
+
+rpc {
+    logs {
+	# Reasonable limit to do not overload your node
```
fmacleal: **nitpick**:

Rephrase to make it more clear: 

```suggestion
	# Reasonable limit to not to overload your node
```

rskj-core/src/main/resources/config/main.conf
fmacleal, line 1: **nitpick**:

Same typos from above apllies here.

rskj-core/src/main/resources/config/regtest.conf
fmacleal, line 1: **nitpick**:

Same typos from above apllies here.

rskj-core/src/main/resources/config/testnet.conf
fmacleal, line 1: **nitpick**:

Same typos from above apllies here.

rskj-core/src/main/resources/config/testnet2.conf
fmacleal, line 1: **nitpick**:

Same typos from above apllies here.

fmacleal (commented):
**praise**:

Nice analysis with the limits! 

**suggestion 1**:
Should we add somehow information that to let it unlimited, we can just set it to 0 in the `expected.conf` file? We usually use that file to explain some configs.

 Something like:
```
# Set to 0 to disable limit (unlimited)
maxBlocksToQuery = 5000
# Set to 0 to disable limit (unlimited)
maxLogsToReturn = 20000
```

**suggestion 2**:
I am missing the `reference.conf` change, actually, if we set it there...we don't need to set it in the others unless we override some of these values. Because all others, inherit the default value from `reference.conf`.

**issue 3**:

We should review the class `LogFilter::checkLimit` ([here](https://github.com/rsksmart/rskj/blob/ba65859cc874ed036bde047125c686b14064fe69/rskj-core/src/main/java/org/ethereum/rpc/LogFilter.java#L344-L350)). Because this is using the `totalToQuery = to - from`, and this counts the number of gaps between block numbers, not the number of blocks. A range `[1, 5001]` spans **5001 blocks** but `totalToQuery = 5000`, so the limit is never triggered. We probably should a `+1` in that line.

**thought**:

I am wondering now, that this can break integration fro Dapps, indexers, or whatever that it's usinga higher limit. We probably should double check this change with our Product Manager. 🤔

fmacleal (approved):
LGTM.

We just need to discuss the impact of these limits now with product.

PR #3484, opened 2026-03-20
Java 21 pegouts wating for confirmation ordering

wlaclaw:
## Description
Change java 21 collection init parameters to make ordering of the elements the same as for Java 17.

## Motivation and Context
Some entries must have same ordering as in previous versions of Java.

## How Has This Been Tested?
Run new  mainnet/testnet nodes with different JVMs - all nodes were able to sync.


rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -88,7 +88,8 @@ public int hashCode() {
     private Set<Entry> entries;
 
     public PegoutsWaitingForConfirmations(Set<Entry> entries) {
-        this.entries = new HashSet<>(entries);
+	this.entries = new HashSet<>(Math.max((int) (entries.size()/.75f) + 1, 16));
```
fmacleal: **question**:

Can you add meaningful names for this max value? Why one of the values chosen is the entries + 25% of it + 1? How about the 16? Was this the change in the  HashSet initialCapacity?

rskj-core/build.gradle
```diff
@@ -42,9 +42,21 @@ java {
 
 application {
     mainClass = 'co.rsk.Start'
-    applicationDefaultJvmArgs = ["-server", "-Xss32m", "-Xms3G", "-Xmx5G", "-XX:+UseCompressedOops", "-XX:-OmitStackTraceInFastThrow"]
+    applicationDefaultJvmArgs = ["-server", "-Xss32m", "-Xms3G", "-Xmx8G",
+    "-XX:+UseCompressedOops", "-XX:-OmitStackTraceInFastThrow"
+    ]
 }
 
+tasks.named('run', JavaExec) {
```
fmacleal, line 50: **praise**:

cool, it makes indeed easier.

nagarev:
:shipit: LGTM

fmacleal (commented):
LGTM, seems a good catch.

My opinion is, very good fix, seems straightforward. Buuut, I think it's still too fragile and we should go with the sorting approach.

I have added a more deep discussion in the slack.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
italo-sampaio: I hate to be that guy but it seems like you used tabs here for indentation =)

italo-sampaio: In the same spirit, I think it deserves a comment explaining what is going on here. I'm guessing by the context that you're explicitly replicating the behavior of that HashSet constructor in Java 17, but it won't be obvious in 3 years for sure.

wlaclaw: Ouch. It should not be like that.

wlaclaw: @fmacleal This is basically copy paste from Java 17 implementation adding "meaningful names" would make it less obvious.

I just need to add some comments.

fmacleal: Whatever makes it clear. But magic numbers isn't welcomed at all.

I would rather to have a constant with meaningful name than comment, but, it's my opinion.

fmacleal (approved):

rskj-core/build.gradle
```diff
@@ -42,9 +42,21 @@ java {
 
 application {
     mainClass = 'co.rsk.Start'
-    applicationDefaultJvmArgs = ["-server", "-Xss32m", "-Xms3G", "-Xmx5G", "-XX:+UseCompressedOops", "-XX:-OmitStackTraceInFastThrow"]
+    applicationDefaultJvmArgs = ["-server", "-Xss32m", "-Xms3G", "-Xmx8G",
+    "-XX:+UseCompressedOops", "-XX:-OmitStackTraceInFastThrow"
```
donequis, line 46: Is this part of the fix?

wlaclaw: This is part of the local debug process. That should be there.

fmacleal (approved):
LGTM

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -88,7 +88,11 @@ public int hashCode() {
     private Set<Entry> entries;
 
     public PegoutsWaitingForConfirmations(Set<Entry> entries) {
-        this.entries = new HashSet<>(entries);
+        // This is a standart code for `new HashSet<>(entries);` in Java 17
```
jeremy-then, line 91:
```suggestion
        // This is a standard code for `new HashSet<>(entries);` in Java 17
```

jeremy-then (approved):
LGTM.

PR #3486, opened 2026-03-23
Removing flag that fails when there is no cache

fmacleal:
Removing the enforcing from the cache in gradle wrapper action step to avoid errors for when there is no cache available.


dlealmac (approved):
LGTM!

marcos-iov (approved):

PR #3495, opened 2026-04-07
fix: debug_traceTransaction failing with callTracer

italo-sampaio:
## Description

Fix `debug_traceTransaction` crashing with `IllegalStateException: Memory was changed during slice lifetime` when using the `callTracer` on transactions that contain internal contract-to-contract calls.

The fix eagerly copies the call data `BytesSlice` into a stable `Bytes` object when constructing a `ProgramInvoke` for internal calls, preventing a stale `MemorySlice` reference from being read after the parent Memory has been modified.

**Changed files:**
- `ProgramInvokeFactoryImpl.java` — one-line fix: copy `dataIn` before storing in `ProgramInvokeImpl`
- `CallTracerTest.java` — regression test with a recursive self-calling contract
- `call_tracer_internal_call.txt` — DSL test fixture for the above

## Motivation and Context

Users reported that `debug_traceTransaction` with `{"tracer": "callTracer"}` fails on the Vetiver release for any transaction involving internal calls.

**Root cause:** The Memory versioning optimization (commit `8aa3ac2529`) introduced `MemorySlice` — a zero-copy view into VM Memory. For internal CALLs, `Program.callToAddress()` creates a `MemorySlice` for the call data and stores it in the child `ProgramInvokeImpl.msgData`. After the child returns, `memorySaveLimited()` writes return data to parent Memory, incrementing its version. When `CallTraceTransformer.buildForCall()` later reads that call data from the trace, the version check fails.

The fix makes the internal-call path consistent with the top-level transaction path, which already uses `Bytes.of(data)` (a safe copy). The `MemorySlice` optimization remains intact for KECCAK256 and CREATE2, where slices are consumed immediately.

## How Has This Been Tested?

- **Integration test:** Added `CallTracerTest.traceTransactionWithInternalCalls()` — deploys a contract with recursive self-calls (`this.recurse(depth - 1)`), executes a transaction with 2 levels of internal calls, and verifies that `callTracer.traceTransaction()` succeeds and returns the expected call hierarchy.
- **k6 test:** Created a test scenario in the k6 test suite to reproduce against a live regtest node (deploys CPUStress contract, calls `runRecursion(3)`, traces with both `callTracer` and `rskTracer`).
- **Local node verification:** Reproduced the original error on a local regtest node, confirmed the fix resolves it.
- All existing `CallTracerTest` tests continue to pass.

## Types of changes

## Checklist:

* **Other information**: The default tracer (`rskTracer`) was never affected — it does not access `ProgramInvoke.getDataCopy()` post-execution. Normal (non-tracing) execution is also unaffected because the child program reads the slice before the parent memory is modified.

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/vm/program/invoke/ProgramInvokeFactoryImpl.java
+++ b/rskj-core/src/main/java/org/ethereum/vm/program/invoke/ProgramInvokeFactoryImpl.java
@@ -164,7 +164,7 @@ public class ProgramInvokeFactoryImpl implements ProgramInvokeFactory {
         long agas = inGas;
         DataWord callValue = inValue;
 
-        BytesSlice data = dataIn;
+        BytesSlice data = dataIn != null ? dataIn.copyBytes() : null;
         DataWord lastHash = program.getPrevHash();
         DataWord coinbase = program.getCoinbase();
         DataWord timestamp = program.getTimestamp();
--- a/rskj-core/src/test/java/co/rsk/rpc/modules/debug/trace/call/CallTracerTest.java
+++ b/rskj-core/src/test/java/co/rsk/rpc/modules/debug/trace/call/CallTracerTest.java
@@ -101,17 +101,14 @@ class CallTracerTest {
     }
 
     /**
-     * Reproduces RSKCORE-5466: debug_traceTransaction with callTracer fails with
-     * "Memory was changed during slice lifetime" on transactions with internal calls.
+     * Regression test for RSKCORE-5466: verifies that debug_traceTransaction with callTracer
+     * succeeds on transactions with internal calls.
      *
-     * The contract calls itself recursively via this.recurse(depth - 1). During execution,
-     * Program.callToAddress() creates a MemorySlice for the call data, which gets stored
-     * in the child ProgramInvokeImpl. After the child returns, memorySaveLimited() writes
-     * return data to parent Memory, incrementing its version and invalidating the slice.
-     * CallTraceTransformer.buildForCall() then tries to read the stale slice and throws.
-     *
-     * TODO: Once the fix is applied, this test should be changed to assert success
-     *       (remove assertThrows, uncomment the assertions below).
+     * The contract calls itself recursively via this.recurse(depth - 1). Previously,
+     * Program.callToAddress() created a MemorySlice for the call data which got stored
+     * in the child ProgramInvokeImpl. After the child returned, memorySaveLimited() wrote
+     * return data to parent Memory, invalidating the slice and causing an IllegalStateException
+     * when CallTraceTransformer.buildForCall() tried to read it.
      */
     @Test
     void traceTransactionWithInternalCalls() throws Exception {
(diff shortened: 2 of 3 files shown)


rskj-core/src/main/java/org/ethereum/vm/program/invoke/ProgramInvokeFactoryImpl.java
```diff
@@ -164,7 +164,7 @@ public ProgramInvoke createProgramInvoke(Program program, DataWord toAddress, Da
         long agas = inGas;
         DataWord callValue = inValue;
 
-        BytesSlice data = dataIn;
+        BytesSlice data = dataIn != null ? dataIn.copyBytes() : null;
```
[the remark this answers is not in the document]
italo-sampaio: In practice, this is a non-issue: `dataIn` is always a `MemorySlice` when coming from `Program.callToAddress()` (the only caller for internal calls), so this suggestion adds complexity with no practical benefits.

rskj-core/src/test/java/co/rsk/rpc/modules/debug/trace/call/CallTracerTest.java
```diff
@@ -100,6 +100,43 @@ void internalOoGErrorTrace() throws Exception {
...
 
+    /**
+     * Regression test for RSKCORE-5466: verifies that debug_traceTransaction with callTracer
+     * succeeds on transactions with internal calls.
+     *
+     * The contract calls itself recursively via this.recurse(depth - 1). Previously,
+     * Program.callToAddress() created a MemorySlice for the call data which got stored
+     * in the child ProgramInvokeImpl. After the child returned, memorySaveLimited() wrote
+     * return data to parent Memory, invalidating the slice and causing an IllegalStateException
+     * when CallTraceTransformer.buildForCall() tried to read it.
```
fmacleal, line 111: **suggestion**:

We can remove the Jira ticket reference here, also, simplify the explanation. No. need to go that deep.

rskj-core/src/test/java/co/rsk/rpc/modules/debug/trace/call/CallTracerTest.java
```diff
@@ -100,6 +100,43 @@ void internalOoGErrorTrace() throws Exception {
...
+     */
+    @Test
+    void traceTransactionWithInternalCalls() throws Exception {
```
fmacleal, line 114: **thought**:

Does it worth to add another level of recursion asserting deeper structure? 🤔 

For example, that recurse(2) produces 2 levels of nested calls?

rskj-core/src/test/resources/dsl/call_tracer_internal_call.txt
```diff
@@ -0,0 +1,57 @@
+comment
+
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+// Minimal contract with self-calling recursion to trigger internal CALL opcodes.
+// Used to reproduce RSKCORE-5466: MemorySlice becomes stale after child call
+// returns and parent memory is written to via memorySaveLimited().
```
fmacleal, line 8: **suggestion**:

I suggest also remove the reference to the Jira ticket number here as well.

fmacleal (commented):
Looks good, I just have a few suggestions.

PR #3502, opened 2026-04-15
Added RSKIP-546 support for Type 1 and Type 2 transactions

dlealmac:
## Description
This pull request implements RSKIP546 (based on Ethereum's EIP-2718). It includes transaction parsing/serialization, activation-gated validation and execution, typed receipt encoding, and RPC updates so typed transactions are correctly exposed through the API.

## Motivation and Context
RSKIP-546 is required to support EIP-2930 and EIP-1559-style transaction formats. This change enables clients to submit, process, and query typed transactions while preserving legacy transaction behavior and keeping pre-fork data compatible across the activation boundary.

## How Has This Been Tested?
Coverage was added and updated with `Rskip546DslTest`, `Rskip546HardforkTest`, and related typed transaction tests such as `TypedTransactionIntegrationTest`, `TypedTransactionReceiptTest`, `TypedTransactionSecurityTest`, `TransactionTypeTest`, and `CallArgumentsParamTest`. These tests cover Type 1 and Type 2 encoding, receipt format, access list handling, RPC serialization, pre-activation rejection, post-activation acceptance, and mixed legacy/typed block behavior.

## Types of changes

## Checklist:

* **Other information**:
Typed transactions are format-only. RSKj does not implement the EIP-1559 dynamic fee market or access list EVM optimization. baseFeePerGas is always 0x0. maxPriorityFeePerGas and maxFeePerGas are accepted but the effective gas price is min(maxPriorityFeePerGas, maxFeePerGas) per RSKIP-546.


rskj-core/src/main/java/org/ethereum/rpc/dto/TransactionResultDTO.java
```diff
@@ -49,6 +62,15 @@ public class TransactionResultDTO {
...
     private String s;
 
+    // EIP-2718 typed transaction fields (null for legacy transactions)
+    private String chainId;
+    private List<AccessListEntryDTO> accessList;
+    private String yParity;
+
+    // EIP-1559 Type 2 only fields (null for legacy and Type 1)
+    private String maxPriorityFeePerGas;
+    private String maxFeePerGas;
```
wlaclaw, line 72: I strongly encourage to split it into several sealed classes do not not have non `null` values that activates on some casese.

rskj-core/src/main/java/org/ethereum/core/Transaction.java
```diff
@@ -110,6 +117,16 @@ public class Transaction {
...
 
+    /** RSKIP546: Access list bytes (RLP-encoded) for Type 1 and Type 2 */
+    private final byte[] accessListBytes;
+
+    /**
+     * RSKIP-546: EIP-1559 fee fields for <em>standard</em> Type 2 only ({@code null} for legacy, Type 1, Type 3/4,
+     * and RSK-namespace Type 2). Effective gas price is {@code min(maxPriorityFeePerGas, maxFeePerGas)}.
+     */
+    private final Coin maxPriorityFeePerGas;
+    private final Coin maxFeePerGas;
```
wlaclaw, line 128: I strongly encourage to split it into several sealed classes do not not have non `null` values that activates on some cases.

wlaclaw (changes requested):
I guess it is a good time to introduce sealed classes for different transaction types. Maybe we could split base Tx to more that 3Tx classes.

https://docs.oracle.com/en/java/javase/17/language/sealed-classes-and-interfaces.html

wlaclaw:
BTW There could be another way to deal with several null propeties in TX by using sealed classes to store internal properties (this is just an assumption).
For example all non constant properties from `Transaction`  class should be moved to `TxProps` that will have sealed classes like `TxLegacyProps`, `Tx1Props`, `Tx2Props` etc. This properties will be stored inside `Transaction` as a `TxProps` class so pattern matchig could be used to work with them. 

The main point here is to have particular type for a transaction with particular set of not null properties.
If something does not exists for some TX type  it should not have any properties and methods to work with it.

meldsun0 (commented):
Suggestion: since transactions now use TransactionEncoder + TransactionEncoderFactory, it might be worth applying the same pattern to receipts.

Right now TransactionReceipt mixes data with encoding logic (including typed vs legacy branching). Introducing a TransactionReceiptEncoder + factory would keep concerns cleaner, align both designs, and make it easier to extend for new receipt types.

Not a blocker, just for consistency and maintainability.

fmacleal:
We are closing this one, because the branch https://github.com/rsksmart/rskj/tree/rskip545 has all the changes from this one and we will use it.

PR #3507, opened 2026-04-28
RSKIP559 deterministic pegouts ordering

wlaclaw:
Adding deterministic ordering for entries into `getNextPegoutWithEnoughConfirmations`.
The patch itself is small but it is require a hardfork with separate RSKIP https://github.com/rsksmart/rskIPs/pull/559


rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +40,127 @@
...
+
+        public boolean removeEntry(Entry entry) {
+            return this.entriesSet.remove(entry);
+        }
+
+        /**
+         * Added entry only if BTC TX is unique.
+         * NOTE: should be used in tests only.
+         */
+        public void addEntryUniqueBtcTx(Entry entry) {
```
italo-sampaio: this seems to be unused, any specific reason to keep it?

italo-sampaio (dismissed):
LGTM! Left a super-minor comment regarding a method that is documented as "for test only", but is not used anywhere

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
wlaclaw: Thanks for noticing. Code removed.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +40,113 @@
...
+        return entries.entriesSet.stream().filter(e -> e.getPegoutCreationRskTxHash() == null).collect(Collectors.toUnmodifiableList());
+    }
+
+    public Collection<Entry> getEntriesWithHash() {
+        return entries.entriesSet.stream().filter(e -> e.getPegoutCreationRskTxHash() != null).collect(Collectors.toUnmodifiableList());
+    }
+
+    public Collection<Entry> getEntries() {
+        return entries.entriesSet.stream().collect(Collectors.toUnmodifiableList());
+    }
```
fmacleal, line 73: **question**:

Why change to a Collection and break the previous contract, was it needed?
The Collection now allows duplicated values, where the Set didn't allow it.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +40,113 @@
...
+     *
+     * @return an optional with an entry with enough confirmations if found. If not, an empty optional.
+     */
+    public Optional<Entry> getNextPegoutWithEnoughConfirmations(Long currentBlockNumber, Integer minimumConfirmations, ForBlock activations) {
+        var rskip559 = activations.isActive(ConsensusRule.RSKIP559);
+        return this.entries.getNextPegoutWithEnoughConfirmations(currentBlockNumber, minimumConfirmations, rskip559);
+    }
+
+    public void add(Entry entry) {
+        this.entries.addEntry(entry);
```
fmacleal, line 95: **issue**:

In the previous code, we had a guard to prevent duplicated transactions:

```
       if (entries.stream().noneMatch(e -> e.getBtcTransaction().equals(transaction))) {
            entries.add(new Entry(transaction, blockNumber, rskTxHash));
        }
```

In this new one, this logic was removed. Any reason to remove it? I could actually replicate the issue adding the sameBtcTx when we have different blockNumbers, or rskHash even with rskip559 active. Maybe worth to check with powpeg the impact of it, I would rather for us to not change any current behaviour. It could lead to some fork.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +40,113 @@
...
+    }
+
+    public boolean removeEntry(Entry entry){
+        return entries.removeEntry(entry);
+    }
+
+    /**
+     * Encapsulate entries while preservin sorting order before fork.
```
fmacleal: **chore**:

Typo here, it should be `preserving`


```suggestion
     * Encapsulate entries while preserving sorting order before fork.
```

rskj-core/src/main/resources/reference.conf
```diff
@@ -103,6 +103,7 @@ blockchain = {
             rskip544 = vetiver900
             rskip551 = vetiver900
             rskip552 = vetiver900
+            rskip559 = -1
```
fmacleal: **chore**:

Trailing whitespace here.


```suggestion
            rskip559 = -1
```

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +40,113 @@
...
+     */
+    public static class EntriesStore {
+
+        // From java SDK
+        private static final float DEFAULT_LOAD_FACTOR = 0.75f;
+
+        private HashSet<Entry> entriesSet;
```
fmacleal: **suggestion**:

It could be `final` no?

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +40,113 @@
  * @author Ariel Mendelzon
  */
 public class PegoutsWaitingForConfirmations {
+
+    private EntriesStore entries;
```
fmacleal: **suggestion**:

It could be `final` no?

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +40,113 @@
...
+
+    public PegoutsWaitingForConfirmations(Set<Entry> entries) {
+        this.entries = new EntriesStore(entries);
+    }
+
+    public Collection<Entry> getEntriesWithoutHash() {
+        return entries.entriesSet.stream().filter(e -> e.getPegoutCreationRskTxHash() == null).collect(Collectors.toUnmodifiableList());
```
fmacleal: **nipick**:

We could use directly `Stream.toList()` here, it's the same result.

```suggestion
        return entries.entriesSet.stream().filter(e -> e.getPegoutCreationRskTxHash() == null).toList();
```

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +40,113 @@
...
+    }
+
+    public Collection<Entry> getEntriesWithHash() {
+        return entries.entriesSet.stream().filter(e -> e.getPegoutCreationRskTxHash() != null).collect(Collectors.toUnmodifiableList());
```
fmacleal: **nipick**:

We could use directly `Stream.toList()` here, it's the same result.


```suggestion
        return entries.entriesSet.stream().filter(e -> e.getPegoutCreationRskTxHash() != null).toList();
```

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +40,113 @@
...
+    }
+
+    public Collection<Entry> getEntries() {
+        return entries.entriesSet.stream().collect(Collectors.toUnmodifiableList());
```
fmacleal: **nipick**:

We could use directly `Stream.toList()` here, it's the same result.


```suggestion
        return entries.entriesSet.stream().toList();
```

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigTest.java
```diff
@@ -156,6 +156,7 @@ class ActivationConfigTest {
         "    rskip544: vetiver900",
         "    rskip551: vetiver900",
         "    rskip552: vetiver900",
+        "    rskip559: 0",
```
fmacleal: **issue**:

Let's follow the pattern and add this RSKIP559 to a specific released version. 
You can change [this](https://github.com/rsksmart/rskj/blob/509f1d870bbfe59a71cafd15c39c0775bac91f78/rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigsForTest.java#L460-L462) method and add the list of RSKIPs to be activated in vetiver. This way, we don't have to manually activate RSKIPs for tests, and we follow the correct order of activations tied to the releases.

fmacleal (commented):
Good job!

Simple and straightforward PR. I just have some small comments, but there is a few important that we should discuss and address. Mainly the one related with duplicated tx addition and the change of the contract from Set to Collection.

If you want, we can discuss in slack or in the daily.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
wlaclaw: I need here exactly immutable list. To avoid any future attempts to add entires there.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
wlaclaw: The same. toList() returns mutable list.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
wlaclaw: It is not secure. Because we already have some code that worked like `getEntries().add(something)` this is not acceptable.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
wlaclaw: We do not share mutable inner state via getEntries. These are immutable collections so you will not be able to add anything there. Thus it is more reasonable to generalize it as a collection

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigTest.java
wlaclaw: Sure, but I do not know yet what a next release name will be.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
wlaclaw: My bad. I guess now know why it happened. I asked AI to made some routine changes with rebasing and removing part of previous draft commit.
It helped but tweaked add_existing test instead adding deduplication.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
fmacleal: Ok, that makes sense. It was a bit of a work to change it throughout the code, but I think it's an understandable and correct change to encapsulate this list and prevent unwanted changes.

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigTest.java
fmacleal: You can add a `tbd1000` for example.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
fmacleal: From what I have checked in the `toList` implementation, the list returned is unmodifieable as well. And if you call .add or .remove on it, it would throw an exception. So, it would be basically the same, but with a better readability.

Image

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigTest.java
wlaclaw: Created `NextRelease` until new name will be defined

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +39,115 @@
...
+    public Collection<Entry> getEntriesWithoutHash() {
+        return entries.entriesSet.stream().filter(e -> e.getPegoutCreationRskTxHash() == null).toList();
+    }
+
+    public Collection<Entry> getEntriesWithHash() {
+        return entries.entriesSet.stream().filter(e -> e.getPegoutCreationRskTxHash() != null).toList();
+    }
+
+    public Collection<Entry> getEntries() {
+        return entries.entriesSet.stream().toList();
```
fmacleal: **thought**:

I am still not fully convinced that it's better change this to Collection. What's the real benefit of it? Do we want this generic enough? Don't we want to avoid duplicated elements? (In this case, Set is better than collection)

Also, I can see that now, in all the tests we are wrapping the collection in a `new HashSet<>()` stating again the fact that we need a Set. 

If someone converts that to a List from the collection, the `contains` becomes O(n) instead of the O(1) provided by the Set. `List.equals` is order sensitive while `Set.equals` is not, which is why a lot of tests had to wrap it in a `HashSet` to compare equality. 🤔 

I won't prevent the approval here, but probably is something to be reviewed by the @rsksmart/rsk-fed team, since this part of the code is owned by them and we are changing the data structure from Set to a more generic one.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +39,115 @@
...
+         * @param withTxComparator turns on deterministic order for entries before filtering.
+         */
+        public Optional<Entry> getNextPegoutWithEnoughConfirmations(Long currentBlockNumber, Integer minimumConfirmations, boolean withTxComparator) {
+            var entries = entriesSet.stream();
+
+            if (withTxComparator) {
+                entries = entries.sorted(Entry.BTC_TX_COMPARATOR);
+            }
+
+            return entries.filter(entry -> hasEnoughConfirmations(entry, currentBlockNumber, minimumConfirmations)).findFirst();
```
fmacleal: **suggestion**:

Here if we filter before the sorting, we will save a bit of resources, since the minimal number of elements will be sorted.

Something like:

```
    var eligible = entriesSet.stream()
        .filter(entry -> hasEnoughConfirmations(entry, currentBlockNumber, minimumConfirmations));
    if (withTxComparator) {
        return eligible.min(Entry.BTC_TX_COMPARATOR);
    }
    return eligible.findFirst();
```

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +39,115 @@
...
+
+        private final HashSet<Entry> entriesSet;
+
+        public EntriesStore() {
+            // this must be equal to new HashSet() call in Java 17
+            this.entriesSet = new HashSet<>(0, DEFAULT_LOAD_FACTOR);
+        }
+
+        public EntriesStore(Collection<Entry> entries) {
+            // This is a standart code for `new HashSet<>(entries);` in Java 17
```
fmacleal: **nitpick**:


```suggestion
            // This is a standard code for `new HashSet<>(entries);` in Java 17
```

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +39,115 @@
...
+
+
     public static class Entry {
-        // Compares entries using the lexicographical order of the btc tx's serialized bytes
+
+        private BtcTransaction btcTransaction;
+
+        private Long pegoutCreationRskBlockNumber;
+
+        private Keccak256 pegoutCreationRskTxHash;
```
fmacleal: **nitpick**:

These probably could be `final` as well.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -76,67 +189,12 @@ public boolean equals(Object o) {
...
                 otherEntry.getPegoutCreationRskBlockNumber().equals(getPegoutCreationRskBlockNumber()) &&
                 (otherEntry.getPegoutCreationRskTxHash() == null && getPegoutCreationRskTxHash() == null ||
-                    otherEntry.getPegoutCreationRskTxHash() != null && otherEntry.getPegoutCreationRskTxHash().equals(getPegoutCreationRskTxHash()));
+                 otherEntry.getPegoutCreationRskTxHash() != null && otherEntry.getPegoutCreationRskTxHash().equals(getPegoutCreationRskTxHash()));
         }
 
         @Override
         public int hashCode() {
-            return Objects.hash(getBtcTransaction(), getPegoutCreationRskBlockNumber());
+            return java.util.Objects.hash(getBtcTransaction(), getPegoutCreationRskBlockNumber());
```
fmacleal: **nitpick**:

static import

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +39,115 @@
...
+                this.entriesSet.add(entry);
+            }
+        }
+
+        public boolean removeEntry(Entry entry) {
+            return this.entriesSet.remove(entry);
+        }
+    }
+
+
```
fmacleal: **nitpick**:

double blank line

```suggestion

```

fmacleal (commented):
Good job.

I am still not fully convinced with the change from `Set` to `Collections`, but I will leave this to the fed team to decide. I will not block the approval, since this is not making the code not working. It's just a matter of approach. If this is how is decided to move on and the others approve, I will approve as well.

Apart of that, there are some small improvements suggestions left.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
wlaclaw: As far as I can see tests are the only place where wrapping into HashSet is required.
So using `Collection` makes sense because it is aligned with a production code.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
wlaclaw: Good catch, will update
_But performance degradation on dozen of pegouts should not be visible anyway_

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
wlaclaw: It was used only once it a file. Sonar has no objections about it :)
WIll update.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +40,111 @@
...
+     * Encapsulate entries while preserving sorting order before fork.
+     */
+    public static class EntriesStore {
+
+        // From java SDK
+        private static final float DEFAULT_LOAD_FACTOR = 0.75f;
+
+        private final HashSet<Entry> entriesSet;
+
+        public EntriesStore() {
```
marcos-iov: Looks like this is not used anywhere

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +40,111 @@
...
+    }
+
+    public boolean removeEntry(Entry entry){
+        return entries.removeEntry(entry);
+    }
+
+    /**
+     * Encapsulate entries while preserving sorting order before fork.
+     */
+    public static class EntriesStore {
```
marcos-iov, line 105: Looks like this can be private, only used internally

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
marcos-iov: Agree that Collection is a bit too generic. Since it is internally being converted to a List I'd suggest we return a List here. And order it post RSKIP activation so we prevent any future surprises.

Basically the same logic that is being followed in BridgeSerializationUtils, could be applied here

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/NetworkUpgrade.java
```diff
@@ -36,7 +36,8 @@ public enum NetworkUpgrade {
     LOVELL700("lovell700"),
     REED800("reed800"),
     REED810("reed810"),
-    VETIVER900("vetiver900");
+    VETIVER900("vetiver900"),
+    NEXTRELEASE("NextRelease");
```
marcos-iov: I'd handle this in a separate PR, probably best when we have a name decided. In the meantime it should be enough to add the RSKIP in ConsensusRule class and in reference.conf with activation `-1`

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
wlaclaw: The point here is that we do not need to convert it in the downstream code.
I've made some updates and BridgeSerializationUtils now use Collection as is.

All the tricks with conversion `getEntries` output to something done only in some tests.

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/NetworkUpgrade.java
wlaclaw: We just decided with a team that having a placeholder is a good thing since it could be easily renamed/refactored into proper name later.

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigTest.java
marcos-iov: Should it say "NextRelease" here?

marcos-iov (dismissed):

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigTest.java
wlaclaw: Fixed

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -33,8 +39,136 @@
...
+
+        private final HashSet<Entry> entriesSet;
+
+        /**
+         * Must be equal to new HashSet() call in Java 17.
+         * Uset it to preserve old behaviour in Java21+.
+         */
+        public static HashSet<Entry> setOfEntries() {
+            return new HashSet<>(0, DEFAULT_LOAD_FACTOR);
+        }
```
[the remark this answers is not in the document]
wlaclaw: > but it currently uses an initial capacity of 0, which is not what HashSet() does (default initial capacity is 16).

This is what exactly hash set does -> stupid AI cant traverse nested calls.

fmacleal (approved):

marcos-iov (approved):

PR #3509, opened 2026-05-07
Transaction pool refactor

meldsun0:
## Description
Refactor TransactionPoolImpl to improve readability, structure, and separation of concerns.

The transaction admission flow has been reorganized into smaller, focused methods, validation logic has been extracted into TransactionPoolValidator, and the transaction lifecycle has been made more explicit (validation → pending/queued → promotion).

The goal of this refactor is to preserve the original transaction admission semantics while improving the internal structure. In addition, this PR includes a few intentional internal improvements that are not strictly refactoring.

## Motivation and Context
The previous implementation mixed validation, transaction admission, state management, and event notification in large methods. This refactor separates those responsibilities into smaller, easier-to-follow methods.

Main structural improvements:

- Extracts transaction admission validation into TransactionPoolValidator.
- Simplifies the transaction flow (validation → pending/queued → promotion).
- Breaks large methods into smaller, more focused methods.
- Centralizes transaction admission and promotion logic.
- Improves readability and maintainability.
- Cleans up obsolete transaction handling (by block/time)
- Centralizes error handling and transaction tracking
- Improves constructor validation and introduces NoOp fallbacks

Behavior preserved:

- Duplicate transaction detection.
- Transaction validation before admission.
- Replacement transaction (gas price bump) validation.
- Queued transaction promotion.
- Quota validation only when a transaction actually becomes pending.
- Aggregate balance validation only when a transaction becomes pending.
- addTransaction() notification semantics.
- addTransactions() batch notification semantics.

Additional internal changes:

- Introduces TransactionPoolValidator as the central component responsible for transaction admission validation.
- Constructor validation is stricter by validating required dependencies and configuration, and by introducing NoOp implementations where appropriate.
- Transaction tracking (transactionBlocks / transactionTimes) is now performed only for transactions that are actually admitted into the pool (pending or queued), instead of also tracking transactions that are later rejected during admission.
- Removing a transaction from the pool now also removes its associated tracking metadata (transactionBlocks and transactionTimes), keeping the cleanup state consistent.

## How Has This Been Tested?
- Existing unit tests.
- New regression tests covering:

## Types of changes

## Checklist:


rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -18,441 +18,482 @@
...
     @java.lang.SuppressWarnings("squid:S107")
-    public TransactionPoolImpl(RskSystemProperties config, RepositoryLocator repositoryLocator, BlockStore blockStore, BlockFactory blockFactory, EthereumListener listener, TransactionExecutorFactory transactionExecutorFactory, SignatureCache signatureCache, int outdatedThreshold, int outdatedTimeout, TxQuotaChecker txQuotaChecker, GasPriceTracker gasPriceTracker) {
-        this.config = config;
-        this.blockStore = blockStore;
-        this.repositoryLocator = repositoryLocator;
-        this.blockFactory = blockFactory;
-        this.listener = listener;
-        this.transactionExecutorFactory = transactionExecutorFactory;
-        this.signatureCache = signatureCache;
+    public TransactionPoolImpl(RskSystemProperties config, RepositoryLocator repositoryLocator, BlockStore blockStore, BlockFactory blockFactory, EthereumListener listener, TransactionExecutorFactory transactionExecutorFactory, SignatureCache signatureCache, int outdatedThreshold, int outdatedTimeout, TxQuotaCheckerImpl txQuotaChecker, GasPriceTracker gasPriceTracker) {
```
wlaclaw: Too much arguments on a single line, format it to have each argument on a single line.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -18,441 +18,482 @@
...
+        cleanerTimer.shutdown();
+        accountTxRateLimitCleanerTimer.shutdown();
+    }
 
-        if (accountTxRateLimitCleanerTimer != null) {
-            accountTxRateLimitCleanerTimer.shutdown();
-        }
+    @Override
+    public synchronized List<Transaction> getPendingTransactions() {
+        this.removeObsoleteTransactions(this.getBestBlockNumber(),this.outdatedThreshold, this.outdatedTimeout);
```
wlaclaw, line 155: Do we have a  strict requrement to return here exactly non obsolete transactions?
Thinking about performance impact.
We still cleaning TXs in background, so should be just perform same process each time be request pending TX. What if there will be many TX in the poll and cleaning could add a delay for this method call?

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolValidator.java
```diff
@@ -0,0 +1,135 @@
...
+
+
+    public TransactionPoolValidator(RskSystemProperties config, SignatureCache signatureCache) {
+        this.config = config;
+        this.validator = new TxPendingValidator(config.getNetworkConstants(), config.getActivationConfig(), config.getNumOfAccountSlots(), signatureCache);
+        this.signatureCache = signatureCache;
+    }
+
+    public Optional<TransactionPoolAddResult> rejectIfTransactionAlreadyKnown(final Transaction tx,  final TransactionSet pendingTransactions,
+                                                                               final TransactionSet queuedTransactions) {
```
wlaclaw: Each argument on a single line please

rskj-core/src/main/java/org/ethereum/listener/NoOpEthereumListener.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.List;
+
+public final class NoOpEthereumListener implements EthereumListener {
+
+    public static final NoOpEthereumListener INSTANCE = new NoOpEthereumListener();
+
+    private NoOpEthereumListener() {}
+
+    @Override public void trace(String output) {}
```
wlaclaw: I expect annotations to be on a separate line before method signature, not on the same line.
It just weird to see it his way for methods

wlaclaw (dismissed):
The are minor syntax notes.
The are many changes and it is hard to follow everything but overall things LGTM.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
meldsun0: Good catch. getPendingState(), getQueuedTransactions() and processBest(Block newBlock)  also performs the same cleanup so the concern is not limited to getPendingTransactions().

The performance impact you mention is valid, especially if these methods are called frequently and the pool grows large, since cleanup scans and mutates the pool while holding the lock. However, the goal of this refactor is not to change behavior. Today these methods guarantee that obsolete transactions are removed before returning data.

If we decide to remove this cleanup from the read paths, we should first verify where these methods are used and make sure no consumer relies on receiving only non-obsolete transactions. Otherwise, we could unintentionally change behavior even if the background cleanup eventually removes them.

So I would prefer to address this possible performance improvement in a separate PR, where we can properly evaluate the impact and validate that changing this behavior is safe.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -18,441 +18,492 @@
...
+        }
+        if (outdatedTimeout <= 0) {
+            throw new IllegalArgumentException("outdatedTimeout must be > 0");
+        }
+        if (config.isAccountTxRateLimitEnabled() && txQuotaChecker == null) {
+            throw new IllegalArgumentException("txQuotaChecker must not be null when account tx rate limit is enabled");
+        }
+        if (this.config.accountTxRateLimitCleanerPeriod() <= 0 ) {
+            throw new IllegalArgumentException("AccountTxRateLimitCleanerPeriod must be > 0");
+        }
```
fmacleal: **issue**:

This is a change in contract and behaviour in comparison with before. I am not sure they are worthy.  Do you think it's necessary keep this in the refactor? Does it bring many benefits?

Because before, any value lower than 0 means disabled, if we change for this approach, any operator that relied on that and had the values set to -1,  would fail in the node startup with this change. 

And what is even more problematic, is that even if the Rate limit is enabled, we are failing if the cleaner period is not set. Which doesn't make sense, because if we are disabling, who cares if we are setting the cleaner properly or not when the feature is disabled? 

And for last, although we have this verification we are still checking if the `cleanerPeriod` is higher than 0 in the method `isAccountTxRateLimitCleanerEnabled`

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -18,441 +18,492 @@
...
-    public synchronized TransactionPoolAddResult addTransaction(final Transaction tx) {
-        TransactionPoolAddResult internalResult = this.internalAddTransaction(tx);
-        List<Transaction> pendingTransactionsAdded = new ArrayList<>();
-
-        if (!internalResult.transactionsWereAdded()) {
-            return internalResult;
-        } else if (internalResult.pendingTransactionsWereAdded()) {
-            pendingTransactionsAdded.add(tx);
-            pendingTransactionsAdded.addAll(this.addSuccessors(tx)); // addSuccessors only retrieves pending successors
+        if (isAccountQuotaExceeded(tx, existingTx, repository, pendingState)) {
```
fmacleal: **issue**:

This verification is consuming the quota twice for queued txs, you are consuming this when you add the tx to the pool and again when you are promoting it to pending calling `promoteNextNonceQueuedTransactionsToPending`, that calls `addTransactionToPoolIfValid` again  which re-runs the full validation chain. This will make the quota be exceeded when the account actually used roughly half of previous throughput. And legitimate transactions will start being rejected with "account exceeds quota".

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -18,441 +18,492 @@
...
+    /**
+     * Applies validation rules before admitting a transaction into the pool.
+     * Includes checks for:
+     * - duplicate transactions
+     * - consensus validation
+     * - sufficient gas price bump for replacement
+     * - account quota limits
+     * - sufficient balance to cover pending and new transactions
+     *
+     * @return an error result if validation fails, otherwise empty
```
fmacleal: **thought**:

Do we really want all these validations at this stage? This changes the behaviour from master, it's not exactly a refactor, it also impacts how the tx are propagated in the network, since you started to doing validations before the tx is actually executed.

On master, all the tx were validated in the `TxPendingValidator` always ran for them, before this PR and after.  And it already includes:

- A per-tx balance check (`TxValidatorAccountBalanceValidator`): sender balance must cover this transaction's gasLimit × gasPrice, so a completely unfunded account can't queue anything.
- The nonce-range / account-slots check (TxValidatorNonceRangeValidator): the nonce must be within accountSlots (default 16) of the account's state nonce, so one account can park at most ~16 transactions.
- Min gas price, intrinsic gas, block gas limit, signature, not-Remasc, etc.**

My concern on doing these validations before ends up rejecting tx too early, comparing a future obligation against a present balance. If someone intend to fill any gas before the gap, common for hot-wallets, would make the tx be rejected way before. And doing these validations too early doesn't mean that we shouldn't do it again when we are executing it. Because we always need to perform the validations with fresh state. 

Did you want to do that, because other chains also does it? Changing this behaviour should be explicit in the PR description.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -18,441 +18,492 @@
...
+     * - Transactions with nonce gaps are added to the queued set
+     * - Otherwise, they are added to the pending set
+     *
+     * @param tx the transaction to process
+     * @return the result of the admission attempt
+     */
+    private TransactionPoolAddResult addTransactionToPoolIfValid(final Transaction tx) {
+        var repository = getRepositoryAtBestBlock();
+        var sender = tx.getSender(signatureCache);
+        var pendingState = getPendingState(repository);
```
fmacleal: **suggestion**:

The state construction it's expensive, copies the entire `TransactionSet`. Would be better to do it, after remove any duplicated tx running `rejectIfTransactionAlreadyKnown` before build the state and repository here.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -18,441 +18,492 @@
...
+        if (!result.transactionsWereAdded()) {
+            return result;
+        }
+
+        List<Transaction> pendingTransactionsAdded = new ArrayList<>();
+        if(result.hasPendingTransactions()) {
+            pendingTransactionsAdded.addAll(collectPendingTransactionsUnlockedBy(tx));
+        }
+
+        this.notifyTransactionPoolUpdated(pendingTransactionsAdded);
```
fmacleal, line 227: **issue**:

This event is now being emitted for every tx, whilst before, we were emitting it for batched txs, after a list of txs were processed. 

This is amplified during fork handling: `processBest` → `reintroduceBlockTransactions` → `addTransactions` can reintroduce a whole block's worth of transactions, and listeners like the miner's (`buildBlockToMine` triggers) and RPC pub-sub now get notified once per tx. The method contract of `onPendingTransactionsReceived(List<Transaction>)` clearly anticipates batches.

You could keep per-tx admission but restore batched notification: have` addTransactions` collect results from a non-notifying internal method and emit a single event at the end (this is exactly what the old internalAddTransaction / emitEvents split achieved).

rskj-core/src/main/java/co/rsk/net/handler/quota/TxQuotaChecker.java
```diff
@@ -1,244 +1,12 @@
...
- * (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
- * GNU Lesser General Public License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public License
- * along with this program. If not, see <http://www.gnu.org/licenses/>.
- */
```
fmacleal, line 17: **nitpick**:

This header shouldn't be removed.

rskj-core/src/main/java/co/rsk/net/handler/quota/TxQuotaChecker.java
```diff
@@ -1,244 +1,12 @@
...
-        private final GasPriceTracker gasPriceTracker;
-
-        public CurrentContext(Block bestBlock, PendingState state, RepositorySnapshot repository, GasPriceTracker gasPriceTracker) {
-            this.bestBlock = bestBlock;
-            this.state = state;
-            this.repository = repository;
-            this.gasPriceTracker = gasPriceTracker;
-        }
-    }
+    boolean acceptTx(Transaction newTx, @Nullable Transaction replacedTx, TxQuotaCheckerImpl.CurrentContext currentContext);
```
fmacleal: **suggestion**:

The idea of creating this interface was very good, but adding the implementation dependency on this interface inverted the DIP (Dependency Inversion Principle). Having the interface depending on the implementation isn't ideal. Because the abstraction can't be described without it's concrete class. 😅 

Suggestion made with the help of Claude:

1. Move `CurrentContext` to a top-level class (or nest it in the interface).
2. Add `TxQuota getTxQuota(RskAddress address)` to the interface (`NoOpTxQuotaChecker` returns `null`), or split a small reader interface for the debug module.
3. Type the pool constructor, `RskContext` field/getter, and `DebugModuleImpl` against `TxQuotaChecker`. Tests can then mock the interface instead of the impl and drop the `TestUtils.setInternalState` reflection.

rskj-core/src/main/java/org/ethereum/core/TransactionPoolAddResult.java
```diff
@@ -51,6 +51,11 @@ public boolean pendingTransactionsWereAdded() {
         return pendingTransactionsAdded != null && !pendingTransactionsAdded.isEmpty();
     }
 
+    //rename for later change it
```
fmacleal: **question**: 

Can't we do the rename on this refactor?

rskj-core/src/main/java/org/ethereum/listener/NoOpEthereumListener.java
fmacleal, line 1: **thought**: 

This is so similar to the class `EthereumListenerAdapter`, could you rename the other and use it if it makes more sense the name from this one? You can add any needed function from this one there.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -18,441 +18,492 @@
...
+                               TxQuotaCheckerImpl txQuotaChecker,
+                               GasPriceTracker gasPriceTracker) {
+        this.config = Objects.requireNonNull(config, "config must not be null");
+        this.repositoryLocator = Objects.requireNonNull(repositoryLocator, "repositoryLocator must not be null");
+        this.blockStore = Objects.requireNonNull(blockStore, "blockStore must not be null");
+        this.blockFactory = Objects.requireNonNull(blockFactory, "blockFactory must not be null");
+        this.transactionExecutorFactory = Objects.requireNonNull(transactionExecutorFactory, "transactionExecutorFactory must not be null");
+        this.signatureCache = Objects.requireNonNull(signatureCache, "signatureCache must not be null");
+        this.gasPriceTracker = Objects.requireNonNull(gasPriceTracker, "gasPriceTracker must not be null");
+        if (outdatedThreshold <0) {
```
fmacleal: **nitpick**:


```suggestion
        if (outdatedThreshold < 0) {
```

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -18,441 +18,492 @@
...
     }
 
-    private long getCurrentBestBlockNumber() {
-        if (bestBlock == null) {
-            return -1;
-        }
-
-        return bestBlock.getNumber();
+    private long getBestBlockNumber() {
+        return bestBlock == null ? -1 :bestBlock.getNumber();
```
fmacleal: **nitpick**:

Spacing
```suggestion
        return bestBlock == null ? -1 : bestBlock.getNumber();
```

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolValidator.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    private boolean isGasPriceBumpSufficient(Transaction newTx, Transaction oldTx) {
+        //oldGasPrice * (100 + priceBump) / 100
+        Coin oldGasPrice = oldTx.getGasPrice();
+        Coin gasPriceBumped = oldGasPrice.multiply(BigInteger.valueOf(config.getGasPriceBump() + 100L)).divide(BigInteger.valueOf(100));
+
+        return oldGasPrice.compareTo(newTx.getGasPrice()) < 0 && gasPriceBumped.compareTo(newTx.getGasPrice()) <= 0;
+    }
+
+    private Coin getTxBaseCost(Transaction tx,   Block bestBlock) {
```
fmacleal: **nitpick**:

Spacing


```suggestion
    private Coin getTxBaseCost(Transaction tx, Block bestBlock) {
```

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
```diff
@@ -63,9 +63,10 @@ class TransactionPoolImplTest {
     private TransactionPoolImpl transactionPool;
     private Repository repository;
     private ReceivedTxSignatureCache signatureCache;
-    private TxQuotaChecker quotaChecker;
+    private TxQuotaCheckerImpl quotaChecker;
+    private int outdatedThreshold;
 
-    @BeforeEach
+   @BeforeEach
```
fmacleal: **nitpick**:

Identation

```suggestion
    @BeforeEach
```

rskj-core/src/test/java/co/rsk/core/bc/TransactionPoolImplTest.java
```diff
@@ -86,7 +87,7 @@ protected RepositoryLocator buildRepositoryLocator() {
 
         RskSystemProperties rskSystemProperties = spy(rskTestContext.getRskSystemProperties());
         when(rskSystemProperties.isAccountTxRateLimitEnabled()).thenReturn(true);
-
+        outdatedThreshold= 10;
```
fmacleal: **nitpick**:

Spacing

```suggestion
        outdatedThreshold = 10;
```

fmacleal (changes requested):
Very nice job!

But I have a few comments below.

Apart of that, since you have changed some important behaviour of validations, we could have more tests to ensure the behaviours you have added. If we decide to keep them, of course, if we revert to fail some validations of tx not at the queue level...we can keep as it is.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
```diff
@@ -18,441 +18,492 @@
...
+     *
+     * @param tx the transaction to process
+     * @return the result of the admission attempt
+     */
+    private TransactionPoolAddResult addTransactionToPoolIfValid(final Transaction tx) {
+        var repository = getRepositoryAtBestBlock();
+        var sender = tx.getSender(signatureCache);
+        var pendingState = getPendingState(repository);
+
+        Optional<TransactionPoolAddResult> rejection =  validateTransactionBeforeAddingToPool(tx, sender, repository, pendingState);
```
dlealmac: Looks like we are changing the behavior for transactions with gaps in nonce.
The validateTransactionBeforeAddingToPool() method runs quota and cumulative balance checks before hasNonceGap / addQueuedTransaction, while on master branch those checks only happen aftet to queue the tx in internalAddTransaction. So future-nonce txs that used to be queued can now be rejected.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
dlealmac, line 227: With the current change, it emits event for each transaction, and before it was accumulated, i think some consumers may see it duplicated or out of order

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
meldsun0: Yes, I agree. This changes the previous contract, where values lower than 0 effectively disabled the cleaner, and that could break operators that currently rely on -1

I think we should only validate accountTxRateLimitCleanerPeriod when account tx rate limiting is enabled. Otherwise, we're failing startup because of a configuration that will never be used.

Also, if we validate it there, the cleanerPeriod > 0 check in isAccountTxRateLimitCleanerEnabled() becomes redundant.

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
meldsun0: Yes.  Good catch!

rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
meldsun0: I think this concern was valid for the previous version, where quota and aggregate balance checks happened before the nonce-gap decision. In the current flow, transactions with a nonce gap are queued before those checks, matching master behavior, and they are revalidated when promoted to pending using fresh state. Also, this refactor does not introduce any new validations—the same validations from the original implementation are still performed, only reorganized to preserve the original behavior.

rskj-core/src/main/java/co/rsk/net/handler/quota/TxQuotaChecker.java
meldsun0: Yes you are right!  Option 1 will be followed.

rskj-core/src/main/java/org/ethereum/listener/NoOpEthereumListener.java
meldsun0: I agree they are functionally similar, but EthereumListenerAdapter is widely used and changing or renaming it would broaden the scope of this transaction-pool-focused refactor. I introduced NoOpEthereumListener as a dedicated null-object singleton to avoid affecting existing consumers; however, I can reconsider reusing the adapter if avoiding duplication is preferred over keeping the change isolated.

PR #3510, opened 2026-05-08
Added changed according to RSKIP555 considering EIP-1884

wlaclaw:


rskj-core/src/main/java/org/ethereum/vm/VM.java
```diff
@@ -58,6 +58,7 @@
 import static org.ethereum.config.blockchain.upgrades.ConsensusRule.RSKIP446;
 import static org.ethereum.config.blockchain.upgrades.ConsensusRule.RSKIP90;
 import static org.ethereum.config.blockchain.upgrades.ConsensusRule.RSKIP91;
+import static org.ethereum.config.blockchain.upgrades.ConsensusRule.RSKIP555;
```
dlealmac: @wlaclaw please, lets remove the unused import.

This is also a blocker from sonar.

dlealmac (changes requested):

meldsun0:
I like the idea of centralizing consensus gas policy inside the GasCost class and keeping the VM focused on execution. However, this introduces coupling with ActivationConfig, which may cause GasCost to accumulate consensus logic over time. If that grows too much, future RSKIPs could become harder to maintain consistently, so we should keep an eye on it.

meldsun0 (approved):
LGTM

rskj-core/src/main/java/org/ethereum/vm/VM.java
wlaclaw: fixed

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigTest.java
```diff
@@ -156,6 +156,7 @@ class ActivationConfigTest {
         "    rskip544: vetiver900",
         "    rskip551: vetiver900",
         "    rskip552: vetiver900",
+        "    rskip555: 0",
```
fmacleal, line 159: **suggestion**:

We could probably make this active since vetiver900 instead genesis, to simulate better the reality. But we can fix that later.

PR #3523, opened 2026-05-22
Improves baseEvent handling

italo-sampaio:
## Description

`BlockHeaderExtensionV2.fromEncoded()` decoded an empty `baseEvent` field with
`getRLPData()`, which returns `null` for the empty RLP element `0x80` (it should
be `byte[0]`). `BlockHeader.addBaseEvent()` then silently drops a `null`
`baseEvent` on re-encode, so any header that travels the **backward body-sync**
path is persisted one RLP element short. On reload, `BlockFactory.canBeDecoded()`
rejects the truncated header and the node crashes with
`IllegalArgumentException: Invalid block header size: 22`.

This PR fixes the root cause **and** makes the decoder tolerant of headers that
were already persisted in the corrupted (short) form, so an affected node can
upgrade by dropping in the new jar — no resync or DB migration required.

**Root-cause fix**
- `BlockHeaderExtensionV2.fromEncoded()` now reads `baseEvent` with
  `getRLPRawData()`, so an empty `baseEvent` decodes to `byte[0]` instead of
  `null` — consistent with how the `edges` field on the next line and
  `BlockFactory.decodeHeader()` already decode it. This alone stops all *new*
  corruption.

**Backward compatibility (headers already persisted short)**
- `BlockFactory.canBeDecoded()` accepts a header that is exactly one RLP element
  short, gated to the only case it can legitimately occur in (`baseEvent`
  enabled, uncompressed header).
- `BlockFactory.decodeHeader()` disambiguates an absent `baseEvent` from the
  optional merged-mining fields using the remaining element count, and treats an
  absent field as an empty event (`byte[0]`), so a reloaded legacy header is
  equal to a freshly built one.

This changes neither block encoding nor hashing — `addElementsEncoded()` already
encodes an empty and a `null` `baseEvent` identically as `0x80`, so the extension
hash and block hash are unchanged.

## Motivation and Context

Some VETIVER-9.0.x testnet nodes crash during sync with
`IllegalArgumentException: Invalid block header size: 22`. The bug corrupts
header persistence on the backward body-sync path whenever a block carries an
empty `baseEvent` (i.e. no Union Bridge event), which is the common case. Without
this fix, affected nodes cannot complete sync; with it, they recover on a simple
jar swap because the tolerant decoder reads the already-corrupted entries in
place.

## How Has This Been Tested?

The first commit adds tests that pin the buggy behavior with `assertThrows`; the
second (fix) commit flips them to successful-decode assertions, so CI moves from
asserting the bug to asserting the fix.

- `BlockHeaderExtensionV2Test` (22 tests, 0 failures) — four decode tests now
  assert an empty/absent `baseEvent` decodes to `byte[0]` rather than `null`.
- `BlockFactoryTest` (47 tests, 0 failures) — regression tests covering the
  backward body-sync store-and-reload path (with and without merged-mining
  fields) and the tolerant decoder reading a header persisted with the
  `baseEvent` element missing (`remaining == 0` and `remaining == miningFieldCount`
  branches).
- Broader regression sweep over `BlockHeader*`, `Block*` and `co.rsk.core.*`
  test classes — the compressed-path and pre-RSKIP351 tests are unaffected, as
  the fix does not touch those paths.
- `spotlessJavaCheck` / `checkstyle` pass on the changed files.

## Types of changes

## Checklist:

**Requires Activation Code (Hard Fork): No.** The fix changes neither block
encoding nor hashing. `BlockHeaderExtensionV2.addElementsEncoded()` already
encodes an empty and a `null` `baseEvent` identically as the RLP byte `0x80`, so
the extension hash and the block hash are unchanged. `canBeDecoded()` /
`decodeHeader()` only change which headers are *accepted* and how an absent field
is *read* — a correctly sized header decodes exactly as before. This is a
storage/decoding-tolerance fix for an already-active feature, not a consensus
change.

* **Other information**:
Legacy 22-element header entries are read in place by the tolerant decoder and
left as-is on disk; they are never rewritten by `IndexedBlockStore.saveBlock`, so
no DB migration is needed.

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
+++ b/rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
@@ -217,9 +217,9 @@ public final class BlockFactory implements BtcHeaderSizeRule {
             }
 
 
-        if (rlpHeader.size() > r && isBaseEventEnabled(blockNumber) && !compressed) {
-            baseEvent = rlpHeader.get(r++).getRLPRawData();
-        }
+        BaseEventDecodeResult baseEventResult = decodeBaseEvent(rlpHeader, r, blockNumber, compressed);
+        baseEvent = baseEventResult.baseEvent;
+        r = baseEventResult.nextIndex;
 
         byte[] bitcoinMergedMiningHeader = null;
         byte[] bitcoinMergedMiningMerkleProof = null;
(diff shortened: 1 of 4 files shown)


rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
```diff
@@ -334,9 +369,20 @@ private BlockHeader createBlockHeader(boolean compressed, boolean sealed, byte[]
     private boolean canBeDecoded(RLPList rlpHeader, long blockNumber, boolean compressed) {
```
donequis, line 369: Did you consider flagging this method for this change? Something like adding a `boolean trustedSource` so only when reading from the db it accepts non standard size?

PR #3531, opened 2026-05-27
Decode always-present extension fields as non-null byte arrays

italo-sampaio:
## Description

This branch contains two commits aligned around how the V2 block header extension reads its always-present fields, plus a narrow, height-bounded testnet decode tolerance for one specific RLP layout.

1. **`Decode an empty baseEvent as a non-null byte array`** — `BlockHeaderExtensionV2.fromEncoded()` now reads the `baseEvent` slot with `getRLPRawData()` so an empty element decodes to `byte[0]`, matching how the edges field and `BlockFactory.decodeHeader` already read it. The `baseEvent` decoding inside `decodeHeader` follows the same inline cursor pattern used for the other optional header fields and treats an absent field as an empty event.

   `BlockFactory.canBeDecoded` additionally accepts an uncompressed testnet header that is one RLP element short of its expected layout, provided `baseEvent` is active at that block and the height is below a bounded threshold; the decoder reconstructs the absent element as an empty event. Outside this height range the strict size check applies as before.

   `BlockFactory` now takes a chain ID; the existing single-arg constructor delegates with `MAINNET_CHAIN_ID` so the additional tolerance is off by default. `RskContext` passes the real chain ID at the single production instantiation site.

2. **`Decode an empty extension logsBloom as a non-null byte array`** — `BlockHeaderExtensionV1` and `BlockHeaderExtensionV2` now read the `logsBloom` slot with `getRLPRawData()` for the same reason as `baseEvent`. After this commit, every always-present extension field decodes empty as `byte[0]` and only the optional `edges` slot remains nullable when absent from the encoded list.

Encoding and hashing are unchanged across both commits, so no hardfork is required.

## Motivation and Context

The V2 extension and the outer header decoder used `RLPItem.getRLPData()` for the always-present extension fields, which returns `null` for an empty element rather than `byte[0]`. That asymmetry — empty round-trips as `null` but the encoder accepts both — can cause an always-present field to disappear from the re-encoded layout when an extension makes a wire round-trip, and the strict size check then rejects the re-encoded payload on disk reload. Switching the always-present fields to `getRLPRawData()` removes the asymmetry; the encoder side was already `null`-safe.

The bounded testnet decode tolerance recovers nodes that already persisted such one-element-short layouts. The combination of testnet-only chain ID, RSKIP-535 + RSKIP-351 active, height below an arbitrary upper bound, and uncompressed restricts the recovery surface to the specific layout the tolerance targets. Misshapen one-element-short payloads whose absent element is a merged-mining field rather than `baseEvent` are caught by the existing BTC merged-mining header size validation immediately downstream of the recovery branch.

## How Has This Been Tested?

**Unit / module tests**
- `./gradlew :rskj-core:test --tests "co.rsk.core.BlockFactoryTest" --tests "org.ethereum.core.BlockHeaderExtensionV2Test" --tests "co.rsk.core.BlockHeaderExtensionV1Test" --tests "co.rsk.core.BlockHeaderV1Test" --tests "co.rsk.core.BlockHeaderV2Test"` — all green.
- `./gradlew :rskj-core:test` — full module suite, BUILD SUCCESSFUL.
- `./gradlew spotlessJavaCheck -PratchetFrom=origin/master` — green.

New tests added in `BlockFactoryTest`:
- `emptyBaseEventPreservedAcrossExtensionRoundTrip{WithMining,NoMining}` — encoder/decoder symmetry for an empty `baseEvent`.
- `headerOneElementShortDecodableOnTestnetWithinBound{WithMining,NoMining}` — decode of a one-element-short header on a testnet factory within the bounded height range.
- `headerOneElementShortRejectedOnMainnetFactory`, `headerOneElementShortRejectedAboveUpperBound`, `headerOneElementShortRejectedOnNonTestnetChainId`, `compressedHeaderRejectedWhenSizeDoesNotMatch` — strict-rejection paths for every combination outside the tolerance.
- `headerOneElementShortWithBaseEventPresentRejectedOnTestnet` — a one-element-short layout whose absent element is a merged-mining field rather than `baseEvent` is rejected at decode time by the BTC merged-mining header size validation immediately downstream of the recovery branch.

`BlockHeaderExtensionV2Test.encodingWithNullFields` was updated to assert the new symmetric shape; the other extension tests use 256-byte logsBloom values and are unaffected by the change.

## Types of changes

## Checklist:

* **Other information**: The testnet decode tolerance in `BlockFactory.canBeDecoded` is intentionally height-bounded by an arbitrary constant (`BASE_EVENT_RECOVERY_UPPER_BOUND = 7_800_000L`). A follow-up will replace the bounded tolerance with a non-sunset solution once the deployment window is past.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/RskContext.java
+++ b/rskj-core/src/main/java/co/rsk/RskContext.java
@@ -377,7 +377,9 @@ public class RskContext implements NodeContext, NodeBootstrapper {
         checkIfNotClosed();
 
         if (blockFactory == null) {
-            blockFactory = new BlockFactory(getRskSystemProperties().getActivationConfig());
+            blockFactory = new BlockFactory(
+                    getRskSystemProperties().getActivationConfig(),
+                    getRskSystemProperties().getNetworkConstants().getChainId());
         }
 
         return blockFactory;
--- a/rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
+++ b/rskj-core/src/main/java/org/ethereum/core/BlockFactory.java
@@ -25,6 +25,7 @@ import co.rsk.core.RskAddress;
 import co.rsk.remasc.RemascTransaction;
 import co.rsk.validators.BtcHeaderSizeRule;
 import org.bouncycastle.util.BigIntegers;
+import org.ethereum.config.Constants;
 import org.ethereum.config.blockchain.upgrades.ActivationConfig;
 import org.ethereum.config.blockchain.upgrades.ConsensusRule;
 import org.ethereum.util.ByteUtil;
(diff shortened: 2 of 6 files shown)


fmacleal (approved):
LGTM,

Good job on making it simple.

PR #3539, opened 2026-06-01
Backward compatible pre RSKIP559 hardcoded pegouts

wlaclaw:
**Context**
- RSKIP559 https://github.com/rsksmart/rskIPs/pull/559
- this is an extension of https://github.com/rsksmart/rskj/pull/3507

**Scope of changes**

Here we are addressing theoretical (not yet proven) assumption that https://github.com/rsksmart/rskj/pull/3484 may somehow somewhere fail while running on Java17/21. 

To mitigate this, we will hardcode the conflicting peg-out outputs in the node now. We are optimistically assuming that, since such events are extremely rare, there is a good chance NO new ones will occur before the hard fork.

Originally, this work was planned for after the hard fork, based on the assumption that all conflicting historical peg-outs could only be reliably acquired once the hard fork was complete.

**Core notion**

This is only about nodes that will re-syncronize historical data. The probable issue could never happen for live nodes or restored from snapshots after the latest conflicting output.

**Code meant to be refactored**

After the hardfork old sorting will not be required. Only new deterministic ordering will stay with hardcoded diffs for historical corner cases. Thus all logic branches, activation conditions, wrapper classes etc. will be removed.

**Configuration based approach**

I think it is logical to put conflicting historical pegouts in `blockchain.config` section.
We already have there `hardforkActivationHeights` and `consensusRules` which are defining hardfork behaviour.
So adding another property `rskip559diff` which defines hardfork behaviour and contains seven entries (for the mainnet) is ok.

The main advantage of this approach is the ability to update the config without changing the node version. This addresses the pessimistic scenario - i.e., "if everything fails" - where new conflicting outputs appear after the node release but before the hard fork, nodes fail to sync historical data, and so on. In that case, we would simply update the node configs instead of releasing a new version

**Testnet configuration**
Has some edge cases described in RSKIP559. 
I do not think it is reasonable to cover them considering that the testnet 3 is near EOL.


gradle/verification-metadata.xml
fmacleal, line 1: **question**:

Why all these changes in the verification-metadata? Also, you need to double check and fix this, it's failing on the RIT (Rootstock Integration Tests) execution.

rskj-core/src/main/resources/config/testnet.conf
```diff
@@ -26,6 +26,46 @@ blockchain.config {
         rskip284 = 2581800,
         rskip290 = 2581800
         rskip535 = 7604200 # To be activated in testnet only, configuring same height as Vetiver activation
+    },
+    rskip559diff = {
+        1120318 = "008e6f56ceb6aed4f2d1e5783d6fde5a50178707f9853021a74547b4a7463a16",
+        1865157 = "9bbdf1215b0afa73d06bf91b03f2081fb388480e1ade47dce4df6ac56499c51e",
+        # 2589068 = "232159dd012a9ff195e7ff018aaab69dbd6f89831610a8c01af926e11fef3d8a",
+        # 2589068 = "bc325057ff3247614896efb64e732c82326952dfb0c14a4cdf429589f41badc0",
```
fmacleal: **issue**:
What are these commented blocks? Do they have conflict or not?
If not, remove this line.

rskj-core/src/main/resources/config/testnet.conf
```diff
@@ -26,6 +26,46 @@ blockchain.config {
...
+        2589071 = "7038f7e065350e6d22e1803dd814b1ba448c18d9647ad7578db9ca36fb027a5d",
+        2589077 = "7212c5d54b3da2b03043b8b3fe57e3b856292a7e01447f8aea4e90ce2c9c0731",
+        2589086 = "3fc16457c2b1cb5288bc5d83fafd603b09fd1fa3dc0d4231e160a5650bb2c693",
+        # 2589112 = "df13869f4f44e010c4a00225e1c37a0bda5cd70d26963670c08ee4bcdc1184ea",
+        # 2589112 = "431fa28be23bc66bb36918ae00def76c046b2d6fb2ff733c2e5f247b627a3449",
```
fmacleal: **issue**:
What are these commented blocks? Do they have conflict or not?
If not, remove this line.

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/PegoutsOverwrites.java
```diff
@@ -0,0 +1,94 @@
...
+                    ActivationConfig.PEGOUTS_OVERWRITES_RULES,
+                    ex
+                ));
+            }
+
+            this.pegouts.put(key, value);
+        }
+    }
+
+    @NonNull
```
fmacleal: **thought**:

Do we really need a whole new library just for this NonNull verification? Why not use the widely used in the codebase `javax.annotation.Nonnull`? Does the jspecify brings any higher value?

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -200,6 +267,23 @@ public Keccak256 getPegoutCreationRskTxHash() {
             return pegoutCreationRskTxHash;
         }
 
+        /**
+         * Returns Sha256 representation for this Entry.
+         */
```
fmacleal: **suggestion**: 

Although it's clear that you are getting the hash bytes plus the block number, it would be good to make clear in the comment adding something like:

```
 * The sha256 is calculated from `btcTxHash || longToBytes(blockNumber)`
```

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/PegoutsOverwrites.java
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * Such approach would allow us to have only one deterministic logic branch
+ * with some corner cases overwrites for historical data.
+ */
+public class PegoutsOverwrites {
+
+    /**
+     * Block number to next pegout waiting for confirmation.
+     */
+    private HashMap<Long, Sha256Hash> pegouts = new HashMap<>();
```
fmacleal: **nitpick**:

make it final to ensure it won't be changed

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/PegoutsOverwrites.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @VisibleForTesting
+    public PegoutsOverwrites() {
+    }
+
+    @VisibleForTesting
+    public PegoutsOverwrites(HashMap<Long, Sha256Hash> overwrites) {
+        this.pegouts = overwrites;
+    }
```
fmacleal: **suggestion**:

If we change to final, we need to make a copy here `Collections.unmodifiableMap(...)`

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/PegoutsOverwrites.java
```diff
@@ -0,0 +1,94 @@
...
+                ));
+            }
+
+            if (this.pegouts.containsKey(key)) {
+                throw new IllegalArgumentException(String.format(
+                    "'blockchain.config.%s' all keys must be unique. Duplicate found for %s.",
+                    ActivationConfig.PEGOUTS_OVERWRITES_RULES,
+                    key
+                ));
+            }
```
fmacleal: **thought**:

Isn't this a dead-code unreachable? I mean, since it's a hash, it's probably impossible to have duplicates no? It would fail during load time, probably.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -200,6 +267,23 @@ public Keccak256 getPegoutCreationRskTxHash() {
...
+        Sha256Hash getSha256() {
+            var digest = Sha256Hash.newDigest();
+            digest.update(this.getBtcTransaction().getHash().getBytes());
+
+            var blockNumber = getPegoutCreationRskBlockNumber();
+            if (blockNumber != null) {
+                            digest.update(ByteUtil.longToBytes(blockNumber));
```
fmacleal: **chore**:

Fix the indentation here.

rskj-core/src/test/java/co/rsk/peg/PegoutsWaitingForConfirmationsTest.java
```diff
@@ -228,6 +242,29 @@ void getNextPegoutWithEnoughConfirmations_rskip559() {
         );
     }
 
+    @Test
```
fmacleal, line 236: **suggestion**:

This tests covers well the happy path, but how about other edge cases scenarios? Consider adding:
- override hash present but **entry already removed** → falls through to normal pegout (the `findRskip559diff` empty branch);
- override hash present but **no entry matches** → falls through;
- a `PegoutsOverwrites(Config)` parsing test for the malformed-key / non-hex / bad-Sha256 branches (currently no test exercises the validation throws);

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ActivationConfig.java
fmacleal, line 1: **question**:

Why add this in a config, that it's changing a contract for a broad and wide used class in lots of places. Apart of that, it's less resilient, a config can be overwritten. It could cause a fork if someone by mistake or because forgot from this small detail, overwrite this piece of configuration.

I think it's safer and with less changes if we add this to some static class, or even an existing one for constants, like [BridgeConstants](https://github.com/rsksmart/rskj/blob/a49e710dbea730008ca800a773605133a3cf5440/rskj-core/src/main/java/co/rsk/peg/constants/BridgeConstants.java#L32). Let's see what people from powpeg thinks about this in the review.

fmacleal (changes requested):
Nice job! 

I have some small comments though that deserves some attention and changes.

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ActivationConfig.java
italo-sampaio: Strongly agree w/ Fred on this. This is a tricky problem, and it's hard to come up with a bulletproof solution, but my main concern is having this list of hard-coded hashes in the config file.

This is an open door for human errors, and it's just impossible to determine whether or not a user/miner has accidentally changed one digit in one of these hashes.

I think anything that is truly hard-coded would be more robust

rskj-core/src/main/resources/config/testnet.conf
wlaclaw: This are corner cases described in RSKIP559

rskj-core/src/main/resources/config/testnet.conf
wlaclaw: The same corner cases

gradle/verification-metadata.xml
wlaclaw: I just added `org.jspecify` dependency. 
This output is what metadata update script produced. 
I guess it removed transient dependencies for the packages that we are not using anymore.

marcos-iov: Not sure this is a good idea, we usually try to stay away from new dependencies since that opens the door for supply chain attacks. 

If there are some unused dependency hashes in this file we should handle that in a separate PR ideally

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ActivationConfig.java
marcos-iov: Agree here, these values shouldn't be part of a config in my opinion. It's not a custom config, it's hardcoded values. It's also adding a responsibility to ActivationConfig class that probably shouldn't have.

BridgeConstants could be a good place to put it I think. Ideally on a separate file but referenced from BridgeConstants

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/PegoutsOverwrites.java
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * RSKIP559 changed behaviour of {@link co.rsk.peg.PegoutsWaitingForConfirmations}.
+ * It is not fully compatible while syncing historical data.
+ * To mitigate this we override some RSKIP559 algorithm outputs
+ * with values provided in the current class.
+ *
+ * Such approach would allow us to have only one deterministic logic branch
+ * with some corner cases overwrites for historical data.
+ */
+public class PegoutsOverwrites {
```
marcos-iov: If we decide to keep the hashes in BridgeConstants guess we won't event need this class

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -87,8 +95,59 @@ public Collection<Entry> getEntries(ForBlock activations) {
      * @return an optional with an entry with enough confirmations if found. If not, an empty optional.
      */
     public Optional<Entry> getNextPegoutWithEnoughConfirmations(Long currentBlockNumber, Integer minimumConfirmations, ForBlock activations) {
```
marcos-iov: I imagine this could be solved very straightforward and simple

```
List<Entry> pegoutsConfirmed = getPegoutsConfirmedOrdered()
if (rskip559IsActive || pegoutsConfirmed.size() == 1) {
  return pegoutsConfirmed.get(0);
} else {
  return bridgeConstants.getPegoutConfirmedForBlock(blockNumber);
}
```
You only need to hardcode cases where there was more than 1 pegout with enough confirmations in the same block, pre RSKIP activation

rskj-core/src/main/resources/config/testnet.conf
fmacleal: And what is the impact of this corner case? What does it mean exactly, that for these corner cases the fix won't work? Will work?

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ActivationConfig.java
wlaclaw: @italo-sampaio about "This is an open door for human errors"
In the `blockchain.config` block we already have `hardforkActivationHeights` and `consensusRules` which also cold be considered as an `open door for` even more critical `human errors` because it affects future node behaviour while my config part is only about historical syncing.

BTW I've added more verbose description to this PR which could address some of your concerns.

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/PegoutsOverwrites.java
wlaclaw: Because `javax.annotation` is deprecated and other products are already removing it

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/PegoutsOverwrites.java
wlaclaw: It could happen during config parsing because from java doc it seems that `config.entrySet()` will execute for any entry in the config.
And it will fail just as a precaution

rskj-core/src/test/java/co/rsk/peg/PegoutsWaitingForConfirmationsTest.java
wlaclaw: Will add more cases.
BUT we can't fail if hardcoded record exists for block but no entry available.
Mainnet node would not sync then on block 5002812.

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/PegoutsOverwrites.java
fmacleal: In this case, I strongly advise to not mix stuff, we could create a PR just to deprecate `javax.annotation` and migrate to `jspecify`. Otherwise we would never remove `javax.annotation` and stay with two libraries.

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ActivationConfig.java
italo-sampaio: @wlaclaw that's a fair point, but let's agree that a list of activation heights is more human-readable than a list of block hashes. Also, activation heights are configurable to attend to a practical need: we use multiple distinct network configs, including temporary federation changes used only for tests, and we need to be able to support all of them without generating a new jar for every network change.

In this case, the list is supposed to be immutable; having this list configurable doesn't really buy us a lot. If we have the opportunity of making the code a little bit safer, despite existing design decisions made years ago, that's preferable to keeping things as they are because it makes logical sense IMO.

wlaclaw: I'm working on different config format. So there wil be no custom hash. Just need to finish collecting proper data from the node.
```
CURRENT_BLK_NUMBER => "BTC_TX_HASH@RSK_BLK_ID"

3345557 = "5965a75e7e56ed4a308cc1bf8d94415c03c6a56f7302ae488e1d3fa05cd70e61@3341556"
```

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
wlaclaw: Well I'm not sure about this part.
if `pegoutsConfirmed.size() == 1` than output will be the same regardless of sorting etc so no need to do wrap it in the if block.

gradle/verification-metadata.xml
wlaclaw: Well this is not an NPM ecosystem. See https://jspecify.dev/about/

rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/PegoutsOverwrites.java
wlaclaw: We will not forget because I created a task `RSKCORE-5487`.

gradle/verification-metadata.xml
wlaclaw: >it's failing on the RIT

@fmacleal It is not failing now. I guess the issue was not related to this particular change.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -80,28 +90,111 @@ public Collection<Entry> getEntries(ForBlock activations) {
...
+        }
+
+        return pegout;
+    }
+
+    @Nullable
+    private Entry getOutputOverwrite(Long blk, ForBlock activations) {
+        if (activations.isActive(ConsensusRule.RSKIP559)) {
+            // The code below will work the same without this early return
+            // But it was requested to put this condition "just in case"
```
fmacleal: **chore**:
These comments here are unnecessary. Please, remove it.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -207,15 +313,134 @@ public boolean equals(Object o) {
...
+        if (networkName == null) {
+            return null;
+        }
+
+        return switch (networkName) {
+            case "main" -> MAINNET;
+            case "testnet" -> TESTNET;
+            default -> null;
+        };
+    }
```
fmacleal: **chore**:

We already have an [enum](https://github.com/rsksmart/rskj/blob/cfe21b67678d298d601bdfcc3c7afe18c971c31b/rskj-core/src/main/java/org/ethereum/config/NetworkName.java#L7) for this, let's use it, so we don't have different ways to do the same thing.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -207,15 +313,134 @@ public boolean equals(Object o) {
...
+                && otherEntry.getPegoutCreationRskBlockNumber().equals(getPegoutCreationRskBlockNumber())
+                && Objects.equals(getPegoutCreationRskTxHash(), otherEntry.getPegoutCreationRskTxHash());
         }
 
         @Override
         public int hashCode() {
             return Objects.hash(getBtcTransaction(), getPegoutCreationRskBlockNumber());
         }
     }
+
```
fmacleal: **suggestion**:

Move all these records and objects to a different class, could be in BridgeConstant or in a different file just for this to leave this class a bit cleaner.

rskj-core/src/main/java/co/rsk/peg/PegoutsWaitingForConfirmations.java
```diff
@@ -80,28 +90,111 @@ public Collection<Entry> getEntries(ForBlock activations) {
...
+        return null;
+    }
+
+    /**
+     * Search entries by pegout reference.
+     */
+    Optional<Entry> getPegoutByRef(PegoutRef pegoutRef) {
+        return this.entries.entriesSet.stream().filter(e -> {
+            return pegoutRef.btcTxHash().equals(e.getBtcTransaction().getHash())
```
fmacleal: **nitpick (optional)**:
Sonar is right here, the curly braces are not necessary and neither is the return.

fmacleal (changes requested):
Thank you for having addressed the comments. 🙏 

There are just a few and small changes needed @wlaclaw.

fmacleal (dismissed):
Thank you for having addressed the issues.

The RIT is failing due to some problem with jspecify, we would have to fix this in order to move forward.

If you don't manage to fix, you can use the non-null from javax and we address the upgrade to jspecify later.

italo-sampaio (dismissed):

marcos-iov:
Converting this to draft as we will probably follow another approach

PR #3546, opened 2026-06-05
Fix incorrect calldata for sub-calls in trace_transaction and trace_block

italo-sampaio:
## Description

`trace_transaction`, `trace_block`, `trace_get` and `trace_filter` reported the **top-level transaction's calldata** as the `input` of **every** sub-call frame, instead of each frame's own calldata.

`TraceTransformer.toAction` sourced the `CALL` action `input` from `txInfo.getReceipt().getTransaction().getData()` (the root transaction data) rather than from the per-frame execution context. This made it impossible to tell which function was invoked at any depth below the root: every nested `CALL`/`DELEGATECALL`/`STATICCALL` reported the root selector.

This change sources the `input` from the per-frame `InvokeData`, identical to what the `debug_traceTransaction` callTracer (`CallTraceTransformer`) already does:

```java
input = HexUtils.toUnformattedJsonHex(invoke.getDataCopy(DataWord.ZERO, invoke.getDataSize()));
```

The now-unused `TransactionInfo` parameter is dropped from `toAction` (and the dead mock setup it required in `TraceTransformerTest`).

## Motivation and Context

Tools that decode sub-call calldata from traces received incorrect results. Concretely, this broke Safe transaction-service indexing: the service identifies a transaction involving a registered Safe singleton via `trace_filter`, but when decoding the `setup()` delegatecall it saw `createProxyWithNonce` calldata instead, failed to recognise the deployment, and never registered the Safe address.

The regression was introduced when a memory-read optimization switched the parity tracer to read the top-level transaction data (sidestepping a stale zero-copy memory slice). The per-frame calldata is now eagerly copied at `ProgramInvoke` construction, so reading it here is safe again — this restores the per-frame source. The transaction itself always executed correctly; only the trace reporting was wrong.

## How Has This Been Tested?

- Added a regression test `TraceModuleImplTest.subCallReportsItsOwnCalldataNotTopLevelInput` driven by a new DSL fixture (`dsl/trace_subcall_calldata.txt`): a contract invoked with calldata `0x1688f0b9` makes an internal `CALL` forwarding a distinct `0xb63e800d`; the test asserts the root frame reports `0x1688f0b9` and the sub-call frame reports `0xb63e800d`. It fails before the fix (`expected: <0xb63e800d> but was: <0x1688f0b9>`) and passes after.
- Full `co.rsk.rpc.modules.trace.*` and `co.rsk.rpc.modules.debug.*` suites pass — no regressions (root-frame and value-transfer output is byte-identical to before).
- `checkstyleMain`, `checkstyleTest`, and `spotlessJavaCheck` all clean.
- Manual QA: against a node with the fix, `trace_transaction` / `trace_block` for a transaction with nested internal calls now report each sub-call's own `input` rather than the root transaction's.

## Types of changes

## Checklist:

* **Other information**: The fix aligns the parity tracer (`trace_*`) with the already-correct Geth callTracer (`debug_traceTransaction`); both now read calldata from the same per-frame source. `CREATE`/`CREATE2` frames are unaffected (they report `init`, not `input`).

### DIFF
--- a/rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceModuleImplTest.java
+++ b/rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceModuleImplTest.java
@@ -222,15 +222,65 @@ class TraceModuleImplTest {
         ArrayNode aresult = (ArrayNode) result;
         Assertions.assertEquals(2, aresult.size());
 
-        // Top-level call carries the transaction's own input.
+        // Top-level call carries the transaction's own input. Its traceAddress is empty (it is the root frame).
         JsonNode topLevel = aresult.get(0);
         Assertions.assertEquals("call", topLevel.get("type").asText());
+        Assertions.assertEquals(0, topLevel.get("traceAddress").size());
         Assertions.assertEquals(TOP_LEVEL_CALLDATA, topLevel.get("action").get("input").asText());
 
-        // The sub-call must carry its OWN calldata, not the top-level input.
+        // The sub-call is the nested frame (traceAddress [0]) and must carry its OWN calldata.
         JsonNode subCall = aresult.get(1);
         Assertions.assertEquals("call", subCall.get("type").asText());
+        Assertions.assertEquals(1, subCall.get("traceAddress").size());
         Assertions.assertEquals(SUB_CALL_CALLDATA, subCall.get("action").get("input").asText());
+
+        // The bug being guarded against: the sub-call must not reuse the root calldata.
+        Assertions.assertNotEquals(
+                topLevel.get("action").get("input").asText(),
+                subCall.get("action").get("input").asText());
+    }
+
+    /**
+     * trace_block shares the per-frame calldata logic with trace_transaction, so the same invariant must
+     * hold when tracing a whole block: the top-level frame reports the transaction input and the nested
+     * frame reports its own sub-call calldata.
+     */
+    @Test
+    void subCallReportsItsOwnCalldataNotTopLevelInputInTraceBlock() throws Exception {
+        ReceiptStore receiptStore = new ReceiptStoreImpl(new HashMapDB());
+        DslParser parser = DslParser.fromResource("dsl/trace_subcall_calldata.txt");
+        World world = new World(receiptStore);
+        WorldDslProcessor processor = new WorldDslProcessor(world);
+        processor.processCommands(parser);
+
+        Block block = world.getBlockByName("b02");
+
+        TraceModuleImpl traceModule = new TraceModuleImpl(world.getBlockChain(), world.getBlockStore(), receiptStore, world.getBlockExecutor(), null, world.getBlockTxSignatureCache(), world.getConfig());
+
+        JsonNode result = traceModule.traceBlock(block.getHash().toJsonString());
+
+        Assertions.assertNotNull(result);
+        Assertions.assertTrue(result.isArray());
+
+        ArrayNode aresult = (ArrayNode) result;
+        Assertions.assertEquals(2, aresult.size());
+
+        // Top-level call carries the transaction's own input. Its traceAddress is empty (it is the root frame).
+        JsonNode topLevel = aresult.get(0);
+        Assertions.assertEquals("call", topLevel.get("type").asText());
+        Assertions.assertEquals(0, topLevel.get("traceAddress").size());
+        Assertions.assertEquals(TOP_LEVEL_CALLDATA, topLevel.get("action").get("input").asText());
+
+        // The sub-call is the nested frame (traceAddress [0]) and must carry its OWN calldata.
+        JsonNode subCall = aresult.get(1);
+        Assertions.assertEquals("call", subCall.get("type").asText());
+        Assertions.assertEquals(1, subCall.get("traceAddress").size());
+        Assertions.assertEquals(SUB_CALL_CALLDATA, subCall.get("action").get("input").asText());
+
+        // The bug being guarded against: the sub-call must not reuse the root calldata.
+        Assertions.assertNotEquals(
+                topLevel.get("action").get("input").asText(),
+                subCall.get("action").get("input").asText());
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceModuleImplTest.java
```diff
@@ -187,6 +187,52 @@ void retrieveMultiContractTraces() throws Exception {
...
+     * Each call frame in a trace must report its own calldata in the action {@code input} field.
+     * A sub-call's input is the data passed to that frame, which is generally different from the
+     * top-level transaction input.
+     *
+     * Here the caller contract is invoked with {@value #TOP_LEVEL_CALLDATA} and makes an internal CALL
+     * passing a distinct {@value #SUB_CALL_CALLDATA} as the sub-call calldata. The top-level frame must
+     * report its own input and the sub-call frame must report the data it received, not the top-level input.
+     */
+    @Test
+    void subCallReportsItsOwnCalldataNotTopLevelInput() throws Exception {
```
fmacleal, line 206: **suggestion**:

This is a very good test addition, but isn't validating the delegate_call sub-calls. This test only validates the plain call. But it would be a very nice addition to make a test to validate the subCall type delegatecall, which was exactly the issue reported.

It would probably require another DSL test, or a new method that performs the delegatecall operation.

rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceModuleImplTest.java
fmacleal, line 1: **suggestion**:

I noticed that you are covering the path of `trace_transaction`, but it would be nice as well cover the `trace_block` path. Adding a test that verifies the `traceModule.traceBlock(...)` for the same block and assert the same calldata invariants on the matching transaction traces.

fmacleal (commented):
Great job, small change the addresses properly the issue.

But I have a few comments of tests that could be expanded. I have added them below.

rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceTransformerTest.java
```diff
@@ -57,16 +48,13 @@ void getActionFromInvokeData() {
...
+        TraceAction action = TraceTransformer.toAction(TraceType.CALL, invoke, CallType.CALL, null, null, null);
 
         Assertions.assertNotNull(action);
         Assertions.assertEquals("call", action.getCallType());
         Assertions.assertEquals("0x0000000000000000000000000000000000000001", action.getTo());
         Assertions.assertEquals("0x0000000000000000000000000000000000000003", action.getFrom());
+        // input is sourced from the per-frame invoke data, not the top-level transaction
         Assertions.assertEquals("0x01020304", action.getInput());
         Assertions.assertEquals("0xf4240", action.getGas());
         Assertions.assertEquals("0x186a0", action.getValue());
```
meldsun0, line 60: You can add assertions that CALL uses input and not init:
```
Assertions.assertNull(action.getInit());
Assertions.assertNull(action.getCreationMethod());
```

rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceTransformerTest.java
meldsun0, line 99: Assertions.assertNull(action.getInput());

meldsun0, line 135: Creation traces should not expose calldata through input:
Assertions.assertNull(action.getInput());

rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceModuleImplTest.java
```diff
@@ -187,6 +187,52 @@ void retrieveMultiContractTraces() throws Exception {
...
+
+        // Top-level call carries the transaction's own input.
+        JsonNode topLevel = aresult.get(0);
+        Assertions.assertEquals("call", topLevel.get("type").asText());
+        Assertions.assertEquals(TOP_LEVEL_CALLDATA, topLevel.get("action").get("input").asText());
+
+        // The sub-call must carry its OWN calldata, not the top-level input.
+        JsonNode subCall = aresult.get(1);
+        Assertions.assertEquals("call", subCall.get("type").asText());
+        Assertions.assertEquals(SUB_CALL_CALLDATA, subCall.get("action").get("input").asText());
```
meldsun0, line 235: Consider adding:

``` 
Assertions.assertNotEquals(topLevel.get("action").get("input").asText(), subCall.get("action").get("input").asText());

```

to make the bug being guarded against explicit ("sub-call must not reuse root calldata"). Optionally, assert the traceAddress values to verify the second trace is actually the nested frame:

```
Assertions.assertEquals(0, topLevel.get("traceAddress").size());
Assertions.assertEquals(1,subCall.get("traceAddress").size());

```

rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceModuleImplTest.java
italo-sampaio: Done!

rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceModuleImplTest.java
italo-sampaio: Done!

rskj-core/src/test/java/co/rsk/rpc/modules/trace/TraceTransformerTest.java
italo-sampaio: Done!

meldsun0 (dismissed):

fmacleal (dismissed):
Thanks for having addressed the issues, looks good now!

fmacleal (approved):
LGTM

PR #3558, opened 2026-06-12
Added RSKIP-545 and RSKIP-546

meldsun0:
## Description
This pull request implements:
- [RSKIP-546 ](https://ips.rootstock.io/IPs/RSKIP546.html)([EIP-2718](https://eips.ethereum.org/EIPS/eip-2718)): It includes transaction parsing/serialization, activation-gated validation and execution, typed receipt encoding, and RPC updates so typed transactions are correctly exposed through the API.
- [RSKIP-545](https://ips.rootstock.io/IPs/RSKIP545.html)([EIP-7702](https://eips.ethereum.org/EIPS/eip-7702)): It includes type4 transaction parsing/serialization, execution logic and delegation resolution on internal EVM calls: CALL / CALLCODE / STATICCALL / DELEGATECALL. It also changes the TransactionPool by restricted only one transaction from a delegated EOA

## Motivation and Context
- RSKIP-546: is required to support EIP-2930 and EIP-1559-style transaction formats. This change enables clients to submit, process, and query typed transactions while preserving legacy transaction behavior and keeping pre-fork data compatible across the activation boundary.
- RSKP-545: Introduces type 4 transactions that allow EOAs to delegate execution to smart contract code while preserving their identity. This enables account-abstraction features such as batching, gas sponsorship, and delegated permissions without requiring smart contract wallets.

## How Has This Been Tested?
- RSKIP545: Coverage was added and updated with Rskip546DslTest, Rskip546HardforkTest, and related typed transaction tests such as TypedTransactionIntegrationTest, TypedTransactionReceiptTest, TypedTransactionSecurityTest, TransactionTypeTest, and CallArgumentsParamTest. These tests cover Type 1 and Type 2 encoding, receipt format, access list handling, RPC serialization, pre-activation rejection, post-activation acceptance, and mixed legacy/typed block behavior.
- RSKP-545: Coverage was added with TransactionPoolImplTest, Type4TransactionExecutorDelegationExecutionTests, Type4TransactionExecutorTests, SetCodeSetCodeAuthorizationTransactionExecutorTest, ProgramTest

## Types of changes

## Checklist:

* **Other information**:
Typed transactions are format-only. RSKj does not implement the EIP-1559 dynamic fee market or access list EVM optimization. baseFeePerGas is always 0x0. maxPriorityFeePerGas and maxFeePerGas are accepted but the effective gas price is min(maxPriorityFeePerGas, maxFeePerGas) per RSKIP-546.


rskj-core/src/main/java/org/ethereum/core/SetCodeAuthorizationTransactionExecutor.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private void verifyAuthorityNonce(byte[] expectedNonce, byte[] currentNonce) {
+        if (!Arrays.equals(currentNonce, expectedNonce)) {
+            throw new IllegalStateException("Authority nonce mismatch");
+        }
+    }
+
+    private long calculateRefund(byte[] code) {
+        boolean isEmpty = code == null || code.length == 0;
+        return isEmpty ? 0L : GasCost.PER_EMPTY_ACCOUNT_COST - GasCost.PER_AUTH_BASE_COST;
```
italo-sampaio, line 114: **The authorization refund is conditioned on the authority's code, not on the authority account**

RSKIP-545 Behavior step 6 adds the refund *"if `authority` is not empty"*, but this decides on code emptiness. `verifyAuthorityCode` runs just above and only admits an authority whose code is either empty or already a delegation indicator — so non-empty code means *already delegated*, and the 9500 refund fires on re-delegation only.

The consequence is that a pre-existing, funded EOA (non-zero nonce, non-zero balance, no code) delegating for the first time receives `0`, where step 6 makes the 9500 owed. That is the primary flow for this feature, so the upfront `PER_EMPTY_ACCOUNT_COST` of 25000 is overcharged by 9500 (38%) on essentially every first delegation. It is deterministic and identical across nodes, so not a divergence risk — but it is baked into gas accounting at activation.

The RSKIP's rationale points the same way: *"`PER_AUTH_BASE_COST` is the cost for each EOA whose code is actually modified successfully... The difference between upfront and ultimate costs (per EOA) is refunded to the sender."* That reading would refund every successfully applied tuple, which also does not match the current condition.

Worth noting that `processAuthorizationTuple` never reads the authority's balance or queries account existence — only code and nonce — so the account-emptiness condition cannot be evaluated as currently structured; whichever way the semantics land, this likely needs an additional repository read.

One caveat on any fix: `processAuthorizationTuple_shouldAllowNonceOneLessThanLimit` and `processAuthorizationTuple_shouldAllowUniversalChainIdWithAnyOuterChainId` both assert `assertEquals(0L, refund)` on a tuple that is applied successfully, so the suite would stay green across the change unless those are updated too.

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
```diff
@@ -0,0 +1,280 @@
...
+        ECDSASignature signature = auth.getSignature();
+        BigInteger r = signature.getR();
+        BigInteger s = signature.getS();
+        if (!signature.validateComponentsWithoutV()) {
+            throw new IllegalArgumentException("Authorization signature components are invalid");
+        }
+        if (r.compareTo(MAX_SIGNATURE_COMPONENT) >= 0 || s.compareTo(MAX_SIGNATURE_COMPONENT) >= 0) {
+            throw new IllegalArgumentException("Authorization signature r and s must be less than 2^256");
+        }
+        if (s.compareTo(SECP256K1N_HALF) >= 0) {
```
italo-sampaio, line 276: question: **Should a single malformed tuple reject the whole transaction?**

RSKIP-545 specifies per-tuple recovery: *"If any step above fails, immediately stop processing the tuple and continue to the next tuple in the list."*

The validation failures raised here (high `s`, `y_parity` outside `{0, 1}`, nonce `2**64 - 1`) reject the entire transaction at parse time, so a transaction carrying one malformed tuple alongside valid ones cannot be included at all.

Separately, `2**64 - 1` is rejected here while step 2 only requires the nonce be *less than* `2**64 - 1` — so the boundary itself differs between the parse and execution layers.

Was that the intended behavior?

italo-sampaio (commented):
Looks good overall! I've left two comments inline, let me know if they make sense (there's a chance it's the RSKIP that needs fixing, and not the code)

PR #3595, opened 2026-06-24
Fix/pegout value when splitting request

julia-zack:


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -1529,30 +1537,38 @@ private void processPegoutsInBatch(
...
                     result.responseCode());
                 return;
             }
 
+            Coin batchedPegoutRequestsValue = pegoutEntries
+                .stream()
+                .map(ReleaseRequestQueue.Entry::getAmount)
+                .reduce(Coin.ZERO, Coin::add);
+            boolean shouldUseTotalPegoutRequestsValue = !activations.isActive(ConsensusRule.RSKIP378);
+            Coin pegoutValueToUse = shouldUseTotalPegoutRequestsValue ? totalPegoutRequestsValue : batchedPegoutRequestsValue;
```
julia-zack: this is the fix

jeremy-then (approved):
LGTM.

PR #3611, opened 2026-07-02
Fix invalid port handling in IpUtils

alaningtrump:
## Description

- Make `IpUtils.parseAddress` return `null` for invalid numeric ports instead of propagating an `IllegalArgumentException`.
- Add regression tests for ports above `65535` and values exceeding the integer range.

## Motivation and Context

`parseAddress` already uses `null` to represent invalid address formats, and `parseAddresses` skips those entries. However, numeric port strings pass the regular expression and can still throw when the port is out of range or cannot be represented as an integer.

This change keeps invalid-address handling consistent and prevents a malformed boot node entry from interrupting peer discovery initialization.

## How Has This Been Tested?

- `IpUtilsTest`: 9 tests passed
- Added coverage for:
  - a port above the maximum valid value
  - a numeric port larger than the integer range

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/util/IpUtils.java
+++ b/rskj-core/src/main/java/co/rsk/util/IpUtils.java
@@ -53,6 +53,10 @@ public class IpUtils {
     }
 
     private static InetSocketAddress parseMatch(Matcher matcher) {
-        return new InetSocketAddress(matcher.group(1), Integer.valueOf(matcher.group(2)));
+        try {
+            return new InetSocketAddress(matcher.group(1), Integer.valueOf(matcher.group(2)));
+        } catch (IllegalArgumentException e) {
+            return null;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


rskj-core/src/main/java/co/rsk/util/IpUtils.java
```diff
@@ -53,6 +53,10 @@ public static List<InetSocketAddress> parseAddresses(List<String> addresses) {
     }
 
     private static InetSocketAddress parseMatch(Matcher matcher) {
-        return new InetSocketAddress(matcher.group(1), Integer.valueOf(matcher.group(2)));
+        try {
+            return new InetSocketAddress(matcher.group(1), Integer.valueOf(matcher.group(2)));
+        } catch (IllegalArgumentException e) {
+            return null;
```
fmacleal, line 65: **suggestion**:

Instead of a silent exception, could you log a warning with a meaningful message here using `logger.warning` log level?

alaningtrump: Thanks for your suggestion.

Added a WARN log containing the invalid host, port, and exception message before returning null.

fmacleal (approved):
LGTM.

Thank you for your contribution! 
🙏

PR #3620, opened 2026-07-08
fix: Fixed max fee fields decoding for Type2 and Type4 transaction parsers

dlealmac:
## Description
Decode zero maxPriorityFeePerGas / maxFeePerGas as Coin.ZERO when wallets encode them as RLP 0x80, aligning Type-2/Type-4 with Type-0/Type-1 and avoiding NPE on raw-tx decode.

## Motivation and Context
Type-2 and Type-4 parsers used Objects.requireNonNull(RLP.parseCoinNonNullZero(...)) for maxPriorityFeePerGas and maxFeePerGas. When those fields are encoded as the canonical RLP empty string (0x80) parseCoinNonNullZero returns null and decoding throws an uncaught NullPointerException.

## How Has This Been Tested?
The unit tests cover canonical zero-fee (0x80) decoding for Type-2 and Type-4 on the raw RLP decode path, plus an integration test for Type-2 over the P2P TransactionsMessage wire path.

## Types of changes

## Checklist:

### DIFF
--- a/rskj-core/src/test/java/org/ethereum/core/TypedTransactionIntegrationTest.java
+++ b/rskj-core/src/test/java/org/ethereum/core/TypedTransactionIntegrationTest.java
@@ -28,6 +28,7 @@ import co.rsk.test.builders.BlockBuilder;
 import com.typesafe.config.ConfigValueFactory;
 import org.ethereum.core.transaction.TransactionType;
 import org.ethereum.db.TransactionInfo;
+import org.ethereum.util.ByteUtil;
 import org.ethereum.util.RLP;
 import org.ethereum.util.RLPList;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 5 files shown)


italo-sampaio (approved):
LGTM!

rskj-core/src/test/java/org/ethereum/core/TypedTransactionIntegrationTest.java
```diff
@@ -211,6 +216,38 @@ void type2Transaction_encodesAndDecodesViaTransactionsMessage() {
                 "Type 2 tx hash must be identical after TransactionsMessage encode/decode");
     }
 
+    @Test
+    void type2CanonicalZeroFees_decodesViaTransactionsMessageP2pPath() {
```
meldsun0, line 223: • Add the equivalent P2P test for Type-4.
• Add a P2P test for an invalid transaction, to verify that invalid input is cleanly rejected rather than causing an uncaught internal exception.

meldsun0 (changes requested):

rskj-core/src/test/java/org/ethereum/core/TypedTransactionIntegrationTest.java
dlealmac: done!

meldsun0 (approved):

PR #3625, opened 2026-07-10
Upgrading version

fmacleal:
Bumping version


italo-sampaio (approved):

fmacleal:
This wont work, because RIT need other changes, we will release with latest master that fixes this issue here: https://github.com/rsksmart/rskj/pull/3627

PR #3632, opened 2026-07-13
fix: allowing type4 transactions to be gas bumped

meldsun0:
## Description
This PR fixes the EIP-7702 delegated-account transaction pool validation so that same-nonce replace-by-fee transactions are allowed.

Delegated accounts are still limited to one transaction in the pool, but a replacement transaction with the same nonce and a sufficient gas price bump can now replace the existing transaction instead of being rejected before the gas-bump validation runs.

## Motivation and Context
Previously, once a delegated account had a transaction in the pool, any second transaction from that account was rejected with:

`delegated account already has a transaction in the pool`

This also rejected valid same-nonce replacement transactions before they reached the existing isBumpingGasPriceForSameNonceTx logic.

As a result, if a delegated account transaction got stuck below the market gas price, the user had no way to bump it.

This change preserves the one-transaction limit for delegated accounts while allowing normal replace-by-fee behavior within that single txpool slot.

## How Has This Been Tested?
Added unit tests covering:

- a delegated account can replace an existing transaction with the same nonce and a valid gas price bump;
- only one transaction remains in the pool after the replacement;
- the original transaction is removed from the pool;
- a delegated account still cannot add a second transaction with a different nonce;
- insufficient gas bumps are still rejected with the normal gas-bump error;
- quota validation receives the replaced transaction when processing a replacement.

## Types of changes

## Checklist:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
+++ b/rskj-core/src/main/java/co/rsk/core/bc/TransactionPoolImpl.java
@@ -264,16 +264,18 @@ public class TransactionPoolImpl implements TransactionPool {
         }
         PendingState pendingState = getPendingState(currentRepository);
 
-        Optional<TransactionPoolAddResult> delegatedAccountResult = rejectIfDelegatedAccountCannotBeAccepted(tx, tx.getSender(signatureCache), currentRepository, pendingState);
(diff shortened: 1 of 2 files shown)


italo-sampaio (approved):
LGTM!

dlealmac (approved):
LGTM

PR #3634, opened 2026-07-14
Add the authorization refund  before computing max refund

meldsun0:
## Description
Fixes EIP-7702 gas refund accounting by adding the authorization refund to the shared future refund counter before applying the refund cap.
This ensures VM refunds and authorization refunds are capped together.
Adds regression coverage for combined refund behavior.

## Motivation and Context
The previous flow capped VM refunds first and added the EIP-7702 authorization refund afterward.
This could make the effective refund exceed gasUsed / 2.
Gas accounting is consensus-critical, so all refund sources must share the same cap.

## How Has This Been Tested?
Added tests for Type 4 transactions with both VM and authorization refunds.
Added coverage for authorization refunds when the total refund is below the cap.
Verified the deducted refund never exceeds the allowed refund ceiling.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
+++ b/rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
@@ -729,12 +729,13 @@ public class TransactionExecutor {
     }
 
     private long refundGas() {
-        // Accumulate refunds for suicides
+        // Accumulate refunds for deleted accounts and authorizations before applying the refund cap
         result.addFutureRefund(GasCost.multiply(result.getDeleteAccounts().size(), GasCost.SUICIDE_REFUND));
+        result.addFutureRefund(authorizationRefund);
 
-        // The actual gas subtracted is equal to half of the future refund
+        // The actual refund is capped to half of the gas used
         long gasRefund = Math.min(result.getFutureRefund(), result.getGasUsed() / 2);
-        gasRefund = GasCost.add(gasRefund, authorizationRefund);
+
         result.addDeductedRefund(gasRefund);
         result.setGasUsedBeforeRefunds(result.getGasUsed());
 
--- a/rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/Type4TransactionExecutorTests.java
+++ b/rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/Type4TransactionExecutorTests.java
@@ -1583,6 +1583,201 @@ class Type4TransactionExecutorTests extends Type4TransactionExecutorHelperTest {
 
     }
 
+    @Test
+    void type4TransactionCapsCombinedVmAndAuthorizationRefundWhenAboveCap() {
+        int storageSlotsToClear = 32;
+
+        long expectedAuthorizationRefund = GasCost.PER_EMPTY_ACCOUNT_COST - GasCost.PER_AUTH_BASE_COST;
+
+        byte[] delegatedCode = DelegationCodeResolver.createDelegatedCode(createRandomAddress());
+
+        MutableRepository cacheTracker = mock(MutableRepository.class);
+        MutableRepository authorizationTracker = mock(MutableRepository.class);
+
+        when(tracker.startTracking()).thenReturn(cacheTracker, authorizationTracker);
+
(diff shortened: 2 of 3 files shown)


rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/Type4TransactionExecutorTests.java
```diff
@@ -1401,4 +1401,196 @@ void shouldRejectTransactionWhenBalanceCannotCoverAuthorizationCosts() {
         verify(authorizationTracker, never()).commit();
     }
 
+    @Test
+    void type4TransactionAppliesCombinedVmAndAuthorizationRefundWhenBelowCap() {
```
italo-sampaio, line 1405: **nit**: These four tests all exercise below-cap scenarios (they assert `deductedRefund <= maxRefund` where the cap doesn't bind). Consider adding one test where the VM refund alone reaches `gasUsed/2` so the authorization refund would push the total over the cap, i.e., a case that fails on the pre-fix ordering (refund added after the cap) but passes here. That's the exact scenario this change guards, and it would catch a future regression that re-adds the refund after the cap.

italo-sampaio (approved):
LGTM, left one comment inline

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/Type4TransactionExecutorTests.java
```diff
@@ -1401,4 +1401,196 @@ void shouldRejectTransactionWhenBalanceCannotCoverAuthorizationCosts() {
...
+
+        var txExecutor = newExecutor(tx);
+
+        assertTrue(txExecutor.executeTransaction());
+
+        long gasUsedBeforeRefunds = txExecutor.getResult().getGasUsedBeforeRefunds();
+        long maxRefund = gasUsedBeforeRefunds / 2;
+        long deductedRefund = txExecutor.getResult().getDeductedRefund();
+        assertEquals(expectedVmRefund + expectedAuthorizationRefund, deductedRefund);
+        assertTrue(deductedRefund <= maxRefund);
```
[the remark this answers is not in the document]
meldsun0: A new test will be created to verify capping.

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/Type4TransactionExecutorTests.java
```diff
@@ -1401,4 +1401,196 @@ void shouldRejectTransactionWhenBalanceCannotCoverAuthorizationCosts() {
...
+    }
```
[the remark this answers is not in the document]
meldsun0: A test will be added to verify this.

italo-sampaio (approved):

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -729,12 +729,13 @@ private TransactionExecutionSummary buildTransactionExecutionSummary(Transaction
     }
 
     private long refundGas() {
-        // Accumulate refunds for suicides
+        // Accumulate refunds for deleted accounts and authorizations before applying the refund cap
         result.addFutureRefund(GasCost.multiply(result.getDeleteAccounts().size(), GasCost.SUICIDE_REFUND));
+        result.addFutureRefund(authorizationRefund);
```
dlealmac, line 734: @meldsun0 please, shouldn't this change be restricted to rskip545 activation?

meldsun0: I don't think an additional RSKIP545 check is needed here. authorizationRefund is initialized to 0 and is only populated for type-4 transactions. Type-4 transactions are already rejected by isTypedTransactionNotAllowed() when RSKIP545 is not active, so a non-zero authorization refund can only reach this code after RSKIP545 activation.

dlealmac: I see, thanks for clarification

dlealmac (approved):

PR #3651, opened 2026-07-22
fix: improve yParity validation in TypedTransactionCodec

dlealmac:
## Description
Fix yParity parser to reject RLP payload longer than one byte.

## Motivation and Context
Typed txs and EIP-7702 auth tuples both use yParity, but only the auth codec enforced a single-byte field. That left an EIP-2718 canonical-encoding inconsistency and a leading-byte misread on padded values. Aligning the parsers improves strictness and robustness.

## How Has This Been Tested?
It has been tested with unit tests from TypedTransactionCodecTest.

## Types of changes

## Checklist:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/TypedTransactionCodec.java
+++ b/rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/TypedTransactionCodec.java
@@ -42,6 +42,9 @@ public final class TypedTransactionCodec {
                                                       int sIndex) {
         byte[] r = txFields.get(rIndex).getRLPData();
         byte[] s = txFields.get(sIndex).getRLPData();
+        // Validated before the unsigned early return so the field is checked on every path.
+        // Callers run requireFieldCount first, so the index always exists.
+        byte yParity = parseTypedYParity(txFields.get(yParityIndex).getRLPData());
 
         if (r == null && s == null) {
             byte chainId = parseTypedTxChainId(txFields.get(chainIdIndex).getRLPData());
@@ -53,7 +56,6 @@ public final class TypedTransactionCodec {
         }
         CommonParsingUtils.requireSignatureComponent(r, "Signature R is not valid");
         CommonParsingUtils.requireSignatureComponent(s, "Signature S is not valid");
-        byte yParity = parseTypedYParity(txFields.get(yParityIndex).getRLPData());
         byte v = (byte) (LOWER_REAL_V + yParity);
         byte chainId = parseTypedTxChainId(txFields.get(chainIdIndex).getRLPData());
         return new SignedSignature(chainId, ECDSASignature.fromComponents(r, s, v));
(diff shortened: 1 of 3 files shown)


rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/TypedTransactionCodec.java
```diff
@@ -68,7 +68,13 @@ public static byte parseRequiredTypedChainId(String chainIdHex) {
     }
 
     private static byte parseTypedYParity(byte[] yParityData) {
-        byte yParity = (yParityData != null && yParityData.length > 0) ? yParityData[0] : 0;
+        if (yParityData == null || yParityData.length == 0) {
```
meldsun0, line 86: Should a signed typed transaction with r and s present but an empty yParity be rejected, rather than implicitly treated as yParity = 0?

meldsun0 (commented):
It may also be useful to share the validation logic with AuthorizationListCodec so both paths remain aligned.

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/TypedTransactionCodec.java
```diff
@@ -68,7 +68,13 @@ public static byte parseRequiredTypedChainId(String chainIdHex) {
...
+            return 0;
+        }
+        if (yParityData.length > 1) {
```
italo-sampaio, line 89: **Apply the same check on the unsigned path.**

The change is correct for signed transactions. A payload of `0x00 0x01` used to read as yParity 0 and produce `V = 27`, so a transaction signed with parity 1 recovered to a different address than the signer. The new check rejects that input. The round trip is safe as well: `RLP.encodeByte` emits `0x80` for zero and a single byte otherwise, so nothing the typed encoders produce can now be rejected.

`parseTypedSignatureState` returns at line 48 when `r` and `s` are both absent. `parseTypedYParity` never runs on that path. An envelope with empty `r` and `s` and a yParity payload of `0x00 0x01` still parses. Each parser calls `CommonParsingUtils.requireFieldCount` before it reaches this method, twelve fields for type 2, so `txFields.get(yParityIndex)` always exists and the check can move above the early return.

The impact is small, because an unsigned envelope is rejected later for the missing signature. The field is still validated on one path and ignored on the other.

One note on scope. RLP encodes the integer 0 as the empty string, so a payload of one `0x00` byte is a non-canonical encoding of 0. Both this method and `AuthorizationListCodec.parseYParity` accept it. That has no effect on the transaction hash, because `Transaction.getEncoded` derives the encoding from the parsed fields through `TransactionEncoderFactory` and never keeps the raw bytes. No change is needed for it. The PR description mentions EIP-2718 canonical encoding, so it helps to say which part of that rule the change covers.

italo-sampaio (commented):
Left a comment inline

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/TypedTransactionCodec.java
dlealmac: empty yParity and zero encoded as the same, RLP encodes both as an empty item, if we reject it we are going to reject valid transactions. This test capture that scenario: parseTypedSignatureState_emptyYParity_decodesAsZero

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/TypedTransactionCodec.java
dlealmac: done

meldsun0 (approved):

italo-sampaio (approved):
LGTM!

dlealmac:
This PR was recreated to target rskip545-security-review: https://github.com/rsksmart/rskj/pull/3686

PR #3652, opened 2026-07-22
Additional test coverage for #3589

italo-sampaio:
This PR targets the `is/fix-bootstrap-exporter` branch so the tests fold directly into #3589. **Tests only — no production changes.** All new tests pass against the branch head, alongside the existing 43 tests in the affected classes.

These close coverage gaps in behaviors this PR introduces that the current tests do not exercise:

### 1. Chunk-length absolute ceiling — `BootstrapImporterV2Test`
- `rejectsChunkLengthAbovePerChunkCeilingBeforeAllocating` — a declared chunk length of `MAX_CHUNK_BYTES + 1` (512 MiB + 1) is rejected **before** `new byte[]`.
- `rejectsNegativeChunkLengthBeforeAllocating` — a negative declared length is rejected.

The existing `rejectsChunkLengthExceedingFileSizeBeforeAllocating` only hits the `len > remaining` branch. The absolute `len > MAX_CHUNK_BYTES` / `len < 0` ceiling — the guard that bounds allocation *regardless of file size*, which is the point of the bounded-memory design — was untested.

### 2. Streamed hash over multiple buffers — `BootstrapFileHandlerTest`
- `retrieveAndUnpack_streamsHashOverMultipleBuffersForLargeFile` — a fixture larger than `HASH_BUFFER_SIZE` (64 KiB) so the streamed digest loop iterates; asserts the streamed digest equals `HashUtil.sha256(full)`.
- `retrieveAndUnpack_rejectsLargeFileWhenOneByteMutated` — same multi-buffer size with a one-byte mutation, asserting rejection (proves every buffer is fed to the digest).

The existing hash tests use ~14-byte fixtures, so the multi-buffer streaming loop (the part that makes the >2 GiB archive hashing bounded-memory) never actually runs.

### 3. Header / framing negatives — `BootstrapImporterV2Test`
- `rejectsEmptyFile`, `rejectsV2WithBadMagic`, `rejectsV2WithUnsupportedVersion`, and `rejectsTruncatedV2MissingEndMarker` (distinct from the existing trailing-bytes test).

### 4. `MultiTrieStore.saveValue` — `MultiTrieStoreTest`
- `saveValueDelegatesToCurrentStoreAndIsRetrievableByHash` — the new `saveValue` writes only to the current store and the value is retrievable by its hash.

No product defects were found while writing these — every test passes against the code as written.

### DIFF
--- a/rskj-core/src/test/java/co/rsk/db/importer/BootstrapImporterV2Test.java
+++ b/rskj-core/src/test/java/co/rsk/db/importer/BootstrapImporterV2Test.java
@@ -345,6 +345,115 @@ class BootstrapImporterV2Test {
         assertTrue(ex.getMessage().toLowerCase().contains("trailing"), ex.getMessage());
     }
 
+    @Test
+    void rejectsChunkLengthAbovePerChunkCeilingBeforeAllocating() throws IOException {
+        // A declared chunk length above the absolute per-chunk ceiling (MAX_CHUNK_BYTES = 2 * CHUNK_MAX =
+        // 512 MiB). validateChunkLength checks this range guard BEFORE the remaining-bytes guard, so a
+        // corrupt length that would otherwise allocate a > 512 MiB byte[] is rejected in a tiny file without
+        // any allocation. This is the absolute-ceiling branch, distinct from the remaining-bytes branch that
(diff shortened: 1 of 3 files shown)


fmacleal (approved):
LGTM

PR #3654, opened 2026-07-23
fix: align low-s sginature validation across authorization paths

dlealmac:
## Description
Aligns RSKIP-545 / EIP-7702 authorization low-s checks so they agree with existing transaction validation (Transaction.acceptTransactionSignature): reject s >= secp256k1n/2.

## Motivation and Context
Two paths applied the low-s malleability check to the same authorization signature with different bounds: verifyLowS accepted s == secp256k1n/2, while AuthorizationListCodec.validateAuthorization (and acceptTransactionSignature) rejected it.

## How Has This Been Tested?
It has been tested with new unit tests around the low-s boundary.

## Types of changes

## Checklist:


rskj-core/src/main/java/org/ethereum/core/SetCodeAuthorizationTransactionExecutor.java
```diff
@@ -67,8 +67,6 @@ private void verifyChainId(BigInteger chainId, BigInteger outerTransactionChainI
     }
 
     private RskAddress checkRecoveredAuthority(SetCodeAuthorization setCodeAuthorization) {
-        setCodeAuthorization.verifyLowS();
-
         byte[] messageHash =  setCodeAuthorization.getSigningHash();
         ECKey key;
```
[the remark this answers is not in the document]
meldsun0: It is not correct to remove setCodeAuthorization.verifyLowS() from checkRecoveredAuthority() because the two validations occur at different protocol layers and have different failure semantics.

The encoding-time check:

auth.s < 2**256

only validates that s fits in the authorization tuple’s encoded field. If this check fails, the transaction itself is malformed, so rejecting the entire transaction is appropriate.

The low-s check:

setCodeAuthorization.verifyLowS();

is part of processing each authorization tuple. Under the EIP-7702 behavior, an authorization with a non-canonical signature must fail that tuple, while processing continues with the remaining authorizations. The specification explicitly says that if any tuple-processing step fails, processing stops for that tuple and continues with the next one.

Therefore, moving verifyLowS() into transaction decoding or general transaction validation would change the consensus behavior

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/SetCodeAuthorizationTransactionExecutorTest.java
```diff
@@ -195,27 +195,6 @@ void processAuthorizationTuple_shouldAllowUniversalChainIdWithAnyOuterChainId()
         assertEquals(0L, refund);
     }
 
-    @Test
```
meldsun0, line 181: This test is okay  as the tuple is well-formed and should be decoded successfully. It should only be rejected later, during authorization tuple execution, by the low-s validation—not during encoding or transaction parsing.

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/SetCodeSetCodeAuthorizationTransactionExecutorTest.java
```diff
@@ -423,11 +402,34 @@ void processAuthorizationTuple_shouldNotSaveOrIncreaseNonce_whenSignatureSIsTooH
...
                 () -> executor.processAuthorizationTuple(repository, ZERO_CHAIN_ID, tuple)
         );
 
+        assertEquals("Signature recovery failed", ex.getMessage());
+        verify(repository, never()).saveCode(any(), any());
+        verify(repository, never()).increaseNonce(any());
+    }
+
+    @Test
+    void processAuthorizationTuple_shouldNotSaveOrIncreaseNonce_whenSignatureSIsHalfCurveOrder() {
```
meldsun0: I don't think this test is validating the intended scenario. Since the signature is mocked, the failure comes from signature recovery rather than the s = n / 2 value itself. As written, it doesn't verify the low-s boundary behavior. Only getS() is stubbed, while the remaining signature fields keep Mockito's default values, resulting in an invalid recoverable signature. Consequently, the test exercises the recovery path instead of the low-s validation.

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
```diff
@@ -94,7 +94,7 @@ public static byte[] encodeList(List<SetCodeAuthorization> authorizations) {
 
     public static byte[] encodeTuple(SetCodeAuthorization auth) {
         validateAuthorization(auth);
```
meldsun0, line 94: That method has twotwo problems:

- The low-s rule belongs to per-tuple execution, because failure should invalidate only that authorization.
- The specification allows equality:  s <= secp256k1n / 2

Therefore, using >= 0 incorrectly rejects s == secp256k1n / 2. If this validation were performed, the comparison would need to be: s.compareTo(SECP256K1N_HALF) > 0

But it should not be performed in this encoding method.

Adapt tests accordingly.

meldsun0 (changes requested):
One important distinction is that a high-s value is not an encoding error. An authorization tuple can be created and propagated with an s value in the valid uint256 range, even if it is not a canonical ECDSA signature. The low-s check is a cryptographic validation performed during authorization processing, where a failure invalidates only that tuple and not the entire transaction. This is why the validation should remain in the execution path rather than being moved to transaction decoding or parsing.

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
```diff
@@ -44,6 +43,7 @@
 public final class AuthorizationListCodec {
 
     private static final int TUPLE_FIELD_COUNT = 6;
+    private static final byte LOWER_REAL_V = 27;
```
italo-sampaio: This constant (and the matching one added to `AuthorizationListCodecTest`) looks like unrelated scope creep — it is not mentioned in the PR description and is not needed for the low-s alignment.

`parser/util` already carries private copies of `27` in `Type0SignatureUtils` and `TypedTransactionCodec`, so replacing `Transaction.LOWER_REAL_V` here brings the count to four definitions of the same protocol constant, one of which is already `public static final` on `Transaction`.

If the intent is to break the `parser.util` -> `Transaction` dependency, that is a reasonable goal but deserves its own pass that introduces one shared constant and removes all the copies at once. Otherwise, suggest reverting this hunk (and the three call sites plus the test constant) so the diff stays scoped to the low-s change.

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/SetCodeSetCodeAuthorizationTransactionExecutorTest.java
```diff
@@ -423,11 +402,34 @@ void processAuthorizationTuple_shouldNotSaveOrIncreaseNonce_whenSignatureSIsTooH
                 signature
         );
 
-        assertThrows(
+        IllegalStateException ex = assertThrows(
                 IllegalStateException.class,
                 () -> executor.processAuthorizationTuple(repository, ZERO_CHAIN_ID, tuple)
         );
 
+        assertEquals("Signature recovery failed", ex.getMessage());
```
italo-sampaio: The vacuity problem flagged on `..._whenSignatureSIsHalfCurveOrder` below applies identically to this test, so both need the same fix. With `verifyLowS()` removed from the executor, nothing between method entry and `signatureToKey` reads `s` at all — the mock only stubs `getS()`, so the failure is driven entirely by the unstubbed `v`/`r`. This assertion would hold with `s = 1`, and it holds whether or not any low-s rule exists anywhere in the codebase.

Once the check is back in the execution path, both tests should assert the specific low-s message (`"Signature s exceeds secp256k1n / 2"`, or whatever wording lands) rather than `"Signature recovery failed"`, so they actually fail if the rule is dropped again.

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/SetCodeSetCodeAuthorizationTransactionExecutorTest.java
```diff
@@ -423,11 +402,34 @@ void processAuthorizationTuple_shouldNotSaveOrIncreaseNonce_whenSignatureSIsTooH
...
+        );
+
+        IllegalStateException ex = assertThrows(
+                IllegalStateException.class,
+                () -> executor.processAuthorizationTuple(repository, ZERO_CHAIN_ID, tuple)
+        );
+
+        assertEquals("Signature recovery failed", ex.getMessage());
         verify(repository, never()).saveCode(any(), any());
         verify(repository, never()).increaseNonce(any());
```
italo-sampaio: Coverage gap worth closing while this area is open: there is no test asserting the invariant that a failed tuple does not sink the transaction.

`TransactionExecutor.processAuthorizationList` wraps each tuple in its own `startTracking()` / `catch` / `rollback()`, which is what implements RSKIP-545's "immediately stop processing the tuple and continue to the next tuple in the list". Nothing currently pins that behaviour for a low-s failure specifically — these executor tests only assert that the single failing tuple writes nothing.

Suggest a `TransactionExecutor`-level test with a type-4 transaction carrying two authorizations, the first with a high-`s` signature and the second valid and properly recoverable: the first must be skipped, the second must be applied (delegation written, authority nonce bumped), and the transaction must still execute. That is the regression guard that would have caught the enforcement moving to decode time, which is where the layering discussion above ultimately lands.

italo-sampaio (commented):
Left some comments inline

rskj-core/src/main/java/org/ethereum/core/SetCodeAuthorizationTransactionExecutor.java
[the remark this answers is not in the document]
dlealmac: done

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/SetCodeSetCodeAuthorizationTransactionExecutorTest.java
dlealmac: done

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
dlealmac: done

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
dlealmac: done

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/SetCodeSetCodeAuthorizationTransactionExecutorTest.java
dlealmac: done

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/SetCodeSetCodeAuthorizationTransactionExecutorTest.java
dlealmac: done

rskj-core/src/main/java/org/ethereum/core/Transaction.java
```diff
@@ -508,6 +508,15 @@ public boolean acceptTransactionSignature(byte currentChainId) {
         return this.getChainId() == 0 || this.getChainId() == currentChainId;
     }
 
+    /**
+     * EIP-2 admits {@code s == secp256k1n/2}. Legacy transactions keep the historical exclusive bound
+     * so consensus over already-validated blocks is unchanged; typed transactions use the spec bound.
+     */
+    private boolean isSignatureSAboveLowSBound() {
+        int comparison = signature.getS().compareTo(SECP256K1N_HALF);
+        return typePrefix.isTyped() ? comparison > 0 : comparison >= 0;
```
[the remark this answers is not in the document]
meldsun0: The s <= secp256k1n/2 rule in EIP-7702 applies specifically to authorization tuple signatures.
It does not change the validation rules for the outer Type 4 transaction signature.
Using isTyped() would also modify the existing behavior of Type 1 and Type 2 transactions.
The inclusive bound should therefore be applied only when validating each authorization item.

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/Type4TransactionValidation.java
meldsun0, line 56: Type 4 should follow the same outer-signature validation used by existing transaction types.
The EIP-7702-specific `s <= secp256k1n/2` check applies to authorization signatures, so preserve existing outer-signature behavior.

meldsun0 (changes requested):

rskj-core/src/main/java/org/ethereum/core/Transaction.java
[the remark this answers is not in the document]
dlealmac: done

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/Type4TransactionValidation.java
dlealmac: done

meldsun0 (approved):

italo-sampaio (approved):
LGTM

PR #3655, opened 2026-07-24
fix: reject invalid eth_call fee parameters with priority fee only

dlealmac:
## Description
Reject RPC call arguments that provide maxPriorityFeePerGas without maxFeePerGas.

## Motivation and Context
Previously, supplying only maxPriorityFeePerGas caused effectiveGasPrice() to return null, which getGasPrice() converted to zero. This could make eth_call and eth_estimateGas return results calculated with an unintended gas price rather than reporting invalid fee parameters.

## How Has This Been Tested?
Added unit tests covering combinations of gasPrice, maxPriorityFeePerGas, and maxFeePerGas, including missing and empty maxFeePerGas values. 

## Types of changes

## Checklist:


meldsun0 (commented):
Consider adding an RPC-level test to verify that eth_call and eth_estimateGas return the expected -32602 JSON-RPC error, ensuring the validation is propagated through the public API.

rskj-core/src/main/java/org/ethereum/rpc/converters/CallArgumentsToByteArray.java
```diff
@@ -64,6 +66,9 @@ private BigInteger effectiveGasPrice() {
         if (hasMaxFee) {
             return HexUtils.strHexOrStrNumberToBigInteger(args.getMaxFeePerGas());
         }
+        if (hasMaxPriority) {
```
italo-sampaio: **suggestion**: Also reject `maxPriorityFeePerGas` greater than `maxFeePerGas`.

This branch closes the case where `maxFeePerGas` is absent. The branch a few lines above still accepts a reversed pair. It returns `maxPriority.min(maxFee)` whenever both fields are present.

`maxFeePerGas = 0x1` with `maxPriorityFeePerGas = 0x64` runs the call at an effective price of 1 and reports nothing. The PR description describes that same outcome: a result calculated with an unintended gas price instead of a report of invalid fee parameters.

The send path already rejects that pair. `Type2RawTransactionParser.validateFeeCapRelationship`, and the Type 4 equivalent, throw "Type 2 transaction maxPriorityFeePerGas (...) must not exceed maxFeePerGas (...)". After this change, `eth_call` and `eth_sendTransaction` still accept different fee pairs.

The same relationship check here closes that difference. Treating an absent `maxFeePerGas` as `0` would let one `maxFeePerGas >= maxPriorityFeePerGas` rule cover both cases, including the case this PR fixes.

italo-sampaio (commented):
Left a comment

rskj-core/src/main/java/org/ethereum/rpc/converters/CallArgumentsToByteArray.java
dlealmac: I created Rskip546FeeValidation.requireFeeCapRelationship to keep validation consistent across all entries. Thanks.

italo-sampaio (approved):

meldsun0 (approved):

PR #3656, opened 2026-07-24
refactor: remove redundant typed transaction activation check from TxPendingValidator

dlealmac:
## Description
Removed a redundant RSKIP543 activation check from TxPendingValidator. Typed-transaction activation validation remains centralized in Transaction.isTypedTransactionNotAllowed.

## Motivation and Context
The removed condition was unreachable because isTypedTransactionNotAllowed already rejects typed transactions when RSKIP543 is inactive using the same activation state.

Removing this duplicate check has no runtime behavior impact and avoids future divergence between two activation-validation paths.

## How Has This Been Tested?
Added tests ensuring that typed transactions are rejected before activation and that validation returns before transaction cost, init-code size, and transaction size checks are performed.

## Types of changes

## Checklist:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/net/handler/TxPendingValidator.java
+++ b/rskj-core/src/main/java/co/rsk/net/handler/TxPendingValidator.java
@@ -73,14 +73,12 @@ public class TxPendingValidator {
     public TransactionValidationResult isValid(Transaction tx, Block executionBlock, @Nullable AccountState state) {
         long executionBlockNumber = executionBlock.getNumber();
         ActivationConfig.ForBlock activations = activationConfig.forBlock(executionBlockNumber);
-        BigInteger gasLimit = activationConfig.areActive(executionBlockNumber, ConsensusRule.RSKIP351, ConsensusRule.RSKIP144)
-                ? BigInteger.valueOf(Math.max(BlockUtils.getSublistGasLimit(executionBlock, true, constants.getMinSequentialSetGasLimit()), BlockUtils.getSublistGasLimit(executionBlock, false, constants.getMinSequentialSetGasLimit())))
-                : BigIntegers.fromUnsignedByteArray(executionBlock.getGasLimit());
-        Coin minimumGasPrice = executionBlock.getMinimumGasPrice();
-        long bestBlockNumber = executionBlock.getNumber();
         if (tx.isTypedTransactionNotAllowed(activations)) {
             return TransactionValidationResult.withError("transaction type " + tx.getTypePrefix() + " is not supported before its activation");
         }
+
+        BigInteger gasLimit = resolveGasLimit(executionBlock, executionBlockNumber);
+        Coin minimumGasPrice = executionBlock.getMinimumGasPrice();
 
         long basicTxCost = tx.transactionCost(constants, activations, signatureCache);
 
@@ -91,7 +89,7 @@ public class TxPendingValidator {
             return TransactionValidationResult.withError("the sender account doesn't exist");
         }
 
-        if (tx.isInitCodeSizeInvalidForTx(activationConfig.forBlock(bestBlockNumber))) {
+        if (tx.isInitCodeSizeInvalidForTx(activations)) {
             return TransactionValidationResult.withError("transaction's init code size is invalid");
         }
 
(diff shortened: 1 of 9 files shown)


meldsun0 (approved):

rskj-core/src/main/java/co/rsk/net/handler/TxPendingValidator.java
meldsun0, line 76: Perform guard validations before calculating gasLimit.

meldsun0: Isn't this variable duplicated?

meldsun0: Can we extract this into a method ?

rskj-core/src/main/java/co/rsk/net/handler/TxPendingValidator.java
```diff
@@ -91,10 +91,6 @@ public TransactionValidationResult isValid(Transaction tx, Block executionBlock,
             return TransactionValidationResult.withError("the sender account doesn't exist");
         }
 
-        if (tx.getTypePrefix().isTyped() && !activations.isActive(ConsensusRule.RSKIP543)) {
-            return TransactionValidationResult.withError("typed transactions are not supported before RSKIP543 activation");
-        }
-
         if (tx.isInitCodeSizeInvalidForTx(activationConfig.forBlock(bestBlockNumber))) {
```
italo-sampaio: **Reconcile the type 4 activation rules while this validation moves to one place.**

The removal is correct. `Transaction.isTypedTransactionNotAllowed` returns `true` for any typed transaction when RSKIP543 is inactive. Line 81 calls it with the same `activations` object, before the code removed here. The removed branch could not run.

One gap is worth closing in the same change, because the stated goal is to keep this validation in one place. `Transaction.isTypedTransactionNotAllowed` gates type 4 on RSKIP543 and RSKIP545. `Type4RawTransactionParser.validate` gates type 4 on RSKIP543, RSKIP546, and RSKIP545. The two rules disagree about RSKIP546.

The RSKIP text points to the parser holding the correct rule. RSKIP-546 states that the type 4 transaction proposed in RSKIP-545 contains the fields introduced successively in type 1 and type 2 transactions. RSKIP-545 lists RSKIP-546 under prerequisites, for `access_list`, `max_fee_per_gas` and `max_priority_fee_per_gas`. RSKIP-545 calls RSKIP-546 helpful rather than required, so the point is worth settling explicitly. If the parser rule is right, `isTypedTransactionNotAllowed` needs the RSKIP546 condition for type 4 as well.

`rskj-core/src/main` contains four calls to `RawTransactionEnvelopeParser.parse`, in `Transaction.fromRaw`, the `Transaction(byte[])` constructor, `Transaction.fromCallArguments`, and `TransactionBuilder`. All four use an overload that skips `validateActivation`. The overloads that call it are used only from tests. So of the two rules, only `isTypedTransactionNotAllowed` runs, here at line 81 and in `TransactionExecutor` at line 175.

rskj-core/src/test/java/co/rsk/net/handler/TxPendingValidatorTest.java
```diff
@@ -209,6 +210,39 @@ void isValid_ShouldBeInvalid_WhenTypedTransactionAndRSKIP543IsNotActivated() {
...
+    @Test
+    void isValid_ShouldRejectRealTypedTransactionBeforeFurtherValidation_WhenRSKIP543IsNotActivated() {
+        long executionBlockNumber = 10L;
+        TestSystemProperties config = new TestSystemProperties(baseConfig -> baseConfig.withValue(
+                "blockchain.config.hardforkActivationHeights.vetiver900",
+                ConfigValueFactory.fromAnyRef(executionBlockNumber + 1)
+        ));
+        ActivationConfig activationConfig = config.getActivationConfig();
+        ActivationConfig.ForBlock activations = activationConfig.forBlock(executionBlockNumber);
+        Transaction tx = Rskip546TestSupport.unsignedType1();
```
italo-sampaio: **Add a type 4 case to this test.**

`Rskip546TestSupport.unsignedType1()` covers one of the three branches in `isTypedTransactionNotAllowed`. The method also gates type 1 and type 2 on RSKIP546, and type 4 on RSKIP545. This PR makes that method the only activation gate in `TxPendingValidator`, so a type 4 case earns its place here. A second transaction from the type 4 support class, with the same config override, reaches the branch the type 1 case does not.

One word on the name. `..._BeforeFurtherValidation` promises an ordering check. The ordering assertions live in the test above, which uses a mock and can call `verify`. This test asserts the error message instead. The message does show that the guard ran first, because a later check reports a different error. A name such as `isValid_ShouldRejectRealTypedTransaction_WhenRSKIP543IsNotActivated` matches what the test asserts.

italo-sampaio (commented):
left comments inline

rskj-core/src/test/java/co/rsk/net/handler/TxPendingValidatorTest.java
dlealmac: Done.

rskj-core/src/main/java/co/rsk/net/handler/TxPendingValidator.java
dlealmac: Done.

rskj-core/src/main/java/co/rsk/net/handler/TxPendingValidator.java
dlealmac: Done

dlealmac: Done.

dlealmac: Done.

italo-sampaio (approved):

meldsun0 (approved):

PR #3658, opened 2026-07-27
Add new checkstyle rules

marcos-iov:
## Description

Adds a set of Checkstyle rules to `config/checkstyle/checkstyle.xml`, which previously enforced only `NewlineAtEndOfFile`, and fixes the pre-existing violations these rules surfaced across `main`, `test`, and `fuzzTest`. Also fixes an unrelated pre-existing bug in the `checkstyleAll` Gradle task.

**Rules added** (grouped by what they catch):

- Correctness / bug-catchers
  - `EqualsHashCode` — `equals()` overridden without `hashCode()`
  - `CovariantEquals` — `equals(SpecificType)` defined without the real `equals(Object)` override
  - `StringLiteralEquality` — `==`/`!=` on a String literal instead of `.equals()`
  - `MissingOverride` — an inherited method overridden without `@Override`
  - `NoClone` / `NoFinalizer` — overriding `Object.clone()`/`finalize()`
  - `FallThrough` — a switch case falls through without a comment marking it intentional
- Style / consistency
  - `ArrayTypeStyle` — Java-style array declarations only (`Type[] name`, not `Type name[]`)
  - `ModifierOrder` — modifiers must follow JLS order (`public static final`, not `static public final`)
  - `RedundantModifier` — redundant `public`/`abstract` on interface methods, redundant `static` on nested interfaces, redundant `final`
  - `UpperEll` — long literals must use uppercase `L` (`1L`, not `1l`)
  - `DefaultComesLast` — a switch `default` clause that isn't physically last
  - `SimplifyBooleanReturn` — `if (x) return true; else return false;` instead of `return x;`
  - `SimplifyBooleanExpression` — boolean expressions/comparisons that can be written more directly, e.g. `if (b != true)` instead of `if (!b)`
  - `MultipleVariableDeclarations` — one variable declaration per statement/line (`int a, b;` or `int a = 0; int b = 0;` on one line, instead of one declaration per line)
- Dead / stray code
  - `EmptyStatement` — stray `;`
  - `UnnecessarySemicolonAfterOuterTypeDeclaration`, `UnnecessarySemicolonInTryWithResources`, `UnnecessarySemicolonInEnumeration` — stray-semicolon variants `EmptyStatement` doesn't cover
  - `EmptyBlock` (scoped to `catch`) — empty catch blocks must at least contain a comment explaining why. Suppressed for `src/fuzzTest/java` via a new `SuppressionFilter` + `config/checkstyle/suppressions.xml` entry, since fuzz tests intentionally catch-and-ignore the exception a malformed/random input is expected to throw — that's the fuzzing idiom, not a swallowed error.
  - `UnusedImports` (strict mode, `processJavadoc=false`) — unused imports, without giving a pass to imports only referenced in a Javadoc tag
  - `RedundantImport` — duplicate imports, or importing from `java.lang` / the current package

Each rule was checked against the actual codebase (all source sets) before being added, specifically to avoid landing a rule with a large pre-existing backlog that would force unrelated rewrites on the next PR that happens to touch an old file. A few candidates were deliberately **not** added for that reason (`NeedBraces`, `FinalClass`, `ExplicitInitialization`, `AvoidStarImport`, `HideUtilityClassConstructor`, `InnerAssignment`, `OneTopLevelClass`, among others) — they have real, sizeable existing backlogs, or (for `OneTopLevelClass`) conflict with an apparently intentional pattern in this codebase, and are left for a dedicated follow-up rather than bundled in here. `MissingSwitchDefault` was also measured and left out for now: 4 of its 6 hits are easy mechanical fixes, but 2 are in `co.rsk.peg.BridgeSupport` (consensus-critical), where adding a `default:` clause is a behavioral question for that code's owner, not a lint fix.

Also fixes: `checkstyleAll` in `rskj-core/build.gradle` depended on a non-existent task `checkstyleFuzz` (should be `checkstyleFuzzTest`, the task name the checkstyle plugin actually generates for the `fuzzTest` source set), so running it failed immediately. Unrelated to the rule changes, found while verifying them.

## Motivation and Context

The repo's Checkstyle config was effectively a no-op (a single newline check), so a number of easy-to-catch Java mistakes (missing `hashCode()`, unused imports, stray semicolons, wrong-case long literals, non-simplified boolean expressions, etc.) had no automated guard rail. This adds one, scoped tightly to rules that are either zero-cost today or required only a small, mechanical, one-time cleanup — not a repo-wide rewrite.

## How Has This Been Tested?

- After each rule (or batch of rules) was added, ran   `./gradlew --no-daemon --stacktrace build -x test`, which mirrors the   PR-blocking `build` job in `build_and_test.yml` (this is what actually gates   merges, since `checkstyleMain`/`checkstyleTest`/`checkstyleJmh`/   `checkstyleIntegrationTest`/`checkstyleFuzzTest` are all wired into `check`).   Confirmed `BUILD SUCCESSFUL` on the final state.
- Additionally ran `./gradlew checkstyleAll` (all five source sets explicitly)   to confirm there are no violations anywhere, including source sets not   exercised by the default `check` lifecycle.
- Before enabling any rule with unknown blast radius, ran it in isolation   first to measure real violation counts across the whole codebase, and only   added it once that count was zero or a small, reviewable, mechanical fix.
- Ran `./gradlew test --tests "org.ethereum.vm.MemoryTest"` after splitting   its multi-declaration lines (`MultipleVariableDeclarations` fix) — all 22   cases pass unchanged.
- No behavior changed anywhere: all fixes are import removals, modifier   reordering/removal, semicolon/literal-suffix corrections, a boolean   comparison simplified to its equivalent (`x != true` → `!x`), variable   declarations split one-per-line, and one added comment on a   previously-empty catch block.

## Types of changes

## Checklist:

* **Other information**:

This is a static-analysis/tooling change with **no consensus or runtime behavior impact** — every fix is cosmetic (unused imports, modifier order, stray semicolons, literal suffix casing, boolean-expression simplification, one-declaration-per-line splits) or purely additive (one explanatory comment). Flagging for extra attention anyway since it touches files under `co.rsk.pcc` (`ExtractPublicKeyFromExtendedPublicKey.java`, `GetMultisigScriptHash.java`, `ToBase58Check.java` — modifier-order/redundant-modifier fixes only) and test files under `co.rsk.peg`, per this repo's sensitive-area convention, even though no logic in either package changed.


rskj-core/build.gradle
fmacleal, line 1: **chore**:

It's missing also the fix in the line 171 for the `checkstyleFuzz` to `checkstyleFuzzTest`.

fmacleal, line 1: **issue**:

Some of the rules you have added are not present in the version of checkstyle we are using (8.45), we need to upgrade it in order to be able to fully applies some of these rules. Like the `UnusedLocalVariable` and `UnnecessaryParenthesesAfterNegation`which is available from the version (9.3+).

We need to do the bump before the rule-expansion pass, so candidates can be evaluated against a current rule set.

config/checkstyle/checkstyle.xml
```diff
@@ -11,4 +11,78 @@
...
+        <!-- Empty catch blocks must at least contain a comment explaining why -->
+        <module name="EmptyBlock">
+            <property name="option" value="text"/>
+            <property name="tokens" value="LITERAL_CATCH"/>
+        </module>
+
+        <!-- Unused imports; strict mode does not give a pass to imports only referenced in Javadoc -->
+        <module name="UnusedImports">
+            <property name="processJavadoc" value="false"/>
+        </module>
```
fmacleal, line 61: **suggestion**:

`UnusedImports processJavadoc=false` left 6 broken Javadoc references

Strict mode treats an import referenced only from a `{@link}`/`@see` tag as unused. Seven such imports were removed here. `HexPrintableBytes` compensated by rewriting the tag to a fully-qualified name, but six others didn't, so those tags no longer resolve:

| File:line | Removed import | Dangling tag |
|---|---|---|
| `co/rsk/core/types/bytes/BytesSlice.java:90` | `java.util.Arrays` | `{@link Arrays#copyOfRange(Object[], int, int)}` |
| `org/ethereum/net/eth/handler/Eth.java:57` | `EthMessageCodes` | `{@link EthMessageCodes#STATUS}` |
| `co/rsk/net/messages/NewBlockHashesMessage.java:31` | `EthMessageCodes` | `@see EthMessageCodes#NEW_BLOCK_HASHES` |
| `org/ethereum/util/FastByteComparisons.java:42` | `UnsignedBytes` | `{@link UnsignedBytes}` |
| `co/rsk/core/bc/BlockExecutor.java:987` | `java.util.concurrent.Executors` | `{@link Executors...}` |
| `co/rsk/peg/bitcoin/BitcoinTestUtils.java:25` (test) | `BridgeConstants` | `{@link BridgeConstants}` |

Steps to reproduce the issue locally

```
$ ./gradlew :rskj-core:javadoc --rerun-tasks 2>&1 | grep "reference not found" | sort -u
.../BytesSlice.java:90: warning: Tag @link: reference not found: Arrays#copyOfRange(Object[], int, int)
.../Eth.java:57: warning: Tag @link: reference not found: EthMessageCodes#STATUS
.../NewBlockHashesMessage.java:31: warning: Tag @see: reference not found: EthMessageCodes#NEW_BLOCK_HASHES
.../FastByteComparisons.java:42: warning: Tag @link: reference not found: UnsignedBytes
.../PegoutsWaitingForConfirmations.java:50: warning: Tag @link: reference not found: Entry.BTC_TX_COMPARATOR
.../PegoutsWaitingForConfirmations.java:57: warning: Tag @link: reference not found: Entry.BTC_TX_COMPARATOR
BUILD SUCCESSFUL
```

The build stays green because `rskj-core/build.gradle:275` disables doclint globally (`tasks.withType(Javadoc) { options.addStringOption('Xdoclint:none', '-quiet') }`). Without it, an unresolvable `{@link}` is `error: reference not found` and `javadoc` exits 1. So nothing in CI will flag these, and `javadocJar` is published from `archives`, so the dead links ship.

**Proposed fix config, not six more Fully Qualified Name (FQN) rewrites:**

```
     <module name="UnusedImports">
         <property name="processJavadoc" value="true"/>
     </module>
```

This keeps all seven imports legitimate, needs zero source edits, and lets the `HexPrintableBytes` Fully Qualified Name change be reverted. The only thing strict mode catches beyond this is "import used solely by documentation", which isn't really an unused import — and the cost is unreadable FQN tags plus this exact failure mode on every future PR. Verified: with `processJavadoc=true` the whole repo is still clean for `UnusedImports`.

config/checkstyle/checkstyle.xml
```diff
@@ -11,4 +11,78 @@
...
+        <module name="UpperEll"/>
+
+        <!-- Stray empty statements (extra ;) -->
+        <module name="EmptyStatement"/>
+
+        <!-- Empty catch blocks must at least contain a comment explaining why -->
+        <module name="EmptyBlock">
+            <property name="option" value="text"/>
+            <property name="tokens" value="LITERAL_CATCH"/>
+        </module>
```
fmacleal, line 47: **suggestion**:

`EmptyBlock`: the rule is narrower, and the fuzz suppression wider, than intended

Test: a file with an empty `catch`, `if`, `while`, `finally` and `synchronized` block, one copy under `src/main/java` and one under `src/fuzzTest/java`, run against this branch's config as-is. Ten possible findings; here's everything reported:

```
[ERROR] src/main/java/demo/MainDemo.java:7:38: Empty catch block. [EmptyBlock]
```

Two separate issues:

- **`tokens=LITERAL_CATCH` only.** An empty `if`, `while`, `finally` or `synchronized` block in `main` passes silently — and those usually mean dropped logic, unlike the deliberate catch-and-ignore idiom.
- **`<suppress checks="EmptyBlock" files=".../fuzzTest/..."/>` matches by check name,** so it disables the whole check for that source set, not just the fuzzing idiom the comment describes. Nothing at all was reported under `fuzzTest` above. That's invisible today because catch is the only token, but the moment the token list is extended the suppression silently swallows the new tokens too.

**Proposed fix** — two modules with `id`s, and suppress by `id`. Note `suppressions.xml` chadeclares Document Type Definition 1.0, which rejects the `id` attribute, so it needs bumping to 1.2 in the DOCTYPE:

Changes in this file `checkstyle.xml`:
```diff
 <module name="EmptyBlock">
+    <property name="id" value="EmptyCatchBlock"/>
     <property name="option" value="text"/>
     <property name="tokens" value="LITERAL_CATCH"/>
 </module>
+<module name="EmptyBlock">
+    <property name="id" value="EmptyNonCatchBlock"/>
+    <property name="option" value="text"/>
+    <property name="tokens" value="LITERAL_TRY,LITERAL_FINALLY,LITERAL_IF,LITERAL_ELSE,LITERAL_FOR,LITERAL_WHILE,LITERAL_DO,LITERAL_SYNCHRONIZED,STATIC_INIT,INSTANCE_INIT,LITERAL_SWITCH"/>
+</module>
```

Changes in this file `supressions.xml`:

```diff
-<!DOCTYPE suppressions PUBLIC "-//Puppy Crawl//DTD Suppressions 1.0//EN" "suppressions_1_0.dtd">
+<!DOCTYPE suppressions PUBLIC "-//Checkstyle//DTD SuppressionFilter Configuration 1.2//EN"
+        "https://checkstyle.org/dtds/suppressions_1_2.dtd">
 <suppressions>
-    <suppress checks="EmptyBlock" files="[\\/]src[\\/]fuzzTest[\\/]"/>
+    <suppress id="EmptyCatchBlock" files="[\\/]src[\\/]fuzzTest[\\/]"/>
 </suppressions>
```

Same test input with that config: all four non-catch blocks are now caught in *both* source sets, the empty catch is still caught in `main`, and the fuzz catch exemption still holds. Measured cost on the real codebase: **exactly one fix**, an intentionally empty loop body that reads better with the comment the rule asks for anyway —

```diff
     private static void processComment(DslParser parser) {
         for (String line = parser.nextLine(); line != null && !line.trim().toLowerCase().equals("end"); line = parser.nextLine()) {
-
+            // skip every line until the closing "end"
         }
     }
```
(`rskj-core/src/test/java/co/rsk/test/dsl/WorldDslProcessor.java:101`)

While in there, `FallThrough` has a `checkLastCaseGroup` option that's off by default; enabling it also costs exactly one fix (a missing `break;` in the `default:` group at `co/rsk/rpc/Web3InformationRetriever.java:89`) and prevents the classic bug where a new `case` gets appended after a `default` that never terminated.

config/checkstyle/checkstyle.xml
```diff
@@ -11,4 +11,78 @@
...
+        <module name="UnusedImports">
+            <property name="processJavadoc" value="false"/>
+        </module>
+
+        <!-- Modifiers that add nothing: public/abstract on interface methods, static on
+             nested interfaces, final on methods of a final class or on private methods -->
+        <module name="RedundantModifier"/>
+
+        <!-- Overriding a method without @Override hides typos in the signature -->
+        <module name="MissingOverride"/>
```
fmacleal, line 72: **issue**:

`MissingOverride` is much narrower than the PR description implies

The description says it catches "an inherited method overridden without `@Override`". Checkstyle's `MissingOverride` only fires when a method has an explicit `{@inheritDoc}` Javadoc tag but no annotation it does no type hierarchy analysis at all. Consistent with that: this PR added **zero** `@Override` annotations, i.e. the rule found nothing.

Keeping it is fine (it's free), but the description overstates it, and the "hides typos in the signature" protection is not actually in place. Real coverage for that needs compiler/IDE settings or ErrorProne. So, maybe we could update the comment to state it better that needs the javadoc tag.

fmacleal (commented):
Very nice additions, good job!

I liked from the approach of verify the rules and change it step by step, with less impact possible. Since it's a shared repo, the approach it's the safest one.

Besides the rules added, I have paired with Claude and reached the conclusion to add a few more rules that could bring value.

Every candidate below was run against the whole repo with Checkstyle 8.45. "Cost" = one-time fixes needed; the enforced-set number (what breaks the build *today*) is smaller in every case, but repo-wide is the number that matters long term.

**Tier 1 — zero violations repo-wide. Pure guard rails, no cleanup at all:**

| Rule | Catches |
|---|---|
| `OuterTypeFilename` | top-level type whose name doesn't match its filename |
| `PackageDeclaration` | file whose `package` doesn't match its directory (or missing entirely) |
| `MutableException` | non-final field in an exception class |
| `NoLineWrap` | `package`/`import` statements split across lines |
| `TypeName`, `PackageName`, `ClassTypeParameterName`, `MethodTypeParameterName`, `LambdaParameterName` | naming conventions that are already 100% respected — free to lock in |

These are the cheapest thing in the whole list: they cost nothing now and prevent regressions forever.

`IllegalIdentifierName` is also worth adding but needs one tweak — the default pattern flags 16 uses of `sealed` (a domain term in `BlockHeader`/`BlockFactory`, 13 in `main`), which should not be renamed. With `sealed`/`permits` allowed it drops to zero while still banning identifiers that break on newer language levels:

```xml
<module name="IllegalIdentifierName">
    <property name="format" value="(?i)^(?!(record|yield|var|_)$).+$"/>
</module>
```

rskj-core/build.gradle
marcos-iov: Good catch, thanks. Fixed in https://github.com/rsksmart/rskj/pull/3658/changes/a578654a677367c521aaa8511e66672bfc8a1721

marcos-iov: Good point. I have updated checkstyle to version 12.2.0, the latest version that still supports java 17 https://github.com/rsksmart/rskj/pull/3658/changes/0b41d007539c4907d654142e2f24f0e1b2336e68

config/checkstyle/checkstyle.xml
marcos-iov: Fixed in https://github.com/rsksmart/rskj/pull/3658/changes/a52444a79fe38b6d0b3654b028389cc25ccba04f

config/checkstyle/checkstyle.xml
marcos-iov: EmptyBlock rule improved in https://github.com/rsksmart/rskj/pull/3658/changes/ee36268b0c4e99501caab940d788fe909681b589

marcos-iov: Added checkLastCaseGroup property to FallThrough in  https://github.com/rsksmart/rskj/pull/3658/changes/db8824df33d3e03530d719c325eedde8fdea40b8 and fix applied in https://github.com/rsksmart/rskj/pull/3658/changes/231c9a454cc6349ea13d5f02a7e10b00042883eb

config/checkstyle/checkstyle.xml
marcos-iov: True, improved comment https://github.com/rsksmart/rskj/pull/3658/changes/34289e23ac657e8d3a1cda18f2a14d6ab5056932

rskj-core/src/main/java/org/ethereum/crypto/cryptohash/Digest.java
```diff
@@ -174,13 +174,13 @@
...
+	int getBlockLength();
 
 	/**
 	 * <p>Get the display name for this function (e.g. {@code "SHA-1"}
 	 * for SHA-1).</p>
 	 *
 	 * @see Object
 	 */
-	public String toString();
+	String toString();
```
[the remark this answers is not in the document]
marcos-iov: Applied in https://github.com/rsksmart/rskj/pull/3658/changes/3b3bbcc437afc26b2b9e30c707dda35370ac816e

rskj-core/src/test/java/co/rsk/remasc/RemascProcessMinerFeesTest.java
```diff
@@ -309,17 +309,17 @@
 
         Block blockAtHeightThree = blocks.get(blocks.size() - 1);
         Block blockWithOneTxA = RemascTestRunner.createBlock(this.genesisBlock, blockAtHeightThree,
-                PegTestUtils.createHash3(), coinbaseA, Collections.emptyList(), minerFee, 0, txValue, cowKey, 2l);
+                PegTestUtils.createHash3(), coinbaseA, Collections.emptyList(), minerFee, 0, txValue, cowKey, 2L);
```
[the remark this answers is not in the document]
marcos-iov: Fixed in https://github.com/rsksmart/rskj/pull/3658/changes/15be8c3bdf282b65d658d346c71baf102c00ca91

rskj-core/src/test/java/co/rsk/remasc/RemascProcessMinerFeesTest.java
```diff
@@ -309,17 +309,17 @@
...
         blocks.add(blockWithOneTxA);
 
         Block blockWithOneTxC = RemascTestRunner.createBlock(this.genesisBlock, blockWithOneTxA,
-                PegTestUtils.createHash3(), coinbaseC, Collections.emptyList(), minerFee, 1, txValue, cowKey, 2l);
+                PegTestUtils.createHash3(), coinbaseC, Collections.emptyList(), minerFee, 1, txValue, cowKey, 2L);
```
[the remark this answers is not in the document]
marcos-iov: Fixed in https://github.com/rsksmart/rskj/pull/3658/changes/15be8c3bdf282b65d658d346c71baf102c00ca91

rskj-core/src/test/java/co/rsk/remasc/RemascProcessMinerFeesTest.java
```diff
@@ -309,17 +309,17 @@
...
         blocks.add(blockWithOneTxC);
 
         Block blockWithOneTxD = RemascTestRunner.createBlock(this.genesisBlock, blockWithOneTxA,
-                PegTestUtils.createHash3(), coinbaseD, Collections.emptyList(), minerFee, 1, txValue, cowKey, 2l);
+                PegTestUtils.createHash3(), coinbaseD, Collections.emptyList(), minerFee, 1, txValue, cowKey, 2L);
```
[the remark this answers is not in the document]
marcos-iov: Fixed in https://github.com/rsksmart/rskj/pull/3658/changes/15be8c3bdf282b65d658d346c71baf102c00ca91

rskj-core/src/test/java/co/rsk/remasc/RemascProcessMinerFeesTest.java
```diff
@@ -309,17 +309,17 @@
...
         Block blockWithOneTxB = RemascTestRunner.createBlock(this.genesisBlock, blockAtHeightThree,
-                PegTestUtils.createHash3(), coinbaseB, Collections.emptyList(), 3 * minerFee, 0, txValue, cowKey, 2l);
+                PegTestUtils.createHash3(), coinbaseB, Collections.emptyList(), 3 * minerFee, 0, txValue, cowKey, 2L);
```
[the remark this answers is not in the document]
marcos-iov: Fixed in https://github.com/rsksmart/rskj/pull/3658/changes/15be8c3bdf282b65d658d346c71baf102c00ca91

fmacleal (approved):
Thanks for having addressed the comments.

julianlen (approved):

jeremy-then (approved):
LGTM.

PR #3670, opened 2026-08-04
Guard the JSON-RPC reference against drift and sync it to devportal

italo-sampaio:
## Description

`doc/rpc` had no enforced relationship to the surface it documents. 105 JSON-RPC methods are dispatchable, 50 were documented, and nothing noticed the gap. This makes that gap a build failure and ships the resulting document to the docs site automatically.

**The guard.** `JsonRpcDocCoverageTest` derives the exposed set by reflection over `co.rsk.rpc.Web3RskImpl` and compares it, in both directions, against the `name` field of every fragment under the repository-root `doc/rpc/methods/`.

The class is the load-bearing choice, so it is a named constant with the call site recorded beside it. `JsonRpcWeb3ServerHandler` hands jsonrpc4j `service.getClass()` — the concrete implementation — as its `remoteInterface`, and that is what backs both the existence gate in `JsonRpcCustomServer.extractMethodNames` and dispatch in `ReflectionUtil.findCandidateMethods`. The `Web3` interface is **not** the dispatch surface: `eth_getBlocksByNumber`, `ext_dumpBlockchain` and `ext_dumpState` are public on the implementation and absent from the interface, and `eth_getBlocksByNumber` is served by a stock node.

The guard reads each fragment's `name` field rather than its filename, because `name` is what reaches the published document. Methods deliberately left out are listed in `doc/rpc/undocumented.json`, each with a one-line reason, so an exclusion is a checked-in decision rather than an absence. Four supporting checks come with it: an allowlist entry must still be exposed and must not also be documented, no two fragments may claim the same method name, every `$ref` in the assembled document must resolve, and no component may be left unreferenced.

**The exclusion rule.** A method a default node does not serve does not appear in the published reference. `sco_`/`debug_`/`trace_` ship `enabled: "false"` in `reference.conf`; `mnr_`, `db_` and `ext_` are absent from the `rpc.modules` block entirely, so `JsonRpcMethodFilter` has no enabled descriptor to match and the methods are unroutable; `rsk_shutdown`, `rsk_flush` and `rsk_getStorageBytesAt` are named in `methods.disabled`. That covers 28 of the 51 allowlist entries. The other 23 are editorial judgements — development helpers, node-side key management, and dead or unimplemented methods — and each says so.

The guard never reads `reference.conf`. The rule decides what a human writes into `undocumented.json`; guarding against config would mean pinning a config, which is a second thing to drift.

**Bringing the reference up to that bar** — coverage now reads 105 = 54 documented + 51 allowlisted:

- 12 new fragments for default-served methods that had none: `rsk_getRawBlockHeaderByHash`, `rsk_getRawBlockHeaderByNumber`, `rsk_getRawTransactionReceiptByHash`, `rsk_getTransactionReceiptNodesByHash`, `rsk_protocolVersion`, `txpool_content`, `txpool_inspect`, `txpool_status`, `eth_accounts`, `eth_bridgeState`, `rpc_modules`, `eth_getBlocksByNumber`.
- `mnr_submitWork.json` deleted. It documented a name that has never existed in any Java source, and the `eth_submitWork` it appears to describe throws `UnsupportedOperationException` on every call — exactly like `eth_submitHashrate` beside it in `Web3Impl`, which this reference allowlists for that reason. It is allowlisted rather than re-homed.
- The six `sco_` fragments and `mnr_getWork.json` removed. All the deletions sit in the first of the two commits, so the hand-written prose stays findable with `git log --diff-filter=D -p -- doc/rpc/methods/` — which is what every affected allowlist entry points at, rather than a commit hash that a rebase would invalidate.
- The eleven component schemas the removals leave unreferenced, dropped. Reachability was computed transitively, not eyeballed: six were orphaned directly, `IPAddress`/`NodeId` only on a second pass through intra-component `$ref`s, and `PowHash`/`MixHash`/`Nonce` only once the `eth_submitWork` fragment went. That last three were missed by hand twice, which is why the orphan direction is now a test rather than a review step.
- `eth_getBalance`'s summary, which was truncated mid-word at `"…0 if not fo"`.

**The sync.** A `genRpcDocs` Gradle `JavaExec` task assembles the document over `sourceSets.main.runtimeClasspath` — it needs only `compileJava` rather than shading every dependency, and it doubles as the local one-liner the tool has never had. `devportal-update.yml` now triggers on `doc/rpc/**`, sets up Java, generates the document with the version it already extracts, copies it to `static/openrpc/rskj-openrpc.json`, stages that path, and no longer fails when a run has nothing to commit.

**Two holes on the publish path this change creates**, both previously absorbed by the human who carried the file across by hand:

- `GenerateOpenRpcDoc` merged fragments without validating references and exited **zero** on a document whose `$ref`s dangled. A typo would have auto-published. The new test assembles the document and fails on any reference that resolves to nothing.
- Nothing checked the *other* direction — a component no method reaches survived silently, and the acceptance criterion covering it was missed by hand in two consecutive rounds. A second test now walks defined→referenced, with the four components that were already unreferenced before this change listed explicitly so the next orphan fails the build.
- The generator parsed every non-directory file under `methods/` while the guard looked only at `*.json`, so a `README.md`, a `.orig` from a conflicted merge or an editor swapfile was invisible to CI and **fatal** to the publish. The generator now skips non-json files, which also protects the `components/` loaders.

No new language, no new build dependency, and no generated artifact is committed.

## Motivation and Context

A developer who adds or removes an RPC method currently gets no signal that the reference is now wrong, and the reference has drifted accordingly: over half the dispatch surface is undocumented, one fragment documents a method that has never existed under that name, and one summary has been truncated mid-sentence in the published document for as long as it has been there. Methods deliberately left out were indistinguishable from methods nobody had got to yet.

The published document also had to be carried to the docs site by hand, so it was only ever as fresh as the last person who remembered — and that same person was the only thing standing between a broken document and the docs site.

Three consequences worth calling out for review, since none is mechanical:

- Removing the six `sco_` fragments and `mnr_getWork.json` deletes hand-written prose and takes `mnr_getWork` out of the published reference. They are recoverable with `git log --diff-filter=D -p -- doc/rpc/methods/`, which every affected allowlist entry names.
- `eth_getBlocksByNumber` is documented rather than allowlisted because the governing rule puts it there: it is a real working method, `eth` is `enabled: "true"` with no `methods.disabled`, and nothing else in the reference exposes non-canonical blocks at a height. If the preference is to keep it out, it moves to the allowlist and the split becomes 53/52.
- Nothing here prevents the *next* truncated summary. The only mechanical signal that would have caught this one is requiring terminal punctuation, and 22 of the 50 existing fragments lack it — three of them legitimately. That was considered and declined; this stays a coverage guard and nothing else.

## How Has This Been Tested?

- `./gradlew genRpcDocs` produces a document whose `info.version` is the real version — `9.1.0-SNAPSHOT` by default, or the value passed via `-PrpcDocVersion` — holding exactly 54 methods, 46 schemas and 15 content descriptors, with no unresolved `$ref` and no component left unreferenced beyond the four pre-existing ones.
- `CliToolsTest` is green, so the generator's own fixture is unaffected by both the repo-root changes and the new file filter.
- `./gradlew --no-daemon --stacktrace build -x test` (the line CI runs), plus `spotlessJavaCheck`, `checkstyleMain` and `checkstyleTest` ratcheted from `origin/master`, are all green. CI runs `gradlew test`, so the guard gates the build.

Not exercised here: the live devportal sync. `devportal-update.yml` triggers on push to `master` and needs org secrets that no branch or fork build can read, so the first real run can only be observed after this merges. Adding `workflow_dispatch` does not change that — GitHub surfaces a dispatch trigger only once the workflow file is on the default branch — but it does let the sync be re-run on demand afterwards.

## Types of changes

## Checklist:


owans (approved):
LGTM! Awesome updated, prevents future drift.

owans:
One question @italo-sampaio Does this update include realistic request and response examples for each method payload?

italo-sampaio:
> One question @italo-sampaio Does this update include realistic request and response examples for each method payload?

Hey @owans  Yes. All 54 documented methods carry an example, and every example pairs the request values with the matching response.

Here are three of them, written out as the call a developer makes and the reply the node sends.

**`eth_getBalance`** — how much RBTC an address holds:

```json
// request
{"jsonrpc": "2.0", "id": 1, "method": "eth_getBalance",
 "params": ["0xc94770007dda54cF92009BFF0dE90c06F603a09f", "latest"]}

// response
{"jsonrpc": "2.0", "id": 1, "result": "0x0234c8a3397aab58"}
```

**`txpool_status`** — how many transactions are waiting to enter a block:

```json
// request
{"jsonrpc": "2.0", "id": 1, "method": "txpool_status", "params": []}

// response
{"jsonrpc": "2.0", "id": 1, "result": {"pending": 3, "queued": 1}}
```

**`rpc_modules`** — which groups of methods this node answers:

```json
// request
{"jsonrpc": "2.0", "id": 1, "method": "rpc_modules", "params": []}

// response
{"jsonrpc": "2.0", "id": 1,
 "result": {"eth": "1.0", "net": "1.0", "rpc": "1.0", "web3": "1.0",
            "evm": "1.0", "personal": "1.0", "txpool": "1.0", "rsk": "1.0"}}
```

The examples sit beside the reference itself, one file per method under `doc/rpc/methods/`. The snippets above add the standard wrapper that every JSON-RPC call carries, and the values come straight from those files.

Two details worth mentioning.

Some methods carry two examples, because the two outcomes look different. `eth_getTransactionCount` shows one example for an address that has sent transactions and one for an address that has not.

Where a response is long, such as a full block or a list of logs, the example points at a shared sample response instead of repeating the same block in twenty places. The reader still sees a complete response.

owans:
> > One question @italo-sampaio Does this update include realistic request and response examples for each method payload?
> 
> Hey @owans Yes. All 54 documented methods carry an example, and every example pairs the request values with the matching response.
> 
> Here are three of them, written out as the call a developer makes and the reply the node sends.
> 
> **`eth_getBalance`** — how much RBTC an address holds:
> 
> ```json
> // request
> {"jsonrpc": "2.0", "id": 1, "method": "eth_getBalance",
>  "params": ["0xc94770007dda54cF92009BFF0dE90c06F603a09f", "latest"]}
> 
> // response
> {"jsonrpc": "2.0", "id": 1, "result": "0x0234c8a3397aab58"}
> ```
> 
> **`txpool_status`** — how many transactions are waiting to enter a block:
> 
> ```json
> // request
> {"jsonrpc": "2.0", "id": 1, "method": "txpool_status", "params": []}
> 
> // response
> {"jsonrpc": "2.0", "id": 1, "result": {"pending": 3, "queued": 1}}
> ```
> 
> **`rpc_modules`** — which groups of methods this node answers:
> 
> ```json
> // request
> {"jsonrpc": "2.0", "id": 1, "method": "rpc_modules", "params": []}
> 
> // response
> {"jsonrpc": "2.0", "id": 1,
>  "result": {"eth": "1.0", "net": "1.0", "rpc": "1.0", "web3": "1.0",
>             "evm": "1.0", "personal": "1.0", "txpool": "1.0", "rsk": "1.0"}}
> ```
> 
> The examples sit beside the reference itself, one file per method under `doc/rpc/methods/`. The snippets above add the standard wrapper that every JSON-RPC call carries, and the values come straight from those files.
> 
> Two details worth mentioning.
> 
> Some methods carry two examples, because the two outcomes look different. `eth_getTransactionCount` shows one example for an address that has sent transactions and one for an address that has not.
> 
> Where a response is long, such as a full block or a list of logs, the example points at a shared sample response instead of repeating the same block in twenty places. The reader still sees a complete response.

Noted, thank you @italo-sampaio

rskj-core/src/main/java/co/rsk/cli/tools/GenerateOpenRpcDoc.java
```diff
@@ -152,6 +152,9 @@ private List<Object> loadFilesInPathAsJson(String basePath, String dirName, Java
         try (Stream<Path> stream = Files.walk(Paths.get(parentPath), 1)) {
```
meldsun0, line 151: Should we sort the discovered JSON files before loading them in GenerateOpenRpcDoc?
The test does this already, but Files.walk() here does not appear to guarantee a stable ordering.
Since the generated file is automatically synced to devportal, deterministic ordering would help avoid noisy diffs.

rskj-core/src/test/java/co/rsk/rpc/JsonRpcDocCoverageTest.java
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    /**
+     * Every JSON-RPC method the node dispatches, deduplicated across overloads.
+     *
+     * <p>Reflects over the implementation, not the {@link Web3} interface — see the class Javadoc. If this
+     * ever needs changing, confirm the new class against the {@code remoteInterface} argument at the
+     * {@code JsonRpcCustomServer} construction site rather than picking the type that reads best.
+     */
+    private static Set<String> exposedMethods() {
```
meldsun0, line 658: Could we point specifically to JsonRpcCustomServer.extractMethodNames() here, and add a reciprocal comment there pointing back to this guard?
These two places intentionally mirror the same getMethods() -> Method::getName discovery.
That cross-reference could help prevent them from drifting if the discovery logic changes later.

meldsun0 (dismissed):
Since the guard intentionally doesn't read reference.conf, could we leave a note near the RPC module configuration to review undocumented.json when availability changes? Otherwise an exclusion reason like debug being disabled could become stale while coverage still passes.

.github/workflows/devportal-update.yml
```diff
@@ -68,14 +83,24 @@ jobs:
...
           git config user.email "github-actions[bot]@users.noreply.github.com"
-          git add docs
-          git commit -m "Automated update from rskj repository"
+          git add docs static/openrpc
+          # A run whose inputs produced no change must not fail the job on `git commit`.
+          if git diff --cached --quiet; then
+            echo "Nothing to commit: devportal already matches this rskj revision."
+          else
+            git commit -m "Automated update from rskj repository"
+          fi
```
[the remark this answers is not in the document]
fmacleal: This is a good point that worth to take a look at.

doc/rpc/public_api_exclusions.json
fmacleal, line 1: **question**:

This is a weird name, why is it called "undocumented.json"? Can you change to methods explicitly not documented and not public?

Maybe `private_methods_undocumented.json`?

.github/workflows/devportal-update.yml
```diff
@@ -68,14 +83,24 @@ jobs:
...
         run: |
           cd devportal
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
-          git add docs
-          git commit -m "Automated update from rskj repository"
+          git add docs static/openrpc
+          # A run whose inputs produced no change must not fail the job on `git commit`.
+          if git diff --cached --quiet; then
+            echo "Nothing to commit: devportal already matches this rskj revision."
```
fmacleal: **suggestions**:

This is good, but isn't being applied in the later steps. Maybe this step here could set a variable.

```yaml
          if git diff --cached --quiet; then
            echo "Nothing to commit: devportal already matches this rskj revision."
            echo "committed=false" >> $GITHUB_OUTPUT
          else
            git commit -m "Automated update from rskj repository"
            echo "committed=true" >> $GITHUB_OUTPUT
          fi
```

With that, you could check this variable in the next steps to skip them using`if: steps.<id>.outputs.committed == 'true'` on the push and PR steps.

.github/workflows/devportal-update.yml
fmacleal: **suggestion**:

Since you might change this to run only if we have changes, we could check the status of this call to print success only if something was created in the line 123. To make it stop to print `Pull request created successfully` in any situation , even if the call fails.

rskj-core/src/main/java/co/rsk/cli/tools/GenerateOpenRpcDoc.java
fmacleal: Good catch, could be good.

rskj-core/src/main/java/co/rsk/cli/tools/GenerateOpenRpcDoc.java
```diff
@@ -162,6 +165,10 @@ private List<Object> loadFilesInPathAsJson(String basePath, String dirName, Java
         }
     }
 
+    private static boolean isJson(Path file) {
+        return file.getFileName().toString().endsWith(".json");
+    }
```
fmacleal: **suggestion**:

Isn't a better way to check that it's a json? What about `.JSON`? This can drift fast. We have the Json5, for example, that it's a new json format that it's also Json. 

This is also a problem in the test `JsonRpcDocCoverageTest.java` that has same method. 

Honestly, I know that we generate the file and that decoding each one might be a too much. Probably, it's better just improve this regex to be case insensitive for now. It's an easy win.

fmacleal (commented):
Good job on the automation, very nice!

I think it improved a lot and prevents the documentation to drift, which is amazing. And automatically updates the devportal, which is amazing.

I have a few comments though from methods that are not being updated, mainly the ones that accept reference for blocks. I will send them offline, because I cant point here in this PR.

rskj-core/src/test/java/co/rsk/rpc/JsonRpcDocCoverageTest.java
```diff
@@ -0,0 +1,777 @@
...
+     *
+     * <p>Keeping it explicit rather than accepting either schema everywhere is what preserves the upper
+     * bound. Were any block-reference schema acceptable for any method of the type, a reading method could be
+     * repointed at the execution variant -- promising a decimal height its retriever rejects -- and nothing
+     * here would notice. An entry naming a method that does not take the mapped parameter type fails, for the
+     * same reason a mapped type matching no method does.
+     */
+    private static final Map<String, String> WIDER_BLOCK_FORM_BY_METHOD = Map.of(
+            "eth_call", "#/components/schemas/ExecutionBlockRef");
+
```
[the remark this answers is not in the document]
italo-sampaio: addressed.

.github/workflows/devportal-update.yml
```diff
@@ -31,11 +39,24 @@ jobs:
...
+        uses: actions/setup-java@1bcf9fb12cf4aa7d266a90ae39939e61372fe520 #v5.4.0
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'gradle'
+
+      - name: Extract version from version.properties
         run: |
           VERSION=$(tr -d "'\"" < rskj-core/src/main/resources/version.properties | cut -d = -f 2- | paste -sd - -)
           echo "RSK_VERSION=${VERSION}" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
italo-sampaio: Not true. `version.properties` has two non-empty lines and no trailing blank line, so `paste -sd - -` joins two fields and yields `9.1.0-SNAPSHOT`. The same expression already builds the jar filename in `build_and_test.yml`; a trailing dash would break that job.

The expression is pre-existing and unchanged here — this PR only renamed the step.

doc/rpc/components/schemas/BlockRef.json
```diff
@@ -2,14 +2,35 @@
...
         "$ref": "#/components/schemas/BlockHash"
       },
       "blockNumber": {
         "$ref": "#/components/schemas/BlockNumberHex"
+      },
+      "requireCanonical": {
+        "title": "requireCanonical",
+        "type": "string",
+        "description": "Whether a block that is not on the canonical chain is acceptable. Send it as the JSON string `\"true\"` or `\"false\"`, not as a JSON boolean: the node reads the object as a map of strings and compares this value against those two words (case-insensitively), so any other value is rejected with `Invalid input: requireCanonical must be a String \"true\" or \"false\"`. It is only consulted alongside `blockHash`; when it is `\"true\"` and the referenced block is not on the canonical chain, the node answers with a block-not-found error stating the block is not canonical.",
+        "enum": [
```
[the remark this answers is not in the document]
italo-sampaio: Both were fixed in `6ea41e6cd0`. Lowercase is now stated as the form to send, so the enum and description agree. The error promise was narrowed to string values, with a note that a JSON boolean fails earlier during deserialisation.

fmacleal (dismissed):
Thank you for having addressed the comments.

LGTM

rskj-core/src/main/java/co/rsk/cli/tools/GenerateOpenRpcDoc.java
italo-sampaio: done in [5073a8806b](https://github.com/rsksmart/rskj/pull/3670/changes/5073a8806b5d4f16ccd21e20cb18945ced5896fe). Adds `.sorted()` to the walk in `loadFilesInPathAsJson` Same commit makes the extension match case-insensitive, so a `.JSON` fragment sorts in rather than being dropped.

rskj-core/src/test/java/co/rsk/rpc/JsonRpcDocCoverageTest.java
italo-sampaio: Both done. The guard's class javadoc already named `JsonRpcCustomServer.extractMethodNames` as what it mirrors; `ae988ceedf` adds the return pointer on `extractMethodNames` itself.

doc/rpc/public_api_exclusions.json
italo-sampaio: Changed to `public_api_exclusions.json` (see https://github.com/rsksmart/rskj/pull/3670/changes/9918a99f587245fa13673d10f056611fd3be69b7)

doc/rpc/methods/txpool_content.json
```diff
@@ -0,0 +1,53 @@
...
+  },
+  "examples": [
+    {
+      "name": "contentExample",
+      "params": [],
+      "result": {
+        "name": "contentResult",
+        "value": {
+          "pending": {
+            "0x7986b3df570230288501eea3d890bd66948c9b79": {
```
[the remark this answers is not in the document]
italo-sampaio: Fixed in `bc31be9ee9` — the example key is bare hex now, matching `RskAddress.toString()`.
The description says the sender appears twice, bare as the map key and `0x`-prefixed as `from`, and to key off the prefixed form.
The serializer change is the better fix, but it alters the response shape — filing it as a follow-up.

doc/rpc/methods/txpool_inspect.json
```diff
@@ -0,0 +1,41 @@
...
+  },
+  "examples": [
+    {
+      "name": "inspectExample",
+      "params": [],
+      "result": {
+        "name": "inspectResult",
+        "value": {
+          "pending": {
+            "0x7986b3df570230288501eea3d890bd66948c9b79": {
```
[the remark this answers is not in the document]
italo-sampaio: Same commit. Both addresses are bare: the map key and the recipient at the head of the summary string.
The description also notes the value and gas price are decimal wei rather than hex.

doc/rpc/components/schemas/TxPoolTransaction.json
```diff
@@ -0,0 +1,42 @@
...
+      "hash": {
+        "$ref": "#/components/schemas/TransactionHash"
+      },
+      "input": {
+        "$ref": "#/components/schemas/Bytes"
+      },
+      "nonce": {
+        "$ref": "#/components/schemas/Nonce"
+      },
+      "to": {
```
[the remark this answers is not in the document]
italo-sampaio: Documented the sentinel in `bc31be9ee9`: `to` now points at `PoolDestinationAddress`, a `oneOf` of `Address` and an enum of `0x00`.
It records that `eth_getTransactionByHash` reports the same transaction's `to` as `null`.
Emitting `null` here is the cleaner answer but changes the response shape — follow-up.

doc/rpc/components/schemas/BlockNumberOrTag.json
```diff
@@ -0,0 +1,14 @@
+{
+  "BlockNumberOrTag": {
+    "title": "blockNumberOrTag",
+    "description": "A block, given as its height in hex or as a tag. A height in decimal is rejected: the block is looked up by `Web3InformationRetriever`, which parses the number as hex and requires the `0x` prefix.",
```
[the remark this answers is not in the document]
italo-sampaio: Took the second option in `6ea41e6cd0` — the description states the hash form is accepted and deliberately outside the union.
The reason is given inline: the node can't distinguish a 32-byte hex hash from a height, so it's a side effect of parsing, and publishing it would make it breaking to narrow later.

doc/rpc/components/schemas/ExecutionBlockNumberOrTag.json
```diff
@@ -0,0 +1,17 @@
+{
+  "ExecutionBlockNumberOrTag": {
+    "title": "executionBlockNumberOrTag",
+    "description": "A block to execute against, given as its height in hex or in decimal, or as a tag. The decimal form is accepted here and nowhere else: the block is resolved by `ExecutionBlockRetriever`, which parses both bases, while the block-reading methods parse hex only.",
```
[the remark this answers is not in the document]
italo-sampaio: Corrected in `bc31be9ee9` — the description names `eth_getBlocksByNumber` as also taking decimal and narrows the claim to methods that take a block parameter.
The test message was corrected in the same commit to "of the methods mapped here, `eth_call` and `eth_estimateGas`".

doc/rpc/methods/eth_getStorageAt.json
```diff
@@ -9,7 +9,7 @@
       "$ref": "#/components/contentDescriptors/Position"
     },
     {
-      "$ref": "#/components/contentDescriptors/BlockNumberOrTag"
+      "$ref": "#/components/contentDescriptors/BlockRefOrNumberOrTag"
```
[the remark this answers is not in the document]
italo-sampaio: Reworded in `bc31be9ee9` to "…a position and a block."
The parameter also moved to `BlockRefOrNumberOrTag` in `42ab48364a`, so the object form is documented rather than left implied.

.github/workflows/devportal-update.yml
italo-sampaio: Done as suggested. The Commit step sets `changed=true|false`, and both Push and Create Pull Request carry `if: steps.commit.outputs.changed == 'true'`.
A run with nothing to publish is now a clean skip rather than a force-push and a rejected PR call.

.github/workflows/devportal-update.yml
italo-sampaio: Done. Success prints only after the API confirms a pull request carrying this commit — HTTP 201 with a URL, or an existing PR on the branch the force-push just updated.
Every other response fails the step instead of logging success.

rskj-core/src/main/java/co/rsk/cli/tools/GenerateOpenRpcDoc.java
italo-sampaio: Took the easy win in `5073a8806b` — the extension match is case-insensitive, so a `.JSON` fragment is a fragment.
json5 is deliberately not accepted: it parses under different rules than the mapper here, so treating it as a fragment would fail at load rather than at the filter. That's written into the javadoc.
The duplicate is gone too — the test now calls `isFragment` rather than keeping its own copy.

.github/workflows/devportal-update.yml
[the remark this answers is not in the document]
italo-sampaio: Agreed, and it's implemented — see the two threads below. The no-change path skips push and PR creation, and success is only claimed once the API confirms one.

fmacleal (approved):
Thank you for having addressed the comments, LGTM.

PR #3672, opened 2026-08-07
Reject contract-originated calls on federation-only Bridge methods

jeremy-then:
## Description

The Bridge exposes a set of methods that are restricted to federation members through two decorators: `Bridge::activeAndRetiringFederationOnly` and `Bridge::activeRetiringAndProposedFederationOnly`. These guard the following methods:

- `updateCollections` — always federation-only
- `addSignature` — always federation-only
- `receiveHeaders` — federation-only branch (when not public)
- `registerBtcTransaction` — federation-only branch (when not public)

This change adds a `BridgeUtils.isContractTx(rskTx)` check inside both decorators. When the call originates from a contract (i.e. the transaction is an `InternalTransaction`), it is rejected with a `VMException` before the federation-membership check runs. Since a federation member always interacts with the Bridge through a top-level, externally signed transaction, a contract-originated call to these methods is never legitimate.

The check is added inside the decorators (not on the `BridgeMethods` enum entries), so the *public* branches of `receiveHeaders` / `registerBtcTransaction` are intentionally left unaffected.

Implemented **without** an RSKIP activation gate, as an initial version pending security-team validation (see *Other information*).

## Motivation and Context

As part of the preparatory work for Account Abstraction (AA) on rskj, we are hardening the Bridge against a potential bypass: a pegnatory could craft a delegated/internal call to the Bridge to reach federation-only methods, potentially bypassing the requirement of having the HSM sign the transaction. Rejecting contract-originated calls on these methods closes that vector as defense-in-depth.

## How Has This Been Tested?

Added unit tests in `BridgeTest` (using a real `InternalTransaction`, since `isContractTx` compares the concrete class):

- `activeAndRetiringFederationOnly_calledFromContract_evenWhenSenderIsFederationMember_throwsVMException` — an internal tx whose sender is a **legitimate active federation member** is still rejected; the decorated method is never executed.
- `activeRetiringAndProposedFederationOnly_calledFromContract_evenWhenSenderIsFederationMember_throwsVMException` — same guarantee for the proposed-federation decorator.
- `activeRetiringAndProposedFederationOnly_calledFromExternalTxByFederationMember_OK` — happy path: an externally signed tx from a federation member still passes.

All existing `activeAndRetiringFederationOnly*`, `updateCollections*`, `addSignature*`, `receiveHeaders*`, and `registerBtcTransaction*` tests continue to pass (0 failures). Verified that no existing integration/performance test invokes these four methods through the Bridge via an internal transaction.

## Types of changes

## Checklist:

* **Other information**:

This is a consensus-relevant area (the Bridge runs on all nodes), so the "Requires Activation Code" box is left unchecked pending security-team review. Analysis:
- A plain `CALL` from a contract already fails today — the internal tx sender is the calling contract's address, never a federation member, so `isFromFederateMember` already rejects it. `isContractTx` rejects nothing new on this path.
- `DELEGATECALL` / `CALLCODE` are already blocked for these methods post-**RSKIP417** by the call-type restriction (all four are `RESTRICTED_TO_CALL`).

On current mainnet the change is therefore most likely a behavioral no-op, which is the argument for shipping ungated. However, any newly-added rejection that could — via an unproven path — reject a previously-accepted transaction would fork upgraded vs. non-upgraded nodes. **Recommendation:** confirm with the security team that no historical/edge path lets a contract-originated tx currently pass the federation check; if there is any doubt, gate this behind an RSKIP activation. The `rejectIfCalledFromContract` helper is trivially wrappable in an `activations.isActive(...)` check if gating is decided.

### DIFF
--- a/rskj-core/src/test/java/co/rsk/peg/BridgeTest.java
+++ b/rskj-core/src/test/java/co/rsk/peg/BridgeTest.java
@@ -46,6 +46,7 @@ import org.ethereum.util.ByteUtil;
 import org.ethereum.vm.DataWord;
 import org.ethereum.vm.MessageCall;
 import org.ethereum.vm.exception.VMException;
+import org.ethereum.vm.program.InternalTransaction;
 import org.junit.jupiter.api.*;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.*;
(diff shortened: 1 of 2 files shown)


rskj-core/src/test/java/co/rsk/peg/BridgeTest.java
```diff
@@ -1014,6 +1015,130 @@ void activeAndRetiringFederationOnly_activeFederationIsNotFromFederateMember_ret
...
+        );
+
+        BridgeSupport bridgeSupportMock = mock(BridgeSupport.class);
+
+        // Even though the sender is a legitimate active federation member, the call must be
+        // rejected because it originates from a contract (e.g. through a delegate call)
+        // rather than an externally signed transaction.
+        Transaction contractTx = createContractTxFromFederationMember();
+
+        ActivationConfig activationConfig = ActivationConfigsForTest.papyrus200();
```
marcos-iov: Why papyrus? Shouldn't it be better to use all activations here?

rskj-core/src/test/java/co/rsk/peg/BridgeTest.java
```diff
@@ -1014,6 +1015,130 @@ void activeAndRetiringFederationOnly_activeFederationIsNotFromFederateMember_ret
...
 
+    @Test
+    void activeAndRetiringFederationOnly_calledFromContract_evenWhenSenderIsFederationMember_throwsVMException() throws Exception {
+        // Given
+        BridgeMethods.BridgeMethodExecutor decorate = mock(
+            BridgeMethods.BridgeMethodExecutor.class
+        );
+        BridgeMethods.BridgeMethodExecutor executor = Bridge.activeAndRetiringFederationOnly(
+            decorate,
+            "updateCollections"
```
marcos-iov: Perhaps we could parameterize this test so that we validate all the methods that only accepts calls from fed members and not just updateCollections

rskj-core/src/test/java/co/rsk/peg/BridgeTest.java
```diff
@@ -1014,6 +1015,130 @@ void activeAndRetiringFederationOnly_activeFederationIsNotFromFederateMember_ret
...
+            "addSignature"
+        );
+
+        BridgeSupport bridgeSupportMock = mock(BridgeSupport.class);
+
+        // Sender is a legitimate active federation member, but the call originates from a
+        // contract (internal transaction), so it must be rejected all the same.
+        Transaction contractTx = createContractTxFromFederationMember();
+
+        ActivationConfig activationConfig = ActivationConfigsForTest.papyrus200();
```
marcos-iov: Same as above

rskj-core/src/test/java/co/rsk/peg/BridgeTest.java
```diff
@@ -1014,6 +1015,130 @@ void activeAndRetiringFederationOnly_activeFederationIsNotFromFederateMember_ret
...
+        assertThrows(VMException.class, () -> executor.execute(bridge, null));
+
+        verify(bridgeSupportMock, never()).getActiveFederation();
+        verify(bridgeSupportMock, never()).getRetiringFederation();
+        verify(bridgeSupportMock, never()).getProposedFederation();
+        verify(decorate, never()).execute(any(), any());
+    }
+
+    @Test
+    void activeRetiringAndProposedFederationOnly_calledFromExternalTxByFederationMember_OK() throws Exception {
```
marcos-iov: Do we need the same for `activeAndRetiringFederationOnly`? Could be a parameterized test, showing that all the methods that are only callable from fed members still accept external calls after this change

rskj-core/src/test/java/co/rsk/peg/BridgeTest.java
jeremy-then: Sure. Fixed: d5575867e88ced64bbb315ee400dd98ac3cd2d9c

rskj-core/src/test/java/co/rsk/peg/BridgeTest.java
jeremy-then: Done: d5575867e88ced64bbb315ee400dd98ac3cd2d9c

rskj-core/src/test/java/co/rsk/peg/BridgeTest.java
jeremy-then: Done: d5575867e88ced64bbb315ee400dd98ac3cd2d9c

rskj-core/src/test/java/co/rsk/peg/BridgeTest.java
jeremy-then: Done: d5575867e88ced64bbb315ee400dd98ac3cd2d9c

marcos-iov (approved):

PR #3676, opened 2026-08-17
docs(node-operators): document import sync, and align the pages that reference --import

italo-sampaio:
## Description

Adds a node-operator page for **import sync**, and updates the pages that already
referred to the `--import` flag so they agree with it.

- `docs/03-node-operators/04-setup/import-sync.md` — new page (JAR, Docker and the
  Ubuntu package; Mainnet and Testnet).
- `doc_config.yaml` — registers it for the devportal sync, `sidebar_position: 4.5`,
  directly after *Node Installation*.
- `02-installation/java.md`, `03-configuration/02-cli.md` — the two places that
  mentioned `--import`: both now carry the erase-on-every-start warning, the
  `VETIVER-9.0.4` floor and a link onward. `java.md` gains a five-step summary and
  sets `-Xmx4G` on the command instead of offering it as a remedy after an OOM.
- `03-configuration/preferences.md`, `03-reference.md`, `04-node-runner/*.md`,
  `09-troubleshooting/index.md` — LevelDB is deprecated and the node logs so on
  every start, but these pages still described switching *to* it. They now advise
  migrating to RocksDB, via `DbMigrate` or import sync.

No functional change to the node.

## Motivation and Context

Import sync was documented as two commands on the Java page, and the behaviour that
matters most was not discoverable from them:

- **It erases the database on every start.** Not "import if needed" — a node with
  `database.import.enabled = true` in a config file wipes and re-downloads on every
  restart, including a reboot or a crash recovery. Hence the one-off command-line
  procedure, and the per-platform variants for Docker and `systemd`, where the node
  restarts itself.
- **It does not leave you at the chain tip**, and the newest entry in any single
  publisher's index is not necessarily the height you get. The page reproduces the
  node's own quorum rule as a `jq` query, so the landing height and the download
  size can be checked before committing to a download.

Worth a reviewer's attention: merging to `master` touches `docs/**`, so the
*Update DevPortal Documentation* workflow will open the devportal PR automatically.

## How Has This Been Tested?

Documentation only, so the testing is of the published output and of the commands
the page tells operators to run.

- **Rendering.** Ran `process_docs.py` over `doc_config.yaml` and served the result
  from a clean checkout of devportal `main`. Every changed page renders; all
  internal anchors resolve, including the three new cross-page targets.
- **The page's own commands**, against the live published indexes: the quorum query
  returns `9198000` (Mainnet) and `8028000` (Testnet), two agreeing signers each;
  the temporary-directory snippet finds a planted leftover on Linux, where the
  `${TMPDIR:-/tmp}` form does not.
- **Docker, end to end.** Testnet import in a container completed in 277 s; a plain
  container started against the same volume served the imported height and synced
  on. This is also how the volume-ownership step in that section was found — without
  it the import dies as the unprivileged `rsk` user.
- **The Ubuntu package** layout, unit file and maintainer scripts were read from the
  installed `9.0.4~jammy` package, not assumed.
- **Failure messages** were taken from the node's source, including which are fatal
  and which are startup warnings that let the run continue and fail later.

## Types of changes

## Checklist:

* **Other information**: Documentation only — no source, build or configuration
change to the node, so no unit tests accompany it. One fix here predates the branch:
the `DbMigrate` argument list in `09-troubleshooting/index.md` carried a stray
`> >`, which the live site still serves.


.github/workflows/scripts/devportal-update/doc_config.yaml
```diff
@@ -31,6 +31,14 @@ files:
     tags: [rsk, rskj, node, node operators, experimental features]
     description: "All the experimental features released in RSKj  are explained in this section, how to enable and what's the purpose of the feature"
 
+  - input: docs/03-node-operators/04-setup/import-sync.md
```
fmacleal, line 34: **suggestion**:

I am in a bit of doubt yet if this is the right place for us to add the information. 

Right now, the mention for the CLI commands are in two places:
1. https://github.com/rsksmart/devportal/blob/main/docs/03-node-operators/04-setup/03-configuration/02-cli.md

Which is the document published here: https://dev.rootstock.io/node-operators/setup/configuration/cli/

2. https://github.com/rsksmart/devportal/blob/main/docs/03-node-operators/04-setup/02-installation/java.md 

Which is published here: https://dev.rootstock.io/node-operators/setup/installation/java/


We should migrate both pages and change the part that refers to the import flag. Probably both pages should contain a banner stating to use the VETIVER-9.0.4 and a reference to the more detailed explanation you are adding. We should also remove the support for `leveldb` there.

Maybe, the more detailed explanation should be under the part https://github.com/rsksmart/devportal/tree/main/docs/03-node-operators/04-setup/02-installation. 

Here comes the question, isn't the import sync flag also useful for people using docker?

docs/03-node-operators/04-setup/import-sync.md
```diff
@@ -0,0 +1,242 @@
...
+| `The file is corrupted or incomplete. Please start again the import process` | The archive downloaded and matched its hash, but could not be unpacked. Remove the temporary files and retry; if it persists, the published archive is faulty — report it rather than working around it. |
+| `Error trying to read bootstrap data contents. Please start again the import process` | The unpacked data could not be read back. Check free space in the temporary directory, then retry. |
+| `Configuration has less trusted sources than the minimum required <n> of 2` | Fewer than two trusted keys are configured. At least two independent signers are always required. Restore the shipped keys for the network. This is a warning at startup, not the failure itself — the run continues and then fails on one of the messages above. |
+| `java.lang.OutOfMemoryError` | The load stage ran out of heap. Raise `-Xmx` and run it again; note that a retry downloads the bootstrap data again. |
+
+## Switching between LevelDB and RocksDB
```
fmacleal: **suggestion**:

We shouldn't support leveldb anymore, due to that, this section could state this more clearly. Stating that if you are using a leveldb database, we strongly advise to migrate to rocksdb and you could use the import feature to help with that.

docs/03-node-operators/04-setup/import-sync.md
fmacleal, line 1: **suggestion**:

I liked this page, but I think it's too much detailed in an advanced mode from the bootstrap. I think it brings value, because it explains in more detail how we generate the snapshots and how reliable it is. But, we could have a less detailed step by step, like TLDR in the java part [here](https://github.com/rsksmart/devportal/blob/main/docs/03-node-operators/04-setup/02-installation/java.md) that now mention the flag `--import`. And in that page, you could direct to this more detailed page here.  

I also think that we could re-write the section from this page to be clear in the steps. I know that each section it's a step, but I think it could be useful to have bullet/numbered step list like:

1. Ensure you have enough disk space to download the snapshot and run the full node
2. Go to the snapshot page and download the snapshot from the network you want to sync (`testnet` or `mainnet`)
3. Run the node with the --import flag pointing to the snapshot file
4. Verify the node is working and syncing running the command blah blah blah
5. Stop the node and run again without the flag
6. Done, you finished the import
7. You can replicate and adapt these steps if you are running the node with docker or ubuntu, making the node use the path from the snapshot visible and running them with the import parameter in the same way you ran using java

Some people are just lazy to read all the page and figure it out the details, I thin if we have a section like TLDR and then expand each step with more detail later, it's better.

docs/03-node-operators/04-setup/import-sync.md
```diff
@@ -0,0 +1,242 @@
...
+
+RSKj leaves the downloaded archive and its extracted contents behind in the temporary directory — together, more than twice the size of the archive. They are not needed once the import has completed.
+
+```shell
+ls -d /tmp/import*
+```
+
+Remove anything left there.
+
+## Common failures
```
fmacleal, line 226: **nitpick**:

Make these failures here like an expandable section.

docs/03-node-operators/04-setup/import-sync.md
```diff
@@ -0,0 +1,242 @@
...
+| `File: <path> does not match with expected hash: <hash>` | The downloaded archive does not match the hash the signers committed to. Usually an incomplete or corrupted download; remove the temporary files and retry. |
+| `Error downloading bootstrap data from <url>. Please start again the import process` | The archive could not be retrieved. Retry; if it persists, the published archive may be missing from the location its index advertises. |
+| `The file is corrupted or incomplete. Please start again the import process` | The archive downloaded and matched its hash, but could not be unpacked. Remove the temporary files and retry; if it persists, the published archive is faulty — report it rather than working around it. |
+| `Error trying to read bootstrap data contents. Please start again the import process` | The unpacked data could not be read back. Check free space in the temporary directory, then retry. |
```
fmacleal: **suggestion**:

`Error trying to read bootstrap data contents` row describes the wrong failure.

The table says this message means "the unpacked data could not be read back". It does not. It is thrown from `checkFileHash`, while reading the **downloaded zip** to compute its SHA-256 — before anything is extracted:

`rskj-core/src/main/java/co/rsk/db/importer/provider/BootstrapFileHandler.java:112-115`

```java
} catch (IOException e) {
    throw new BootstrapImportException(
            "Error trying to read bootstrap data contents. Please start again the import process", e);
}
```

The message for the case the row describes is a different one, and it is missing from the table entirely — `BootstrapFileHandler.java:67-68`:

```java
throw new BootstrapImportException(String.format(
        "Failed reading data after decompression %s", bootstrapPath), e);
```

<p>Suggested rows:</p>

Message | What it means
-- | --
Error trying to read bootstrap data contents. Please start again the import process | The downloaded archive could not be read while its hash was being checked. The download is likely incomplete or the temporary filesystem is failing. Remove the temporary files and retry.
Failed reading data after decompression <path> | The extracted data could not be read back. Check free space in the temporary directory, then retry.

docs/03-node-operators/04-setup/import-sync.md
```diff
@@ -0,0 +1,242 @@
...
+| Regtest | Not supported. No bootstrap data is published for Regtest, and because the database is erased before the import is attempted, running `--import` there deletes the Regtest database and then fails |
+
+## Before you start
+
+- **Java 17 JDK** and **RSKj `VETIVER-9.0.4` or later** — see [Setup node using Java](/node-operators/setup/installation/java/). Earlier versions may not be able to read the published bootstrap data.
+- **Disk for the database.** Size the data directory for a full node, per the [minimum requirements](/node-operators/setup/requirements/) — not for the size of the download.
+- **Disk for temporary files.** RSKj downloads the bootstrap archive and extracts it into the JVM's temporary directory (`/tmp` on Linux and macOS by default), and both exist there at the same time. The extracted contents are larger than the archive, so allow **about three times the size of the bootstrap archive** in temporary space, on top of the database itself. To check that size before committing to the download, see [Check where import sync will land you](#check-where-import-sync-will-land-you).
```
fmacleal: **nitpick**:

On macOs the tmp folder isn't actually on `/tmp`.  It's a per-user launchd path, not `/tmp`:

```
$ java -XshowSettings:properties -version 2>&1 | grep tmpdir
    java.io.tmpdir = /var/folders/1n/5xyy8pjd6v727ckkwcn6lvv80000gn/T/
```

Two consequences for a macOS operator: they size the wrong filesystem when following *Before you start*, and the cleanup command `ls -d /tmp/import*` at line 202 finds nothing, leaving ~2 GB stranded in `/var/folders`.

Suggested change at line 30: *(`/tmp` on Linux; on macOS a per-user directory under `/var/folders`, the value of `$TMPDIR`)*, and at line 202:

```shell
ls -d "${TMPDIR:-/tmp}"/import*
```

docs/03-node-operators/04-setup/import-sync.md
```diff
@@ -0,0 +1,242 @@
...
+  <TabItem value="index-testnet" label="Testnet">
+    ```text
+    https://import.testnet.rskcomputing.net/dbs/testnet/<TRUSTED-PUBLIC-KEY>/index.json
+    ```
+  </TabItem>
+</Tabs>
+
+Every entry in an index carries a block `height`, the path to the archive, its `hash`, and the signer's signature over that entry.
+
+RSKj will only import bootstrap data that **at least two trusted signers agree on**: the same height, with the same hash, correctly signed by each. A height offered by only one signer is ignored, however recent it is. Among the heights that meet that bar, RSKj takes the highest.
```
fmacleal: **suggestion**:

The rule it's actually, `majority of the configured keys, floored at 2`.

co/rsk/RskContext.java:292-295`:

```java
int minimumRequired = publicKeys.size() / 2 + 1;
if (minimumRequired < 2) {
    logger.warn("Configuration has less trusted sources than the minimum required {} of 2", minimumRequired);
    minimumRequired = 2;
}
```

With the three keys shipped for each network this evaluates to 2, so everything the page says is correct for a default installation — and the `jq`'s `>= 2` matches. But the page states it as an absolute ("at least two trusted signers agree on", "At least two independent signers are always required"), and line 212 explicitly contemplates an operator editing `database.import.trusted-keys`. Configure five keys and the node requires three; the page's query would then report an agreement the node rejects.

Suggested change at line 58, after the existing sentence: *The threshold is a majority of the keys you have configured, and never fewer than two — with the three keys shipped for each network, that is two. If you change `database.import.trusted-keys`, change the `>= 2` in the query below to match.*

docs/03-node-operators/04-setup/import-sync.md
```diff
@@ -0,0 +1,242 @@
...
+    ```shell
+    java -cp <PATH-TO-THE-RSKJ-JAR> co.rsk.Start --import
+    ```
+  </TabItem>
+  <TabItem value="import-testnet" label="Testnet">
+    ```shell
+    java -cp <PATH-TO-THE-RSKJ-JAR> co.rsk.Start --testnet --import
+    ```
+  </TabItem>
+</Tabs>
```
fmacleal, line 168: **suggestion**:

`java.md:88` currently recommends `-Xmx4G` for exactly this command. The new page mentions `-Xmx` only in the OOM row, after the fact — and by the page's own admission there, "a retry downloads the bootstrap data again", so on Mainnet that is another 705 MB.

This is not hypothetical for the data actually published. Both live archives are v1, and the v1 import path reads the whole `.bin` into a single `byte[]` and decodes it into in-memory queues; `rskj-core/src/main/java/co/rsk/db/importer/bootstrap-data-format.md` states that "peak memory scales with total state size" for v1. Worth putting `-Xmx4G` in the *Run the import* commands directly, the way `java.md` does, rather than leaving operators to discover it from the failure table.

docs/03-node-operators/04-setup/import-sync.md
```diff
@@ -0,0 +1,242 @@
...
+## Check where import sync will land you
+
+**Import sync does not leave you at the chain tip.** It leaves you at the newest height that enough signers agree on, and the node then syncs the remaining blocks from peers in the normal way. Depending on how recent the published data is, that remainder can still be substantial: import sync shortens the initial sync, it does not remove it.
+
+You can see exactly which height you would land on before downloading anything. Substitute the three trusted keys for your network, then:
+
+<Tabs>
+  <TabItem value="land-mainnet" label="Mainnet" default>
+    ```shell
+    URL=https://import.mainnet.rskcomputing.net/dbs/mainnet/
```
fmacleal, line 96: **suggestion**:

Add here a `mkdir` specifically for `mainnet` and below in the `testnet` part add another for that network. This way, the `index-*.json` command doesn't mix the networks for an operator that has both mainnet and testnet configured.

docs/03-node-operators/04-setup/import-sync.md
```diff
@@ -0,0 +1,242 @@
...
+
+```shell
+ls -d /tmp/import*
+```
+
+Remove anything left there.
+
+## Common failures
+
+| Message | What it means |
```
fmacleal, line 231: **suggestion**:

- **`Failed to create a temporary directory. Please start again the import process`** (`BootstrapFileHandler.java:132-133`) is missing from the table, which is a shame on a page that spends a bullet on sizing temporary space. It is the message you get when that sizing was wrong.

fmacleal (changes requested):
Good job in the documentation!

I think we need to migrate the other two pages that refer the import flag as well, since they touch this topic.

Also, we can't publish this before the snapshots doesn't finish, otherwise we will be lying. 😅 
Since it could take days instead of hours for someone that use the import flag, due to the absence of an updated and recent snapshot.

I have left some important comments, if you want, we can sync to discuss it better.

docs/03-node-operators/04-setup/import-sync.md
```diff
@@ -0,0 +1,324 @@
...
+  <TabItem value="index-testnet" label="Testnet">
+    ```text
+    https://import.testnet.rskcomputing.net/dbs/testnet/<TRUSTED-PUBLIC-KEY>/index.json
+    ```
+  </TabItem>
+</Tabs>
+
+Every entry in an index carries a block `height`, the path to the archive, its `hash`, and the signer's signature over that entry.
+
+RSKj will only import bootstrap data that **enough trusted signers agree on**: the same height, with the same hash, correctly signed by each. The threshold is a majority of the keys you have configured, and never fewer than two — with the three keys shipped for each network, that is two. A height offered by too few signers is ignored, however recent it is. Among the heights that meet that bar, RSKj takes the highest.
```
core-baguira, line 68: **One unreachable signer stops the import, even when the rest agree.**

RSKj downloads every configured signer's index before it compares anything. If one index fails to download, it throws right there. The quorum rule only decides who has to *agree*; it does not make a missing publisher optional.

I tested it. I served 2 of the 3 real Testnet indexes from a local server and let the third one return 404. Both served indexes agreed on height 8,046,000, so the threshold of 2 was already met. The node still failed after about 4 seconds with `Failed to download and parse index from <third-key>/index.json`, and never requested the archive.

Source: `BootstrapIndexRetriever.retrieve()` loops over every key, and `readJson` throws `BootstrapImportException` on the first `IOException`.

This is worth stating on the page because it is the failure operators are most likely to meet. Stalled publishers are what blocked this ticket for six weeks.

````suggestion
RSKj will only import bootstrap data that **enough trusted signers agree on**: the same height, with the same hash, correctly signed by each. The threshold is a majority of the keys you have configured, and never fewer than two — with the three keys shipped for each network, that is two. A height offered by too few signers is ignored, however recent it is. Among the heights that meet that bar, RSKj takes the highest.

Every configured signer's index must also be **reachable**. RSKj fetches all of them before it compares anything, so one index that cannot be fetched or parsed stops the import, even when the signers that did respond would have met the threshold. The threshold decides how many signers must agree; it does not make a missing publisher optional. If a publisher is down, wait for it to come back and run the import again.
````

File: docs/03-node-operators/04-setup/import-sync.md [hunk 11]
```diff
@@ -0,0 +1,324 @@
...
+
+## Common failures
+
+<details>
+<summary>Expand the message you are seeing</summary>
+
+| Message | What it means |
+|---|---|
+| `Failed to download and parse index from <url>` | A signer's index could not be fetched or parsed. Check network access to the import URL, and that you have not overridden `database.import.url` with something unreachable. |
+| `Downloaded files doesn't contain enough entries for a common height` | No single height is offered, with a matching hash, by the required majority of trusted signers. This is what you see if `database.import.trusted-keys` has been narrowed to fewer working signers than the threshold, or if the indexes have no height in common. |
```
core-baguira: **These two rows send the operator to the wrong place when a publisher is down.**

Following on from the note at line 68: a dead publisher produces the *first* message, not the second. Right now the first row tells the operator to check their own network, when the fault is usually upstream. And the second row says "fewer working signers than the threshold", which suggests a dead signer shows up here. It does not, because the run never gets that far.

Both rows are also worth saying the database is already gone by this point, since the page's own danger callout promises that.

````suggestion
| `Failed to download and parse index from <url>` | A signer's index could not be fetched or parsed. **This stops the import even if the other signers already agree**, so it is often an outage at the publisher rather than a problem on your side. Your database has already been erased at this point. Check network access to the import URL, and that you have not overridden `database.import.url` with something unreachable. If the URL is right and reachable for the other keys, wait for that publisher to recover and retry. |
| `Downloaded files doesn't contain enough entries for a common height` | No single height is offered, with a matching hash, by the required majority of trusted signers. This is what you see if `database.import.trusted-keys` has been narrowed below the threshold, or if the indexes have no height in common. An unreachable signer does **not** produce this message; it fails earlier, with `Failed to download and parse index from <url>`. |
````

**Review by core-baguira — COMMENTED** [review]
I tested this page by following it. The procedure works: a VETIVER-9.0.4 JAR imported Testnet in 227 seconds and landed at block 8,046,000, which is exactly the height the page's own `jq` query predicts. Every error message I triggered matched the failure table word for word. The Docker section, the Ubuntu paths, and the devportal transform all check out, and the page builds clean in Docusaurus with no broken links or anchors.

So the notes below are about the paths where it does not work. One of them stops the import cold and is not described anywhere on the page.

Comments are inline, with suggested text you can apply directly.

One thing worth flagging: the ten review threads from the earlier round look addressed in your commits, but none of them are marked resolved, so the PR still reads as CHANGES_REQUESTED. Sweeping them would make the remaining work clearer.

---

**Two notes that could not go inline, because the lines are outside this diff.**

**1. `03-reference.md` lines 112-115 still describe the old feature, and this PR links straight to them.** The new page links there twice, from its Related list and from `java.md`. A reader who follows those links is told the feature is experimental v0.1 and served from an S3 bucket, which does not match the page they just came from. The URL is not a bucket anywhere in the code either: `BootstrapURLProvider` just joins a base string and opens a plain `java.net.URL`, and the shipped hosts are `import.mainnet.rskcomputing.net` and `import.testnet.rskcomputing.net`. Suggested replacement for those four lines:

> Options related to import sync. See [Bootstrap a node using Import Sync](/node-operators/setup/import-sync/) for the full procedure.
>
> * `database.import.url = URL`
>   is the base URL under which the trusted signers publish their indexes and archives.

**2. A non-blocking nit on temporary space.** The page gives the figure three times and two of them disagree: line 40 says "about three times the size of the bootstrap archive", line 217 says "more than twice", and line 301 repeats "roughly three times" for the container's `/tmp`. I measured it on the current Testnet archive: the zip is 1,823,354,365 bytes and the extracted file is 2,324,750,125 bytes, so the peak is 2.28x. There is no other temporary copy, because the download, the hash check and the unzip all stream. Rather than swap in a new number that changes with every snapshot, it may be simpler to state the rule: both files sit there at once, so allow the archive plus the extracted file, which is the larger of the two.

File: docs/03-node-operators/09-troubleshooting/index.md [hunk 12]
```diff
@@ -45,16 +45,18 @@ If what you need is not in this section, **contact us** without hesitation throu
...
   The tool class is: `co.rsk.cli.tools.DbMigrate`
 
 **Required CLI arguments:**
 - `args[0]` - database target where we are going to insert the information from the current selected database.
-    - Note: You cannot migrate to the same database or an error will be thrown. It is highly recommended to turn off the node in order to perform the migration since latest data could be lost.> > - Example migrating from leveldb to rocksdb:
-- `java -cp rsk-core-<VERSION>.jar co.rsk.cli.tools.DbMigrate rocksdb`
+    - Note: You cannot migrate to the same database or an error will be thrown. It is highly recommended to turn off the node in order to perform the migration since latest data could be lost.
+    - Example migrating from `leveldb` to `rocksdb`:
+
+      `java -cp rsk-core-<VERSION>.jar co.rsk.cli.tools.DbMigrate rocksdb`
```
core-baguira: **This command cannot work as written.**

`DbMigrate` takes the target as a required option, not a positional argument. `co.rsk.cli.tools.DbMigrate rocksdb` fails immediately with:

```
Missing required option: '--targetDb=<targetdb>'
```

I ran both forms. The `-t` form works. The positional form fails every time, on any platform, because it is picocli argument parsing.

`02-cli.md` in this same PR already documents the correct `-t rocksdb`, so the two pages currently disagree.

The `args[0]` label is also not a real parameter name.

(`DbMigrate.java`'s own Javadoc says `args[0]` too, which is where this probably came from. That is a separate rskj bug and I can file it.)

````suggestion
- `-t, --targetDb` - the database engine to migrate to (`rocksdb` or `leveldb`). Use `rocksdb`; `leveldb` is deprecated.
    - Note: You cannot migrate to the same database or an error will be thrown. It is highly recommended to turn off the node in order to perform the migration since latest data could be lost.
    - Example migrating from `leveldb` to `rocksdb`:

      `java -cp rsk-core-<VERSION>.jar co.rsk.cli.tools.DbMigrate -t rocksdb`
````

File: docs/03-node-operators/04-setup/import-sync.md [hunk 13]
```diff
@@ -0,0 +1,324 @@
...
+| `Error trying to read bootstrap data contents. Please start again the import process` | The downloaded archive could not be read back while verifying its hash — this happens before extraction, not after. Usually a truncated or removed temporary file, or an I/O error on the filesystem holding the temporary directory. Retry the import. |
+| `Error reading bootstrap data from <path>` | The extracted `bootstrap-data.bin` could not be read. Check that the temporary directory still holds it and that nothing is cleaning temporary files while the node runs. |
+| `Configuration has less trusted sources than the minimum required <n> of 2` | Fewer than two trusted keys are configured, and two is the floor however few you configure. Restore the shipped keys for the network. This is a warning at startup, not the failure itself — the run continues and then fails on one of the messages above. |
+| `java.lang.OutOfMemoryError` | The load stage ran out of heap. Raise `-Xmx` above the 4G used above and run the import again; note that a retry downloads the bootstrap data again. |
```
core-baguira: **This row points at the wrong log, and the advice does not work on older nodes.**

Two problems.

First, the location. `Start.main` catches `Exception`, and `OutOfMemoryError` is an `Error`, so it never reaches the logger. It only appears on the terminal. The page says at line 172 that the import "prints nothing to the console" and "writes to `logs/rsk.log`", so this is the one message in the table that cannot be where the page sends people to look. What the operator sees is a log that simply stops after `Bootstrap data extracted`.

Second, the advice. On a node older than VETIVER-9.0.4, raising `-Xmx` cannot help. Those versions read the whole payload into a single byte array, and today's payload is about 2.3 GB, which is over Java's maximum array size. I ran the 9.0.3 JAR against the current Testnet archive: it downloaded, hash-checked and extracted fine, then died with `OutOfMemoryError: Required array size too large` on stderr, with nothing in the log.

Since the page already requires 9.0.4 or later, this is exactly what a reader who skipped that line will hit.

````suggestion
| `java.lang.OutOfMemoryError` | Note this one appears **on the terminal, not in `logs/rsk.log`** — if the log stops after `Bootstrap data extracted`, check the terminal before assuming the node is still working. `Java heap space` means the load stage ran out of heap: raise `-Xmx` above the 4G used above and run the import again. `Required array size too large` means the node is older than `VETIVER-9.0.4` and cannot read the bootstrap data published today; raising `-Xmx` will not help, so upgrade instead. Either way, a retry downloads the bootstrap data again. |
````

File: docs/03-node-operators/04-setup/import-sync.md [hunk 14]
```diff
@@ -0,0 +1,324 @@
...
+
+The archive is downloaded and unpacked inside the container, under `/tmp`, which is the container's own writable layer rather than the volume. Allow roughly three times the archive size there, on top of the space the database needs in the volume.
+
+### Ubuntu package
+
+The package installs the JAR at `/usr/share/rsk/rsk.jar` and its configuration in `/etc/rsk`, and runs the node as the `rsk` user. Stop the service, run the same JAR once with `--import`, then start the service again:
+
+```shell
+sudo service rsk stop
+sudo -u rsk java -Xmx4G -cp /usr/share/rsk/rsk.jar co.rsk.Start --import
```
core-baguira: **This command writes its log somewhere the operator will not look.**

The `rsk` service starts the node with `-Dlogback.configurationFile=/etc/rsk/logback.xml`, and that file writes to `/var/log/rsk/rsk.log`. The command here does not pass that option, so it falls back to the JAR's default of `./logs/rsk.log`, relative to whatever directory the admin happened to run it from.

So the third bullet below tells the operator to wait for `Bootstrap data has successfully been imported`, but the file they will check, `/var/log/rsk/rsk.log`, never gets it. If the current directory is not writable by `rsk`, the log may not be written at all.

`postinst` chowns `/var/log/rsk` to `rsk:rsk`, so passing the option is enough to fix it.

````suggestion
sudo -u rsk java -Xmx4G -Dlogback.configurationFile=/etc/rsk/logback.xml -cp /usr/share/rsk/rsk.jar co.rsk.Start --import
````

Worth adding to the third bullet too: "The `-Dlogback.configurationFile` option above is what the service uses, so the import logs to `/var/log/rsk/rsk.log`. Without it the log goes to `./logs/rsk.log` in whatever directory you ran the command from.".

File: docs/03-node-operators/04-setup/import-sync.md [hunk 15]
```diff
@@ -0,0 +1,324 @@
...
+| `The file is corrupted or incomplete. Please start again the import process` | The archive downloaded and matched its hash, but could not be unpacked. Check free space in the temporary directory first: running out during extraction produces this message even though the archive itself is fine. If space is adequate and it persists, the published archive is faulty — report it rather than working around it. |
+| `Error trying to read bootstrap data contents. Please start again the import process` | The downloaded archive could not be read back while verifying its hash — this happens before extraction, not after. Usually a truncated or removed temporary file, or an I/O error on the filesystem holding the temporary directory. Retry the import. |
+| `Error reading bootstrap data from <path>` | The extracted `bootstrap-data.bin` could not be read. Check that the temporary directory still holds it and that nothing is cleaning temporary files while the node runs. |
```
core-baguira: **This row's message has a second spelling that is not covered, and the v2 errors are missing.**

`Error reading bootstrap data from <path>` is thrown when RSKj reads the first byte to detect the format, not when it reads the extracted file back. A failure during the actual streaming read produces a different string, `Error reading bootstrap-data v2 from <path>`.

Separately, the v2 reader has about ten messages that all start with `Bootstrap-data v2 ` (bad magic, unsupported version, chunk length, section order, truncation, trailing bytes). They all mean the same thing to an operator: the file is not intact. Rather than a row each, one line covers them.

````suggestion
| `Error reading bootstrap data from <path>` or `Error reading bootstrap-data v2 from <path>` | The extracted `bootstrap-data.bin` could not be read. Check that the temporary directory still holds it and that nothing is cleaning temporary files while the node runs. |
| Any message starting `Bootstrap-data v2 ` (bad magic, unsupported version, chunk length, section order, truncated or trailing bytes) | The extracted file is not intact. Check first that you are on `VETIVER-9.0.4` or later, then delete the temporary files and run the import again. If it repeats, the published archive is faulty — report it rather than working around it. |
````

File: docs/03-node-operators/04-setup/import-sync.md [hunk 16]
```diff
@@ -0,0 +1,324 @@
...
+           | [.[].dbs[] | {height, hash}] | group_by(.height)
+           | map({height: .[0].height, agreeing: ([group_by(.hash)[] | length] | max)})
+           | map(select(.agreeing >= $required)) | max_by(.height)' index-*.json
+    ```
+  </TabItem>
+</Tabs>
+
+This reproduces the node's own selection rule and prints the height it would choose, along with how many signers agree on it. The threshold is derived from the number of index files you fetched, so the query stays correct if you configure a different set of keys.
+
+Each network gets its own directory because the final `jq` reads every `index-*.json` it finds. Index files are named after the signer's key, and the two networks use different keys, so running both checks in one directory leaves six files there rather than overwriting three — and the query then takes the highest height across both networks. On a Testnet check that had Mainnet files alongside it, the answer would be a Mainnet height, reported with two agreeing signers and no error.
```
core-baguira: **This describes what the query used to do, not what it does now.**

The threshold is derived from the number of index files, as the paragraph above says. With six files it becomes four. Neither network has four signers, so nothing passes and the query prints `null`.

I ran it: three Testnet plus three Mainnet index files gives `null`. The "two agreeing signers" result it describes cannot happen with the query directly above.

The hazard itself is real, just at a different count: with four or five files in the directory, the wrong network can win.

````suggestion
Each network gets its own directory because the final `jq` reads every `index-*.json` it finds. Index files are named after the signer's key, and the two networks use different keys, so running both checks in one directory leaves six files there rather than overwriting three. That breaks the query rather than giving a wrong answer: the threshold is derived from the file count, so six files require four agreeing signers, more than either network has, and the query prints `null`. With a partial mix of the two networks in one directory, it can instead return a height from the wrong network.
````

File: docs/03-node-operators/04-setup/import-sync.md [hunk 17]
```diff
@@ -0,0 +1,324 @@
...
+The `--import` flag sets `database.import.enabled` for that run only. It supplies no URL and no keys of its own — those come from the network configuration shipped in the JAR. `-Xmx4G` is the heap recommended for this command in [Setup node using Java](/node-operators/setup/installation/java/); left out, the JVM sizes the heap from physical RAM instead.
+
+The import prints nothing to the console. It writes to `logs/rsk.log`, relative to the directory you ran the command from. A successful import logs there, in order:
+
+```text
+Bootstrap data downloaded
+Bootstrap data hash checked
+Bootstrap data extracted
+Bootstrap data has successfully been imported in <n> mills
+```
```
core-baguira, line 183: **A v2 import logs two more lines than this.**

The four lines here are all correct and appear in this order. But since both networks now publish v2 data, every import also logs `Detected bootstrap-data v2 (chunked) format` after "extracted", and a summary line before the last one. An operator matching the log against this block will see extra lines and wonder if something is wrong.

This is what my Testnet import actually wrote:

```
Bootstrap data downloaded
Bootstrap data hash checked
Bootstrap data extracted
Detected bootstrap-data v2 (chunked) format
Bootstrap-data v2 imported 4000 blocks, 6905624 long values and 17393585 state nodes in 135832 ms
Bootstrap data has successfully been imported in 227353 mills
```

(There is also an `Inserting blocks and state...` line at DEBUG level. I left it out below since it depends on the log level.)

`````suggestion
```text
Bootstrap data downloaded
Bootstrap data hash checked
Bootstrap data extracted
Detected bootstrap-data v2 (chunked) format
Bootstrap-data v2 imported <blocks> blocks, <values> long values and <nodes> state nodes in <n> ms
Bootstrap data has successfully been imported in <n> mills
```

Other subsystems log in between; what matters is that these six appear in this order. Older bootstrap data logs `Detected bootstrap-data v1 (legacy) format` in place of the two `v2` lines.
`````

File: docs/03-node-operators/04-setup/import-sync.md [hunk 18]
```diff
@@ -0,0 +1,242 @@
...
+      <TRUSTED-KEY-1> \
+      <TRUSTED-KEY-2> \
+      <TRUSTED-KEY-3>
+    do
+      curl -sS -o "index-$KEY.json" "$URL$KEY/index.json"
+    done
+
+    jq -s '[.[].dbs[] | {height, hash}] | group_by(.height)
+           | map({height: .[0].height, agreeing: ([group_by(.hash)[] | length] | max)})
+           | map(select(.agreeing >= 2)) | max_by(.height)' index-*.json
```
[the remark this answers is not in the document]
core-baguira: Checked this against the live indexes. The query is correct as written, and no `sort_by` is needed.

`group_by/1` sorts by the grouping key internally, so `group_by(.height)` and `group_by(.hash)` already receive sorted input by definition.

I also ran the page's query unmodified against the three live index files for each network. It returned `{"height": 8046000, "agreeing": 3}` for Testnet and `{"height": 9216000, "agreeing": 3}` for Mainnet. I then ran the actual import: the node selected 8,046,000, the same height the query predicted.

docs/03-node-operators/09-troubleshooting/index.md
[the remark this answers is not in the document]
core-baguira: I could not reproduce this. I built the site to check.

I ran the devportal transform over this branch, copied the output into a clone of `rsksmart/devportal` (Docusaurus 3.10.2) and ran `npm run build`. The line renders as a code element, not as literal backticks:

```html
<code>java -cp rsk-core-&lt;VERSION&gt;.jar co.rsk.cli.tools.DbMigrate rocksdb</code>
```

The build also reported no broken links or anchors.

Note that this line changes anyway if the `-t rocksdb` correction suggested elsewhere in this review is applied.

docs/03-node-operators/09-troubleshooting/index.md
```diff
@@ -45,16 +45,18 @@ If what you need is not in this section, **contact us** without hesitation throu
...
 
 **Required CLI arguments:**
-- `args[0]` - database target where we are going to insert the information from the current selected database.
-    - Note: You cannot migrate to the same database or an error will be thrown. It is highly recommended to turn off the node in order to perform the migration since latest data could be lost.> > - Example migrating from leveldb to rocksdb:
-- `java -cp rsk-core-<VERSION>.jar co.rsk.cli.tools.DbMigrate rocksdb`
+- `-t, --targetDb` - the database engine to migrate to (`rocksdb` or `leveldb`). Use `rocksdb`; `leveldb` is deprecated.
+    - Note: You cannot migrate to the same database or an error will be thrown. It is highly recommended to turn off the node in order to perform the migration since latest data could be lost.
+    - Example migrating from `leveldb` to `rocksdb`:
+
+      `java -cp rsk-core-<VERSION>.jar co.rsk.cli.tools.DbMigrate -t rocksdb`
```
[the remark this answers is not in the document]
fmacleal: Context: this thread is about one example command. The command is on line 59 of the troubleshooting page. It shows an operator how to run `DbMigrate`.

We keep `rsk-core-<VERSION>.jar`. There are three reasons.

- Line 44 of this page shows the `RewindBlocks` example. That example uses `rsk-core-<VERSION>.jar`. This PR does not change line 44. The two examples on this page now agree with each other.
- Comment https://github.com/rsksmart/rskj/pull/3676#discussion_r3935956784 is your earlier comment on line 59. That comment asks us to make this example agree with the `RewindBlocks` example above it. The example now agrees.
- `rsk.jar` in the Docker section and the Ubuntu section is not a placeholder. The Ubuntu package installs the JAR at `/usr/share/rsk/rsk.jar`. The operator types that path.

A change to `<PATH-TO-THE-RSKJ-JAR>` makes the two examples on this page disagree.

.github/workflows/scripts/devportal-update/doc_config.yaml
```diff
@@ -31,6 +31,14 @@ files:
...
+    output: transformed/03-node-operators/04-setup/import-sync.md
+    sidebar_label: Import Sync
+    sidebar_position: 4.5
+    title: Bootstrap a node using Import Sync
```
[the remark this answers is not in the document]
fmacleal: Context: this thread is about the sidebar position of the new page.

The description is now correct. It gives `sidebar_position: 4.5`.

4.5 is the value we want. The *Node Installation* page has position 4. Position 4.5 puts the new page directly after *Node Installation*. An operator meets the `--import` flag in the installation steps first. The earlier value 5.5 put the new page after *Configure the node*. That position is further from the installation steps.

.github/workflows/scripts/devportal-update/doc_config.yaml
```diff
@@ -31,6 +31,14 @@ files:
...
+    tags: [import, sync, bootstrap, rskj, node, node operators, rootstock, mainnet, testnet, how-to]
+    description: "Bootstrap an RSKj node from published, signed bootstrap data instead of syncing from genesis."
+
```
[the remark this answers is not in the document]
fmacleal: Context: this thread is about the file list in the description.

The description is now correct. It gives all ten files.

- `04-setup/import-sync.md` — the new page.
- `doc_config.yaml` — the devportal entry for the new page.
- `02-installation/java.md` and `03-configuration/02-cli.md` — the two pages that already mention the `--import` flag. Each page now gives the erase warning, the `VETIVER-9.0.4` floor and a link to the new page.
- `03-configuration/preferences.md`, `03-configuration/03-reference.md`, `04-node-runner/linux.md`, `04-node-runner/macos.md`, `04-node-runner/windows.md` and `09-troubleshooting/index.md` — the LevelDB deprecation.

fmacleal (approved):
Thanks for having addressed the comments, this is way clearer now!

LGTM

core-baguira (approved):

PR #3678, opened 2026-08-19
fix: preserve EIP-7702 authorization refunds on failed execution

meldsun0:
## Description
Authorization refunds are now preserved when subsequent EVM execution fails, and the final post-refund gas amount is used consistently when calculating sender refunds, paid fees, and receipt gas usage.

The transaction execution finalization logic was also simplified by removing the dependency on TransactionExecutionSummary for gas settlement calculations.

As a side effect of touching the same code path, a related bug is also fixed: a direct call to a precompiled contract that throws now correctly reports `receipt.status = Failed` (previously reported `Success`).

**Gated behind a new consensus rule (`RSKIPXXX`) — requires a hard fork.** `receipt.status` and `receipt.gasUsed` are hashed into the receipts trie and block header, so old and new nodes would compute different block hashes for the same block without a coordinated activation. 

The code that fixes the authorization refund (in calculateFee()/calculateRefund()) only checks whether RSKIP560 is active — it never checks RSKIP545 (the flag that enables type-4/EIP-7702 transactions). Today this causes no issue because both are mapped to the same hardfork (vetiver900) in the config, so they activate at exactly the same time. But if someone ever decides to separate them — for example, activating RSKIP545 at one block height and RSKIP560 at a later one — a window would open where type-4 transactions are already enabled on the network, but the refund fix isn't yet. During that window, the original bug (the authorization refund gets discarded instead of returned to the sender) would silently reappear, with nothing flagging it.

Regression tests were added for exceptional halts and REVERT.

## Motivation and Context
Exceptional halts could preserve the authorization refund in receipt gas while still charging the sender the full gas limit, causing inconsistent fee accounting.

## How Has This Been Tested?

- Direct precompile-call failure: receipt status/gasUsed reconcile with paidFees
- Pre/post activation parity (legacy behavior preserved when `RSKIP560` is off)

Added regression tests using real VM execution covering:
- Invalid opcode
- Out-of-gas
- REVERT
- Authorization persistence after execution failure
- Authorization refunds and fee/receipt consistency

## Types of changes

## Checklist:

* **Other information**:


rskj-core/src/main/java/org/ethereum/config/blockchain/upgrades/ConsensusRule.java
```diff
@@ -119,6 +119,7 @@ public enum ConsensusRule {
     RSKIP546("rskip546"), // Type 1 and Type 2 transaction formats (EIP-2930/EIP-1559 encoding)
     RSKIP551("rskip551"), // Disable RSKIP459
     RSKIP552("rskip552"), // Blake2F improvements in data handling
+    RSKIP560("rskip560"), // Precompiled contract exception handling on direct-call path -- number provisional pending RSKIP acceptance
```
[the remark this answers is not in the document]
meldsun0: N/A

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -417,6 +417,11 @@ private void call() {
                     track.setupContract(targetAddress);
                 }
             } catch (VMException | RuntimeException e) {
+                if (activations.isActive(ConsensusRule.RSKIP560)) {
+                    gasLeftover = 0;
+                    gasUsed = txGasLimit;
+                    execError(e);
+                }
                 result.setException(e);
             }
```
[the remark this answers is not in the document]
meldsun0: This is a known issue and will be solved in another PR.

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -416,6 +416,11 @@ private void call() {
...
```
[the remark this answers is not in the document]
meldsun0: This fix will be added on another PR.

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -675,60 +671,39 @@ private void localCallFinalization() {
...
-
-        if (result.getException() != null) {
-            summaryBuilder.markAsFailed();
-        }
-
-        logger.trace("Building transaction execution summary");
+    private Coin calculateFee() {
+        BigInteger chargedGas = toBI(tx.getGasLimit()).subtract(BigInteger.valueOf(gasLeftover));
+        return tx.getGasPrice().multiply(chargedGas);
+    }
```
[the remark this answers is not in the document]
meldsun0: This was already solved

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -675,60 +671,39 @@ private void localCallFinalization() {
...
 
-        return summaryBuilder.build();
+    private Coin calculateRefund() {
+        return tx.getGasPrice().multiply(BigInteger.valueOf(gasLeftover));
     }
```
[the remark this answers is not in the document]
meldsun0: This was already solved

rskj-core/src/main/resources/reference.conf
```diff
@@ -108,6 +108,7 @@ blockchain = {
             rskip551 = vetiver900
             rskip552 = vetiver900
             rskip559 = cardamom1000
+            rskip560 = vetiver900
```
italo-sampaio, line 111: question: shouldn't this be cardamom1000 instead? Or are we treating everything as vetiver900 to update in one pass in the future? Seems a bit awkward to define the activation heigh in the past in any case.

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -674,60 +671,45 @@ private void localCallFinalization() {
...
-    private TransactionExecutionSummary buildTransactionExecutionSummary(TransactionExecutionSummary.Builder summaryBuilder, long gasRefund) {
-        summaryBuilder
-                .gasUsed(toBI(result.getGasUsed()))
-                .gasRefund(toBI(gasRefund))
-                .deletedAccounts(result.getDeleteAccounts())
-                .internalTransactions(result.getInternalTransactions());
-
-        if (result.getException() != null) {
-            summaryBuilder.markAsFailed();
+    private Coin calculateFee() {
```
italo-sampaio, line 697: question: should this really be gated on RSKIP560 or RSKIP545 instead?

the delta between branches is non-zero only when `result.getException() != null` and `gasLeftover != 0` after `refundGas()`, and the only surviving contributor there is `authorizationRefund`, which is assigned only under `tx.isType4()` — admitted only when RSKIP545 is active. So RSKIP560-before-RSKIP545 is a no-op, but RSKIP560-*after*-RSKIP545 silently loses the refund.

Gating on `RSKIP545` here would make the dependency match reality and remove the hazard the PR description already describes. Worth noting `Type4TransactionExecutorFailingTests.authorizationRefundOnPlainVmExceptionBehavesPerActivation` currently runs the 545-on/560-off combination and asserts the divergence as expected, which pins the bad window as correct behaviour.

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigTest.java
```diff
@@ -162,6 +162,7 @@ class ActivationConfigTest {
         "    rskip551: vetiver900",
         "    rskip552: vetiver900",
         "    rskip559: cardamom1000",
+        "    rskip560: vetiver900",
```
italo-sampaio, line 165: The config wiring here is right, but the test preset next door wasn't updated: `ActivationConfigsForTest.getVetiverRskips()` still lists only `{RSKIP535, RSKIP540, RSKIP544, RSKIP551, RSKIP552}`. Since `enableTheseDisableThose` defaults every unlisted rule to `-1`, any test using `vetiver900()` / `vetiver900AllRskips()` silently runs with RSKIP560 **disabled**, contradicting `reference.conf`. RSKIP559 was added to its equivalent list, so this one looks like an oversight rather than intent.

rskj-core/src/main/java/org/ethereum/core/transaction/parser/TransactionInput.java
```diff
@@ -132,11 +132,11 @@ public static TransactionInput fromCallArguments(CallArguments args, Supplier<St
 
         return new TransactionInput(
                 typePrefix,
-                nonce == null ? null : nonce.toByteArray(),
+                nonce == null ? null : CommonParsingUtils.unsignedBytes(nonce),
                 gasPrice,
                 maxPriorityFeePerGas,
                 maxFeePerGas,
-                gasLimit.toByteArray(),
+                CommonParsingUtils.unsignedBytes(gasLimit),
```
italo-sampaio, line 139: This changes the encoding for both `nonce` and `gasLimit`, and it's a real fix — `BigInteger.toByteArray()` prepends a `0x00` sign byte for any value with the top bit set, so `gasLimit = 12_000_000` (`0xB71B00`) was being emitted as `00 B7 1B 00`.

No test in this PR touches `TransactionInput`, `CommonParsingUtils`, or `fromCallArguments`, despite this being the head commit's stated subject. Boundary cases at 127/128, 255/256 and 32767/32768 for both fields would pin it.

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/Type4TransactionExecutorFailingTests.java
```diff
@@ -0,0 +1,398 @@
...
+
+        if (rskip560Active) {
+            assertEquals(authorizationRefund, txExecutor.getResult().getDeductedRefund(), "authorization refund should be fully applied (well under the half-of-gasUsed cap)");
+            assertFalse(receipt.isSuccessful(), "post-activation, both gates fire: status must be FAILED");
+
+            Coin expectedFee = Coin.valueOf(100_000L - authorizationRefund); // 90_500
+
+            assertEquals(BigInteger.valueOf(expectedGasUsed), reportedGasUsed, "receipt.getGasUsed should be gasLimit minus the authorization refund");
+
+            assertEquals(expectedFee, txExecutor.getPaidFees());  // effective gasPrice = 1
```
italo-sampaio, line 205: The `// effective gasPrice = 1` note here points at a gap across the whole new suite: every new test uses `gasPrice = 1`, which makes `paidFees == gasUsed × gasPrice` indistinguishable from `paidFees == gasUsed`.

Verified by mutation — changing `calculateFee` to `return new Coin(chargedGas)` and `calculateRefund` to drop the price factor entirely, so the gas price disappears from both the sender's charge and the fee, leaves **all 139 tests in this package passing**. Only pre-existing state-root assertions in `ParallelExecutionStateTest` catch it. This is exactly the arithmetic `TransactionExecutionSummary` owned before it was deleted, and it had no dedicated test either. One case with `gasPrice > 1` would close it.

italo-sampaio (approved):
Looks good overall. Left a few comments / suggestions. Approving regardless.

italo-sampaio (approved):

rskj-core/src/main/resources/reference.conf
meldsun0: Yes you are right, it should be the next coming release but we are treating everything as vetiver900 to update in one pass in the future as you mentioned.

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -416,6 +416,11 @@ private void call() {
                     track.setupContract(targetAddress);
                 }
             } catch (VMException | RuntimeException e) {
+                if (activations.isActive(ConsensusRule.RSKIP560)) {
```
dlealmac: Was the missing `!localCall` here intentional? The out-of-gas pre-check at :393 has it, and for
a local call `txGasLimit` is the `gasEstimationCap` (6,800,000) rather than a real limit, so
`gasUsed = txGasLimit` at :421 propagates through `result.spendGas` into `maxGasUsed`, and
`internalEstimateGas` returns it.

Called `eth_estimateGas` against a throwing precompile going from 47,000 to 6,800,000 across
the gate, and `handleTransactionRevertIfHappens` doesn't intercept it because a precompile
`VMException` isn't a revert. If that's not intended, `!localCall && activations.isActive(...)`
covers it.

dlealmac (changes requested):

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
meldsun0: I think RSKIP560 should remain separate from RSKIP545 because it fixes a pre-existing exceptional-halt accounting issue unrelated to Type 4 transactions.
On precompile failures, the full gas limit was charged while only the partial pre-failure gas was reported, which also caused the refund cap to be calculated from the wrong gas usage.
Moreover, any refund calculated by refundGas() was ultimately discarded because failed execution summaries returned zero refund and leftover.
RSKIP545 only introduces authorizationRefund, exposing another instance of this existing issue rather than being the reason for the fix.

rskj-core/src/test/java/org/ethereum/config/blockchain/upgrades/ActivationConfigTest.java
meldsun0: RSKIP545 (and also 543/546) are already mapped to vetiver900 in reference.conf but are not included in getVetiverRskips(), so this does not seem to be a complete mirror of the production hardfork configuration today.

I think this should be updated once the target hardfork for these features is definitively decided. At that point, the corresponding test preset should be updated to match the final activation mapping, whether that is Vetiver or a later hardfork.

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/Type4TransactionExecutorFailingTests.java
meldsun0: Good catch! I will add that test!

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
```diff
@@ -674,60 +671,45 @@ private void localCallFinalization() {
...
 
-        logger.trace("Building transaction execution summary");
-
-        return summaryBuilder.build();
+    private Coin calculateRefund() {
+        if (result.getException() != null && !activations.isActive(ConsensusRule.RSKIP560)) {
+            return Coin.ZERO;
+        }
+        return tx.getGasPrice().multiply(BigInteger.valueOf(gasLeftover));
     }
```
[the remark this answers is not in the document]
meldsun0: Rather than changing the gate to RSKIP545, we should ensure that the hardfork scheduling/configuration does not create an incompatible activation window between the two rules.

rskj-core/src/test/java/co/rsk/core/bc/transactionexecutor/PrecompiledContractExceptionActivationTest.java
```diff
@@ -0,0 +1,279 @@
...
+            Assertions.assertFalse(receipt.isSuccessful(), "post-activation, receipt must correctly report FAILED status");
+            Assertions.assertEquals(gasLimit, reportedGasUsed, "post-activation, receipt.gasUsed must equal the full gasLimit charged");
+            Assertions.assertEquals(expectedReportedFee,  executor.getPaidFees(), "paidFees must equal receipt.gasUsed * gasPrice");
+        } else {
+            Coin expectedLegacyFee = Coin.valueOf(gasLimit.longValueExact() * gasPrice);
+            Assertions.assertEquals(expectedLegacyFee, executor.getPaidFees());
+            Assertions.assertTrue(receipt.isSuccessful(), "pre-activation, legacy (buggy) SUCCESS status must be preserved");
+            long requiredGas = 23_000L; // BridgeMethods.RELEASE_BTC fixedCost fallback
+            long basicTxCost = tx.transactionCost(config.getNetworkConstants(), activationConfig.forBlock(block.getNumber()), blockTxSignatureCache);
+            BigInteger expectedLegacyGasUsed = BigInteger.valueOf(requiredGas + basicTxCost);
```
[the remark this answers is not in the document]
meldsun0: Not applicable. The 23_000L value is intentional here as part of the expected legacy behavior this regression test is preserving. Deriving it from BridgeMethods.RELEASE_BTC.getCost() would make the expected value depend on production logic, while this test is meant to explicitly capture the behavior at the time of the RSKIP560 activation.

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
meldsun0: Nice find!

rskj-core/src/main/java/org/ethereum/core/transaction/parser/TransactionInput.java
meldsun0: This is covered in PR #3696. It adds the requested boundary tests for both nonce and gasLimit, and they fail without the unsignedBytes fix.

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
dlealmac, line 400: The authorization refund is still destroyed on the precompile out-of-gas pre-check. 

The early `return` at :400 skips `result.spendGas(gasUsed)` at :426, so `result.getGasUsed()` is
0 at `refundGas()` and the cap at :718 collapses to `min(9500, 0) = 0` — the authorization refund
is discarded even though the authorization committed. Measured on this branch with the real
`Identity` precompile at `gasLimit == basicTxCost == 46_000`: `deductedRefund = 0`, sender charged
the full 46,000; at 46,015 the same transaction is charged 36,515.

It is pre-existing and inert with respect to RSKIP560 (this path never sets
`result.getException()`, so the gate doesn't apply), should we handle it now or create a separate ticket for it?

dlealmac (changes requested):

rskj-core/src/main/java/org/ethereum/core/TransactionExecutor.java
meldsun0: The bug already exists in the original code: the precompile OOG path returns before result.spendGas(), leaving result.getGasUsed() at 0.
This causes the refund cap to collapse to 0, regardless of any available refund.
RSKIP-545 did not introduce this issue; it only made the existing bug affect authorization refunds as well.
Since it is pre-existing and outside this PR’s scope, it should be fixed in a separate PR.

dlealmac: fine, it works!

dlealmac (approved):
LGTM!

italo-sampaio (approved):
LGTM!

PR #3679, opened 2026-08-20
refactor: extracted markTxIdAsProcessed from markTxAsProcessed 

julianlen:
## Description

markTxAsProcessed(BtcTransaction btcTx) marks a BTC transaction as processed. It only uses one thing from the full BtcTransaction it receives: btcTx.getHash(false) (the no-witness txid, the actual key persisted via provider.setHeightBtcTxhashAlreadyProcessed(...)). Then, btcTx.getHash(true)(the witness hash/wtxid) is used only in a debug log line.

However, by changing the method markTxAsProcessed to receive a txid, we are increasing the risk of misusing it. This is because, from the dev’s point of view, markTxAsProcessed can receive a txid or a wtxid. Having markTxAsProcessed as is deletes this risk because the method itself is computing the txid and not the wtxid. 

Therefore, we decided to create a method markTxIdAsProcessed that will be called from markTxAsProcessed once it computes the txid.

In the future, markTxIdAsProcessed will be called only from registerPegoutTransaction.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
+++ b/rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
@@ -761,19 +761,19 @@ public class BridgeSupport {
 
     private void markTxAsProcessed(BtcTransaction btcTx) throws IOException {
         // Mark tx as processed on this block (and use the txid without the witness)
-        long rskHeight = rskExecutionBlock.getNumber();
-        Sha256Hash btcTxId = btcTx.getHash(false);
-        markTxIdAsProcessed(btcTxId, rskHeight);
+        Sha256Hash btcTxHash = btcTx.getHash(false);
+        markBtcTxHashAsProcessed(btcTxHash);
         logger.debug(
             "[markTxAsProcessed] Mark btc transaction {} (wtxid: {}) as processed at height {}",
-            btcTxId,
+            btcTxHash,
             btcTx.getHash(true),
-            rskHeight
+            rskExecutionBlock.getNumber()
         );
     }
 
-    private void markTxIdAsProcessed(Sha256Hash btcTxId, long rskHeight) throws IOException {
-        provider.setHeightBtcTxhashAlreadyProcessed(btcTxId, rskHeight);
+    private void markBtcTxHashAsProcessed(Sha256Hash btcTxHash) throws IOException {
+        long rskHeight = rskExecutionBlock.getNumber();
+        provider.setHeightBtcTxhashAlreadyProcessed(btcTxHash, rskHeight);
     }
 
     private boolean shouldProcessPegInVersionLegacy(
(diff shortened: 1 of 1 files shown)


rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -762,15 +762,20 @@ private void refundTxSender(
     private void markTxAsProcessed(BtcTransaction btcTx) throws IOException {
         // Mark tx as processed on this block (and use the txid without the witness)
         long rskHeight = rskExecutionBlock.getNumber();
-        provider.setHeightBtcTxhashAlreadyProcessed(btcTx.getHash(false), rskHeight);
+        Sha256Hash btcTxId = btcTx.getHash(false);
+        markTxIdAsProcessed(btcTxId, rskHeight);
```
julia-zack:
```suggestion
        markTxHashAsProcessed(btcTxId, rskHeight);
```

since the provider method says `...BtcTxHash`, ill keep that naming.

jeremy-then: In that case, we can also rename the `btcTxId` variable to `btcTxHash`. The method were we get it from is called `btcTx.getHash` anyways.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
```diff
@@ -762,15 +762,20 @@ private void refundTxSender(
...
         logger.debug(
             "[markTxAsProcessed] Mark btc transaction {} (wtxid: {}) as processed at height {}",
-            btcTx.getHash(),
+            btcTxId,
             btcTx.getHash(true),
             rskHeight
         );
     }
 
+    private void markTxIdAsProcessed(Sha256Hash btcTxId, long rskHeight) throws IOException {
```
julia-zack: also not sure about having the `rskHeight` as a param, right? is just a long number. Much better to obtain it from the rskExecutionBlock

jeremy-then:
```suggestion
    private void markTxIdAsProcessed(Sha256Hash btcTxHash, long rskHeight) throws IOException {
```

Similar to what @julia-zack suggested.

jeremy-then (approved):
LGTM.

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julianlen: done https://github.com/rsksmart/rskj/pull/3679/commits/73ca22bad6832315aa8bca40c97a1912549c1269

julianlen: done https://github.com/rsksmart/rskj/pull/3679/commits/73ca22bad6832315aa8bca40c97a1912549c1269

rskj-core/src/main/java/co/rsk/peg/BridgeSupport.java
julianlen: done https://github.com/rsksmart/rskj/pull/3679/commits/73ca22bad6832315aa8bca40c97a1912549c1269

julia-zack (approved):

marcos-iov (approved):

PR #3696, opened 2026-08-27
Structured rpc transactions emit non canonical nonce encodings

meldsun0:
## Description
Ensures nonce and gas limit values use canonical unsigned byte encoding when transactions are constructed from structured RPC inputs.

## Motivation and Context
BigInteger.toByteArray() may add a leading sign byte for values such as 128, producing non-canonical RLP.
This change ensures structured RPC transactions match canonical builder and raw transaction encodings.

## How Has This Been Tested?
Added boundary tests for nonce and gas limit across Legacy and transaction Types 1, 2, and 4.
Tests verify canonical bytes, signing preimages, signed encodings, and transaction hashes.

## Types of changes

## Checklist:

### DIFF
--- a/rskj-core/src/test/java/org/ethereum/core/transaction/parser/NonceGasLimitCanonicalEncodingTest.java
+++ b/rskj-core/src/test/java/org/ethereum/core/transaction/parser/NonceGasLimitCanonicalEncodingTest.java
@@ -24,10 +24,12 @@ import org.ethereum.core.transaction.TransactionType;
 import org.ethereum.core.transaction.parser.util.AuthorizationListCodec;
 import org.ethereum.core.transaction.parser.util.CommonParsingUtils;
 import org.ethereum.rpc.CallArguments;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
+import javax.annotation.Nullable;
 import java.math.BigInteger;
 import java.util.List;
 import java.util.stream.Stream;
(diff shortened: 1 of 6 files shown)


dlealmac (approved):
LGTM

rskj-core/src/test/java/org/ethereum/core/transaction/parser/NonceGasLimitCanonicalEncodingTest.java
```diff
@@ -0,0 +1,225 @@
...
+    void legacyNonceOmittedAlsoResolvesToCanonicalZero() {
+        Transaction rpcTx = Transaction.fromCallArguments(callArguments(TransactionType.LEGACY, null, DEFAULT_GAS_LIMIT), null, REGTEST_CHAIN_ID);
+        assertArrayEquals(CommonParsingUtils.unsignedBytes(BigInteger.ZERO), rpcTx.getNonce());
+    }
+
+    private static void assertCanonical(TransactionType type, BigInteger nonce, BigInteger gasLimit) {
+        Transaction builderTx = build(type, nonce, gasLimit);
+        Transaction rpcTx = Transaction.fromCallArguments(callArguments(type, nonce, gasLimit), null, REGTEST_CHAIN_ID);
+
+        byte[] expectedNonce = CommonParsingUtils.unsignedBytes(nonce);
```
italo-sampaio, line 108: The expected values here are produced by `CommonParsingUtils.unsignedBytes`, which is the function the fix calls, and the remaining assertions compare one construction path against another. That makes this a consistency check rather than a correctness one: nothing in the file states which byte string is actually canonical, so if `unsignedBytes` were wrong, all three paths would agree on the wrong answer and the suite would stay green.

Could we pin at least one independent literal? The values are short enough to derive by
hand from the RLP spec:

| nonce | minimal magnitude | RLP element |
| --- | --- | --- |
| 0 | `{}` | `80` |
| 127 | `7f` | `7f` |
| 128 | `80` | `8180` |
| 255 | `ff` | `81ff` |
| 256 | `0100` | `820100` |
| 32767 | `7fff` | `827fff` |
| 32768 | `8000` | `828000` |

You can reuse the same idea that currently exists in  `org.ethereum.core.transaction.encoder`: `GoldenTransactionVectors` holds pinned hex constants with a comment forbidding regeneration from production output, and `ReferenceRlpEncoder` is a second RLP implementation written from the spec. Every vector there currently uses nonce `{0x01}`, so no boundary value is covered. Adding three
boundary-nonce vectors would reuse the machinery and would catch the case above.

italo-sampaio (commented):
Looks good. Left one suggestion for a test, let me know if that makes sense

PR #3698, opened 2026-08-28
fix: apply authorization tuple checks per tuple instead of at parse time

dlealmac:
Fixes validations to ensure not invalidate whole transaction for tuples processing failures.

## Description
RSKIP-545 has two validation tiers: decode bounds (`nonce < 2**64`, `y_parity < 2**8`, `r`/`s < 2**256`, ...) invalidate the whole transaction, while processing-step failures skip just that tuple. Four checks were enforced in the wrong tier.

`AuthorizationListCodec` now decodes to the actual bounds (nonce `< 2^64`, any single-byte `y_parity`, `r`/`s < 2^256` without the curve-range check). `SetCodeAuthorization` gains `verifyYParity()` and `verifySignatureComponents()`, run during processing in `checkRecoveredAuthority` where the executor already rolls back and continues per tuple.

Also catches `IllegalArgumentException` during recovery (Bouncy Castle's error for a non-curve `r`), and masks the authorization-list `y_parity` in `TransactionResultDTO` so values above 127 aren't sign-extended.

## Motivation and Context
Malformated tuple discarded the whole transaction along with valid siblings. The changes fix it.

## How Has This Been Tested?
Updated unit tests assert the previously-rejected values now decode, that processing still rejects and skip them, and that two-tuple transaction with a bad tuple first still executes and applies the good tuple.

## Types of changes

## Checklist:

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
+++ b/rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
@@ -45,7 +45,7 @@ public final class AuthorizationListCodec {
 
     private static final int TUPLE_FIELD_COUNT = 6;
     private static final BigInteger MAX_CHAIN_ID = BigInteger.ONE.shiftLeft(256);
-    private static final BigInteger MAX_NONCE = BigInteger.ONE.shiftLeft(64).subtract(BigInteger.ONE);
+    private static final BigInteger MAX_NONCE = BigInteger.ONE.shiftLeft(64);
     private static final BigInteger MAX_SIGNATURE_COMPONENT = BigInteger.ONE.shiftLeft(256);
 
     private AuthorizationListCodec() {}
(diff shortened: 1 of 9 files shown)


rskj-core/src/test/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodecTest.java
meldsun0, line 1: Add a decodeTuple_highS_decodes test so we can verify that an authorization tuple with a high-s value is still accepted during decoding as long as s < 2^256.

rskj-core/src/test/java/org/ethereum/core/transaction/parser/Type4RawTransactionParserTest.java
meldsun0, line 1: Add a test that verifies that a high-s authorization fails during tuple processing.

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
meldsun0, line 121: Could we simplify decodeTuple by extracting the decode + validation logic for each field into dedicated methods?
For example, decodeAndValidateNonce(), decodeAndValidateYParity(), and decodeAndValidateSignatureComponent().
This would make the method map more directly to the authorization tuple fields in the spec.
It would also make the distinction between decode-time and processing-time validation clearer.

meldsun0 (changes requested):

rskj-core/src/test/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodecTest.java
dlealmac: done!

rskj-core/src/test/java/org/ethereum/core/transaction/parser/Type4RawTransactionParserTest.java
dlealmac: done!

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
dlealmac: done!

italo-sampaio (approved):

meldsun0 (approved):

PR #3705, opened 2026-08-31
fix: enforce canonical RLP scalar encodings in authorization tuples

dlealmac:
## Description
Enforces canonical RLP scalar encodings for the six-field RSKIP545 authorization tuple on both the decode and the encode side.

## Motivation and Context
decodeTuple validated each field's length and value but never that its encoding was minimal, so five of the six fields accepted non-canonical RLP. On the encode side nothing owned the nonce's canonical form: encodeTuple byte-preserved it through RLP.encodeElement, which returns a single zero byte verbatim, and getSigningHash read the same unnormalized bytes. One root cause — the tuple's scalars had no single point that decided what their minimal form is.

## How Has This Been Tested?
New tests was added covering non-minimal chain_id, nonce, y_parity, r and s.

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.
- [X] Requires Activation Code (Hard Fork)

* **Other information**:

### DIFF
--- a/rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
+++ b/rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
@@ -31,6 +31,7 @@ import org.ethereum.util.RLPList;
 
 import java.math.BigInteger;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
@@ -124,6 +125,7 @@ public final class AuthorizationListCodec {
 
         byte[] chainIdData = inner.get(0).getRLPData();
         CommonParsingUtils.requireDataWordBytes(chainIdData, "Authorization chain_id is not valid");
+        CommonParsingUtils.requireCanonicalScalar(chainIdData, "Authorization chain_id");
         BigInteger chainId = decodeChainId(chainIdData);
         RskAddress address = decodeAddress(inner.get(1).getRLPData());
         byte[] nonce = inner.get(2).getRLPData();
@@ -133,21 +135,37 @@ public final class AuthorizationListCodec {
             nonce = ByteUtil.cloneBytes(nonce);
         }
         CommonParsingUtils.requireDataWordBytes(nonce, "Authorization nonce is not valid");
+        CommonParsingUtils.requireCanonicalScalar(nonce, "Authorization nonce");
         validateNonceValue(decodeNonce(nonce));
-        byte yParity = parseYParity(inner.get(3).getRLPData());
+        byte[] yParityData = inner.get(3).getRLPData();
+        CommonParsingUtils.requireCanonicalScalar(yParityData, "Authorization y_parity");
+        byte yParity = parseYParity(yParityData);
         byte[] r = inner.get(4).getRLPData();
         byte[] s = inner.get(5).getRLPData();
         if (r == null || s == null) {
             throw new IllegalArgumentException("Authorization list tuple signature is incomplete");
         }
-        CommonParsingUtils.requireSignatureComponent(r, "Authorization signature r is not valid");
-        CommonParsingUtils.requireSignatureComponent(s, "Authorization signature s is not valid");
+        CommonParsingUtils.requireDataWordBytes(r, "Authorization signature r is not valid");
+        CommonParsingUtils.requireDataWordBytes(s, "Authorization signature s is not valid");
+        CommonParsingUtils.requireCanonicalScalar(r, "Authorization signature r");
+        CommonParsingUtils.requireCanonicalScalar(s, "Authorization signature s");
         byte v = (byte) (Transaction.LOWER_REAL_V + yParity);
         ECDSASignature signature = ECDSASignature.fromComponents(r, s, v);
 
         SetCodeAuthorization auth = new SetCodeAuthorization(chainId, address, nonce, signature);
         validateAuthorization(auth);
+        requireCanonicalTupleEncoding(tupleBytes, auth);
         return auth;
+    }
+
+    /**
+     * Rejects a tuple whose RLP framing is not minimal, by requiring that re-encoding it reproduces
+     * the bytes received.
+     */
+    private static void requireCanonicalTupleEncoding(byte[] tupleBytes, SetCodeAuthorization auth) {
+        if (!Arrays.equals(tupleBytes, encodeTuple(auth))) {
+            throw new IllegalArgumentException("Authorization list tuple is not canonically encoded");
+        }
     }
 
     private static SetCodeAuthorization parseCallArgumentsEntry(CallArguments.AuthorizationListEntry entry, int index) {
--- a/rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
+++ b/rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
@@ -93,6 +93,11 @@ public final class AuthorizationListCodec {
 
     public static byte[] encodeTuple(SetCodeAuthorization auth) {
         validateAuthorization(auth);
+        return encodeTupleUnchecked(auth);
+    }
+
+    /** Pure encoder. The round-trip check in decodeTuple has already validated the authorization. */
+    private static byte[] encodeTupleUnchecked(SetCodeAuthorization auth) {
         byte yParity = (byte) (auth.getSignature().getV() - Transaction.LOWER_REAL_V);
         return RLP.encodeList(
                 RLP.encodeBigInteger(auth.getChainId()),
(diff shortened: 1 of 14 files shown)


rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/TypedTransactionCodec.java
```diff
@@ -41,7 +41,9 @@ public static SignatureState parseTypedSignatureState(RLPList txFields,
         byte[] s = txFields.get(sIndex).getRLPData();
         // Validated before the unsigned early return so the field is checked on every path.
         // Callers run requireFieldCount first, so the index always exists.
-        byte yParity = parseTypedYParity(txFields.get(yParityIndex).getRLPData());
+        byte[] yParityData = txFields.get(yParityIndex).getRLPData();
+        CommonParsingUtils.requireCanonicalScalar(yParityData, "Typed transaction yParity");
+        byte yParity = parseTypedYParity(yParityData);
```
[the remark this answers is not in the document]
dlealmac: this is fixed

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/CommonParsingUtils.java
```diff
@@ -52,6 +52,16 @@ public static void requireDataWordCoin(Coin coin, String message) {
         }
     }
 
+    /**
+     * Rejects a scalar field that was received with a leading zero byte.
+     */
+    public static void requireCanonicalScalar(byte[] field, String fieldLabel) {
+        if (field != null && field.length > 0 && field[0] == 0) {
+            throw new IllegalArgumentException(
+                    fieldLabel + " must not have leading zero bytes; zero is encoded as the empty string");
```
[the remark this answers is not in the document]
dlealmac: this is fixed.

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
meldsun0, line 121: Simplify this method as mentioned in this [PR](https://github.com/rsksmart/rskj/pull/3698) and check the following:
validateAuthorization(auth) runs at line 156, then requireCanonicalTupleEncoding (line 157) re-encodes via encodeTuple(auth) (line 166), whose  first statement (line 95) calls validateAuthorization again — same chainId/nonce/signature checks, twice per tuple. The canonical round-trip check itself is correct and should stay; the duplication is just that encodeTuple bundles validate+encode instead of being a pure encoder.
A fix could be to  extract a pure encode-only method for the round-trip to call, keeping 
validateAuthorization only where it's actually decided (line 156).

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/CommonParsingUtils.java
```diff
@@ -52,6 +52,16 @@ public static void requireDataWordCoin(Coin coin, String message) {
...
+        }
+    }
+
     public static void requireSignatureComponent(byte[] component, String message) {
```
meldsun0: This is not only validating the encoding but also normalizing it. Please rename the method name so we avoid using it on RLP path.

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
meldsun0: Should this be renamed to decodeUnsignedBigInteger? Is not nonce specific.

meldsun0: Maybe is better to name it to requireNonceInRange to show better the intention of the method.

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
```diff
@@ -133,23 +135,39 @@ public static SetCodeAuthorization decodeTuple(RLPElement element) {
...
         validateAuthorization(auth);
+        requireCanonicalTupleEncoding(tupleBytes, auth);
         return auth;
     }
 
+    /**
+     * Rejects a tuple whose RLP framing is not minimal, by requiring that re-encoding it reproduces
+     * the bytes received.
+     */
+    private static void requireCanonicalTupleEncoding(byte[] tupleBytes, SetCodeAuthorization auth) {
```
meldsun0: It should be clear that this validate the complete RLP frame so maybe we can name it requireCanonicalAuthorizationTubleRlp

rskj-core/src/test/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodecTest.java
```diff
@@ -673,6 +673,163 @@ void encodeTuple_oversizedR_throws() {
...
+        IllegalArgumentException ex = assertThrows(IllegalArgumentException.class,
+                () -> decodeSingleTuple(tuple));
+        assertTrue(ex.getMessage().contains("signature s must not have leading zero bytes"), ex.getMessage());
+    }
+
+    @Test
+    void decodeTuple_oversizeSignatureRWithLeadingZero_reportsLengthError() {
+        SetCodeAuthorization reference = Rskip545TestSupport.minimalAuthorization((byte) 33);
+        byte[] tuple = rebuildTupleField(reference, 4,
+                RLP.encodeElement(withLeadingZero(BigIntegers.asUnsignedByteArray(reference.getSignature().getR()))));
```
italo-sampaio: This test might be flaky and fail roughly 1 run in 256.

`minimalAuthorization` signs with a fresh `new ECKey()` on each call, so `r` is a new random scalar every time. `asUnsignedByteArray` returns 32 bytes only while `r >= 2**248`; when the draw is smaller (p ≈ 1/256) the minimal encoding is 31 bytes, `withLeadingZero` yields 32 rather than 33, `requireDataWordBytes` no longer fires, and `requireCanonicalScalar` throws first instead:

```
Authorization signature r must not have leading zero bytes; zero is encoded as the empty string
```

which fails the `"Authorization signature r is not valid"` assertion. Looping this exact construction 5,000 times gave 19 failures (0.38%, against 1/256 = 0.39%), and the count of short-`r` draws matched the failure count exactly.

Since the point of the test is the *ordering* — length rule ahead of minimality rule — pinning the payload to 33 bytes preserves that and drops the randomness. Mirrors the neighbouring `zeroPaddedToDataWord`, padding up instead of down:

```java
private static byte[] zeroPaddedBeyondDataWord(BigInteger value) {
    byte[] magnitude = BigIntegers.asUnsignedByteArray(value);
    byte[] padded = new byte[Transaction.DATAWORD_LENGTH + 1];
    int taken = Math.min(Transaction.DATAWORD_LENGTH, magnitude.length);
    System.arraycopy(magnitude, magnitude.length - taken, padded, padded.length - taken, taken);
    return padded;
}
```

```java
byte[] tuple = rebuildTupleField(reference, 4,
        RLP.encodeElement(zeroPaddedBeyondDataWord(reference.getSignature().getR())));
```

That is always 33 bytes with at least one leading zero regardless of the draw, so the length rule fires first every time. I ran the patched version over 5,000 draws — 9 of them short-`r`, which would each have failed the current version — with zero failures.

For what it's worth, the other new cases here look fine: `zeroPaddedToDataWord` builds a fixed 32 bytes from the low 31, so those stay deterministic whatever `r` comes out as. This is the only one that inherits the signature's own length.

italo-sampaio (commented):
Looks good overall! Posted one comment, there's one test that might be flaky (in theory), worth checking.

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/CommonParsingUtils.java
meldsun0, line 31: Would be nice to review the Transaction entity for duplicated validation logic.
Some of its data-word and signature checks overlap with validations already available in CommonParsingUtils.
Reusing those helpers would keep validation rules centralized and avoid maintaining the same checks in multiple places.

meldsun0 (changes requested):
The rule should be: normalize values at the RPC boundary, but never normalize already-encoded RLP.
parseFromCallArguments may normalize inputs because it is converting RPC representations into numeric values.
decodeTuple must validate the bytes exactly as received and reject any non-canonical scalar encoding.
So, RPC → normalize; RLP → validate and reject if non-canonical.

rskj-core/src/test/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodecTest.java
dlealmac: done!

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
dlealmac: done!

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/CommonParsingUtils.java
dlealmac: done!

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
dlealmac: done!

dlealmac: done!

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/AuthorizationListCodec.java
dlealmac: done!

rskj-core/src/main/java/org/ethereum/core/transaction/parser/util/CommonParsingUtils.java
dlealmac: I totally agree, but this has started taking longer than expected. I suggest we leave it for now, and keep it for future refactoring.

meldsun0 (approved):

italo-sampaio (approved):

PR #3712, opened 2026-09-08
docs(node-operators): apply the import-sync review findings from RSKCORE-5740

core-baguira:
Applies the review I left on #3676 ([review](https://github.com/rsksmart/rskj/pull/3676#pullrequestreview-5137186890)) so the fixes are not blocked on your return. **Base is `is/docs-import-sync`, not `master`** — merging this brings the corrections into #3676, and #3676 stays yours to merge.

Documentation only; ten commits, one per finding, so you can drop or reorder any single fix. Every item was found while testing the page end to end: a `VETIVER-9.0.4` JAR imported Testnet in 227 s and landed at 8,046,000, exactly the height the page's own `jq` query predicts, and the failure messages I triggered matched the table word for word. These are the paths where it did not hold.

| Commit | Finding | Review comment |
|---|---|---|
| say that every signer's index must be reachable | `BootstrapIndexRetriever` fetches every configured key's index and throws on the first failure, before any quorum comparison. Serving 2 of the 3 real Testnet indexes and 404ing the third aborted the import in ~4 s, though both served indexes agreed at 8,046,000 and the threshold is 2. The page implied the opposite twice. | [#discussion_r3954206285](https://github.com/rsksmart/rskj/pull/3676#discussion_r3954206285) |
| give DbMigrate its required target option | `co.rsk.cli.tools.DbMigrate rocksdb` fails with `Missing required option: '--targetDb=<targetdb>'`. `-t rocksdb` runs. `02-cli.md` in this branch already had the correct form, so the two pages disagreed. | [#discussion_r3954206302](https://github.com/rsksmart/rskj/pull/3676#discussion_r3954206302) |
| point the out-of-memory row at the terminal | `Start.main` catches `Exception`, so `OutOfMemoryError` never reaches the logger — the only row in the table that cannot appear in `logs/rsk.log`. Also splits the two causes: 9.0.3 dies with `Required array size too large` on today's 2.3 GB payload, where raising `-Xmx` cannot help. | [#discussion_r3954206307](https://github.com/rsksmart/rskj/pull/3676#discussion_r3954206307) |
| log the Ubuntu import where the service logs | Adds `-Dlogback.configurationFile=/etc/rsk/logback.xml`, which the unit passes and which writes to `/var/log/rsk/rsk.log`. Without it the import logged to `./logs/rsk.log` under the admin's cwd, while the bullets say to watch the logs. | [#discussion_r3954206312](https://github.com/rsksmart/rskj/pull/3676#discussion_r3954206312) |
| cover the v2 framing errors in the failure table | `Error reading bootstrap data from <path>` comes from the one-byte format sniff; the streaming read spells it `-data v2`. One combined row covers the v2 reader's framing errors rather than ten. | [#discussion_r3954206319](https://github.com/rsksmart/rskj/pull/3676#discussion_r3954206319) |
| describe what mixing the two networks really does | Six index files derive a threshold of four, which neither network meets, so the query prints `null` rather than a Mainnet height with two agreeing signers. Confirmed with 3 Testnet + 3 Mainnet indexes; the wrong-network answer needs a partial mix. | [#discussion_r3954206323](https://github.com/rsksmart/rskj/pull/3676#discussion_r3954206323) |
| tell the two index rows apart | A dead publisher produces the index message, not the quorum one, so the first row should not point at the reader's own network. Both rows now note the database is already erased. | [#discussion_r3954206297](https://github.com/rsksmart/rskj/pull/3676#discussion_r3954206297) |
| show the two v2 lines a real import writes | Both networks publish v2 now, so an import logs the format detection and a v2 summary as well. Taken from my Testnet run; the DEBUG `Inserting blocks and state...` line is deliberately left out. | [#discussion_r3954206328](https://github.com/rsksmart/rskj/pull/3676#discussion_r3954206328) |
| stop calling import sync experimental v0.1 | `03-reference.md` lines 112-115 still described experimental v0.1 served from an S3 bucket, and this branch routes readers there twice. Could not be an inline suggestion — the lines are outside the diff. | review body, note 1 |
| state the temporary-space rule instead of a multiplier | The figure appeared three times and two disagreed (3x, "more than twice", 3x). Measured peak on the current Testnet archive is 2.28x, and there is no third copy since download, hash check and unzip all stream — so the rule replaces the constant. | review body, note 2 |

Verified with `.github/workflows/scripts/devportal-update/process_docs.py` over the corrected pages: exit 0, no errors, front matter and H1 handling unchanged, code fences balanced, the `<details>` failure table still two columns on every row, tabs and admonitions intact.

Wording is the same text the suggestion blocks carried, so nothing here should be new to read. Rationale and measurements for each item are in the commit bodies.


fmacleal (approved):
LGTM

Good findings, congrats!

PR #3713, opened 2026-09-08
Adding typed transaction to eth_call and eth_estimateGas with tests

meldsun0:
## Description
 Extends eth_call and eth_estimateGas to correctly simulate Type 1 (RSKIP-546/EIP-2930), Type 2 (RSKIP-546/EIP-1559) and Type 4 (RSKIP-545/EIP-7702) transactions, instead of always treating the call as legacy. 

## Motivation and Context
RSKIP-545 and RSKIP-546 added Type 1/2/4 transaction support to the protocol, but eth_call/eth_estimateGas still simulated every call as legacy, ignoring authorizationList, accessList, and fee caps. This left it unable it to estimate gas or simulate EIP-7702 delegations before sending them on-chain.

## How Has This Been Tested?
  - Existing unit tests updated for the new signatures.                                                                                                                                                                                  
  - Two new integration test suites against a real test blockchain (no mocks): EthModuleType4CallSimulationTest, EthModuleTypedCallSimulationTest.

## Types of changes

## Checklist:

* **Other information**:
RPC/simulation-only change — no consensus or block-validation rules touched, so no activation code is needed.

### DIFF
--- a/rskj-core/src/main/java/co/rsk/core/ReversibleTransactionExecutor.java
+++ b/rskj-core/src/main/java/co/rsk/core/ReversibleTransactionExecutor.java
@@ -27,6 +27,7 @@ import org.ethereum.core.TransactionBuilder;
 import org.ethereum.core.TransactionExecutor;
 import org.ethereum.core.transaction.SetCodeAuthorization;
 import org.ethereum.core.transaction.TransactionType;
+import org.ethereum.core.transaction.parser.util.CommonParsingUtils;
 import org.ethereum.util.ByteUtil;
 import org.ethereum.util.RLP;
 import org.ethereum.vm.PrecompiledContracts;
@@ -97,7 +98,7 @@ public class ReversibleTransactionExecutor {
                                                     ReversibleTransactionParams params) {
         Repository track = snapshot.startTracking();
 
-        ReversibleTransaction tx = new ReversibleTransaction(track.getNonce(params.fromAddress()).toByteArray(), params);
+        ReversibleTransaction tx = new ReversibleTransaction(CommonParsingUtils.unsignedBytes(track.getNonce(params.fromAddress())), params);
 
         TransactionExecutor executor = transactionExecutorFactory
                 .newInstance(tx, 0, coinbase, track, executionBlock, 0, precompiledContracts)
@@ -168,12 +169,12 @@ public class ReversibleTransactionExecutor {
         }
 
         /**
-         * Type 2/4 transactions need both maxPriorityFeePerGas and maxFeePerGas set. If the caller
-         * gave us one or both we use those; if not, the default is to reuse gasPrice for the
-         * missing one.
+         * Type 2/4 transactions need both maxPriorityFeePerGas and maxFeePerGas set. CallArgumentsToByteArray
+         * rejects gasPrice mixed with either fee-cap field and a single fee-cap field given alone, so by the
+         * time we get here it's always both caps present, or neither — never just one. When neither was given
+         * (Type 4 via authorizationList with no fee info at all), both default to gasPrice.
          */
-        private static void applyFees(TransactionBuilder builder, TransactionType type, Coin gasPriceCoin,
-                                       byte[] maxPriorityFeePerGas, byte[] maxFeePerGas) {
+        private static void applyFees(TransactionBuilder builder, TransactionType type, Coin gasPriceCoin, byte[] maxPriorityFeePerGas, byte[] maxFeePerGas) {
             if (type == TransactionType.TYPE_2 || type == TransactionType.TYPE_4) {
                 Coin priority = maxPriorityFeePerGas != null ? RLP.parseCoinNonNullZero(ByteUtil.cloneBytes(maxPriorityFeePerGas)) : gasPriceCoin;
                 Coin maxFee = maxFeePerGas != null ? RLP.parseCoinNonNullZero(ByteUtil.cloneBytes(maxFeePerGas)) : gasPriceCoin;
--- a/rskj-core/src/main/java/co/rsk/core/ReversibleTransactionExecutor.java
+++ b/rskj-core/src/main/java/co/rsk/core/ReversibleTransactionExecutor.java
@@ -23,118 +23,80 @@ import co.rsk.db.RepositorySnapshot;
 import org.ethereum.core.Block;
 import org.ethereum.core.Repository;
 import org.ethereum.core.Transaction;
+import org.ethereum.core.TransactionBuilder;
 import org.ethereum.core.TransactionExecutor;
+import org.ethereum.core.transaction.SetCodeAuthorization;
+import org.ethereum.core.transaction.TransactionType;
+import org.ethereum.util.ByteUtil;
+import org.ethereum.util.RLP;
 import org.ethereum.vm.PrecompiledContracts;
 import org.ethereum.vm.program.ProgramResult;
+
+import java.util.List;
 
 /**
  * Encapsulates the logic to execute a transaction in an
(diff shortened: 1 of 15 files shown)


rskj-core/src/main/java/co/rsk/core/ReversibleTransactionExecutor.java
```diff
@@ -145,20 +109,88 @@ private TransactionExecutor reversibleExecution(RepositorySnapshot snapshot, Blo
...
+         * resolve the type ourselves and fill in the missing fee-cap defaults (see #applyFees)
+         * before building.
+         */
+        private static Transaction buildReversibleTransaction(byte[] nonce, ReversibleTransactionParams params) {
+            TransactionType type = params.type();
+            Coin gasPriceCoin = RLP.parseCoinNonNullZero(ByteUtil.cloneBytes(params.gasPrice()));
+
+            TransactionBuilder builder = Transaction.builder()
+                    .type(type)
+                    .isLocalCall(true)
```
dlealmac, line 153: ReversibleTransaction runs with `.isLocalCall(true)`, while the TransactionExecutor.execute() puts the nonce increment inside the `!localCall` block, before the type-4 verification:

```java
if (!localCall) {
    track.increaseNonce(tx.getSender(signatureCache));
    ...
}
if(tx.isType4()){
    authorizationRefund = processAuthorizationList(tx.getAuthorizationList(), track, tx);
```

So simulation skips the increment but still processes the tuples, the execution can run with nonce+1 while simulation runs with nonce+0.

Failure scenario: a user signs the authorization they will actually broadcast (senderNonce+1) and calls eth_estimateGas. The tuple mismatches and the failure is swallowed at TransactionExecutor.java:345 (catch → rollback → logger.debug). The call runs against the bare EOA, so the estimate is 21000 + 25000·n with no delegated execution gas. They broadcast and run out of gas — exactly "producing a gas limit below the minimum required by the real Type-4 transaction", the failure the issue was filed to prevent.

rskj-core/src/main/java/co/rsk/core/ReversibleTransactionExecutor.java
```diff
@@ -23,118 +23,82 @@
...
-    }
+            List<SetCodeAuthorization> authorizationList,
+            byte chainId,
+            TransactionType type,
+            byte[] accessListBytes,
+            byte[] maxPriorityFeePerGas,
+            byte[] maxFeePerGas
+    ) {}
 
     private TransactionExecutor reversibleExecution(RepositorySnapshot snapshot, Block executionBlock, RskAddress coinbase,
```
dlealmac, line 97: TransactionExecutor only checks the rskip545 activation (line 174) after the localCall check (line 170), so that simulation can be performed before AA activation, then eth_call will returns 0x as a success and eth_estimateGas returns 6,800,000 for a transaction the node would refuse at all.

rskj-core/src/main/java/co/rsk/core/ReversibleTransactionExecutor.java
```diff
@@ -23,118 +23,82 @@
...
-                value,
-                data,
-                fromAddress,
-                precompiledContracts
-        );
+    /**
+     * Estimates gas against the provided snapshot using the executor's
+     * configured precompiled contracts.
+     */
+    public TransactionExecutor estimateGas(Block executionBlock, RskAddress coinbase, RepositorySnapshot snapshot,
```
dlealmac, line 59: EthModule.java:245 always passes gasEstimationCap as the tx gas limit; the local-call short-circuit skips the txGasLimit < basicTxCost guard at TransactionExecutor.java:300; Program.spendGas then throws without spending, so maxGasUsed stays 0 and localCallFinalization returns early.

Measured with 300 authorizations (intrinsic ≈ 7.5M > 6.8M cap): 0x0. The real transaction needs ≥ 7.5M. That's the issue's central rule not merely violated but inverted — and before this PR authorizationList was ignored, so reaching it took ~100 KB of calldata.

Test that captured that scenario:

```java
    @Test
    void estimateGas_mustNotReturnZeroWhenIntrinsicGasExceedsTheCap() {
        boot(0);
        int tupleCount = 300;
        long intrinsic = 21_000L + GasCost.PER_EMPTY_ACCOUNT_COST * tupleCount;
        assertTrue(intrinsic > GAS_ESTIMATION_CAP, "precondition: intrinsic must exceed the cap");

        List<SetCodeAuthorization> tuples = new ArrayList<>();
        for (int i = 0; i < tupleCount; i++) {
            tuples.add(Rskip545TestSupport.createSignedAuthorization(new ECKey(), const42, BigInteger.ZERO, CHAIN_ID));
        }

        long estimate;
        try {
            estimate = estimateGas(const42, GAS_CAP_HEX, tuples, null);
        } catch (RskJsonRpcRequestException e) {
            return; // reporting "gas required exceeds allowance" is correct
        }
        assertTrue(estimate >= intrinsic,
                "eth_estimateGas must never return less than the transaction's intrinsic cost; "
                        + "intrinsic=" + intrinsic + " returned=" + estimate);
    }
```

dlealmac (changes requested):

rskj-core/src/test/java/co/rsk/rpc/modules/eth/EthModuleGasEstimationDSLTest.java
```diff
@@ -72,7 +72,7 @@ void testEstimateGas_basicTests() throws FileNotFoundException, DslProcessorExce
 
         assertEquals(0, eth.getEstimationResult().getDeductedRefund());
 
-        ProgramResult callConstantResult = eth.callConstant(args, block);
```
italo-sampaio, line 76: These tests now exercise `simulateTransactionExecution`, which duplicate production code and re-implement `EthModule.call`. This means that a divergence in `call` would still pass the tests.

Before this PR, `call` delegated to `callConstant`, and these tests called it directly. Could we keep that behavior? Move the typed-params execution into `callConstant` (both overloads), have `call` delegate to it, and point these tests back at `eth.callConstant(args, block)`. That also removes the copied fields in `EthModuleTestUtils` and lets `buildParams` go back to private.

rskj-core/src/main/java/co/rsk/core/ReversibleTransactionExecutor.java
meldsun0: This bug pre-dates type4 — gasIsValid() has always been skipped for localCall, so any tx type could already hit intrinsic-gas-exceeds-cap via large calldata, authorizationList just makes it much cheaper to trigger. Let's fix it separately, keeping this PR scoped to type4.

rskj-core/src/main/java/org/ethereum/rpc/converters/CallArgumentsToByteArray.java
```diff
@@ -127,6 +137,51 @@ public RskAddress getFromAddress() {
...
+        if (args.getAuthorizationList() != null) {
+            return TransactionType.TYPE_4;
+        }
+        if (!isAbsent(args.getMaxFeePerGas()) || !isAbsent(args.getMaxPriorityFeePerGas())) {
+            return TransactionType.TYPE_2;
+        }
+        if (args.getAccessList() != null) {
+            return TransactionType.TYPE_1;
+        }
+        return TransactionType.LEGACY;
```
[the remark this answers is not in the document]
meldsun0: Shape is inferred from the provided fields same as what other clients are doing.  So this comment is not valid.

dlealmac (approved):
LGTM

italo-sampaio (approved):
LGTM!

PR #3717, opened 2026-09-11
fix: report gas usage for stored typed transactions

dlealmac:
## Description
Parity-style traces (trace_transaction, trace_block, trace_filter, trace_get) reported "gasUsed": "0x0" on the successful root frame of every confirmed Type 1, Type 2 and Type 4 transaction. TraceModuleImpl already re-executes the block to build the program trace, but discarded the returned BlockResult — which carries a fresh receipt per transaction with gasUsed intact. The root frame's gas now comes from there.

## Motivation and Context
RSKIP-545/546 and the corresponding Ethereum EIPs intentionally omit individual gas usage from typed receipts and store only cumulativeGasUsed. After a typed receipt is persisted and decoded, its gasUsed field is therefore empty, but TraceTransformer was reading that field directly and converts it to zero. 

## How Has This Been Tested?

- New TraceModuleTypedReceiptGasTest (17 tests): all four trace RPCs × Type 1 / 2 / 4, on a DSL fixture whose four transactions span three RSKIP-144/351 sublists (edges {1,2}, asserted). Preconditions assert the stored receipts genuinely carry a zero-length gasUsed, and one Type 2 transaction sits mid-sublist so cumulative gas differs from per-transaction gas. 

- New TraceGasCharacterizationTest (7 tests): frame-by-frame legacy gas pins for nested creation, nested call, revert, suicide and subcall. Green before and after, to show legacy output did not move.

- TraceModuleImplTest: one added pending-block test, using a legacy transaction.

## Types of changes

## Checklist:


rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -102,15 +105,16 @@ public JsonNode traceTransaction(String transactionHash) {
         txInfo.setTransaction(tx);
 
         ProgramTraceProcessor programTraceProcessor = new ProgramTraceProcessor();
-        this.blockExecutor.traceBlock(programTraceProcessor, VmConfig.LIGHT_TRACE, block, parent.getHeader(), false, false);
+        BlockResult blockResult = this.blockExecutor.traceBlock(programTraceProcessor, VmConfig.LIGHT_TRACE, block, parent.getHeader(), false, false);
```
meldsun0: Low-priority: As we only only need one transaction, building Map for all the block might be unnecessary.

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -102,15 +105,16 @@ public JsonNode traceTransaction(String transactionHash) {
...
 
         SummarizedProgramTrace programTrace = (SummarizedProgramTrace) programTraceProcessor.getProgramTrace(tx.getHash());
 
         if (programTrace == null) {
             return null;
         }
 
-        List<TransactionTrace> traces = TraceTransformer.toTraces(programTrace, txInfo, block.getNumber());
+        long rootGasUsed = rootGasUsed(gasUsedByTx(blockResult), txInfo, tx.getHash());
```
meldsun0: Low-priority: As we only only need one transaction, building a Map for all the block might be unnecessary.

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -287,6 +293,35 @@ private List<TransactionTrace> buildBlockTraces(Block block, TraceFilterRequest
...
+     * Per-tx gas from the re-execution, keyed by tx hash: stored Type 1/2/4 receipts omit the field
+     * (RSKIP-545/546).
+     */
+    private static Map<Keccak256, Long> gasUsedByTx(BlockResult blockResult) {
+        Map<Keccak256, Long> gasUsedByTx = new HashMap<>();
+
+        for (TransactionReceipt receipt : blockResult.getTransactionReceipts()) {
+            Transaction tx = receipt.getTransaction();
+            if (tx != null) {
+                gasUsedByTx.put(tx.getHash(), new BigInteger(1, receipt.getGasUsed()).longValue());
```
meldsun0: Low: BigInteger.longValue() silently truncates values that exceed the long range, which could hide an unexpected or corrupted gasUsed value.
Consider using longValueExact() so an invalid out-of-range value fails explicitly instead of producing an incorrect result. Anyway, there is no expected valid or user-controlled path where gasUsed can exceed the long range.

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
```diff
@@ -287,6 +293,35 @@ private List<TransactionTrace> buildBlockTraces(Block block, TraceFilterRequest
...
+    private static long rootGasUsed(Map<Keccak256, Long> freshGasUsed, TransactionInfo txInfo, Keccak256 txHash) {
+        Long fresh = freshGasUsed.get(txHash);
+
+        if (fresh != null) {
+            return fresh;
+        }
+
+        logger.warn("No re-execution receipt for tx {}, root trace gas falls back to the stored receipt", txHash);
+
+        return new BigInteger(1, txInfo.getReceipt().getGasUsed()).longValue();
```
meldsun0: Since re-execution is now the source of truth for root gas usage, should we fail when the fresh receipt is missing instead of falling back to the stored receipt? For typed receipts that fallback can silently turn an unavailable value into gasUsed = 0, recreating the original issue.

meldsun0 (approved):

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: you are right, done!

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: done!

rskj-core/src/main/java/co/rsk/rpc/modules/trace/TraceModuleImpl.java
dlealmac: done!

italo-sampaio (approved):
LGTM!

meldsun0 (approved):

